PR #1038, opened 2014-11-07
Workflow menus

jakirkham:
Adds support for workflow menus.


jakirkham:
@stuarteberg, I recently noticed that we never merged this. Is there something else we need here before merging?

PR #1054, opened 2014-12-03
Cleanup convenience methods applet serializer

jakirkham:
Changed the convenience methods so that `getOrCreateGroup` takes advantage of the method `require_group` from h5py, which does the same thing. Also, this ensures that an existing dataset might not be returned instead of a group. Now, an error will be raised. Changed `deleteIfPresent` to simply attempt deleting and handle an exception in the case that the item does not exist as opposed to performing the search twice.


jakirkham:
Rebase on current master.

jakirkham:
Added tests to ensure helper methods behave correctly.

jakirkham:
@stuarteberg If you don't have any strong opposition, I'd like to merge this.

PR #1059, opened 2014-12-05
Ipcknime

nielsbuwen:
HDF5/CSV Exporter and IPC Server

- adds the option to export object features and raw/labeling images to hdf5 or csv in object classification
- adds a TCP server for interprocess communication that sends out (un/clear)hilite commands and receives and processes commands from KNIME

all features are only available in debug mode

possible features ( because specifying a port number is not user friendly ):
- IPC over named pipes on windows and domain sockets on linux ( connection by file name )
- Use only one named pipe for other programms to ask for a TCP connection over that pipe


nielsbuwen:
Closing this PR.  Cleaned-up version is submitted as ilastik/ilastik#1079

PR #1086, opened 2015-02-25
[READY] Changed memUsageDialog to just refresh instead of recreate the treeView

burgerdev:
Before the change, all collapsed caches would expand every five seconds.


burgerdev:
This is ready.

PR #1091, opened 2015-03-20
Travis workaround

jakirkham:


stuarteberg:
We now use the conda binaries for CI, so this PR isn't necessary.  Also, I switched to using circle-ci, which has a few benefits over Travis-CI.

Issue #933, opened 2014-07-03
Color table list index out of range in labeling gui


martinsch:
When adding more than 16 label classes, the labeling gui throws a "list index out of range" exception. Tested with object classification.

Quick workaround: disable "add label" after 15 label classes.

<pre>
ERROR 2014-07-03 14:11:58,912 log_exception 15669 140300557887296 Logged the above exception just in case PyQt loses it.
ERROR 2014-07-03 14:11:58,912 excepthooks 15669 140300557887296 Traceback (most recent call last):
  File "/home/mschiegg/software/ilastik06/ilastik/ilastik/utility/bind.py", line 64, in __call__
    self.f(*(self.bound_args + args[0:self.numUnboundArgs]))
  File "/home/mschiegg/software/ilastik06/ilastik/ilastik/applets/labeling/labelingGui.py", line 570, in _addNewLabel
    self._labelingSlots.labelNames.setValue( operator_names, check_changed=False )
  File "/home/mschiegg/ilastik06/lazyflow/lazyflow/slot.py", line 89, in call_in_setup_context
    return func(self, *args, **kwargs)
  File "/home/mschiegg/ilastik06/lazyflow/lazyflow/graph.py", line 136, in __exit__
    sig_setup_complete()
  File "/home/mschiegg/ilastik06/lazyflow/lazyflow/utility/orderedSignal.py", line 73, in __call__
    f(*args, **kw)
  File "/home/mschiegg/software/ilastik06/ilastik/ilastik/utility/gui/threadRouter.py", line 70, in routed
    val = func(*args, **kwargs)
  File "/home/mschiegg/software/ilastik06/ilastik/ilastik/applets/layerViewer/layerViewerGui.py", line 401, in updateAllLayers
    newGuiLayers = self.setupLayers()
  File "/home/mschiegg/software/ilastik06/ilastik/ilastik/applets/objectClassification/objectClassificationGui.py", line 590, in setupLayers
    self._setPredictionColorTableForRow( predictLayer, row )
  File "/home/mschiegg/software/ilastik06/ilastik/ilastik/applets/objectClassification/objectClassificationGui.py", line 661, in _setPredictionColorTableForRow
    oldcolor = self._colorTable16_forpmaps[row+1]
IndexError: list index out of range
</pre>

akreshuk:
[...]

PR #1095, opened 2015-03-27
Set the maximum label count to 15

agarwalayush:
Fixes the issue #933

### DIFF
--- a/ilastik/applets/labeling/labelingGui.py
+++ b/ilastik/applets/labeling/labelingGui.py
@@ -92,7 +92,6 @@ class LabelingGui(LayerViewerGui):
         return self._maxLabelNumber
     @maxLabelNumber.setter
     def maxLabelNumber(self, n):
-        self._maxLabelNumber = n
         while self._labelControlUi.labelListModel.rowCount() < n:
             self._removeLastLabel()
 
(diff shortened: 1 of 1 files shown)


ilastik/applets/labeling/labelingGui.py
```diff
@@ -579,7 +579,7 @@ def _updateLabelList(self):
             self._labelControlUi.labelListModel[i].name = n
                 
         if hasattr(self._labelControlUi, "AddLabelButton"):
-            self._labelControlUi.AddLabelButton.setEnabled(numLabels < self.maxLabelNumber)
+            self._labelControlUi.AddLabelButton.setEnabled(numLabels < 15)
```
burgerdev: I would prefer setting `self._maxLabelNumber = 15` over a [magic number](http://en.wikipedia.org/wiki/Magic_number_%28programming%29#Unnamed_numerical_constants) here.

agarwalayush:
okk

agarwalayush:
Please review and merge

stuarteberg:
Sorry, I can't merge this for multiple reasons.

- This breaks the `@maxLabelNumber.setter` function
- The bug referenced here (#933) pertains specifically to object classification.  The right place to fix it is in the `objectClassificationGui.py`.  This change would affect other workflows, like pixel classification.
- Even if we did keep the change in `labelingGui.py`, I'd want to see a comment explaining why the number 15 was chosen.

agarwalayush:
ok. I will find the original problem rather than quick workaround, and will send the pull request soon.

Issue #1083, opened 2015-02-20
CSV/HDF5 Export should be found in the corresponding data-export-applet


martinsch:
Instead of adding a menu item in the object-classification or tracking applet, the csv- and hdf5-export options by @nielsbuwen should be provided in the corresponding data export applet for consistency reasons. Of course, this involves connecting the two applets through many additional slots, but this should be the way to go.

martinsch:
Additionally, for the communication with knime, the export dialog should still be reachable from the object-classification and tracking applet.

nielsbuwen:
Here is a screenshot of what it is currently looking like.
I added an export button to the "Object Prediction Export" applet and renamed the tab.

![inline-export](https://cloud.githubusercontent.com/assets/9267463/6849368/1a3844ae-d3d3-11e4-8b33-8f75273c7a47.png)

I didn't want to add a new entry for the ComboBox, because those exporters seem to work differently from the "KNIME"-exporter.

That button will open the same dialog as the "Export->Export Object Information" menu item.
The menu item also remains where it is.

If this is the way you meant it to be I will create a pull request for that.

stuarteberg:
I think this UI is fine.  Can you rename the heading from "Export Features" to, I dunno, "Export Object Feature Table", or something similar?

Please go ahead and create a PR.  I'd like to take a look at the implementation.

@martinsch, How do you like this placement?

martinsch:
That's better now, thanks.

PR #1097, opened 2015-03-31
Fixing ilastik/ilastik-#1083

nielsbuwen:
This PR moves the KNIME-export into the DataExportApplet for ObjectClassification and ConservationTracking ( see https://github.com/ilastik/ilastik/issues/1083#issuecomment-86564313 ).
It also includes Export for ManualTracking ( only as a menu item, debug only )


stuarteberg:
This branch works for conservation tracking, but it breaks the manual tracking workflow.  Can you fix it?

1. The manualTrackingWorkflow never calls `set_exporting_operator()`
2. If you fix that, the workflow can open, but the export can't be used because `OpManualTracking` is not a subclass of `OpTrackingBase`. (Maybe it should be?  @martinsch?)

Otherwise, these changes look okay.

Side note: Instead of fixing typos with a special "cleanup" commit, `git rebase --interactive` can be a really useful tool for "changing history" to removing typos from the offending commit.  It can help you keep the git history nice and clean (and easier to review).

nielsbuwen:
1. I was a bit confused, because the manualTrackingOperator is not a subclass of the baseTrackingOperator. But Manual Tracking uses the BaseTrackingExportApplet/Gui.

So now the manTrackWorkflow sets the exporting operator ( but does not use it. It will, when I integrate the exporter in its ExportApplet )

2. In debug mode you can use the exporter using the menu entry.

I will try to use interactive rebase more often. But in this case I shouldn't remove the "Cleanup 8d630d8" because it is already here, right?

stuarteberg:
>But in this case I shouldn't remove the "Cleanup 8d630d8" because it is already here, right?

I think github can handle the case of a rebased PR branch, but don't worry about it this time.

stuarteberg:
@nielsbuwen, this still crashed for me when I tried a manual tracking project (I used the mitocheck example from our website).  I fixed it with a minor change before I merged.  Pleas take a quick look at the merge (now in master), and make sure it's satisfactory as far as you're concerned.

For reference, the exception was this:

```
ERROR 2015-06-19 14:51:48,522 excepthooks 29333 140735234900736 Unhandled exception in thread: 'MainThread'
ERROR 2015-06-19 14:51:48,523 excepthooks 29333 140735234900736 Traceback (most recent call last):
  File "/Users/bergs/Documents/workspace/ilastik-meta/ilastik/ilastik/utility/exportingOperator.py", line 123, in show_export_dialog
    feature_names = self.get_feature_names()
  File "/Users/bergs/Documents/workspace/ilastik-meta/ilastik/ilastik/applets/tracking/base/trackingBaseDataExportGui.py", line 44, in get_feature
_names
    return self.get_exporting_operator().ComputedFeatureNamesWithDivFeatures([]).wait()
  File "/Users/bergs/Documents/workspace/ilastik-meta/ilastik/ilastik/utility/operatorSubView.py", line 135, in __getattribute__
    return getattr(self.__innerOp, name)
AttributeError: 'OpManualTracking' object has no attribute 'ComputedFeatureNamesWithDivFeatures'
```

chaubold:
Thanks @stuarteberg, I was going to look into that PR this morning but you were faster :). I just ran it once for an automated tracking project and inspected the resulting exported files. All looked fine to me and worked seamlessly. I also checked the manual tracking workflow, and there the list of features in the selection dialog is empty even though the default features (count, mean, var, region center) do get exported. @nielsbuwen is that displayed list supposed to be empty?

Issue #1099, opened 2015-03-31
Blockwise object classification: Applying new block size takes too long


christophdecker:
After changing the block size and halo in the "Blockwise Object Classification" applet and clicking "Apply" it takes several hours until you can continue. As the problem appeared with a txy dataset of dimensions 1800x1024x1024, I figure it might be due to ilastik computing the result with the new block size for all time frames instead of only the one currently displayed.

stuarteberg:
Can you try the profiler to see where the slowdown resides?

1. Install the `yappi` module from PyPI.
2. Load a smaller dataset (but one that still exhibits at least some delay).  Be sure to use debug mode.
3. Select menu item: Debug > Profiling > Start
4. Reproduce the issue you described (click "Apply").
5. Debug > Profiling > Stop
6. Debug > Profiling > Save Sorted Stats... (I would guess 'cumulative' or 'time' may be the most informative.)

burgerdev:
You are correct, the preview is in fact computing all time frames at once. There are a couple of lines in `OpBlockwiseObjectClassification` that seem to be using odd / invalid cache block shapes, but the following did the trick for me:

```diff
diff --git a/ilastik/applets/blockwiseObjectClassification/opBlockwiseObjectClassification.py b/ilastik/applets/blockwiseObjectClassification/opBlockwiseObjectClassification.py
index 5f156c6..754135d 100644
--- a/ilastik/applets/blockwiseObjectClassification/opBlockwiseObjectClassification.py
+++ b/ilastik/applets/blockwiseObjectClassification/opBlockwiseObjectClassification.py
@@ -429,7 +429,7 @@ class OpBlockwiseObjectClassification( Operator ):
             elif k == 'c':
                 shape[i] = 1
             elif k == 't':
-                shape[i] = self.RawImage.meta.shape[i]
+                shape[i] = 1
             else:
                 assert False,  "Unknown axis key: '{}'".format( k )
```

@stuarteberg: Perhaps it would be worthwhile taking a closer look at this operator again.

stuarteberg:
Nice find, @burgerdev.  I don't think there is a good reason to make the blocks span across all time slices.  Your fix seems exactly correct to me.

Issue #1075, opened 2015-01-19
Memory leak and speed issues with TXY data in Pixel & Object Classification workflow


christophdecker:
I'm currently trying to process a TXY data set of dimensions 800 x 2000 x 2000 (400 MB) with 2 classes in pixel classification and 5 classes in object classification (roughly 50 objects in each of the 800 frames). Since starting the object prediction export 6 hours ago, ilastik greedily allocated 134 GB of RAM but only finished 14% of the job (i.e. only 120 frames or 4 seconds of the input video). 

From this I see several issues:
1. In terms of processing time, this is in the same range as the numbers mentioned in issue #856 - so we still have a major problem with TXY data.
2. There must be a memory leak somewhere; how else could ilastik need so much memory to process such a tiny dataset?
3. ilastik itself claims to use "only" 31 GB of RAM - so there is a discrepancy of factor 4-5 between what ilastik thinks it uses and what the OS sees.

christophdecker:
PS: Scrolling through the object predictions by hand in the object classification applet takes much less time even for "yet unvisited" frames. In numbers: The export takes ~3 minutes per frame, but when jumping to a new frame in ilastik, it only takes ~5 seconds to show the object predictions.

stuarteberg:
Can you upload your data and project file somewhere for me to have a look?

stuarteberg:
>In terms of processing time, this is in the same range as the numbers mentioned in issue #856 - so we still have a major problem with TXY data.

Do you observe this behavior in pixel classification alone?  Or does it only occur when using Pixel Clasification + Object Classification in the same workflow?

>There must be a memory leak somewhere; how else could ilastik need so much memory to process such a tiny dataset?

The following points come to mind:

1. If the workflow contains caches, those caches will accumulate until the system gets low on RAM.  At that point, the `ArrayCacheMemoryMgr` thread will kick in, and it will attempt to expel data from eligible caches.  The expected behavior for any "cacheful" workflow is that ilastik will continue to consume RAM until it approaches the RAM limit.

2. Not all workflows are "cacheful".  Indeed, the PixelClassification **batch** workflow does not use any caches at all.  In that workflow, the RAM usage should be limited solely to the total of intermediate arrays used in during the (parallel) processing of each block.

[...]

PR #1104, opened 2015-04-09
[Ready] fix for #1099 (and possibly others) - 5d-ified object extraction

burgerdev:
fix for #1099. I'm sure this helps with #1075, too.


burgerdev:
@christophdecker @akreshuk @(other people involved with OC) can you give this branch a try? Object classification with time series could be much better now...

christophdecker:
So from a very first test with TXY data, I can tell that it's much faster now, but RAM usage is still pretty high. I'll run it on a bigger file tomorrow to be able to give some real numbers. Thanks anyway, this makes ilastik much more useful for people with TXY data!

burgerdev:
I found another source for slowness: `OpObjectCenterImage` was splitting up its dirty propagation in time, although it has no reason to - it always gets a range as roi. Fixed this in db091a4.

stuarteberg:
This PR looked fine to me, but I did make one change before I merged it: I parallelized the computation across time slices (with a `RequestPool`).

PR #1119, opened 2015-06-17
Added support for importing label masks

holstgr-kaust:
I've added support for an Import... option to the Label Layer.  It enables the user to read in a set of images (or image sequence), and insert them into the specified labels at the given z-plane.

I have tested it successfully with both the Carving and PixelClassification workflows. 

I would welcome a careful review, since I might not have all the lazyflow idioms correct.  In particular how I map from the input image stack into the desired label channel (I use numpy directly instead of lazyflow operators); and the use of the OpOutputProvider to create a simple slot wrapper for the MetaInfoWidget.

### DIFF
--- a/ilastik/applets/labeling/dataImportOptionsDlg.ui
+++ b/ilastik/applets/labeling/dataImportOptionsDlg.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>692</width>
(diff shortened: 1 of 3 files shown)


stuarteberg:
I've done some work on top of this PR, but first I decided to rebase it on the current master.  The result is a new PR: #1128.  I'm closing this PR; let's continue the discussion over there.

PR #1124, opened 2015-06-22
Tracking merger resolving

chaubold:
Introduces proper merger resolving including updating the label image and right-click-information for both 2D and 3D.

screen shot 2015-06-22 at 15 10 27

The old "Merger" layer is still included, but depending on whether "Merger Resolution" is selected or not, it contains a color coded cell count indicator, or the relabeled objects. The LabelImage ("Objects") layer is also updated for display and export.

**TODO:** @nielsbuwen's HDF5/CSV export functionality still works on the old label image, because after updating the segmentation, we need to compute the region features for all new objects. What would be the best way to achieve this? Could we temporarily change the input slot of the OpObjectExtraction to be the label image (it currently takes raw and/or binary), extract the features for exporting to HDF5/CSV and then change it back? But that might break the KNIME bridge I guess. Any hints are highly appreciated.


chaubold:
btw: the Circle CI tests are failing due to missing libhdf5? (and the VM is set up to install ilastik-everything-without-tracking)

stuarteberg:
>btw: the Circle CI tests are failing due to missing libhdf5? (and the VM is set up to install `ilastik-everything-but-tracking`)

Yeah, sorry about that.  There was a defect in my package metadata for vigra.  When the conda folks added a new version of the `hdf5` package to their servers a few days ago, our vigra package tried to use it.  Unfrotunately, we're going to be exposed to issues like this every now and then, until conda/conda#918 is resolved.

stuarteberg:
I need a little help to evaluate this branch and provide advice:

1. What dataset should I use when testing this branch?  Do you have one that has nice examples of merges that need to be resolved?
2. I tried to test this branch with our example mitocheck 2D dataset, but I'm running into an exception when pressing the "Track!" button:

```
ERROR 2015-06-22 20:50:05,026 log_exception 69904 4675784704 Traceback (most recent call last):
  File "/Users/bergs/Documents/workspace/ilastik-meta/ilastik/ilastik/applets/tracking/conservation/conservationTrackingGui.py", line 220, in _track
    force_build_hypotheses_graph = False
  File "/Users/bergs/Documents/workspace/ilastik-meta/ilastik/ilastik/applets/tracking/conservation/opConservationTracking.py", line 248, in track
    "none" # dump traxelstore
ArgumentError: Python argument types in
    ConsTracking.__init__(ConsTracking, int, bool, float, float, bool, float, str, FieldOfView, str)
did not match C++ signature:
    __init__(_object*, int max_number_objects, double max_neighbor_distance, double division_threshold, std::string detection_rf_filename, bool size_dependent_detection_prob, double forbidden_cost, double ep_gap, double avg_obj_size, bool with_tracklets, double division_weight, double transition_weight, bool with_divisions, double disappearance_cost, double appearance_cost, bool with_merger_resolution, int number_of_dimensions, double transition_parameter, double border_width, pgmlink::FieldOfView fov, bool with_constraints, double cplex_timeout, std::string event_vector_dump_filename)

ERROR 2015-06-22 20:50:05,026 log_exception 69904 4675784704 Error during tracking.  See above error traceback.
```

chaubold:
Oh excuse me, I meant to add "**Requires newest commit of https://github.com/martinsch/pgmlink.git**". This should fix your point 2.

Regarding 1., any tracking dataset with merged objects would do. The conservation tracking project of the mitocheck 2D+t example dataset for instance contains - if you use the shipped labels for each class - one (false) merger from the first frame on, which divides in frame 40 or so, and a proper merger in frame 23. Compare running tracking with and without `Merger Resolving` enabled, the latter results in what ilastik did before.

A bit offtopic: I tried to save a small test-project and ilastik complained as seen below. It turns out that the labels and random forests of object classification (here in the Division and Object Count classifiers) were not saved. Does the following traceback tell you anything about why it refused to save the annotations and RFs?

```
ERROR 2015-06-23 08:52:13,827 log_exception 30014 5289996288 Traceback (most recent call last):
  File "/Users/chaubold/hci/code/ilastik/src/ilastik/ilastik/ilastik/shell/projectManager.py", line 234, in saveProject
    item.serializeToHdf5(self.currentProjectFile, self.currentProjectPath)
  File "/Users/chaubold/hci/code/ilastik/src/ilastik/ilastik/ilastik/applets/base/appletSerializer.py", line 935, in serializeToHdf5
    ss.serialize(topGroup)
  File "/Users/chaubold/hci/code/ilastik/src/ilastik/ilastik/ilastik/applets/base/appletSerializer.py", line 195, in serialize
    self._serialize(group, self.name, self.slot)
  File "/Users/chaubold/hci/code/ilastik/src/ilastik/ilastik/ilastik/applets/base/appletSerializer.py", line 568, in _serialize
    classifier.serialize_hdf5( classifier_group )
  File "/Users/chaubold/hci/code/ilastik/src/ilastik/lazyflow/lazyflow/classifiers/parallelVigraRfLazyflowClassifier.py", line 170, in serialize_hdf5
    h5py_group['feature_names'] = self._feature_names
  File "/Users/chaubold/hci/code/ilastik/src/h5py-2.3.1/build/lib.macosx-10.9-x86_64-2.7/h5py/_hl/group.py", line 270, in __setitem__
    ds = self.create_dataset(None, data=obj, dtype=base.guess_dtype(obj))
  File "/Users/chaubold/hci/code/ilastik/src/h5py-2.3.1/build/lib.macosx-10.9-x86_64-2.7/h5py/_hl/group.py", line 94, in create_dataset
    dsid = dataset.make_new_dset(self, shape, dtype, data, **kwds)
  File "/Users/chaubold/hci/code/ilastik/src/h5py-2.3.1/build/lib.macosx-10.9-x86_64-2.7/h5py/_hl/dataset.py", line 53, in make_new_dset
    raise TypeError("Either data or shape must be specified")
TypeError: Either data or shape must be specified
```

stuarteberg:
>Does the following traceback tell you anything about why it refused to save the annotations and RFs?

Looks like the `feature_names` variable was `None`.  I just pushed a fix to avoid the error.  That explains why the RF wasn't saved correctly, but I'm not sure about the annotations.

stuarteberg:
>Requires newest commit of https://github.com/martinsch/pgmlink.git

OK, I'm having trouble building the latest version of `pgmlink` on my machine.  Maybe you can give me a hint.

Everything compiles, but it can't link.  I get duplicate symbol errors like this:

```
[ 48%] Linking CXX shared library libpgmlink.dylib
duplicate symbol opengm::LPDef::default_gomoryCutLevel_      in:
    CMakeFiles/pgmlink.dir/src/merger_resolving.cpp.o
    CMakeFiles/pgmlink.dir/src/pgm_chaingraph.cpp.o
duplicate symbol opengm::LPDef::default_disjunctCutLevel_      in:
    CMakeFiles/pgmlink.dir/src/merger_resolving.cpp.o
    CMakeFiles/pgmlink.dir/src/pgm_chaingraph.cpp.o
duplicate symbol opengm::LPDef::default_flowpathCutLevel_      in:
    CMakeFiles/pgmlink.dir/src/merger_resolving.cpp.o

[...]

ld: 108 duplicate symbols for architecture x86_64
collect2: error: ld returned 1 exit status
make[2]: *** [libpgmlink.dylib] Error 1
make[1]: *** [CMakeFiles/pgmlink.dir/all] Error 2
make: *** [all] Error 2
Command failed: /bin/bash -x -e /Users/bergs/Documents/workspace/ilastik-build-conda/pgmlink/build.sh
```

The linker seems to be confused by the fact the symbols in this opengm header are defined more than once:
[opengm/inference/auxiliary/lp_solver/lp_solver_cplex.hxx](https://github.com/opengm/opengm/blob/master/include/opengm/inference/auxiliary/lp_solver/lp_solver_cplex.hxx)

But I don't see anything wrong -- all of the functions in that file are correctly declared as `inline`.  Have you seen this error before?

PS -- I verified that I'm using the latest commit of opengm...

chaubold:
Thanks for pointing out that I didn't update OpenGM in a while :/ I get those duplicate symbols already when building the OpenGM python wrapper using the latest opengm commit... I'll investigate and get back to you. Do you build opengm with the python wrapper?

chaubold:
Okay, the duplicate symbols error is now fixed by https://github.com/opengm/opengm/commit/a30ffb461726f7de1ebf4066a53ebbb498af350f and https://github.com/martinsch/pgmlink/commit/e46aaf4526ec23bde79d699252245eff3599aab3, please update both repos and let me know whether that works better. (I'm compiling with GCC-4.8 btw)

stuarteberg:
After updating pgmlink and opengm, I can use this branch without errors.

>TODO: @nielsbuwen's HDF5/CSV export functionality still works on the old label image, because after updating the segmentation, we need to compute the region features for all new objects. What would be the best way to achieve this?

If the object features need to be recomputed anyway (not pulled from a cache), then the easiest thing to do might be to just instantiate a temporary `OpRegionFeatures` operator (see `opObjectExtraction.py`) and hook up the `RawVolume`, `LabelVolume`, and `Features` slots on-the-spot.

My first attempt would probably be something like this:

- Rename `OpTrackingBase.do_export()` to `_do_export_impl` or something like that.
- Modify `_do_export_impl()` so that it doesn't rely on `self.LabelImage`, `self.ObjectFeatures`, etc (instead, they will need to be passed in as parameters).
- Implement `OpTrackingBase.do_export()` as a simple pass-through to `_do_export_impl()` that just passes in `LabelImage`, etc.  Now everything just works as it did previously.
- In `OpConservationTracking`, overload `do_export()`, which will set up a little operator pipeline with an instance of `OpRegionFeatures` (which is fed with the correct label image), and then call `_do_export_impl()`, passing in the output slots from the little pipeline.
- Don't forget to `cleanUp()` any temporary operators you create.

I'm not sure if I'm missing something important here.  If you want to chat about this tomorrow, we can set up a Skype call or just chat on IRC.

chaubold:
I think I have fixed all remaining problems regarding result export, just recompuing features for the CSV/HDF5 tables is still missing. Perhaps @Steffen-Wolf, @nielsbuwen and @ilastikdev could double check whether this works fine and try to break it?

pgmlink needs to be updated to latest master again!

chaubold:
Recomputing new features after merger resolving for export was added in 9b6cd62.

Steffen just found the following issues:
* if the children of a division merge (same lineage), they have the same color, and can not be distinguished after merger resolving. It is correct that they are shown in the same color, but perhaps add a boundary in between? Also right clicking does not work in that case (meaning it does not know which of the objects was clicked)
* one debug output message of pgmlink's resolved to events was not removed yet
* without merger resolving -> right click on a merger does not show list of contained tracks but just one of them

chaubold:
> * OpRelabeledMergerFeatureExtraction::merge_features does not work yet if no new features were computed (i.e. no mergers are present in frame)

fixed in 0bfecf7ff6ab26ccd78b5db5f769734dfd9af010

> * one debug output message of pgmlink's resolved to events was not removed yet

fixed in https://github.com/martinsch/pgmlink/commit/ce86cf509b60aaceb4e4bcf392698c2fc6e7e1fd

chaubold:
> * if the children of a division merge (same lineage), they have the same color, and can not be distinguished after merger resolving. It is correct that they are shown in the same color, but perhaps add a boundary in between? Also right clicking does not work in that case (meaning it does not know which of the objects was clicked)

Right clicking is fixed (dc389ff) in ilastik's release mode now, as mergers are resolved by default, and right click information shows the same lineage, but different track and object IDs. BUT a visual hint would still be nice.

chaubold:
@stuarteberg : I find the new feature table export behaviour highly unintuitive:

* The export is only issued once the user has clicked the "configure feature table export" button, which is different to the behaviour of the default lane that is to be exported (because it uses some defaults).
* The name of the button is misleading in the tracking sense, because it conveys much more information than just the features. The tables also contain all linking information, as well as divisions.
* The progress bar at the bottom of ilastik only reflects the status of the image lane that is being exported, but the feature table export can take quite some time in large datasets. On the command line the progress is shown, though.
* It is not possible to export the feature table without exporting another export source. Is that necessary?

Issue #1061, opened 2014-12-08
Graph cut segmentation disabled


akreshuk:
in the current binary (1.1.3), linux.

burgerdev:
I tried this on a clean Ubuntu 14.10 virtual machine, seems like cplex is not optional any more...

```bash
burger@vbox:~/testing $ ilastik-1.1.3-Linux/bin/python -c 'import opengm'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/burger/testing/ilastik-1.1.3-Linux/lib/python2.7/site-packages/opengm/__init__.py", line 12, in <module>
    from _inference_interface_generator import _inject_interface , InferenceBase
  File "/home/burger/testing/ilastik-1.1.3-Linux/lib/python2.7/site-packages/opengm/_inference_interface_generator.py", line 8, in <module>
    from _inference_injector import _injectGenericInferenceInterface
  File "/home/burger/testing/ilastik-1.1.3-Linux/lib/python2.7/site-packages/opengm/_inference_injector.py", line 3, in <module>
    from inference  import InferenceTermination
  File "/home/burger/testing/ilastik-1.1.3-Linux/lib/python2.7/site-packages/opengm/inference/__init__.py", line 1, in <module>
    from _inference import *
ImportError: libilocplex.so: cannot open shared object file: No such file or directory
```

burgerdev:
@stuartberg: Is it possible that we need a binary compiled for use with cplex and another one compiled without it?

akreshuk:
You mean, not optional for opengm? That shouldn't really be so,  but maybe we accidentally built the opengm binary this way.

stuarteberg:
Fixed by @chaubold via #1126

PR #1126, opened 2015-06-23
Include opengm before everything else to enable graph cut thresholding

chaubold:
See #1061. (Including vigra before opengm yields ImportError: cannot import name IndexVector)

This only has an affect if OpenGM is built with the python wrapper, but that should be the case in our binaries.

### DIFF
--- a/ilastik/shell/gui/startShellGui.py
+++ b/ilastik/shell/gui/startShellGui.py
@@ -33,7 +33,7 @@ import splashScreen
 import ilastik.config
 shell = None
 
-def startShellGui(workflow_cmdline_args, eventcapture_mode, playback_args, *testFuncs):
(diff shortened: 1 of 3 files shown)


stuarteberg:
In general, we should try to minimize the number of `import` statements that execute before the splash screen is visible.  Otherwise, it looks like the app isn't even starting.

In `ilastik_main.main()`, there already existed a list of initialization functions that are executed (in order) during the splash screen, but (unfortunately) they were running as a post-initialization step (most of ilastik was already imported).

Now I just pushed a commit to this branch that also runs "pre-init" functions during the splash screen.  I moved the opengm import into this pre-init list.  Does it still work as you intended?

chaubold:
Yes it does GraphCut stays enabled, thanks a lot! Fixes #1061

screen shot 2015-06-24 at 16 15 23

PR #1128, opened 2015-07-10
Label Import: Rebased from #1119

stuarteberg:
I took the commits from #1119 and rebased them against the latest master.  Then I made some changes on top of it.  See the commit log for details.  BTW, to use this branch you will need the latest versions of lazyflow and volumina (master branches).

Summary of the changes I added on top of #1119:
- The new GUI for label import options is always shown to the user -- no defaults are assumed.
- The user may specify the axis-order of the file she selected (in case the `OpInputDataReader` guessed wrong)
- Don't assume 3D (or 5D)
- Show a busy indicator while the data is loading

### DIFF
--- a/ilastik/applets/labeling/labelingImport.py
+++ b/ilastik/applets/labeling/labelingImport.py
@@ -1,0 +1,436 @@
+###############################################################################
+#   ilastik: interactive learning and segmentation toolkit
+#
+#       Copyright (C) 2011-2015, the ilastik developers
+#                                <team@ilastik.org>
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# In addition, as a special exception, the copyright holders of
+# ilastik give you permission to combine ilastik with applets,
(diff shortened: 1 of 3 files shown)


ilastik/applets/labeling/labelingImport.py
```diff
@@ -0,0 +1,534 @@
...
+            return
+
+        # Expect x, y shape to match original shape
+        labels_tagged_shape = labelingSlots.labelInput.meta.getTaggedShape()
+        file_tagged_shape = opReorderAxes.Output.meta.getTaggedShape()
+        if (  labels_tagged_shape['x'] != file_tagged_shape['x']
+           or labels_tagged_shape['y'] != file_tagged_shape['y'] ):
+            QMessageBox.critical(parent_widget, "Shape does not match",
+                                 "X,Y shape must match original input stack.")
+            return
```
stuarteberg: @holstgr-kaust, Is there a particular reason that the XY shape of the imported label data must match the raw data?

stuarteberg:
In a separate email, @holstgr-kaust approved my changes above, so I'll go ahead and merge this PR.  Thanks, Glendon, this feature will be very useful for lots of Carving and Pixel Classification users!

PR #1141, opened 2015-08-28
Use the new pgmlink API in opConservationTracking

chaubold:
After the recent confusion about which version of pgmlink is needed for ilastik master (again sorry about that), here is the change to use the current pgmlink master (tested with https://github.com/martinsch/pgmlink/commit/9c4dadb14dd97a4dd1b6d8aca408809978616c45) from ilastik.

@stuarteberg could you please verify that tracking also works for you with this combination? If it does, I'll also create a PR for ilastik-build-conda to use the most up to date versions of OpenGM and pgmlink and update the build numbers.

### DIFF
--- a/ilastik/applets/tracking/conservation/opConservationTracking.py
+++ b/ilastik/applets/tracking/conservation/opConservationTracking.py
@@ -170,36 +170,6 @@ class OpConservationTracking(OpTrackingBase):
         if ndim == 2:
             assert z_range[0] * z_scale == 0 and (z_range[1]-1) * z_scale == 0, "fov of z must be (0,0) if ndim==2"
 
-        # tracker = pgmlink.ConsTracking(maxObj,
-        #                                  float(maxDist),
-        #                                  float(divThreshold),
-        #                                  "none",  # detection_rf_filename
-        #                                  sizeDependent,   # size_dependent_detection_prob
-        #                                  0,       # forbidden_cost
-        #                                  float(ep_gap), # ep_gap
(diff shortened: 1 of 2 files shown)


ilastik/applets/tracking/conservation/opConservationTracking.py
```diff
@@ -174,43 +170,130 @@ def track(self,
...
+        #                                  cplex_timeout,
+        #                                  "none" # dump traxelstore
+        #                                  )
+        #
+        #
+        # try:
+        #     eventsVector = tracker(ts, coordinate_map.get())
+        # except Exception as e:
+        #     raise Exception, 'Tracking terminated unsuccessfully: ' + str(e)
+
```
stuarteberg: Is it necessary to leave these old source lines in the file?

chaubold: No that's not needed any more. Removed.

chaubold:
Update on build scripts:
* https://github.com/ilastik/ilastik-build-win64 is already building the most up to date version of ```pgmlink```as well as ```opengm```. Which also means that **tracking will fail in the current Windows builds**! We should build new binaries *soon*.
* ilastik-build-conda has a new PR (https://github.com/ilastik/ilastik-build-conda/pull/9) to update the opengm and pgmlink versions to the leatest head.

By merging both PRs, all repositories master-branches should work together smoothly (fingers crossed...)

chaubold:
Fixed by merging #1124

Issue #1147, opened 2015-09-22
Tracking: Unintuitive Table Export UI


stuarteberg:
(Copying @chaubold's [comments](https://github.com/ilastik/ilastik/pull/1124#issuecomment-142231969) PR #1124 into a new issue here.)

>I find the new feature table export behaviour highly unintuitive:
>
>   - The export is only issued once the user has clicked the "configure feature table export" button, which is different to the behaviour of the default lane that is to be exported (because it uses some defaults).
>   - The name of the button is misleading in the tracking scenario, because the table conveys much more information than just the features. The tables also contain all linking information, as well as divisions.
>   - The progress bar at the bottom of ilastik only reflects the status of the image lane that is being exported, but the feature table export can take quite some time in large datasets. On the command line the progress is shown, though.
>   - It is not possible to export the feature table without exporting another export source. Is that necessary?

stuarteberg:
>- The export is only issued once the user has clicked the "configure feature table export" button, which is different to the behaviour of the default lane that is to be exported (because it uses some defaults).

I suppose it would be nice to choose some defaults for the user.  We need to choose a default file name, and also the default set of features to export, and the file format (csv vs hdf5).  The default settings can be configured by calling [`configure_table_export_settings()`](https://github.com/ilastik/ilastik/blob/1b540bd1694ba591ef029351361e8fc0d3eee104/ilastik/applets/tracking/base/opTrackingBase.py#L71).  I guess the right time to call that function might be in either `OpConservationTracking.__init__()`, or maybe `ConservationTrackingWorkflowBase.onProjectLoaded()`.

>- The name of the button is misleading in the tracking scenario, because the table conveys much more information than just the features. The tables also contain all linking information, as well as divisions.

Fair enough.  It can be fixed [here](https://github.com/ilastik/ilastik/blob/4662daee721f30025715ab2e9e5df2122dbb8f93/ilastik/applets/tracking/base/trackingBaseDataExportGui.py#L68).

>- The progress bar at the bottom of ilastik only reflects the status of the image lane that is being exported, but the feature table export can take quite some time in large datasets. On the command line the progress is shown, though.

[...]

PR #1151, opened 2015-09-24
Fix tracking export behaviour

chaubold:
See #1147


stuarteberg:
This all looks good to me.  (Thanks for keeping the commits cleanly separated so it was easy to follow.)  Unless you have any more changes to make, merge away!

PR #1152, opened 2015-09-29
Added color codes to the feature selection dialog based on individual computation times

JaimeIvanCervantes:
Added color codes to feature selection dialog based on individual computation times.

This is a screenshot with the color codes deactivated:
noclors

And this is a screenshot with the color codes activated:
colors


k-dominik:
I do like the idea, but with multiple supported architectures and 2Dvs3D computation we'd have to assess a lot of data. For now the intuition that larger sigmas take more time seems to be good enough. General recommendation still is to use all features.

PR #1180, opened 2016-01-25
Fixed problem with axis ranges stored in parameters

JaimeIvanCervantes:
The axis ranges (stored in the `Parameters` slot) of the first lane were being used incorrectly for all the lanes during batch-processing. In order to fix the problem, I removed the `withBatchProcessing` flag from `_generate_traxelstore`, and restore the original axis ranges in `post_process_lane_export`.

There is a fundamental problem with the `Parameters` slot since it is a broadcasted slot, and therefore the same for all the lanes, but it also contains information on axis ranges that is different for every lane. The ideal fix would be to split `Parameters` into two slots, `Parameters`  and `AxisRanges`. This change however, requires a lot of modifications to the code which will make the merging of structured learning more difficult. For now I fixed the problem in a more simple way, by doing the least changes, but after structured learning is changed, I will split `Parameters` as described.

### DIFF
--- a/ilastik/applets/tracking/base/opTrackingBase.py
+++ b/ilastik/applets/tracking/base/opTrackingBase.py
@@ -397,20 +397,18 @@ class OpTrackingBase(Operator, ExportingOperator):
                               max_traxel_id_at=None,
                               with_opt_correction=False,
                               with_coordinate_list=False,
-                              with_classifier_prior=False,
-                              with_batch_processing = False):
+                              with_classifier_prior=False):
 
         if not self.Parameters.ready():
             raise Exception("Parameter slot is not ready")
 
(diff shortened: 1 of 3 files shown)


chaubold:
IIRC we discussed splitting the parameter object while you were here, but you are making a good point regarding the merge difficulties! Looks good to me.

stuarteberg:
>For now I fixed the problem in a more simple way, by doing the least changes, but after structured learning is changed, I will split Parameters as described.

Would it be worth opening a new issue, just as a reminder to follow-up on this after the merge is completed?

JaimeIvanCervantes:
@stuarteberg I just submitted the issue.

PR #1186, opened 2016-02-15
Structured Learning Tracking

ilastikdev:
Structured learning tracking code from learning branch is based on:

https://github.com/ilastik/ilastik
(tracking M.Schieg and C. Haubold; structured learning workflow, training applet, and GUI J. Ales)

https://github.com/ilastik/pgmlink
(tracking M.Schieg; merger resolving C.Haubold; structured learning J.Ales)

https://github.com/ilastik/volumina
(cropping markers S. Berg; structured learning tracking workflow GUI J. Ales)

and learning-experimental branch of

https://github.com/opengm/opengm
(J. Kappes, T.Bier, J.Funke, C.Haubold, J.Ales, plus Dec. 2014 hackathon crew)

User trains division and object count classifier and
inputs tracking training.

Best structured learning weights for the given data and training are calculated and
used to obtain a tracking solution.

Object count and Division classifiers demand the same amout of training as in the ilastik binary.
Small tracking training sets have lead to good results in tracking (see bellow).

Future work:
- transition classifier incorporating object features
- Convex Hull and Skeleton features (2D) in tracking (also for conservation tracking) (DONE)
- Tests for determining the number of sufficient training points for weights that produce good tracking solutions.


ilastik/applets/tracking/structured/opStructuredTracking.py
```diff
@@ -0,0 +1,723 @@
...
+            sigmas.append(0.0)
+        uncertaintyParams = pgmlink.UncertaintyParameter(1, pgmlink.DistrId.PerturbAndMAP, sigmas)
+
+        self.detectionWeight = self.parent.parent.trackingApplet._gui.currentGui()._drawer.detWeightBox.value()
+        self.divisionWeight = self.parent.parent.trackingApplet._gui.currentGui()._drawer.divWeightBox.value()
+        self.transitionWeight = self.parent.parent.trackingApplet._gui.currentGui()._drawer.transWeightBox.value()
+        self.appearanceWeight = self.parent.parent.trackingApplet._gui.currentGui()._drawer.appearanceBox.value()
+        self.disappearanceWeight = self.parent.parent.trackingApplet._gui.currentGui()._drawer.disappearanceBox.value()
+
+        print "detectionWeight=",self.detectionWeight
```
akreshuk: please replace prints with logging messages

akreshuk:
Pixel classification from this branch fails when loading data:
  File "/home/akreshuk/miniconda2/envs/ilastik-tracking-test/ilastik-meta/ilastik/ilastik/applets/dataSelection/dataSelectionGui.py", line 623, in _configureOpWithInfos
    self.topLevelOperator.DatasetGroup[laneIndex][roleIndex].setValue( info )
  File "/home/akreshuk/miniconda2/envs/ilastik-tracking-test/ilastik-meta/lazyflow/lazyflow/slot.py", line 110, in call_in_setup_context
    return func(self, *args, **kwargs)
  File "/home/akreshuk/miniconda2/envs/ilastik-tracking-test/ilastik-meta/lazyflow/lazyflow/graph.py", line 136, in __exit__
    sig_setup_complete()
  File "/home/akreshuk/miniconda2/envs/ilastik-tracking-test/ilastik-meta/lazyflow/lazyflow/utility/orderedSignal.py", line 76, in __call__
    f(*args, **kw)
  File "/home/akreshuk/miniconda2/envs/ilastik-tracking-test/ilastik-meta/ilastik/ilastik/utility/gui/threadRouter.py", line 75, in routed
    val = func(*args, **kwargs)
  File "/home/akreshuk/miniconda2/envs/ilastik-tracking-test/ilastik-meta/ilastik/ilastik/applets/layerViewer/layerViewerGui.py", line 438, in updateAllLayers
    if not (self.editor.cropModel._crop_extents[0][0]  == None or self.editor.cropModel.cropZero()):
AttributeError: 'CropExtentsModel' object has no attribute 'cropZero'

stuarteberg:
>Pixel classification from this branch fails when loading data:

That's my bad.  The needed volumina changes hadn't been merged yet.  But they're merged now.  Update your volumina repo and try again.

ilastikdev:
cropZero was not in the master branch, but it is in learning branch...

akreshuk:
After volumina changes all starts up very nicely. But when changing to the Crop Selection applet, I get "No module named logger" error.
  File "/home/akreshuk/miniconda2/envs/ilastik-tracking-test/ilastik-meta/ilastik/ilastik/applets/cropping/cropSelectionGui.py", line 37, in <module>
    import logger
ImportError: No module named logger

ilastikdev:
forgotten to push the latest change,

pull again at it will work

akreshuk:
And now I can track! I ran through the workflow without problems, trained two crops and it even produced meaningful tracks in the end, even though I haven't trained the classifiers much! I have a lot of comments on the UI, but haven't looked in the code yet. Keep in mind that i) I did navigate through it successfully, so my comments are just to make it more user friendly, ii) I'm a good approximation of a first time tracking user.

1. Crop selection:
1.1 What is the start->end slider at the bottom right doing? Just goes through the time slices? Sets the time interval for the crop? 
1.2 What does Set Crop do? Save a crop? A tooltip would be helpful here
2. Training:
2.1 What does Initialize Training do? Why is it necessary? Why is it not initialized with the first track?
2.2 Where is the "Go to the next unlabeled object" supposed to take me? It threw me to some location in the block, I have no idea why I got there and what I was supposed to do
2.3 Why does "Mark as false detection" freeze everything else?
3. Training Export
3.1 What are the two sources? What should I choose to save my crop labels so that I can load them afterwards?
4. Structured learning
4.1 What are weights if there is a "Calculate Weights" right above? Initial weights? A tooltip would be helpful, also to explain which one to choose.
4.2 Shouldn't "Tracking" be "Track!"?
4.3 Why is using training annotations optional?
4.4 What does "Max. Nearest Neighbor" set?
4.5 Why is there an "export results as h5 sequence" at the very bottom, if there is an export applet right after? 
4.6 What is in the Filter section - is it a crop again or something else? If it's a crop, call it a crop.
4.7 If the big classifiers are optional, why should they be trained by default? Better yet, add to the error message what the user should do if they don't want to train this classifier. E.g. "Division classifier has not been trained! Uncheck divisible objects if your objects don't divide or go back to the division training applet and train it". 
4.8 For this error message, why are there words like "exception"? That's for the log, not for the friendly dialogues.  Also, it's printed into my terminal and scares me. Can't you check the consistency of the chosen option with the trained classifiers in a way that wouldn't throw?
4.9 There is non-logger output produced by the optimization routine and printed into the terminal too. Does it come from OpenGM? Is there a way we could redirect most of it to the log file? 
5. Object Count classification
5.1 When you add labels, it's "1 Objects". Please make it "1 Object"
6. Serialization
6.1 For the input raw data of 75 MBytes (mitocheck) and prob maps of 65 MBytes I get a tracking project of 1.2 GBytes. What takes so much? Could a Compressed Cache help?

To summarize: great work on bringing it all together! After the inevitable polishing, smoothing of the rough edges and molding it into ilastik, I'm sure the users will appreciate simply training in the little crops and having it do the rest of the work.

ilastikdev:
I am glad you liked the number crunching part - the whole point is in this:
"... it even produced meaningful tracks in the end, even though I haven't trained the classifiers much!"

Most of your comments are on my TODO list and will be done in due time with foreseeable not too much work involved. 6) Memory issue may take more time.

 4.iii) Using training for learning in a tracking solution is a great idea that can help the algorithm find a correct tracking result in difficult cases that are problematic for the current model - that said, it should definitely stay optional. 

First because tracking solutions can not be evaluated fairly (important for research purposes) if we fix a part of the solution to training points.

Second, even if we use it only as an option we are facing problems that will address the limitations of the conservation tracking model (this has nothing to do with learning, just the conservation tracking that runs in our released binaries under Automatic Tracking) as it is posed and implemented now.
This is a topic we should move off-line and discuss it at a team meeting.

chaubold:
2.iii) This behaves as in manual tracking: clicking it once goes into "I only label false detections now"-mode, clicking it again goes back to normal mode. But I agree, this is not intuitive.

To clarify on 4.iii): if you turn this option on, then we do not only use the supplied labels to train for weights that give us the tracking solution closest to these labels, but we also introduce hard constraints in the ILP that fix the respective variables to the values supplied by the user. That being said, I agree with Janez that we should have an option to turn this off, but I also see Anna's point that for an end user it might be more convenient if this option is enabled by default.

4.ix) This has always been the case for tracking, and now there is some more output coming from the structured learning side as well (even though Janez already removed most of it). The logs are coming from OpenGM as well as from CPLEX / Gurobi directly. We switch off the ILP solver's verbosity, but apart from that I don't know whether we're able to capture logs to STDOUT printed by shared libraries/modules on the python side?

akreshuk:
I get point 4.iii) now, before it was like "why did I spend all this time labelling if you are not going to use it"? How about changing the wording to "Turn your labels into hard constraints". Many users won't get it, but at least they won't get it wrong, like I did. If this option is on, but the optimization doesn't converge, we can show an error message saying "not all your tracks could be reproduced in our best solution" and return the solution with the option off.

ilastikdev:
3.i) We can export all labels and divisions (so all of your training) if you choose "Training" as a source.

Crop specific training export is not supported (yet). 
It would only be useful to merge different training sets from various projects on the same data and/or
to do statistics on training per crop.
(Note that tracking training on different data sets is not compatible, even just re-running thresholding
one needs to re-train for tracking.
 
It is a part of a ToDo task:
a) import training (possibly per crop) needed for testing on data with ground truth;
b) export training per crop.

akreshuk:
Can the user re-import the exported labels and divisions? I mean, from the GUI, not hacking them manually into the project file. Can you rename this source to something more obvious, like "User labels"?

ilastikdev:
Import is not implemented yet in any of the three tracking workflows.

ilastikdev:
Max Nearest Neighbour sets the number of neighbours considered in forward direction as candidates for a tracking transition. Also in backward direction (same as in automatic tracking).

When using "Training to Hard constraints" this may have to be increased (gets done automatically) in order to have ALL training transitions in the hypothesis graph.

ilastikdev:
4.v) is the same as in Automatic tracking - it exports a sequence of tracking labels IMAGES as hdf5 files.
One can do this also in the Tracking Result Export, so this button can be removed in the structured learning tracking applet.

akreshuk:
We should try to find a better name or write a really clear tooltip for the Max Nearest Neighbor. You see, conservation tracking had a lot of mysterious parameters, but users were always told not to mess with them, so I suppose most people didn't even try to understand. Now that most of the interface is actually understandable by an application domain expert, the remaining parameters are quite puzzling.

ilastikdev:
Further to our discussion, structured learning workflow is now ready to be merged.
pgmlink/learning will get merged on the coming Tuesday, so all dependencies will have learning in master (opengm, lazyflow, volumina, and pgmlink).

Code base additions (import training. tests for structured learning) and memory/speed optimization are on todo list with highest priority.

akreshuk:
It seems like the only thing standing in the way is now building pgmlink on windows.

Steffen-Wolf:
Currently conservation tracking does not produce the same results as it did in 1.1.9! See #1232

ilastikdev:
#1232 has been fixed

chaubold:
I've tried to use the Structured Learning workflow on my own for the first time today. I was not as lucky as @akreshuk and did not manage to get learned weights out yet (see below). Here's the dump of things that I noticed. More to come in the next days.

### Serious:

* In "Training": Right click + "Run automatic tracking for object XX" does not care about crop boundaries.
* I have annotated a single crop, no divisions, no mergers (even though the Mitocheck dataset does contain both and settings allow it), should be straightforward, but: `DatasetConstraintError: Constraint of 'Structured Learning' applet was violated: Your tracking annotations contradict this model assumptions! All tracks must be continuous, tracks of length one are not allowed, and mergers may merge or split but all tracks in a merger appear/disappear together.`?

### Usability improvements

* Why do you need to "Save Crop" in "Crop Selection" before it is available in "Training"? Very unintuitive.
* I like that in "Crop Selection" there is a gray highlight of the selected area. 
 But how about graying out everything BUT the selected crop? Then we could use this grayed out background also in Training.
* "Training Export"??? What should "object identities" do that the final "Tracking Export" cannot do? I think this can be removed.
* Jumping with tab between the weight-fields is in random order (low priority, probably doesn't work anywhere else either...)

chaubold:
I figured out what the problem was why I could not learn weights with my annotations: 
I had annotations that were purely outside of any crop, but still within the time range of the crop.
Also I noticed that I was fooled by my screen about the crop "highlight".

See this updated list of issues, which superseeds the one from yesterday.

### Serious:

1. In "Training": Right click + "Run automatic tracking for object XX" does not care about crop boundaries.
1. Manual annotations of objects that lie completely outside of the boundaries of crops (but still within the time range of a crop!) lead to a
  ```
  DatasetConstraintError: Constraint of 'Structured Learning' applet was violated: Your tracking annotations contradict this model assumptions! All tracks must be continuous, tracks of length one are not allowed, and mergers may merge or split but all tracks in a merger appear/disappear together.
  ``` 
  This should not be the case, so I see two possible options:
    * do not allow any annotations outside of crops (would also fix 1. )
    * filter annotations properly when writing them to the hypotheses graph for learning the weights
1. There are no warnings/errors if I learn weights with divisions enabled but have not annotated any divisions in "Training". 
 This would mean the division weight can be arbitrarily high, but is only reduced by the regularizer. 
 If there are still divisions present in the dataset, they will not necessarily be recognized properly. 
1. There are no warnings/errors if the selected crops do not contain any appearance/disappearance. 
 I guess appearing in the first frame (here of a crop) and disappearing in the last does not count here, same as in "standard" conservation tracking?
 Then the same argument as above holds: This would mean the appearance/disappearance weights can be arbitrarily high, but are only reduced by the regularizer.  

### Usability improvements (ordered by my personal prioritization):

3. Why is there a "Training Export" applet? What should the "object identities" export-slot do that the final "Tracking Export" cannot do? 
 If this was meant to allow import/export of annotations, this would be a nice feature. But as long as that does not work, this applet should be removed/hidden.
1. Why do you need to "Save Crop" in "Crop Selection" before it is available in "Training"? Seems unintuitive.
2. I like that in "Crop Selection" and "Training" the background gets darker (reduced opacity). 
 But why is the opacity of the cropped out region itself also reduced? 
 (Not as much, but still noticeable when one toggles the eye of the "Crop" layer)
2. Why is there no "Crop" layer in training that one could enable/disable?
2. When one zooms out, the crop boundaries are also becoming thinner, until they are no longer visible at all. 
 Wouldn't it be nicer to always show the boundaries at least as 1px lines? (I know this would probably be a pain to implement, so low priority).   
4. Jumping with tab between the weight-fields in "Tracking - Structured Learning" is in random order (low priority, probably doesn't work anywhere else either...)

chaubold:
I just added a new crop, but did not click "Save Crop", and saved the project. I pulled the las commit and reopened ilastik. When I try to load the project again, I get this error:
screen shot 2016-06-29 at 14 36 45

(The log does not say anything else than this.)
@ilastikdev can you reproduce this error? Do you have an idea what could go wrong there?

chaubold:
Regarding da30da8 : The info boxes for annotating outside the crop boundaries are good, but when one of these annotations is added by the semi-automatic tracking it is rather confusing. 
Can we use the "object left crop" criterion to simply stop the semi-automatic track extension?

chaubold:
Adding on to the list:

### Serious:

* [ ] I'd have expected "Delete Training" to only delete training of the selected crop, but it deletes everything?! Warn about that!
* [ ] Interesting, I can run "Calculate Weights" even after I've removed all annotations and it runs structured learning. Is it trying to predict all zeros then? Shouldn't we warn if one learns without any annotations?
* [ ] If one deletes the training annotations but the crops are still there, and presses "Calculate Weights", please give a more useful error message than
 ```
  ERROR 2016-06-29 15:05:04,569 excepthooks 48096 140735309665024 Traceback (most recent call last):
    File "/Users/chaubold/opt/miniconda/envs/ilastikdev/ilastik-meta/ilastik/ilastik/applets/tracking/structured/structuredTrackingGui.py", line 444, in _onRunStructuredLearningButtonPressed
      crop = self.mainOperator.Annotations.value[cropKey]
  KeyError: u'Crop 1'
 ``` 
* [ ] After running "Calculate Weights", **all training annotations are gone!** I did not press "Delete training" or "Delete crop". Why is that? 
 (No annotations are shown any more, and right click shows "object XX contains no track ids"). 
  Note: I'm still working on a project based on the one that complained about `QAbstractSlider.setValue` as mentioned above, 
  but now I do not get that error any more on save & load. 

### Usability improvements (ordered by my personal prioritization):

* [ ] It is nice that the min-slider is synchronized with the max-slider to prevent that `max < min`. 
 But when one enters the `max` value by hand and `min > 0`, and one first deletes everything to type a number afresh, 
 the `min` slider automatically gets set to zero (or another low number if one leaves one digit e.g. leaves the 1 when changing from 19 to 15).

ilastikdev:
Any annotated object does contain appearance/disappearance nodes, so these weights do participate in the decision making (they have to be sufficently big so that making a long track is cheaper than making many short tracks). I would omit a message for missing appearance/disappearance objects inside the time boundary.

chaubold:
> Any annotated object does contain appearance/disappearance nodes, so these weights do participate in the decision making (they have to be sufficently big so that making a long track is cheaper than making many short tracks). I would omit a message for missing appearance/disappearance objects inside the time boundary.

Okay!

ilastikdev:
I can not reproduce this error:

"I just added a new crop, but did not click "Save Crop", and saved the project..."

chaubold:
Could you please also change the info message for semi-automatic tracking when the end of the crop is reached? Thanks.

(located in https://github.com/ilastik/ilastik/blob/learning/ilastik/applets/tracking/annotations/annotationsGui.py#L898)

Steffen-Wolf:
## I have tried to use structured learning on **3d data**. ilastik crashes when I try to **save the project**.

Steps taken:
- open ilastik structured learning workflow
- load 3d drosophila dataset and prediction
- switch to division detection applet
- save project

The error message I get is:
VLEN strings do not support embedded NULLs

with a log:


Starting ilastik from "/storage/data/Machine_Learning/miniconda/envs/ilastik-learning-test".
ERROR 2016-07-04 14:30:18,085 log_exception 6812 140412430030592 Traceback (most recent call last):
  File "/storage/data/Machine_Learning/miniconda/envs/ilastik-learning-test/ilastik-meta/ilastik/ilastik/shell/projectManager.py", line 278, in saveProject
    serializer.serializeToHdf5(self.currentProjectFile, self.currentProjectPath)
  File "/storage/data/Machine_Learning/miniconda/envs/ilastik-learning-test/ilastik-meta/ilastik/ilastik/applets/base/appletSerializer.py", line 971, in serializeToHdf5
    ss.serialize(topGroup)
  File "/storage/data/Machine_Learning/miniconda/envs/ilastik-learning-test/ilastik-meta/ilastik/ilastik/applets/base/appletSerializer.py", line 195, in serialize
    self._serialize(group, self.name, self.slot)
  File "/storage/data/Machine_Learning/miniconda/envs/ilastik-learning-test/ilastik-meta/ilastik/ilastik/applets/base/appletSerializer.py", line 227, in _serialize
    self._serialize(subgroup, subname, slot[i])
  File "/storage/data/Machine_Learning/miniconda/envs/ilastik-learning-test/ilastik-meta/ilastik/ilastik/applets/base/appletSerializer.py", line 222, in _serialize
    self._saveValue(group, name, slot(()).wait())
  File "/storage/data/Machine_Learning/miniconda/envs/ilastik-learning-test/ilastik-meta/ilastik/ilastik/applets/base/appletSerializer.py", line 807, in _saveValue
    dset = group.create_dataset(name, data=pickled)
  File "/home/wolf/Machine_Learning/miniconda/envs/ilastik-learning-test/lib/python2.7/site-packages/h5py/_hl/group.py", line 103, in create_dataset
    dsid = dataset.make_new_dset(self, shape, dtype, data, **kwds)
  File "/home/wolf/Machine_Learning/miniconda/envs/ilastik-learning-test/lib/python2.7/site-packages/h5py/_hl/dataset.py", line 126, in make_new_dset
    dset_id.write(h5s.ALL, h5s.ALL, data)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper (--------src-dir--------/h5py-2.5.0/h5py/_objects.c:2458)
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper (--------src-dir--------/h5py-2.5.0/h5py/_objects.c:2415)
  File "h5py/h5d.pyx", line 217, in h5py.h5d.DatasetID.write (--------src-dir--------/h5py-2.5.0/h5py/h5d.c:3345)
  File "h5py/_proxy.pyx", line 163, in h5py._proxy.dset_rw (--------src-dir--------/h5py-2.5.0/h5py/_proxy.c:1937)
  File "h5py/defs.pyx", line 2800, in h5py.defs.H5Tconvert (--------src-dir--------/h5py-2.5.0/h5py/defs.c:25621)
  File "h5py/_conv.pyx", line 424, in h5py._conv.str2vlen (--------src-dir--------/h5py-2.5.0/h5py/_conv.c:3793)
  File "h5py/_conv.pyx", line 113, in h5py._conv.generic_converter (--------src-dir--------/h5py-2.5.0/h5py/_conv.c:1626)
  File "h5py/_conv.pyx", line 249, in h5py._conv.conv_str2vlen (--------src-dir--------/h5py-2.5.0/h5py/_conv.c:2470)
ValueError: VLEN strings do not support embedded NULLs

ERROR 2016-07-04 14:30:18,516 log_exception 6812 140412430030592 Project Save Action failed due to the exception shown above.

Steffen-Wolf:
Calling automatic tracking on objects produces the info:
"object xxx in timeframe yyy left the current crop time boundary."

chaubold:
> Calling automatic tracking on objects produces the info: "object xxx in timeframe yyy left the current crop time boundary."

@Steffen-Wolf: This is the behaviour we've now settled for, but we should probably add "stopping automatic tracking at crop boundary".

Steffen-Wolf:
## usability:
 * [ ] show object and binary layer in crop selection

chaubold:
* [ ] @ilastikdev you've changed the message for semi-automatic tracking stopping at the spatial crop borders, but not for the time-boundary.

chaubold:
I wanted to check whether all issues have been resolved, and started a new project. I can add and save crops just fine, started to do some training annotations, but at some point it stopped displaying them. Unfortunately I don't know what triggered it. I tried to save the project and reopen it, but was faced with the following useless setup:
screen shot 2016-07-05 at 19 49 30

Note that (see red arrows):
1. the "active track" is greyed out (I've never seen that before)
2. there are colored objects which -- on right click -- report that they do not contain any labels.

And since then I cannot manage to show any labels from training annotations at all. I've tried to create new crops and delete the old one in that project. I also created completely new projects (with and without restarting ilastik), and was always faced with missing labels as seen below.

screen shot 2016-07-05 at 20 02 42

I'll keep trying to get my label colors back again. @ilastikdev could it be that this weird behaviour is caused by the new way of propagating dirtyness of the labels?

Steffen-Wolf:
After training, when I want to calculate the weights I get the following error:

> You have not trained or saved your training for Crop 1. Go back to the Training applet and save all your training!

It obviously tells me what to do. For some reason I can not figure out which button to press to satisfy the unrelenting beast. This might be a severe case of PEBKAC but I find this saving highly unintuitive!

Question: Why do we require the user to save here. Can we not just simply save everything automatically when the "calculate weights" button is pressed.

Steffen-Wolf:
> Note that (see red arrows):
1. the "active track" is greyed out (I've never seen that before)
2. there are colored objects which -- on right click -- report that they do not contain any labels.

> And since then I cannot manage to show any labels from training annotations at all. I've tried to create new crops and delete the old one in that project. I also created completely new projects (with and without restarting ilastik), and was always faced with missing labels as seen below (always with 2 crops).

I get the same behaviour!

ilastikdev:
Errors have been fixed. Usability improvements will be worked on after the workshop.

stuarteberg:
So... is this ready, then?  Who wants to click the "merge" button?

akreshuk:
@chaubold , go ahead if you think it's time.

Issue #1188, opened 2016-02-18
Conservation tracking and Animal Tracking do not use 2D convex hull and skeleton features


ilastikdev:
These features were called via Select Features dialog but not via Subset Features used in these two workflows.

JaimeIvanCervantes:
I stumble across this issue when I was trying to load my ilastik project file from Hytra, and I discovered that the classifier was trained using only the `Standard Object Features`. 

One of the problems occurs in `objectClassificationGui.py` where the `SelectedFeatures` slot is being modified straight from the GUI code. I'm working on a fix for this issue.

stuarteberg:
To add some details: The correct features are only used iff you use the "subset features" window before training the classifier for the first time.  Once the classifier is trained, the features are cached and it is too late to change them.

The reason for this is that the GUI is (attempting to allow the user to provide "plugin" features in the "subset features" UI, instead of in the original "select features" UI in object extraction.  The only reason it accidentally works is that in [`handleSubsetFeaturesClicked()`](https://github.com/ilastik/ilastik/blob/master/ilastik/applets/objectClassification/objectClassificationGui.py#L282), the feature dictionary from the object extraction applet is [retrieved](https://github.com/ilastik/ilastik/blob/master/ilastik/applets/objectClassification/objectClassificationGui.py#L284), and [modified directly](https://github.com/ilastik/ilastik/blob/master/ilastik/applets/objectClassification/objectClassificationGui.py#L324), without going through the proper procedure of copying it and setting it on a slot, so that dirty state is propagated through object extraction.

Essentially `OpObjectExtraction` is not "aware" that it's `Features` slot contents have changed.

I think the proper solution should be:

1. By default, the ObjectExtraction `Features` slot should be populated with all available features (including plugins), even before the user has clicked the "Select Features" button.

2. In `ObjectClassificationGui`, don't modify the `computedFeatures` dictionary directly -- use a copy of it.

akreshuk:
From what I recall, opObjectExtraction shouldn't be aware of what's happening in the object classification applet. The original idea was that you compute all possible features and experiment with different subsets in the classification applet. It looks like your proposed solution will actually bring it to this stage.

PR #1189, opened 2016-02-18
add 2D convexhull and skeleton features to conservation tracking and …

ilastikdev:
…animal tracking

This has also been added in Structured Learning Workflow (branch: learning)

Issue: #1188

### DIFF
--- a/ilastik/workflows/tracking/conservation/animalConservationTrackingWorkflow.py
+++ b/ilastik/workflows/tracking/conservation/animalConservationTrackingWorkflow.py
@@ -32,7 +32,7 @@ class AnimalConservationTrackingWorkflowBase( ConservationTrackingWorkflowBase )
            
         super(AnimalConservationTrackingWorkflowBase, self).connectLane(laneIndex) 
         
-    def _createDivisionDetectionApplet(self,selectedFeatures=None):
+    def _createDivisionDetectionApplet(self,selectedFeatures=dict()):
         return None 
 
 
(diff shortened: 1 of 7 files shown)


ilastik/workflows/tracking/conservation/animalConservationTrackingWorkflow.py
```diff
@@ -32,7 +32,7 @@ def connectLane(self, laneIndex):
            
         super(AnimalConservationTrackingWorkflowBase, self).connectLane(laneIndex) 
         
-    def _createDivisionDetectionApplet(self):
+    def _createDivisionDetectionApplet(self,selectedFeatures=None):
```
akreshuk: shouldn't that be selectedFeatures=dict()?

ilastikdev: Code calls the proc in conservationTracking, where it is set to dict() so in the current code either works.

akreshuk:
I tried object classification with this branch. When I choose features from "Standard" plugin (all for a 2d image) in object extraction and then open the object classification applet and click "Subset features", other plugins (like convex hull features, etc) are still visible, although they don't have any features listed under them. Would it be possible to remove them?

ilastikdev:
This was already discussed  a year ago with Uli and it leads to users forgetting that there are other features that they could use once they limit only to one of the sets.

Also, when, for example, selecting a subset of Standard features we also display all of them with selected checked, so here too, we display all possible and mark the selected ones.

PR #1211, opened 2016-04-19
opObjectClassification: Object extraction and prediction can now be done in parallel

JaimeIvanCervantes:
On the `execute()` function of `OpObjectPredict`, object feature computation and object prediction was previously done one frame at a time, and now it can be done in parallel.

### DIFF
--- a/ilastik/applets/objectClassification/opObjectClassification.py
+++ b/ilastik/applets/objectClassification/opObjectClassification.py
@@ -1115,14 +1115,8 @@ class OpObjectPredict(Operator):
             return n
 
         # Keep a list of times that are not in the cache
-        times_not_cached = []
-        
         with self.lock:
-            for t in times:
-                if t in self.prob_cache:
-                    continue
-
-                times_not_cached.append(t)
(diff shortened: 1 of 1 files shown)


chaubold:
Nice work! Do you already have any feeling for how well this scales with the number of cores? Is caching getting in the way of full linear scaling? Does it block when there is no more RAM available to load the raw data and prepare the features for one of the RF prediction requests, or are all features computed beforehand? (Forgive me for not having a full grasp on what lazyflow handles automatically anyways...)

JaimeIvanCervantes:
It doesn't scale too well since there are still other bottlenecks in the code, like this one on `opObjectExtraction` where the object features are requested per frame.

https://github.com/ilastik/ilastik/blob/master/ilastik/applets/objectExtraction/opObjectExtraction.py#L208-L216

Overall, the object extraction and prediction operators doesn't scale very well, and more work has to be done on this end.

On another news, swapping the old compressed cache with the blocked array cache did make a big difference: https://github.com/ilastik/lazyflow/pull/214

PR #1219, opened 2016-05-11
Replaced opCompressedCache with opBlockedArrayCache

JaimeIvanCervantes:
Replaced `opCompressedCache` with `OpBlockedArrayCache` (with the compression flag enabled).

### DIFF
--- a/ilastik/applets/objectExtraction/objectExtractionSerializer.py
+++ b/ilastik/applets/objectExtraction/objectExtractionSerializer.py
@@ -98,7 +98,7 @@ class ObjectExtractionSerializer(AppletSerializer):
             SerialBlockSlot(operator.LabelImage,
                             operator.LabelImageCacheInput,
                             operator.CleanLabelBlocks,
-                            name='LabelImage',
+                            name='LabelImage_v2',
                             subname='labelimage{:03d}',
                             selfdepends=False,
                             shrink_to_bb=False,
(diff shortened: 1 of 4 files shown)


ilastik/applets/trackingFeatureExtraction/trackingFeatureExtractionSerializer.py
```diff
@@ -9,10 +9,14 @@
...
         slots = [
-            SerialHdf5BlockSlot(operator.LabelOutputHdf5,
-                                operator.LabelInputHdf5,
-                                operator.CleanLabelBlocks,
-                                name="LabelImage"),
+            SerialBlockSlot(operator.LabelImage,
+                            operator.LabelImageCacheInput,
+                            operator.CleanLabelBlocks,
+                            name='LabelImage',
+                            subname='labelimage{:03d}',
```
stuarteberg: We need to avoid backwards-compatibility issues here.  (Sorry, I didn't think of this until just now.)

I'm worried about the scenario in which someone tries the following:

1. Save a project file with the old version of ilastik
2. Upgrade to a new version (which includes your changes)
3. Open the project file

We need to make sure the new version doesn't attempt to deserialize the old compressed data (the new format isn't identical to the old one).  The easiest way to avoid this is to just to change this parameter from `name='LabelImage'` to `name='LabelImage_v2'` or something like that.  That way the the old data will be ignored.  (The labels will need to be recomputed, but that's not a big deal, right?)

ilastik/applets/objectExtraction/objectExtractionSerializer.py
```diff
@@ -95,10 +95,14 @@ def deserialize(self, group):
...
     def __init__(self, operator, projectFileGroupName):
         slots = [
-            SerialHdf5BlockSlot(operator.LabelOutputHdf5,
-                                operator.LabelInputHdf5,
-                                operator.CleanLabelBlocks,
-                                name="LabelImage"),
+            SerialBlockSlot(operator.LabelImage,
+                            operator.LabelImageCacheInput,
+                            operator.CleanLabelBlocks,
+                            name='LabelImage',
```
stuarteberg: Same backwards-compatibility concern for this slot, too.  Change the `name` parameter here.

PR #1231, opened 2016-06-16
fix resolved label image export in tracking h5 sequence export

Steffen-Wolf:
During export of h5 sequence (in conservation tracking) the original label image (without resolved labels) was exported
but move and division ids are taken from the resolved events.

Fix:
- check if self.mainOperator.RelabeledImage exists (to be compatible with chaingraphTrackingApplet)
- if it exists use resolved label image for export

### DIFF
--- a/ilastik/applets/tracking/base/trackingBaseGui.py
+++ b/ilastik/applets/tracking/base/trackingBaseGui.py
@@ -316,19 +316,24 @@ class TrackingBaseGui( LayerViewerGui ):
 
             t_from = int(t_from)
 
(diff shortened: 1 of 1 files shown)


stuarteberg:
@JaimeIvanCervantes @chaubold @ilastikdev  -- Can one of you review this?

JaimeIvanCervantes:
It seems correct to me. @Steffen-Wolf note that `_export()` will also change when `HyTra-integration` is merged, since we will not be using the `label2color` dictionary anymore (see this commit: https://github.com/ilastik/ilastik/commit/dda0df9807d9c619fdded2d66e0bef76d640c0dc)

PR #1247, opened 2016-07-07
Fix 1188: Convex hull and skeleton features are now being used in the tracking workflow

### LINKED ISSUE #1188: Conservation tracking and Animal Tracking do not use 2D convex hull and skeleton features

JaimeIvanCervantes:
Fix #1188

The conservation tracking workflow was training the RF classifier only with the `Standard Object Features`, and it was ignoring the skeleton and convex hull features. In order to solve this problem the following changes were made:

1. The `opObjectExtraction` `Features` slot was populated with the `Standard Object Features`, `Skeleton Features` and `Convex Hull Features` on the initialization of the conservation tracking workflow.

2. In `objectClassificationGui` the values of the `computedFeatureNames` and `selectedFeatureNames` slots were obtained using a deep copy in order to prevent modfying them directly from the GUI.

### DIFF
--- a/ilastik/applets/objectClassification/objectClassificationGui.py
+++ b/ilastik/applets/objectClassification/objectClassificationGui.py
@@ -31,6 +31,7 @@ from ilastik.applets.objectExtraction.opObjectExtraction import default_features
 from ilastik.applets.objectClassification.opObjectClassification import OpObjectClassification
 
 import os
+import copy
 import vigra
 
 import numpy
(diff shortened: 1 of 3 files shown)


chaubold:
Cool thanks! (Haven't tested it yet though)

stuarteberg:
@JaimeIvanCervantes, Is this ready to go now?

JaimeIvanCervantes:
This is ready to merge

PR #1248, opened 2016-07-07
Structured learning tracking test

Steffen-Wolf:
test if structured learning can load a the included ilp project file


akreshuk:
Thanks, @Steffen-Wolf !

PR #1300, opened 2016-07-21
do not take 2d features for 3d data in tracking

akreshuk:
If you switch to the object count classification applet, it still shows 25 features selected, but runs with 5.


JaimeIvanCervantes:
I checked the fix with drosophilia 3d+t and it works fine. I did notice the message below the `Subset Features` button stating that 25 features were selected like you mentioned, but it is set correctly after the button is pressed. I'm going to go ahead and merge the request.

PR #1303, opened 2016-07-22
VTK replacement - ilastik version

nielsbuwen:
This is the second part of the vtk replacement pull request. The other part is located in ilastik/volumina#164


chaubold:
changes are incorporated through #1640

Issue #1268, opened 2016-07-12
Tracking (manual and learning): "active track" number in training tab is unreadable for dark colors


FredHamprecht:
plot the number next to the color

Issue #1270, opened 2016-07-12
Tracking with Learning workflow: disallow negative parameters in the solution


FredHamprecht:
or, more generally: add an l2-bias towards parameters that are sensible -- for example the parameters that we published in past papers. 

Bias should be strong when there are few annotations and vice versa.

chaubold:
@ilastikdev: as an initial solution we could simply take the numbers that get out, check whether they are negative, and if not, set them to some small epsilon + warn that they are wrong. Or we fall back to the default parameters the Conservation Tracking workflow always used. @akreshuk said that she would like to have this check in the binary as soon as possible, because negative weights lead to tracking solutions that will just confuse the users.

ilastikdev:
Default weights used to be
det 10
div 10
tra 10
app 500
disapp 500
normalized this translates to
.01
.01
.01
.7
.7
which is very far off from some typical weights from learning:
.92 
.11
.01
.27
.21
and
.77
.49
.01
.34
.19

So I would suggest:
.8
.4
.01
.3
.3
for the bias.

ilastikdev:
Adding non-negativity constraints for the weights to the structured learning would be the best solution, 
but would require modifying opengm.

ilastikdev:
Initial solution has been coded.

"... initial solution we could simply take the numbers that get out, check whether they are negative, and if not, set them to some small epsilon + warn that they are wrong."

ilastikdev:
will open an issue in opengm:
"Learning needs a flag to trigger non-negativity constraints for weights" #472

PR #1313, opened 2016-08-01
Learning: fixes #1276

### LINKED ISSUE #1276: Structured Learning Tracking: retraining a classifier removes tracking annotations
While presenting the structured learning workflow today I went back to the division classifier and object count classifiers, added the "2 objects" class to the latter, added labels and retrained both. Then I went to go to the "Tracking -SL" applet again and tried to compute weights, and only when the results looked weird (negative transition weight) I noticed that the training annotations were completely gone.

The annotations should persist unless one changes the threshold/segmentation, and then it should warn about the annotation invalidation (probably already does, because that also invalidates the labels for the classifiers).

ilastikdev:
It would be good if this makes it into the next release (rc8)

### DIFF
--- a/ilastik/applets/tracking/annotations/annotationsGui.py
+++ b/ilastik/applets/tracking/annotations/annotationsGui.py
@@ -79,6 +79,8 @@ class AnnotationsGui(LayerViewerGui):
         self._drawer.gotoLabel.pressed.connect(self._onGotoLabel)
         self._drawer.saveAnnotations.pressed.connect(self._onSaveAnnotations)
         self._drawer.initializeAnnotations.pressed.connect(self._onInitializeAnnotations)
+        self._drawer.trackColorButton.setToolTip("Active track colour.")
+        self._drawer.activeTrackBox.setToolTip("Active track label.")
 
         self.editor.showCropLines(True)
         self.editor.cropModel.editableChanged.emit (False)
--- a/ilastik/applets/tracking/annotations/annotationsGui.py
+++ b/ilastik/applets/tracking/annotations/annotationsGui.py
@@ -979,12 +979,12 @@ class AnnotationsGui(LayerViewerGui):
                     uniqueLabels.remove(0)
                 if len(uniqueLabels) != 1:
                     self._log('tracking candidates at t = ' + str(t) + ': ' + str(uniqueLabels))
-                    self._gotoObject(oid_prev, t-1, True)
+                    self._gotoObject(oid_prev, t-1, keepXYZ=True)
                     t_end = t-1
                     break            
                 if numpy.count_nonzero(li_product) < 0.2 * numpy.count_nonzero(li_prev_oid):
                     self._log('too little overlap at t = ' + str(t))
-                    self._gotoObject(oid_prev, t-1, True)
+                    self._gotoObject(oid_prev, t-1, keepXYZ=True)
                     t_end = t-1
                     break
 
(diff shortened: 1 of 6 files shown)


ilastikdev:
This is the initial solution to #1270

ilastikdev:
Learning errors are now fixed and we could make a new release
(see Anna's idea in #1270)

ilastikdev:
fixes #1268

stuarteberg:
Just checking in: Is this branch ready for review, or are you still working on it?  If it's ready for review, who would be the best choice to review it?  (If you need any feedback from me, let me know.)

ilastikdev:
There are two fixes (#1270 and #1276) that should make it into the next release asap.
There are no backend C++ changes.

That said, I am continuing to work on learning, but it is enhancements only.

So go ahead and review/release.

chaubold:
This feels very sturdy now. I've tried to make it break in different ways but it did very well, no crashes, no lost labels, no weird resulting weights.

@stuarteberg do you want to skim over the code changes to make sure this is good before we merge? 
I think it's fine and could be merged as is.

**Comments:**
* I like the `set active track to label XXX` functionality!
* Something we should discuss offline: the user may still explicitly set weights to negative numbers, do we want to allow that?

**Suggestion for UI improvement:**

* you changed the display of the active track color, but now one cannot see the color of other track IDs any more. Could we please switch to the same display style that @akreshuk used for the new channel selection in the thresholding applet? That would be *a)* consistent and *b)* get rid of the unreadable label IDs but still show the color.

stuarteberg:
Cool.  Sure, I'll have a look later today.

stuarteberg:
>Could we please switch to the same display style that @akreshuk used for the new channel selection in the thresholding applet?

Just to cross-reference, @chaubold is talking about this change to `thresholdTwoLevelsGui.py`: https://github.com/ilastik/ilastik/commit/8788b971b9f8b615bcef484d01610db25981b7d2

ilastik/applets/tracking/annotations/annotationsGui.py
```diff
@@ -1064,15 +1072,15 @@ def _onDivEventPressed(self):
         self._enableButtons(exceptButtons=[self._drawer.divEvent], enable=(not self.divLock))                      
 
 
-    def _setStyleSheet(self, widget, qcolor):                         
+    def _setStyleSheet(self, widget, qcolor, type="QComboBox"):
         values = "{r}, {g}, {b}, {a}".format(r = qcolor.red(),
```
stuarteberg: Small nit-pick: Here, you're overwriting the meaning of the built-in `type` definition.  I think it's better to avoid that.  (Fun fact: Python 2.7 even allows you to overwrite `False` and `True` if you want -- but don't do that, either! :-)

ilastik/workflows/tracking/structured/structuredTrackingWorkflow.py
```diff
@@ -180,16 +180,17 @@ def connectLane(self, laneIndex):
...
+            opDivDetection.BinaryImages.connect( op5Binary.Output )
+            opDivDetection.RawImages.connect( op5Raw.Output )
+            opDivDetection.SegmentationImages.connect(opTrackingFeatureExtraction.LabelImage)
+            opDivDetection.ObjectFeatures.connect(opTrackingFeatureExtraction.RegionFeaturesAll)
+            opDivDetection.ComputedFeatureNames.connect(opTrackingFeatureExtraction.ComputedFeatureNamesAll)
+            opDivDetection.SelectedFeatures.setValue(configConservation.selectedFeaturesDiv)
+            opDivDetection.LabelNames.setValue(['Not Dividing', 'Dividing'])
+            opDivDetection.AllowDeleteLabels.setValue(False)
+            opDivDetection.AllowAddLabel.setValue(False)
+            opDivDetection.EnableLabelTransfer.setValue(False)
```
stuarteberg: Instead of calling `setValue()` here (in `connectLane()`) on level-0 slots  (`SelectedFeatures`, `LabelNames`, `AllowDeleteLabels`, `AllowAddLabel`, `EnableLabelTransfer`), I think it would be better to do that in `__init__()`, above.  The reason is that `connectLane()` is called repeatedly (for every image loaded), so you'll be calling `setValue()` with the same values over and over.

IIUC, the current code won't lead to wrong results, but if we ever decide to make these slots configurable by the user, then these lines will definitely need to be moved.  Otherwise, we'll be overwriting their settings every time they load a new image.

ilastikdev:
there is also a PR in volumina: 
github.com/ilastik/volumina/pull/166
that improves the time slider

stuarteberg:
LGTM.

PR #1329, opened 2016-08-31
Py2to3 ilastik

njase:
**Tag:** py2to3_v1.0
**Details:** PEP238 support for integer division using __future__ import division.
**Test status**:
 - Nearly all changes tested - list attached
      -- unit tests
      -- manual invocation of application and/or python -3 switch
 - Some changes could not be reached and is asserted
 - Automatic tracking has to be tested by tracking guys
[v1.0_Overall_result.xlsx](https://github.com/ilastik/ilastik/files/447335/v1.0_Overall_result.xlsx)

### DIFF
--- a/ilastik/applets/autocontextClassification/autocontextClassificationSerializer.py
+++ b/ilastik/applets/autocontextClassification/autocontextClassificationSerializer.py
@@ -36,6 +36,8 @@ traceLogger = logging.getLogger("TRACE." + __name__)
 
 from lazyflow.utility import Tracer
 
+from __future__ import division
+
 class Section():
     Labels = 0
     Classifiers = 1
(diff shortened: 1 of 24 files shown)


chaubold:
Thanks a lot! 

Regarding the tracking workflows: I assume you didn't test it because you don't have CPLEX available? @akreshuk or I can give you the installer. But we're switching to a newer tracking pipeline anyways (in the branch `ilastik/HyTra-integration`, using https://github.com/chaubold/hytra), so it might be more important to check that this works fine with Python 3.

@akreshuk, do you think this would be a good thing for @njase to look at, too? I'll be on vacation most of September though (leaving on Tuesday 6th), so you'd have to consult @Steffen-Wolf or @JaimeIvanCervantes for details. And in that case I should try to build the C++ package [`dpct`](https://anaconda.org/chaubold/dpct) with python-3 bindings before I leave.

njase:
@chaubold : Yes, the reason was due to CPLEX. Thanks for inputs - I'll also discuss with @akreshuk

PR #1354, opened 2016-09-16
fixes #1352 clear label code added

### LINKED ISSUE #1352: "Clear dividing" does not work
It is not possible to remove a label in the object classification applets of the tracking workflow and clearing the labels also does not do what's expected. We should provide users a way to clear the labels while experimenting, without going through the dataset and unclicking everything.

ilastikdev:
triggers live update mode for the  user to see the changes

remove lablel also triggers live update mode


ilastikdev:
@akreshuk done

akreshuk:
Thanks!

PR #1367, opened 2016-11-08
Object extraction optimization

JaimeIvanCervantes:
Object extraction is now done in parallel. 

This PR also includes a script to compare the running-time of the object extraction operator against a simplified version of the same operator in order to figure out the running-time of the overhead (caches, plugin system, labelling, etc). This script is found in the `benchmarks` folder.

### DIFF
--- a/benchmarks/objectExtractionRunningTime.py
+++ b/benchmarks/objectExtractionRunningTime.py
@@ -68,9 +68,6 @@ FEATURES = {
     }
 }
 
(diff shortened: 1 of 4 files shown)


akreshuk:
Should we merge that or are you planning more changes? Looks very good to me!

JaimeIvanCervantes:
I would like to do one small change before merging, and then we're set.

PR #1375, opened 2016-11-29
New Conservation Tracking

JaimeIvanCervantes:
An important update to the new conservation tracking workflow that includes the following changes:

* Serialization of the hypotheses graph and resolved mergers dict
* Removed serialization of tracking volumes
* Replaced the old compressedCaches by opBlockedArrayCaches
* Added option to export csv only
* Added ILP solver to the solvers available on the dropdown menu
* Removed unused slots and member variables
* Other minor fixes and optimizations

This pull request follows the `New conservation tracking workflow` PR: https://github.com/ilastik/ilastik/pull/1363

Also, see `new-conservation-tracking` branch on HyTra: https://github.com/chaubold/hytra/tree/new-conservation-tracking


ilastik/applets/tracking/conservation/opConservationTracking.py
```diff
@@ -406,12 +408,37 @@ def constructFov(shape, t0, t1, scale=[1, 1, 1]):
...
+                                neighbor = edge[1]                        
+                                if  originalGraph.hasNode(neighbor) and 'value' in originalGraph._graph.node[neighbor] and  originalGraph._graph.node[neighbor]['value'] > 1:
+                                    withCoords = True
+                                    break
+                                                 
+                        if withCoords:    
+                            pool.add(Request(partial(mergerResolver.getCoordinatesForObjectId, coordinatesForIds, labelImage[0, ..., 0], objectId)))
+                            poolIsNotEmpty = True                   
+                    
+                    if poolIsNotEmpty:
```
stuarteberg: Tiny note: `RequestPool` has a `__len__` method, so you could simplify this logic with `if len(pool):`.  (**Edit:** I see that this code is removed in a later commit anyway, so there's nothing to do here...)

ilastik/applets/tracking/conservation/conservationTrackingGui.py
```diff
@@ -96,25 +123,35 @@ def _loadUiFile(self):
...
+                if dpct:
+                    solvers.append('Flow-based')
+            except:
+                pass
+            try:
+                if mht:
+                    solvers.append('ILP')
+            except:
+                pass
+        else:
```
stuarteberg: Writing `except: pass` is always super-sketchy.  It's best to catch the specific error you're expecting to see.

ilastik/applets/tracking/conservation/opConservationTracking.py
```diff
@@ -498,62 +525,96 @@ def stackExtraColumnWithZeros(array):
...
+        
+        # Reduce labels to the ones that contain mergers
+        if onlyMergers:
+            if resolvedMergersDict:
+                if time not in resolvedMergersDict:
+                    idxs = []
+                else:
+                    newIds = [nodeDict['newIds'] for _, nodeDict in resolvedMergersDict[time].items()]
+                    newIds = [id for ids in newIds for id in ids]
+                    idxs = [id for id in idxs if id in newIds]
```
stuarteberg: Style note: `id` is the name of a builtin function, so it's best not to use it as a variable name.

ilastik/applets/tracking/conservation/opConservationTracking.py
```diff
@@ -132,84 +125,79 @@ def setupOutputs(self):
...
 
             for t in trange:
-                if (self.mergerResolver 
-                        and 'time_range' in parameters 
-                        and t <= parameters['time_range'][-1] 
-                        and t >= parameters['time_range'][0]):
-                    self.mergerResolver.relabelMergers(result[t-roi.start[0],...,0], t)
+                if 'time_range' in parameters and t <= parameters['time_range'][-1] and t >= parameters['time_range'][0]:
+                    if self.ResolvedMergers.value:
+                        self._labelMergers(result[t-roi.start[0],...,0], t, offset)
```
stuarteberg: It might be better to pre-save the result of `self.ResolvedMergers.value` outside of this loop, since `.value` is probably relatively slow compared to normal property.  (Just a note for future reference; no need to change this PR if you'd rather not.)

ilastikdev:
@JaimeIvanCervantes 

can we serialize with a (custom) dict slot with group/subroup hdf5 structure 

insted of 

SerialPickledValueSlot(mainOperator.ExportSettings),
 +                     SerialPickledValueSlot(mainOperator.HypothesesGraph),
 +                     SerialPickledValueSlot(mainOperator.ResolvedMergers)

in d111ee5f4113dce609d5577e65583e1912fc6943

ilastikdev:
... it makes any kind of test case debuging related to tracking solutions possible

JaimeIvanCervantes:
@ilastikdev yes, in fact customizing the serialization to something more structured is on my list of things to do. I am pickling those slots for now, only because of time constraints.

chaubold:
I reverted the "color table fix" commit that I added before because it introduced the merge conflict. But now the history has 2 commits that we could get rid of completely if we rewrite the history of this PR branch. But we merged the `new-conservation-tracking` branch already into `hytra-learning` and thus rewriting the history could mess up the history of the other branch. 
@stuarteberg what do you think? Does it matter?

stuarteberg:
>this branch has 2 commits that we could get rid of completely if we rewrite the history of this PR

IMHO, it's better to leave it like it is.  I'm a fan of using `rebase`, etc. to maintain a clean git history, but only when it's simple to do so.  In more complicated cases, or if I've already shared my commits with other people, then it's less confusing to avoid rewriting history.

Issue #1387, opened 2016-12-09
export fails if a division only has one child


ilastikdev:
```
[====================] 100%
[====================] 100%
[====================] 100%
	contains 7366 segmentation hypotheses
	contains 17302 linking hypotheses
Log: Initializing Residual Graph ...
Log:  done in 0.474798 secs
Log: Beginning tracking ...
Log: 	Cycle returned to source with negative cost -0.796828 in iteration 1
Log: 	Cycle returned to source with negative cost -1.11022e-16 in iteration 2
Log: Tracking took 1.49085 secs and 314 iterations
Log: Final energy: -241652
	contains 189 segmentation hypotheses
	contains 316 linking hypotheses
Log: Running min cost max flow on a graph with 380 nodes and 567 edges
Log: Found MaxFlow value: 31
Log: MaxFlow Tracking took 0.000416 secs
ERROR 2016-12-09 11:57:31,106 log_exception 4322 139728917706496 Traceback (most recent call last):
  File "/home/jales/software/miniconda/envs/ilastik-devel/ilastik-meta/lazyflow/lazyflow/request/request.py", line 306, in _execute
    self._result = self.fn()
  File "/home/jales/software/miniconda/envs/ilastik-devel/ilastik-meta/ilastik/ilastik/applets/tracking/conservation/opConservationTracking.py", line 665, in do_export
    self._do_export_impl(settings, selected_features, progress_slot, object_feature_slot, label_image, lane_index, filename_suffix)
  File "/home/jales/software/miniconda/envs/ilastik-devel/ilastik-meta/ilastik/ilastik/applets/tracking/conservation/opConservationTracking.py", line 729, in _do_export_impl
    assert len(hypothesesGraph._graph.out_edges(node)) != 2, "Division node {}, does not contain 2 children".format(node)
AssertionError: Division node (4, 41), does not contain 2 children

ERROR 2016-12-09 11:57:31,107 log_exception 4322 139728917706496 Export failed
ERROR 2016-12-09 11:57:31,108 log_exception 4322 139726689896192 Traceback (most recent call last):
  File "/home/jales/software/miniconda/envs/ilastik-devel/ilastik-meta/ilastik/ilastik/applets/dataExport/dataExportGui.py", line 392, in exportSlots
    self.parentApplet.post_process_lane_export(lane_index)
[...]

PR #1388, opened 2016-12-09
ConservationTracking: fix table export

chaubold:
Suggested fix for #1387, @ilastikdev please check whether it works for the dataset where you saw the error.


chaubold:
the last commit fixes https://github.com/chaubold/hytra/issues/17 by assigning resolved mergers the id `0` and false detections the id `1` as it was in ilastik before the switch to hytra.

Issue #1217, opened 2016-05-06
carving: visual cue needed when preprocessing is done


FredHamprecht:
The visual change at the end of (lengthy) preprocessing is very subtle. It would be good to have a clearer change upon completion, e.g. by displaying the watershed superpixels.

k-dominik:
fixed in #1391

PR #1391, opened 2016-12-13
Fix for issue #1217, carving workflow: watershed shown after successful preprocessing

k-dominik:
In preprocessing for the carving workflow, watershed layer is set visible iif none of the result layers are visible

Fix for issue #1217 @akreshuk


k-dominik:
this is ready to be merged;

additional changes (as discussed with @akreshuk):
 * supervoxel layer with opacity .5
 * supervoxel layers (and colored completed segment layers) with randomizable colors

PR #1401, opened 2016-12-24
Py3base ilastik

njase:
This pull request is only for the purpose of review


ilastik/applets/layerViewer/layerViewerGui.py
```diff
@@ -69,7 +69,7 @@ def __call__(cls, *args, **kwargs):
         instance._after_init()
         return instance
 
-class LayerViewerGui(QWidget):
+class LayerViewerGui(QWidget, metaclass=LayerViewerGuiMetaclass):
     """
```
stuarteberg: FWIW, it looks like `six` has a solution `with_metaclass` that would let us keep Py2 and Py3 support simultaneously.

ilastik/applets/tracking/annotations/opAnnotations.py
```diff
@@ -293,7 +293,7 @@ def do_export(self, settings, selected_features, progress_slot, lane_index, file
             ott = partial(self.lookup_oid_for_tid, oid2tid)
             divs = [(value[1], ott(key, value[1]), key, ott(value[0][0], value[1] + 1), value[0][0],
                      ott(value[0][1], value[1] + 1), value[0][1])
-                    for key, value in sorted(divisions.iteritems(), key=itemgetter(0))]
+                    for key, value in sorted(iter(divisions.items()), key=itemgetter(0))]
```
stuarteberg: I don't understand why `iter()` is needed here, but I guess it can't hurt...

ilastik/applets/tracking/structured/structuredTrackingGui.py
```diff
@@ -67,37 +67,37 @@ def _loadUiFile(self):
...
+        if 'borderAwareWidth' in list(parameters.keys()):
             self._drawer.bordWidthBox.setValue(parameters['borderAwareWidth'])
-        if 'cplex_timeout' in parameters.keys():
+        if 'cplex_timeout' in list(parameters.keys()):
             self._drawer.timeoutBox.setText(str(parameters['cplex_timeout']))
-        if 'appearanceCost' in parameters.keys():
+        if 'appearanceCost' in list(parameters.keys()):
             self._drawer.appearanceBox.setValue(parameters['appearanceCost'])
-        if 'disappearanceCost' in parameters.keys():
+        if 'disappearanceCost' in list(parameters.keys()):
```
stuarteberg: FWIW, I think there's no need for conversion to `list()` here -- `in` works with iterators, too.

ilastik/ilastik_logging/default_config.py
```diff
@@ -263,9 +263,9 @@ def init(format_prefix="", output_mode=OutputMode.LOGFILE_WITH_CONSOLE_ERRORS, l
...
     # Custom format for warnings
-    def simple_warning_format(message, category, filename, lineno, line=None):
-        filename = os.path.split(filename)[1]
-        return filename + "(" + str(lineno) + "): " + category.__name__ + ": " + message[0]
+    #def simple_warning_format(message, category, filename, lineno, line=None):
+    #    filename = os.path.split(filename)[1]
+    #    return filename + "(" + str(lineno) + "): " + category.__name__ + ": " + message[0]
 
-    warnings.formatwarning = simple_warning_format
+    #warnings.formatwarning = simple_warning_format
```
stuarteberg: I don't understand why this had to be removed.  If it is somehow not working in Py3, we need to come up with a replacement.

ilastik/shell/gui/startShellGui.py
```diff
@@ -73,7 +73,7 @@ def _applyStyleSheet(app):
     Apply application-wide style-sheet rules.
     """
     styleSheetPath = os.path.join( os.path.split(__file__)[0], 'ilastik-style.qss' )
-    with file( styleSheetPath, 'r' ) as f:
+    with open( styleSheetPath, 'r' ) as f:
```
stuarteberg: Nice catch.  From a quick grep, it looks like the following files need the same change (`file` -> `open`).  (You might have caught these alreadly; this is a note for me to check.)

-  `ilastik/clusterConfig.py`
- `lazyflow/utility/io_util/RESTfulVolume.py`
- `lazyflow/utility/io_util/RESTfulVolume.py`

(Weird thing is, I didn't even know that `file()` can substitute for `open()`, but apparently I wrote all these lines... Strange that I made this mistake without even knowing it was legal.)

ilastik/utility/autoEval.py
```diff
@@ -21,6 +21,6 @@
 def autoEval(x, t):
     if type(x) is t:
         return x
-    if type(x) is str or type(x) is unicode and t is not str:
+    if type(x) is str or type(x) is str and t is not str:
```
stuarteberg: This is now redundant...

tests/helpers/__init__.py
```diff
@@ -18,5 +18,13 @@
...
+submodule_dir = '/Users/saurabh/miniconda3/envs/ilastik-py3/ilastik-meta'
+# Add all submodules to the PYTHONPATH
+from . import expose_submodules
+expose_submodules.expose_submodules(submodule_dir)
+
+#import sys
+#print(sys.path)
+
+from .shellGuiTestCaseBase import ShellGuiTestCaseBase
+from . import mainThreadHelpers
```
stuarteberg: This will need to be fixed, of course...

tests/test_applets/thresholdTwoLevels/testThresholdTwoLevels.py
```diff
@@ -43,8 +43,9 @@
 ## for testing ThresholdOneLevel
 class Generator1(unittest.TestCase):
 
-    def generateData(self, (nx, ny, nz)):
+    def generateData(self, xxx_todo_changeme):
 
+        (nx, ny, nz) = xxx_todo_changeme
```
stuarteberg: :smile:

ilastik/__init__.py
```diff
@@ -29,10 +29,10 @@
 this_file = os.path.realpath( this_file )
 ilastik_package_dir = os.path.dirname(this_file)
 ilastik_repo_dir = os.path.dirname(ilastik_package_dir)
-submodule_dir = os.path.join( ilastik_repo_dir, 'submodules' )
+submodule_dir = os.path.dirname(ilastik_repo_dir)
```
njase: Yes, relative imports were otherwise causing problem. The submodule_dir was having incorrect dir path, and the same were "exposed". 
Python3 was unable to find the needed modules during imports.

ilastik/applets/cropping/cropSelectionGui.py
```diff
@@ -295,12 +295,12 @@ def newCrop(self):
...
-            unicode("starts"): self.editor.cropModel.get_roi_3d()[0],
-            unicode("stops"): self.editor.cropModel.get_roi_3d()[1],
-            unicode("cropColor"): (color1.red(), color1.green(),color1.blue()),
-            unicode("pmapColor"): (color2.red(), color2.green(),color2.blue())
+        self.topLevelOperatorView.Crops.value[str(self._cropControlUi.cropListModel[selectedRow].name)] = {
+            str("time"): (self.topLevelOperatorView.MinValueT.value, self.topLevelOperatorView.MaxValueT.value),
+            str("starts"): self.editor.cropModel.get_roi_3d()[0],
+            str("stops"): self.editor.cropModel.get_roi_3d()[1],
+            str("cropColor"): (color1.red(), color1.green(),color1.blue()),
+            str("pmapColor"): (color2.red(), color2.green(),color2.blue())
```
njase: Sounds nice. If you are thinking of generalizing this to complete code, I just read a word of caution here - https://pythonhosted.org/six/#six.u
"Note On Python 2, u() doesn’t know what the encoding of the literal is. Each byte is converted directly to the unicode codepoint of the same value. Because of this, _**it’s only safe to use u() with strings of ASCII data**_"

ilastik/applets/layerViewer/layerViewerGui.py
njase: This is an automatic change by python utility - 2to3
I agree with your remark.

ilastik/applets/tracking/annotations/opAnnotations.py
njase: Automatic change by 2to3.
Answer is - principles laid down by PEP469
See here: http://legacy.python.org/dev/peps/pep-0469/#id9

ilastik/applets/tracking/structured/structuredTrackingGui.py
njase: Same as previous

ilastik/ilastik_logging/default_config.py
njase: I don't remember making this change except maybe during some debugging. Definitely, it should not be removed. Sorry for this and thanks you noticed. 
I am sure there won't be another one like this.

ilastik/shell/gui/startShellGui.py
njase: Actually file has been removed in python 3 so runtime execution will fail.  
yes, I missed to fix other instances!

ilastik/utility/autoEval.py
njase: :)

tests/helpers/__init__.py
njase: Actually, the whole way of relative imports in "tests".
I will be really interested to learn the strategy you might define for this.

tests/test_applets/thresholdTwoLevels/testThresholdTwoLevels.py
njase: Oh!

k-dominik:
Closing this as the python3-change is already done

PR #1417, opened 2017-02-15
learning/tracking workflow

ilastikdev:
changing from region centers for traxel generations to lowers/uppers
(at least one pixel needs to be in the crop for the object to be in the crop - so that what the user sees in the crop and annotates in the crop gets taken into learning and tracking

this does not change the behaviour on the boundary of the dataset


ilastikdev:
ready to be merged

ilastik/applets/tracking/conservation/opConservationTracking.py
```diff
@@ -435,7 +436,7 @@ def track(self,
         trackingGraph = hypothesesGraph.toTrackingGraph()
         trackingGraph.convexifyCosts()
         model = trackingGraph.model
-        model['settings']['allowLengthOneTracks'] = False
+        model['settings']['allowLengthOneTracks'] = True
```
chaubold: We do need an option for users to filter these length-one tracks from the output, or are they filtered away by default? The reason is that having single-frame tracks popping up from time to time will confuse any analysis steps build for ilastik versions up to now. Users must be made aware of this change, or we should produce output (= images, CSV table, other output plugins) without length one tracks by default.

ilastik/utility/gui/progress.py
```diff
@@ -0,0 +1,81 @@
...
+        self.trackProgress.newStep.connect(self.__onNewStep, Qt.BlockingQueuedConnection)
+        self.trackProgress.done.connect(self.onTrackDone)
+        self.trackProgress.start()
+
+    def onTrackDone(self):
+        self.trackProgress.wait() # Join the extractor thread so its safe to immediately destroy the window
+        self.finished.emit()
+        self.close()
+
+class GuiProgressVisitor(DefaultProgressVisitor):
```
chaubold: I don't see where you are using this. Shouldn't you inject that into the hypotheses graph in `opConservationTracking.track()`?

ilastik/applets/tracking/structured/structuredTrackingGui.py
```diff
@@ -107,7 +105,14 @@ def _loadUiFile(self):
             self._drawer.appearanceBox.setValue(parameters['appearanceCost'])
         if 'disappearanceCost' in parameters.keys():
             self._drawer.disappearanceBox.setValue(parameters['disappearanceCost'])
-        
+
+        solverName = self.topLevelOperatorView._solver
```
chaubold: Side note: Did you already provide a solver selection drop down box as in the normal tracking workflow?

ilastikdev: on todo list

ilastik/utility/gui/progress.py
ilastikdev: have not commited all of the code yet, but now you can see it (see bellow)

chaubold:
I find it weird that we now have a command line option to choose whether we want a command line or GUI progress bar. Shouldn't we use the command line progress bar when running headless, and the GUI one if a GUI is present?

ilastikdev:
we can do that by default, but we should give the user at least an option of turning it off

ilastikdev:
this is ready to be merged

it contains a few minor fixes and a lot of enhancements,

including a mamut export and load of a learning workflow even if cplex is not present
(so users with no cplex can annotate and run tracking with user's input of weights)

akreshuk:
I also don't see the point of turning progress bar off. I don't recall seeing such option in any other software. Is it more resource-consuming than I think?

ilastikdev:
For small to medium size projects progress bar slows down the execution because of pyqt signals that facilitate communication between the main thread and progress bar thread.

Progress bar is useful for new users and running it on new datasets in gui mode, but for rerunning on similar data sets it is not needed, also in headless python print on command line (from HyTra) takes time.

stuarteberg:
Merging this now so it can be included in the release.  But feel free to continue the discussion...

akreshuk:
To continue the discussion, pyqt signals have been freed of all charges. 
The current theory is that the gui progress bar is updated too frequently and slows everything down. @ilastikdev plans to reduce the update rate, let's see if that helps.

PR #1427, opened 2017-03-14
Bypassing cache when conservation tracking runs on headless mode

JaimeIvanCervantes:
Bypassing cache (used in opObjectExtraction/opLabelVolume) when conservation tracking is running on headless mode.


stuarteberg:
Instead of disabling the cache in headless mode, I think it would be a more general fix to bypass it whenever the user is exporting data in batch mode (including with the GUI).  This can be done by changing the flag in `prepare_for_entire_export()` and `post_process_entire_export()`.  For example, the [Autocontext workflow does this](https://github.com/ilastik/ilastik/blob/master/ilastik/workflows/newAutocontext/newAutocontextWorkflow.py#L430).

PR #1434, opened 2017-03-20
Tracking export plugin system

chaubold:
This PR introduces a plugin system for the export of tracking solutions.
Plugins are located in `ilastik/plugins_default` and use ilastik's yapsy plugin manager.

In the tracking export applet, there is a new export source called `Plugin`, and when that is selected the `Configure Image Export Settings` button will bring up a specialized view to select an export plugin, select a filename, and provides an explanation what the plugin does.

screen shot 2017-03-20 at 16 35 07

They can also be used in batch mode on the command line e.g. by:

```sh
./run_ilastik.sh --project mytracking.ilp --headless --export_source="Plugin" --export_plugin="CSV-Table" --raw_data="some.h5/data" --segmentation_image="someother.h5/exported_data"
```

**Currently available plugins:**

* HDF5 sequence export
* Fiji's MaMuT (relies on our ilastik-fiji plugin to be installed in Fiji to load the HDF5 volume in our format. That plugin is not released yet, but should come soon)
* CSV Table

**Other changes:**

* This PR gets rid of the "configure export table" option that ilastik had before, as that functionality is implemented as plugin now. This is inconsistent with the ObjectClassification workflow.
* all of these changes only apply if the *hytra* backend is used for tracking. Once we no longer need the pgmlink backend for Windows we can get rid of more code.

**TODO:**

* [x] Test whether that still works with the *pgmlink* backend. 
* [x] Add tests

**Next steps:**

* [ ] Put the required [MaMutExporter](https://github.com/Beinabih/MaMutConverter) conda package into our ilastik channel
* [ ] Release [ilastik-fiji](https://github.com/Beinabih/ilastik-Fiji-Plugin) plugin
* [ ] Add output format of the ISBI Cell Tracking Challenge

(edited to split ToDos into immediate things and next steps)


ilastik/workflows/tracking/conservation/conservationTrackingWorkflow.py
```diff
@@ -353,27 +355,56 @@ def prepare_lane_for_export(self, lane_index):
             withBatchProcessing = True
         )
 
-    def post_process_lane_export(self, lane_index):
+    def post_process_lane_export(self, lane_index, time=1):
```
JaimeIvanCervantes: Could you add a comment about why you added the `time` parameter and why does it starts at 1? It seems a bit counter-intuitive to me. You're also changing the name of a function that is called from the `dataExportApple` but you probably already thought of this.

JaimeIvanCervantes:
Good job guys, I briefly went over the code and left one comment already. I will be doing some stress testing in the next days since I have to process a bunch of data soon. I will also use add a plugin to export larvae contours soon.

ilastik/workflows/tracking/conservation/conservationTrackingWorkflow.py
LetiP: This was my perhaps unexperienced way to make sure, that we check only once for files that could be overwritten, pop up the MessageBox and then don't export. (time=0). For the next round we click the export button, we really want it to export, so time=1. The default parameter is 1, so everything but not 0, in order to ensure writing out even in headless mode.
Perhaps there is a more elegant way?

chaubold: We could maybe change the parameter name to `overwriteExistingFiles` and set it to `True` by default?

@JaimeIvanCervantes I agree that changing the signature of `post_process_lane_export` is not the most elegant way, but as a default parameter is specified that shouldn't be a problem, right?

@stuarteberg: Is the export of other files from the export applets simply overwriting existing results, or is there some other way the user could be notified about files that are about to be overwritten?

LetiP: @chaubold yes, very good idea with the new parameter name. I changed it.

ilastik/workflows/tracking/conservation/conservationTrackingWorkflow.py
```diff
@@ -391,9 +391,9 @@ def post_process_lane_export(self, lane_index, time=1):
                     logger.error("Cannot export from plugin with empty output filename")
                     return
 
-                if time == 0 and selectedPlugin == 'H5-Event-Sequence' and os.path.exists(filename + '/H5-Event-Sequence/'):
+                if overwriteExistingFiles == False and selectedPlugin == 'H5-Event-Sequence' and os.path.exists(filename + '/H5-Event-Sequence/'):
```
chaubold: You could even write `if not overwriteExistingFiles and ...` now 😉 

Now that I think about it, the workflow file is not the right place for these checks. The bad thing is that we need an if statement for every new plugin here. The plugin itself knows best where it tries to save the results, so we could push this `exportStatus` one level further: 
Plugins should get an extra method to check whether they would overwrite any files, and that could be called from within `self.trackingApplet.topLevelOperator.getLane(lane_index).exportPlugin(filename, exportPlugin` **`, overwriteExistingFiles)`**. What do you think about that @LetiP?

LetiP: @chaubold thanks for the idea, I hope my last commit implements it the way you thought of it.

chaubold: Yes thanks! My changes below only made it a little nicer.

ilastik/plugins_default/tracking_csv_export.py
```diff
@@ -0,0 +1,96 @@
...
+                                    else:
+                                        table[rowIdx, colIdx] = 0
+                                colIdx += 1
+                        else:
+                            table[rowIdx, colIdx] = features[frame][category][feature][label]
+                            colIdx += 1
+
+        # sort table by frame, then labelImage
+        table = table[table[:,1].argsort()]
+        table = table[table[:,0].argsort(kind='mergesort')]
```
chaubold: I wanted the table to be ordered as in the old CSV export, as the comment says, first key is frame, second is label ID. AFAIK numpy doesn't provide a simple way to sort by two keys. So: I first sort by label ID and then I use `mergesort` to sort by frames. `mergesort` because it is stable, and hence leaves the order of label IDs within each frame untouched (default is Quicksort, not stable).

ilastik/plugins_default/tracking_csv_export.py
```diff
@@ -0,0 +1,96 @@
...
+        headers = ['frame', 'labelimageId', 'trackId', 'lineageId', 'parentTrackId', 'mergerLabelId']
+        excludedFeatures = ['Histogram']
+        
+        # check which features are present and construct table of the appropriate size
+        frame, _ = graph.nodes_iter().next()
+        for category in features[frame].keys():
+            for feature in features[frame][category].keys():
+                if feature not in excludedFeatures:
+                    if (np.array(features[frame][category][feature])).ndim == 2:
+                        for column in range(np.array(features[frame][category][feature]).shape[1]):
```
chaubold: Thanks for the hint, I'll replace those!

chaubold:
One word about the `mamutexport` conda package. We don't necessarily need that plugin for the merge, the plugin infrastructure is more important. In general, if a plugin fails to load, we do show an `ERROR` in the log but continue. So even if we ship the plugin but not `mamutexport`, we will get an error in the log, but that's it. The plugin should not show up in the list of available export options.

That's why I put this under "next steps", which are not necessary for the merge of this PR in my opinion, but would be nice to have. The plugin system does anyway provide HDF5 sequence and CSV exports, so similar functionality as before.

ilastik/applets/tracking/base/trackingBaseDataExportGui.py
```diff
@@ -101,10 +216,25 @@ def show_export_dialog(self):
...
+        if self.topLevelOperator.SelectedExportSource._value == OpTrackingBaseDataExport.PluginOnlyName and not self.pluginWasSelected:
+            QMessageBox.critical(self, "Choose Export Plugin",
+                                 "You did not select any export plugin. \nPlease do so by clicking 'Choose Export Image Settings'")
+            return
+        else:
+            super(TrackingBaseDataExportGui, self).exportResultsForSlot(opLane)
+
+    def exportAllResults(self):
+        if self.topLevelOperator.SelectedExportSource._value == OpTrackingBaseDataExport.PluginOnlyName and not self.pluginWasSelected:
+            QMessageBox.critical(self, "Choose Export Plugin",
```
chaubold: nope, fixed.

chaubold:
Last big todo: 

* [ ] update the [documentation](http://ilastik.org/documentation/tracking/tracking#tocAnchor-1-1-6)!

ilastik/plugins_default/tracking_csv_export.py
stuarteberg: Ah, I see.

>AFAIK numpy doesn't provide a simple way to sort by two keys.

You can do it with [np.lexsort()](https://docs.scipy.org/doc/numpy/reference/generated/numpy.lexsort.html).

```python
# sort table by frame, then labelImage
table = table[ np.lexsort(table[:,:2].transpose()[::-1]) ]
```

(The `[::-1]` is necessary because lexsort considers the *last* position to be the primary key.)

chaubold: Thanks!

chaubold:
@stuarteberg: We're good to go I think. The conda package for `mamutexport` is also ready, so if you want to put this into your list of things to build, that'd be awesome. See https://github.com/ilastik/ilastik-build-conda/pull/28

Who wants to click on **merge**? 😉

PR #1437, opened 2017-03-24
Watershed workflow

shaker6691:
1. Documentation expansion to get a faster introduction in implementing an applet and a workflow
2. Watershed Workflow including
  * data selection
  * seeds applet
  * watershedSegmentation applet
  * data export
3. seeds applet: 
includes the generation of seeds
4. watershedSegmentation applet:
the manual manipulation of seeds ist possible. the seeds can be reset to the initial seeds, so that all changes are removed. 
The watershed algorithm can be executed on boundaries and seeds (if seeded)
The pixel value of the LabelCache can be displayed


ilastik.py
```diff
@@ -77,6 +77,7 @@ def main():
     # Provide your command-line args here. See examples below.
     
     ## Auto-open an existing project
+    parsed_args.project='/home/andreas/MyProject.ilp'
```
akreshuk: Could we remove all of these lines? Or leave just one in as an example, but not add more?

ilastik/applets/layerViewer/layerViewerGui.py
```diff
@@ -310,6 +356,72 @@ def _create_grayscale_layer_from_slot(cls, slot, n_channels):
         return layer
 
     @classmethod
+    def createGrayscaleLayer(cls, slot):
```
akreshuk: Do we really need these functions? There is already an addGrayscaleLayer in the viewer, what's the advantage of this function? I don't mind adding it if you think it makes developer life easier, just want to avoid too many functions doing almost the same thing.

shaker6691:
I guess, one line would be nice as an example but more are not needed.

If you would like to remove it, the nice2know.rst file contains some information on the as well.

Am 27.03.2017 um 14:14 schrieb akreshuk:

@akreshuk commented on this pull request.

________________________________

In ilastik.py<https://github.com/ilastik/ilastik/pull/1437#discussion_r108153542>:

> @@ -77,6 +77,7 @@ def main():
     # Provide your command-line args here. See examples below.

     ## Auto-open an existing project
+    parsed_args.project='/home/andreas/MyProject.ilp'


Could we remove all of these lines? Or leave just one in as an example, but not add more?

ilastik.py
shaker6691: I guess, one line would be nice as an example but more are not needed. If you would like to remove it, the nice2know.rst file contains some information on the as well.

ilastik/applets/layerViewer/layerViewerGui.py
shaker6691: The function createGrayscaleLayer seems to be redundant as you said. I found just the method GrayscaleLayer and this needs a source/lazyflowsource. Maybe you add a comment on how you use the addGrayscaleLayer and replace the one I used in the code. 
So that next programmers easily find the function and know how to use it. 

I guess, _create_single_color_layer_from_slot and create_8bit_ordered_random_colortable_layer_from_slot are not needed either. but the one with transparent zero is used.

PR #1445, opened 2017-03-27
Added option to use simpler configuration to get features of relabeled mergers for animal tracking

JaimeIvanCervantes:
Added option to use simple  configuration to get features of relabeled mergers for animal tracking, instead of using `opRelabeledMergerFeatureExtraction`.

@chaubold I'm using your operator on the `Tracking Workflow` (cell tracking), what do you think?


JaimeIvanCervantes:
I'll merged these changes, but we have to address issue: #1447

PR #1449, opened 2017-03-29
Modified run_each_unit_test script to not stop on first failure

k-dominik:
currently, `run_each_unit_test.sh` will exit on the first failed test.

A code change might effect multiple tests. All those tests offer the possibility to track down the bug/error, so testing should not stop after encountering the first fail.
With this commit, the script will continue to run each test, accumulate the number of test runs that fail (NOT the number of single test cases, but the number of nosetest executions that return with with an `errorcode != 0`. At the end of the script, the failing tests are printed to `stdout`.

### DIFF
--- a/tests/run_each_unit_test.sh
+++ b/tests/run_each_unit_test.sh
@@ -24,6 +24,9 @@ SKIP_GUI_TESTS=${SKIP_GUI_TESTS-0}
 
 SKIP_UNTIL=${2-"RUN_ALL"}
 
+FAILURES=0
+BROKEN=()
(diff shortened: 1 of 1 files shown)


akreshuk:
how about adding this as a command line option for run_each_unit_test? Like --all or something.

k-dominik:
ready to be merged

didn't go for the command line flag, but you can simply start the test with

```
CONTINUE_ON_FAILURE=1 run_each_unit_test.sh
```

to get the new behaviour

Issue #1451, opened 2017-03-31
Tracking - NotImplementedError in get_exporting_operator


Confusezius:
```
ERROR 2017-03-31 11:06:52,173 excepthooks 19717 140118287963968 Traceback (most recent call last):
  File "/home/kroth/miniconda2/envs/ilastik-devel/ilastik-meta/ilastik/ilastik/utility/exportingOperator.py", line 151, in show_export_dialog
    op = self.get_exporting_operator()
  File "/home/kroth/miniconda2/envs/ilastik-devel/ilastik-meta/ilastik/ilastik/utility/exportingOperator.py", line 134, in get_exporting_operator
    raise NotImplementedError
NotImplementedError

```

LetiP:
So I guess you used:
![screenshot from 2017-03-31 11-27-32](https://cloud.githubusercontent.com/assets/16118202/24544844/1fb22140-1605-11e7-9751-ca31e0041c68.png)
I do not know yet, what this button is intended for.

But for tracking result export, you should click in this area:
![screenshot from 2017-03-31 11-26-52](https://cloud.githubusercontent.com/assets/16118202/24544868/3838ac02-1605-11e7-9230-d63d6beeb6b9.png)

:)
What t

chaubold:
Thanks @LetiP for finding that relict. That was a very old implementation of exporting a CSV table of features (even before it ever made it into the **Tracking Export Applet**). So that button should just not be there at all.

@Confusezius yup, what @LetiP said 😉

k-dominik:
menu entry is gone..

PR #1452, opened 2017-03-31
Remove Export Tracking Information Relict from menu

LetiP:
This should fix Issue #1451 Tracking - NotImplementedError in get_exporting_operator

### DIFF
--- a/ilastik/applets/tracking/base/trackingBaseDataExportGui.py
+++ b/ilastik/applets/tracking/base/trackingBaseDataExportGui.py
@@ -188,33 +188,6 @@ class TrackingBaseDataExportGui( DataExportGui, ExportingGui ):
         # TODO: remove once tracking is hytra-only
         self._default_export_filename = filename
 
-    # override this ExportingOperator function so that we can pass the default filename
-    def show_export_dialog(self):
-        """
-        Shows the ExportObjectInfoDialog and calls the operators export_object_data method
-        """
-        # TODO: remove once tracking is hytra-only
-        # Late imports here, so we don't accidentally import PyQt during headless mode.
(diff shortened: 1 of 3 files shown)


ilastik/applets/tracking/conservation/conservationTrackingGui.py
```diff
@@ -349,16 +349,6 @@ def _handle_failure( exc, exc_info ):
         req.notify_finished( _handle_finished )
         req.submit()
 
-    def menus(self):
-        menus = super( ConservationTrackingGui, self ).menus()
-        
-        m = QtGui.QMenu("&Export", self.volumeEditorWidget)
-        m.addAction("Export Tracking Information").triggered.connect(self.show_export_dialog)
```
chaubold: I think we can safely drop this menu item on all versions of ilastik, so I agree, this is the fix to the bug #1451

ilastik/applets/tracking/base/trackingBaseDataExportGui.py
```diff
@@ -188,33 +188,6 @@ def set_default_export_filename(self, filename):
         # TODO: remove once tracking is hytra-only
         self._default_export_filename = filename
 
-    # override this ExportingOperator function so that we can pass the default filename
-    def show_export_dialog(self):
-        """
-        Shows the ExportObjectInfoDialog and calls the operators export_object_data method
-        """
-        # TODO: remove once tracking is hytra-only
```
chaubold: Thanks a lot for attempting this fix! There's just one problem: This code must still work with pgmlink so that windows users still get a working tracking workflow. This is why the **TODOs** are in there and the functions still exist.

So please do not remove them yet. (You can try that by installing the `pgmlink` package in your environment and temporarily moving the `hytra` folder from your conda environment's `site-packages` folder.)

Issue #1432, opened 2017-03-20
Error when starting feature selection in object classification


alexanderbrown:
Hi, 

I'm running Ilastik 1.2.0. When I try to select features in object classification I get the following error message

`Python argument types in
    vigra.analysis.extract2DConvexHullFeatures(VigraArray)
did not match C++ signature:
    extract2DConvexHullFeatures(vigra::NumpyArray<2u, vigra::Singleband<unsigned int>, vigra::StridedArrayTag> labels, boost::python::api::object ignoreLabel=None, bool list_features_only=False)
 (Advanced information about this error may be found in the log file: /home/abrown/ilastik_log.txt)`

Is this related to an undocumented requirement or is it a bug?
I can send the log file if that's useful.
Running Ubuntu 16.04

msschwartz21:
I am also experiencing this problem on my computer running Windows 10.

akreshuk:
We are planning a new release later this week, this problem should be gone.

imagejan:
Is there a new release available somewhere already? We just stumbled upon this very same issue as well 🙂

chaubold:
We are currently testing our release candidate, it will be out *very* soon!

PR #1465, opened 2017-04-12
TrackingFeatureExtraction: do no overwrite input slot value...

chaubold:
... depending on data shape. Adresses #1432 in tracking workflows.

Overwriting the list of pre-selected features for tracking depending on the raw shape is too restrictive. If the user first loads a 3D dataset but then, by adjusting the axestags, changes that to 2D+t, the 2D features should become available again. If the feature slot value is overwritten when loading the 3D data but the original value is lost, this cannot be recovered. 

Hence with this (admittedly hacky) PR the initial value is stored as member variable of the operator, and the input slot is overwritten with the correct value.

This does fix the problem in #1432, but it seems wrong to overwrite the `FeatureNamesVigra` input slot value (which is also connected to input slots of other operators..?). @stuarteberg what is the suggested solution, to create a `FilteredFeatureNamesVigra` **output** slot which is then connected to the other operators?

Btw, the Object Classification Workflow does **not** suffer from that problem, because it does not have a hard-coded preselection of features that would have to be filtered according to the dataset dimensionality.

Thanks to @k-dominik for debugging this with me.

### DIFF
--- a/ilastik/applets/trackingFeatureExtraction/opTrackingFeatureExtraction.py
+++ b/ilastik/applets/trackingFeatureExtraction/opTrackingFeatureExtraction.py
@@ -168,7 +168,7 @@ class OpTrackingFeatureExtraction(Operator):
         
     def __init__(self, parent):
         super(OpTrackingFeatureExtraction, self).__init__(parent)
-        
+        self._default_features = None
         # internal operators
         self._objectExtraction = OpObjectExtraction(parent=self)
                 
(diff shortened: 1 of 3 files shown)


stuarteberg:
>This does fix the problem in #1432, but it seems wrong to overwrite the `FeatureNamesVigra` input slot value (which is also connected to input slots of other operators..?).

You're right, it's generally not recommended for an operator to configure it's own input slots.  But given that this operator already behaved that way, think your fix is reasonable.  I think it's okay for such configuration to happen in a `notifyReady` handler.  It's worse to put it in `setupOutputs()`.

>what is the suggested solution, to create a FilteredFeatureNamesVigra output slot which is then connected to the other operators?

I don't think that would be right, either.  The "other operators" in this case are 'children' operators, so it isn't right for them to receive inputs from the output of the parent operator.

Perhaps another solution would be to just allow `FeatureNamesVigra` to be optional, in which case reasonable defaults are chosen if the slot isn't `ready()`.  If the slot has an explicit value that is incompatible with the volume axes, then a `DatasetConstraintError` could be raised, telling the user to select different features first.  That's a little annoying to the user, though.

Anyway, since you say this works, I'm merging as-is.

PR #1467, opened 2017-04-13
Let DataExportApplets specify whether they provide checkOverwriteFiles

### LINKED ISSUE #1466: Error with checkOverwriteFiles on dataExportGui
I created a new `pixel classification` project and as I was trying to export the pixel probabilities, I got this error:

```
ERROR 2017-04-12 21:18:49,871 log_exception 23195 139669920544512 Traceback (most recent call last):
  File "/opt/local/miniconda/envs/ilastik-tracking/ilastik-meta/ilastik/ilastik/applets/dataExport/dataExportGui.py", line 396, in exportSlots
    exportStatus = self.parentApplet.post_process_lane_export(lane_index, checkOverwriteFiles=True)
TypeError: post_process_lane_export() got an unexpected keyword argument 'checkOverwriteFiles'
```
@chaubold and @LetiP do you think this might be related to the plugin system? Maybe you added the `checkOverwriteFiles` parameter only to the overrided `post_process_lane_export` function of the tracking operator, and you forgot to add it to the rest of the workflows.

chaubold:
... which fixes #1466.

An alternative option would have been to change all signatures to e.g. `post_process_lane_export(self, lane_index, *args, **kwargs)`, but I think the way this PR handles it is more explicit and thus less prone to mistakes as the one we introduced, leading to #1466.

### DIFF
--- a/ilastik/applets/dataExport/dataExportApplet.py
+++ b/ilastik/applets/dataExport/dataExportApplet.py
@@ -76,6 +76,18 @@ class DataExportApplet( Applet ):
             self._gui = DataExportGui( self, self.topLevelOperator )
         return self._gui
 
+    @staticmethod
+    def postprocessCanCheckForExistingFiles():
+        '''
+        While exporting, we can check whether files would be overwritten.
+        This is handled by an additional parameter to post_process_lane_export called "checkOverwriteFiles",
+        and that method should return True if the export files do NOT exist yet.
+
+        By default we do not want to check for existing files,
+        and most operators don't provide checkOverwriteFiles, so return False.
+        '''
+        return False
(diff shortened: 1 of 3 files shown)


JaimeIvanCervantes:
Alternatively we can changes the signature of all the `post_process_lane_export` overrided functions for all workflows, and adopt the new parameter, since we would want to check for overwrite in all workflows anyways right?

chaubold:
Exactly. Or we could introduce it one by one... But that's also possible with the check that we have in there right now.

JaimeIvanCervantes:
Ok perfect, you probably can just go ahead and merge it.

Issue #1473, opened 2017-04-19
Thresholding does not re-normalize prob maps anymore


akreshuk:
Before the thresholding applet refactoring, prob maps could be provided both as float32, [0, 1] and as uint8, [0, 255] via an admittedly ugly hack. It looks like now only the [0,1] version works.

akreshuk:
Fixed in #1474

PR #1474, opened 2017-04-19
Polishing the new thresholding applet

akreshuk:
- Add prob normalization (copied from multicut applet) and close issue #1473 
- Cosmetic improvements in the ui file
- Hide the word "Final" when single threshold is used


stuarteberg:
LGTM

Issue #1494, opened 2017-05-12
Structured Learning cannot be run in headless mode due to QtGui import


chaubold:
We should only import the **GUI** part of the progress bar for tracking when we are not running in headless mode. The easy way to reproduce that error is to run the following:

```
nosetests test_applets/structuredLearningTracking/testStructuredLearningTrackingHytraWithDivisions.py
```

Gives that error:

```pytb
======================================================================
ERROR: testStructuredLearningTrackingHytraWithDivisions.TestStructuredLearningTrackingHeadless.testStructuredLearningTrackingHeadless
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/chaubold/opt/miniconda/envs/ilastik12/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/chaubold/opt/miniconda/envs/ilastik12/ilastik-meta/lazyflow/lazyflow/utility/timer.py", line 142, in wrapper
    return func(*args, **kwargs)
  File "/Users/chaubold/opt/miniconda/envs/ilastik12/ilastik-meta/ilastik/tests/test_applets/structuredLearningTracking/testStructuredLearningTrackingHytraWithDivisions.py", line 87, in testStructuredLearningTrackingHeadless
    import ilastik.workflows.tracking.structured
  File "/Users/chaubold/opt/miniconda/envs/ilastik12/ilastik-meta/ilastik/ilastik/workflows/__init__.py", line 59, in <module>
    import tracking.conservation
  File "/Users/chaubold/opt/miniconda/envs/ilastik12/ilastik-meta/ilastik/ilastik/workflows/tracking/conservation/__init__.py", line 4, in <module>
    from conservationTrackingWorkflow import ConservationTrackingWorkflowFromPrediction, ConservationTrackingWorkflowFromBinary
  File "/Users/chaubold/opt/miniconda/envs/ilastik12/ilastik-meta/ilastik/ilastik/workflows/tracking/conservation/conservationTrackingWorkflow.py", line 6, in <module>
    from ilastik.applets.tracking.conservation.conservationTrackingApplet import ConservationTrackingApplet
  File "/Users/chaubold/opt/miniconda/envs/ilastik12/ilastik-meta/ilastik/ilastik/applets/tracking/conservation/conservationTrackingApplet.py", line 14, in <module>
    from ilastik.applets.tracking.conservation.opConservationTracking import OpConservationTracking
[...]

PR #1495, opened 2017-05-12
tests in learning

ilastikdev:
check for ILP solver added (needed for learning)
non-gui progress bar methods moved to ilastik/utility/progress.py


ilastik/utility/gui/progress.py
```diff
@@ -4,76 +4,7 @@
 from PyQt4.QtCore import *
 from PyQt4.QtGui import *
```
akreshuk: Could you please not import *?

chaubold:
The error from #1494 is fixed for me.

Issue #96, opened 2012-11-28
Label class list widget improvements


stuarteberg:
Janelia users suggested:
- Respond to the del key for deleting labels
- Provide a context menu that includes "erase all" as an option.
- 'x' button disappears on windows in live mode

akreshuk:
Right-clicking on the color, still displays "None" instead of the label name. Clear all is provided. "x" button should disappear in live mode. Del key is too ambiguous for a shortcut.

k-dominik:
sorry for the cluttered references here ;) only the PR #1516 is relevant

akreshuk:
Final point about label names when right-clicking on color solved in #1516

PR #1516, opened 2017-05-24
LabelListView: show label name in context menu regardless of selected column

k-dominik:
Fixes issue #96


akreshuk:
Good start of the day!

PR #1521, opened 2017-05-26
minor improvements to learning tests

ilastikdev:
... and conservation tracking test

### DIFF
--- a/ilastik/applets/tracking/base/opTrackingBase.py
+++ b/ilastik/applets/tracking/base/opTrackingBase.py
@@ -644,9 +644,11 @@ class OpTrackingBase(Operator, ExportingOperator):
             else:
                 logger.debug("No divisions occurred. Division Table will not be exported!")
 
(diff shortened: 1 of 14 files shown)


chaubold:
@ilastikdev please list all the important changes when creating a PR (or even better, create separate PRs for different issues). Only when looking at the changes I noticed that this PR also fixes the problem of the progress bar not showing up in *Tracking with Learning*.

ilastikdev:
progress bar Qt import error you are reffering to was already fixed in PR #1509

chaubold:
But what is https://github.com/ilastik/ilastik/pull/1521/commits/87fd86a2babe2d3a0b2576c2f9b4420628d74340 doing then? That's exactly why I asked you do document all changes.

ilastikdev:
it makes sure that these calls execute in gui threads
see https://github.com/ilastik/ilastik/blob/master/ilastik/utility/gui/threadRouter.py

PR #1526, opened 2017-05-31
stack along channels UI-part.

### LINKED ISSUE #1202: Stack across c should be an option too
We have those helpful radio buttons below the stack file selection. We should also make it possible to stack by channel.

FynnBe:
Fixes #1202


FynnBe:
while writing the tests I noticed that the stacking along 'c' is still buggy...
Could I rewrite the OpStackLoader in a way that it returns an almost arbitrary axis order? I think this would simplify the problem a whole lot and keeping specific axis orders for its output will soon be redundant anyway...

stuarteberg:
Sure, go ahead.  It's not obvious to me how adding flexibility to `OpStackLoader`'s output axes will simplify the problem, but maybe it will become clear when I review the PR.

PR #1531, opened 2017-06-01
removing crops - training for tracking in the the whole data set can …

ilastikdev:
…be sparse

(tracks should still be continuous and mergers appear/disappear together)

this will not work with pgmlink because pgmlink requires full training per model (crop)
so merging to hytra-learning (or master) should wait until we can run hytra on windows


ilastikdev:
should be merged at the same time as
https://github.com/ilastik/volumina/pull/183

ilastikdev:
superseeded by the qt5py3 version

PR #1532, opened 2017-06-01
Parallel global features

### LINKED ISSUE #347: Parallelize global feature computation
... with a RequestPool. Clamp the default vigra features in there as well.

akreshuk:
Compute global object features in parallel (one plugin per thread). Fixes #347 .


chaubold:
[LGTM](http://www.lgtm.in/i/Yu8lDUqWp)

PR #1537, opened 2017-06-08
Hytra learning

ilastikdev:
fixes the structured learning test


ilastikdev:
from dominik:

"
Hi Janez,

I have merged your most recent commit with the modified ilp file and ran the test 40 
times without any error. I think we can assume now it works again.

Cool, that you fixed it!

Cheers
Dominik
"

Issue #1513, opened 2017-05-23
Change/enhance the font of the applet name


akreshuk:
We should make the applets stand out more in the vertical list of widgets

k-dominik:
I would argue that https://github.com/ilastik/ilastik/pull/1900 fixed that. Closing.

PR #1545, opened 2017-06-13
highlight selected applet drawer item

k-dominik:
addresses #1513
In ilastik, the currently selected (workflow) item in the app drawer doesn't stand out particularly well:

ilastik-drawer-old

now the title of the current drawer is highlighted in bold font as follows:

ilastik-drawer-new

### DIFF
--- a/ilastik/shell/gui/ilastik-style.qss
+++ b/ilastik/shell/gui/ilastik-style.qss
@@ -14,7 +14,7 @@ QToolBox::tab {
 }
 QToolBox::tab:selected {
     font: bold;
-    border-bottom: 0px;
+    border-bottom: 1px solid darkgray;
     background-color: -1;
 }
 /********************************************
--- a/ilastik/shell/gui/ilastik-style.qss
+++ b/ilastik/shell/gui/ilastik-style.qss
@@ -15,9 +15,7 @@ QToolBox::tab {
 QToolBox::tab:selected {
     font: bold;
     border-bottom: 0px;
-}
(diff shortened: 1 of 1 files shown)


akreshuk:
Looks much better! Do you think this could be extended to fix #1513?

k-dominik:
ok, this should be ready to be merged, looks like this on my machine:

next-try

k-dominik:
latest update, added a thin line below the active applet heading:

thinline

k-dominik:
@akreshuk @stuarteberg @chaubold : any more thoughts on this?

akreshuk:
Should be merged as far as I'm concerned

k-dominik:
hmmm, or should we wait for this until the master->py3 switch? Making it look different after the switch also has a nice ring to it.

chaubold:
I would rather not introduce new _features_ on master anymore, it was supposed to be only a few bug fixes before we release the last Py 2.7 version.

This doesn't mean I don't like the improved highlighting of the current applet, I'd just postpone it to the Py3 switch.

chaubold:
I just tried this on macOS and am not so convinced about the white backgrounds...

I guess I'd prefer for bold font of the selected applet only, or additionally a different color to highlight the active one, but not changing the color of the other tabs...

Before:
screen shot 2017-08-24 at 13 53 08

After:
screen shot 2017-08-24 at 13 51 57

k-dominik:
superseeded by https://github.com/ilastik/ilastik/pull/1900

Issue #1582, opened 2017-08-28
Remove old pgmlink-based tracking files.


chaubold:
Now that we've switched to the Python 3 code base, windows will also use the new hytra-based tracking pipeline.

Hence we can get rid of all pgmlink legacy code.

PR #1583, opened 2017-08-28
Tracking: Remove pgmlink

chaubold:
Now that we've switched to the Python 3 code base, windows also uses the new hytra-based tracking pipeline. Hence, in this PR, a lot of legacy code is removed and some minor cleanups are performed.

**References:**

* addresses #1582
* includes (and hence supersedes) #1452


chaubold:
Looks like the layers showing the tracking result are not populated properly. Don't merge yet...

chaubold:
Should be good now. @akreshuk would you mind taking a look?

akreshuk:
Good to see chaingraph tracking gone, along with all the backend ifs. It's hard for me to see if something could go very wrong after this, I hope there are enough tests in tracking by now! Should I click through it once to be sure?

chaubold:
Well we have no tracking GUI tests, so it would be great if you could click through it, yes. Thanks

akreshuk:
Immediately found an error in thresholding, will fix that and come back

akreshuk:
Export plugin dialog on my screen:
image
Can we possible make this line not so narrow?

akreshuk:
I changed some parameter before exporting, so it had to recompute the tracking to export. For some reason, the terminal progress updates stop at 92%:
image
The progress bar in the gui has disappeared already and the machine is not doing anything, so i think it's done.

akreshuk:
with the commit above i can actually track and export results, so works for me (and mitocheck)

chaubold:
> Can we possible make this line not so narrow?

@akreshuk I made it a little nicer, could you please check whether it looks ok on your screen, too?

akreshuk:
Looks good now. I also tried batch processing for the sake of completeness, it works fine, but the terminal progress reporting also stops at 92%. And before doing anything, it informs us that

QObject::connect: Cannot queue arguments of type QVector<int>
(Make sure QVector<int> is registered using qRegisterMetaType().)>

This could be an overall problem in batch processing, I haven't checked.

akreshuk:
I tried with drosophila now and 3D also seems to be fine, so looks good to me.
One very nitpicky remark: after the last thing is done in the command line progress bar (computing lineage), the gui progress bar in the export applet keeps doing something for a while more. Do we need another entry into the CLI progress bar? Or, could we have some "Done!" print after the last point so that it's clear nothing else is coming?

chaubold:
Thanks for checking!

For the export: I think what it's doing after the GUI progress bar is hidden is to actually save the file to disk, and that's something we can't hook into from within the "tracking" operator that populates the progress bar.

PR #1592, opened 2017-09-19
No crops qt5py3

ilastikdev:
removing crops - training for tracking in the the whole data set can be sparse
(tracks should still be continuous and mergers appear/disappear together)


ilastikdev:
needs:
https://github.com/ilastik/volumina/pull/183

PR #1600, opened 2017-09-26
uncertainty layer added to object classification

ilastikdev:
uncertainty of an object =

0 if numClasses <=1
1-(maxProb-secondMaxProb) otherwise

background is also an object - always with uncertainty 0


akreshuk:
Rebased on master and added (trivial) tests. LGTM now.

Issue #956, opened 2014-07-11
Counting color overflow in LabelPreview layer


burgerdev:
Steps to reproduce:
- open counting, click through obvious applets
- in counting applet, switch too a large sigma (e.g. 20)
- click a few times into a small region of the image (see attached)
- result: color overflow

When trying to export this layer, the datatype is shown as float64, but the drange is [0.0, 255.0]. Is this correct?

![snapshot1](https://cloud.githubusercontent.com/assets/2814259/3555352/415b5af4-0916-11e4-8d69-291c7ff44afa.png)

akreshuk:
I tried this and also got a strange image - the gaussian is not drawn outside the original tile
![ilastik - -home-akreshuk-myproject ilp - cell density counting_003](https://cloud.githubusercontent.com/assets/901064/26394687/090a678e-406e-11e7-8bc6-b53f8039714d.png)

akreshuk:
And the mapping from floats in the actual gaussians to the displayed values is also wrong

k-dominik:
![counting_q4](https://user-images.githubusercontent.com/24434157/30481774-9d8d7dfe-9a20-11e7-9b7c-efdaa8a1dc78.png)
![counting_q3](https://user-images.githubusercontent.com/24434157/30481782-a1e0f552-9a20-11e7-8319-d1c58a1cff80.png)

just to pitch in some more examples - and confirm that this is still an issue

k-dominik:
So what we are seeing with the overlapping gaussians are definitely overflows. The gaussians are not normalized (correctly), but only with respect to a single one here in `opCounting.OpUpperBound`:

https://github.com/ilastik/ilastik/blob/c170860c3a544bb546a6a0889c72ae175f8f365d/ilastik/applets/counting/opCounting.py#L140-L145

PR #1603, opened 2017-09-26
Fix counting gui test freeze

k-dominik:
The main commit is "solves" the problem of the gui test halting is https://github.com/ilastik/ilastik/commit/533006600dc185135cd7c6473ecfdb952a79bf09 . Unfortunately it is not a real solution, but rather a mere hack: Due to not yet clearly identified problems in lazyflow, `OpCacheFixer` will block the main thread if `OpCacheFixer.fixAtCurrent` is changed, while at the same time someone pulls from its output.

In any case, this will let the tests run through for now.
Note, that there will still be an error in the tests, which I will address in a different PR (out of scope).

This PR includes a few minor changes that I have introduced to make it more like pixelClassification.

Summary of other changes:

* didn't see a reason why `OpDenseLabelArray` was used, so changed it to `OpCompressedLabelArray`, as in pixel classification
* there was a problem serializing to hdf5 with pickled objects. these should be saved as `numpy.void`
* the callback added in https://github.com/ilastik/ilastik/commit/be2fb4c0528085a951ab7a62908cd2eacc08590d, works around the a part of the problem mentioned in https://github.com/ilastik/ilastik/issues/956, where the gaussian would only be drawn in the current tile and not across tile borders

### DIFF
--- a/ilastik/applets/counting/countingGui.py
+++ b/ilastik/applets/counting/countingGui.py
@@ -127,8 +127,8 @@ class CountingGui(LabelingGui):
         labelSlots = LabelingGui.LabelingSlots()
         labelSlots.labelInput = topLevelOperatorView.LabelInputs
         labelSlots.labelOutput = topLevelOperatorView.LabelImages
-        labelSlots.labelEraserValue = topLevelOperatorView.opLabelPipeline.opLabelArray.EraserLabelValue
-        labelSlots.labelDelete = topLevelOperatorView.opLabelPipeline.opLabelArray.DeleteLabel
+        labelSlots.labelEraserValue = topLevelOperatorView.opLabelPipeline.opLabelArray.eraser
+        labelSlots.labelDelete = topLevelOperatorView.opLabelPipeline.opLabelArray.deleteLabel
         labelSlots.maxLabelValue = topLevelOperatorView.MaxLabelValue
         labelSlots.labelNames = topLevelOperatorView.LabelNames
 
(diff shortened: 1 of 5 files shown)


k-dominik:
Ok, hang on, I have pushed a fix for the failing test as well, because it's such a tiny one.

I think this test has never really worked (have confirmed this in python2 as well). The only difference, is that, with the cache in place, those values were never evaluated, hence `0.0 == 0.0`. Without the cache, however the actual computation is triggered. The box label is rounded to 1 digit precision. So in the test, one cannot compare that value to the output of the operator.

k-dominik:
Yeah, the discussions with @akreshuk were extremely helpful!

Am 26.09.2017 5:34 nachm. schrieb "Carsten Haubold" <
notifications@github.com>:

akreshuk:
Oh the pleasure of the green check!

stuarteberg:
>Oh the pleasure of the green check!

+1

>didn't see a reason why OpDenseLabelArray was used, so changed it to OpCompressedLabelArray, as in pixel classification

Sorry about that.  The reason it exists is that the `OpCompressedLabelArray` adds a little lag to the UI since labels need to be compressed/decompressed, and it uses HDF5 to do it instead of vigra.  Since Counting only supports 2D data, simply storing a dense array of labels is not RAM-intensive, and it is faster.

BUT... I completely agree with your decision to make the switch.  It's better to just be consistent with the other workflows.  And if there weren't proper comments explaining the purpose of the strange operator, then it deserves to die. :-)

k-dominik:
Thanks @stuarteberg for the insight. And good to know that we can expect a little less performance when annotating. If that gets too annoying, we at least know what to do.

k-dominik:
Okay, I'll merge it in then.

PR #1609, opened 2017-10-06
Adapted appveyor.yml for python 3 testing

k-dominik:
I have updated the appveyor config to use our latest recipe based on python 3.

There is a failure on appveyor is a real failure and should be fixed. This PR re-enables appveyor testing and should be merged.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,29 +1,42 @@
 clone_folder: c:\projects\ilastik
 
-cache:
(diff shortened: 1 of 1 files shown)


chaubold:
Tests are failing because `networkx` is too new. We should restrict its version until the tracking side was update for the networkx >= 2.0...

chaubold:
Nice, the tests are running properly now, the fix for the failing tracking test is coming in #1612.

PR #1611, opened 2017-10-13
add option 'c' to stackFileSelectionWidget

FynnBe:


FynnBe:
As discussed before the rules of how axes are interpreted are not trivial yet to add a catchy comment. I will attend to this while simplifying axis order handling in general. For now I think we should merge this (needs lazyflow part to make sense, this is pure UI)

chaubold:
I assume that makes #1526 obsolete?

PR #1615, opened 2017-10-17
2D and 3D convex Hull features

Beinabih:
added a 3D case for the convex hull features thats shows up in the object feature selection


chaubold:
Thanks for the PR, Jonas! The code looks good, but as you see in the github status above, it cannot be merged into master due to some conflicts - because most certainly `ilastik/master` has changed since you started. Please merge the current `ilastik/master` into your branch and resolve the conflicts.

chaubold:
Oh, one more task before we merge this: Could you please add a test in `ilastik/tests/objectClassification` that checks that for a 2D input image, the 2D (but not the 3D) convex hull features are available and can be computed, and similarly the 3D but no 2D features for 3D images.

tests/test_applets/objectClassification/testConvexHull.py
```diff
@@ -0,0 +1,126 @@
+import numpy as np
+import vigra
+import unittest
+from lazyflow.graph import Graph
+from ilastik.plugins_default.vigra_objfeats_convex_hull3D import VigraConvexHullObjFeats3D
+from ilastik.plugins_default.vigra_objfeats_convex_hull import VigraConvexHullObjFeats
```
chaubold: Those imports are not necessary for anything below, right?

akreshuk:
What would now happen with the old projects which had (2D) convex hull features already selected? Like all tracking projects?

chaubold:
Old projects that use 3D data but still wrongly used the 2D convex hull features (because we somehow also allowed those to be used in 3D) and have a trained classifier might not work out of the box, that's right. 

I created a ticket for this: #1638

PR #1630, opened 2017-10-26
Remove all evals

k-dominik:
In this PR I removed all `eval` occurences in the code.

Rationale: `eval` is just not that safe and not really needed for our purposes.

This PR should be thoroughly reviewed, as I am not sure everything is tested - so a second pair of eyes would be nice.

* Conversion from strings representing lists is now mostly handled while parsing the arguments and based on `json.loads`-> `ilastik.utility.commandLineProcessing.ParseListFromString`
* in the two scripts residing in `/bin` I have changed the behaviour slightly. Single sumbers are not allowed anymore for `scale` and `offset`. Also I have made optional arguments `required=False`, rather than `nargs='?', default=None`. Behaviour should be the same.

### DIFF
--- a/bin/downsample_pointcloud.py
+++ b/bin/downsample_pointcloud.py
@@ -6,6 +6,8 @@ import csv
 
 import numpy
 
+from ilastik.utility.commandLineProcessing import ParseListFromString
+
 import logging
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


k-dominik:
Hmmm that appveyor fail looks puzzling ...

k-dominik:
I think the failing tests might be related to a recent dependency upgrade on windows only(?).:
```
33c33
<     ilastikrag:                      0.1.3.post5-py36_0_gaddf55b       ilastik-forge
---
>     ilastikrag:                      0.1.3.post4-py36_0_gd026251       ilastik-forge
36d35
<     intel-openmp:                    2018.0.0-hcd89f80_7                            
47c46
<     mkl:                             2017.0.4-h6d528fc_0                            
---
>     mkl:  
```

chaubold:
Hmm weird. I don't think that `ilastikrag` is to blame here, as the only added commit (`post5` instead of `post4` was Anna's [fix to iterate over a dict](https://github.com/ilastik/ilastikrag/commit/addf55bab031af910aadddeb28c94206cfaa1932)).

`MKL` on the other hand could mess up everything, as it is the fast math implementation backing `numpy` on windows. On linux and mac we explicitly _enforce_ a `numpy` version built with `openblas` instead, because the `conda-forge` MKL version conflicts with one used inside `CPLEX` leading to immediate crashes if you import both. On Windows that was no problem up to now, and there is no `openblas` based `numpy` on conda-forge yet IIRC...

chaubold:
I've hardcoded the required MKL version for the time being, see https://github.com/ilastik/ilastik-build-conda/commit/4c05ccb90464158f42f3b9541e10c161b5895daf.

@k-dominik, @FynnBe and @beinabih, if you trigger AppVeyor rebuilds of your respective PRs, they should turn green now!

k-dominik:
ughh! good point. I don't think this works right now, but sure is an easy fix...

k-dominik:
@chaubold seems to be ready now. Thanks for the excellent suggestion

chaubold:
You're welcome!

PS: AppVeyor still needs a little convincing from time to time (had to trigger a rebuild because it got stuck for 1h again)...

Issue #1631, opened 2017-10-27
autocontext: suggestFeatures fails to execute @ Stage 2


ilastikdev:
```
  File "/home/jales/software/miniconda/envs/p3-ilastik/ilastik-meta/ilastik/ilastik/applets/pixelClassification/pixelClassificationGui.py", line 489, in show_feature_selection_dialog
    self.featSelDlg.exec_()
  File "/home/jales/software/miniconda/envs/p3-ilastik/ilastik-meta/ilastik/ilastik/applets/pixelClassification/FeatureSelectionDialog.py", line 238, in exec_
    self._add_grayscale_layer(self.raw_xy_slice, "raw_data", True)
  File "/home/jales/software/miniconda/envs/p3-ilastik/ilastik-meta/ilastik/ilastik/applets/pixelClassification/FeatureSelectionDialog.py", line 554, in _add_grayscale_layer
    self.editor.dataShape = list(data_shape)
  File "/home/jales/software/miniconda/envs/p3-ilastik/ilastik-meta/volumina/volumina/volumeEditor.py", line 136, in dataShape
    self.parent.volumeEditorWidget.quadview.ensureMinimized(i)
AttributeError: 'FeatureSelectionDialog' object has no attribute 'volumeEditorWidget'

```

suggestFeatures does execute @ Stage 1

ilastikdev:
fixed in 
https://github.com/ilastik/ilastik/pull/1632

chaubold:
I cannot reproduce that error. I created a new autocontext workflow and ran feature selection in both stages and there was no problem at all. Is there anything special that I need to do?

I assume your `volumina` branch is up to date as it hasn't changed for over a month.

ilastikdev:
Run a 3d+t dataset

chaubold:
I tried 2D, 3D and 3D+t now, and it works perfectly everywhere, even without your fix...

ilastikdev:
Lines that fail are 135-136:

            for i in range(3):
                self.parent.volumeEditorWidget.quadview.ensureMinimized(i)

in volumeEditor.py

which clearly fail if called from FeatureselectionDialog.py

because of conflicting

            self.viewer = volumeEditorWidget.VolumeEditorWidget()

The error is triggered when dataShape is called.

chaubold:
Yes, I saw that in your traceback. But it just doesn't occur for me, so I was wondering when you ran into this issue, or basically: what did you have to do as a user so that `dataShape` is called?

[...]

PR #1632, opened 2017-10-27
non-standard ilastik variable name (viewer) used for volumeEditorWidget

ilastikdev:
This is a typical example why we should have (in OO fashion) get functions,
in this case, getVolumeEditorWidget()
fix is done in line with an ilastik coding practice


ilastikdev:
fixes https://github.com/ilastik/ilastik/issues/1631

PR #1636, opened 2017-11-06
Test all workflows in headless mode with blank project files

k-dominik:
The code in this PR tests all workflows that are discovered with `ilastik.workflows.getAvailableWorkflows()`.

The tests are very basic:
* a blank project file is created for each workflow using `ilastik.shell.projectManager.ProjectManager`
* this project file is then opened in the headless shell.

Any unexpected errors (uncaught exceptions) will make this test fail.

Furthermore, project file creation via the command line is also tested.

Currently, the ObjectClassificationWorkflow will complain about `Couldn't obtain a classifier from your project file` and `Was not able to understand the batch mode command-line arguments.`. Those messages are properly caught errors, and should be okay for our purpose.

Open to suggestions to improve this test @akreshuk @chaubold

### DIFF
--- a/tests/test_workflows/test_all/testAllHeadless.py
+++ b/tests/test_workflows/test_all/testAllHeadless.py
@@ -71,8 +71,11 @@ class TestHeadlessWorkflowStartupProjectCreation(object):
 
         # Start up the ilastik.py entry script as if we had launched it from the command line
         parsed_args, workflow_cmdline_args = ilastik_main.parser.parse_known_args()
+
         shell = ilastik_main.main(
             parsed_args=parsed_args, workflow_cmdline_args=workflow_cmdline_args, init_logging=False)
+
+        shell.closeCurrentProject()
 
         # now check if the project file has been created:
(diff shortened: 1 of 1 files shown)


tests/test_workflows/test_all/testAllHeadless.py
```diff
@@ -0,0 +1,123 @@
...
+        """Tests opening project files in headless mode via the command line
+        Args:
+            workflow_class_tuple (tuple): tuple returned from getAvailableWorkflows
+              with (workflow_class, workflow_name, workflow_class.workflowDisplayName)
+        """
+        workflow_class, workflow_name, display_name = workflow_class_tuple
+        logger.debug(f'starting {workflow_name}')
+        project_file = os.path.join(
+            temp_dir,
+            f'test_project_{"_".join(workflow_name.split())}.ilp'
```
chaubold: To prevent mismatches between project filenames, it might be safer to create a method for the filename construction and use it here as well as above in the project creation.

Looks good otherwise!

k-dominik: thanks :)

PR #1639, opened 2017-11-06
Fix tracking bugs

chaubold:
This PR fixes a bunch of tracking bugs and problems that came up when I processed a larger set of data.

* properly close the progress bar window when an exception occurred
* allow batch processing in the _Tracking with Learning_ workflow (was looking for annotations)
* adds per-track color as feature in the MaMuT export
* fixes a bug in appletSerializer that was expecting lists to always only contain strings

### DIFF
--- a/ilastik/plugins_default/tracking_mamut_export.py
+++ b/ilastik/plugins_default/tracking_mamut_export.py
@@ -70,38 +70,64 @@ class TrackingMamutExportFormatPlugin(TrackingExportFormatPlugin):
         graph = hypothesesGraph._graph
 
         features = objectFeaturesSlot([]).wait() # this is a dict of structure: {frame: {category: {featureNames}}}
-        firstPass = True
 
+        # first loop over all nodes to find present features and highest trackId
+        firstNode = True
+        presentTrackIds = set([])
+        for node in graph.nodes_iter():
+            frame, label = node
+            # print(f"Node {node} has value={graph.node[node]['value']} and trackId={graph.node[node]['trackId']}")
+            if graph.node[node]['value'] > 0:
+                t = graph.node[node]['traxel']
+                presentTrackIds.add(int(graph.node[node]['trackId']))
+
+                if firstNode:
+                    radius = 2*features[frame]['Standard Object Features']['RegionRadii'][label, 0]
+                    for category in list(features[frame].keys()):
+                        for key in list(features[frame][category].keys()):
+
+                            if firstNode:
+                                feature_string = key
+                                feature_string = convertKeyName(feature_string)
+                                if len(feature_string) > 15:
+                                    shortname = getShortname(feature_string).replace('_', '')
+                                else:
+                                    shortname = feature_string.replace('_', '')
+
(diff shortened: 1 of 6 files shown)


ilastik/plugins_default/tracking_mamut_export.py
```diff
@@ -111,30 +138,29 @@ def export(self, filename, hypothesesGraph, objectFeaturesSlot, labelImageSlot,
                                 if (np.asarray(features[frame][category][key])).ndim == 2:
                                     for j in range((np.asarray(features[frame][category][key])).shape[1]):
                                         try:
-                                            exists = features[frame][category][key][label, 0]
+                                            _ = features[frame][category][key][label, 0]
                                         except IndexError:
                                             if 'SquaredDistances' in key:
                                                 featureDict[convertKeyName(key) + '_{}'.format(str(j))] = 9999.
```
k-dominik: I know it is not part of your changes, but while being at it.... `covertKeyName(key)` is used 5 times in these ~15 lines. Would make a good candidate to save it in a variable.

I also wonder what would happen if `np.asarray(features[frame][category][key])).ndim > 2`? or is that impossible?

ilastik/plugins_default/tracking_mamut_export.py
```diff
@@ -70,38 +70,65 @@ def export(self, filename, hypothesesGraph, objectFeaturesSlot, labelImageSlot,
...
 
+        # first loop over all nodes to find present features
+        presentTrackIds = set([])
         for node in graph.nodes_iter():
             frame, label = node
+            # print(f"Node {node} has value={graph.node[node]['value']} and trackId={graph.node[node]['trackId']}")
             if graph.node[node]['value'] > 0:
                 t = graph.node[node]['traxel']
 
                 radius = 2*features[frame]['Standard Object Features']['RegionRadii'][label, 0]
```
k-dominik: `radius` is not used in this scope. Also I'm kind of wondering, why radius is multiplied by 2...

ilastik/plugins_default/tracking_mamut_export.py
chaubold: Thanks for the detailed review! You're helping a lot to get our code clean! 😄  
... I guess you've noticed that this was originally a Python 2 script that I've shaped more and more into a Python 3 plugin here, but now it looks much better :) ...

About the `np.array` dimensionality: this is the return value of vigra's object features. Up to now, they have been at most 2D per object, and AFAIK we have many places where this is implicitly assumed. I added a check nevertheless!

chaubold:
Thanks for the detailed review! You're helping a lot to get our code clean! 😄
... I guess you've noticed that this was originally a Python 2 script that I've shaped more and more into a Python 3 plugin here, but now it looks much better :) ...

About the np.array dimensionality: this is the return value of vigra's object features. Up to now, they have been at most 2D per object, and AFAIK we have many places where this is implicitly assumed. I added a check nevertheless!

k-dominik:
LGTM :)

PR #1640, opened 2017-11-07
replicate changes from Niels

akreshuk:
enable the 3d widget in carving


chaubold:
Works fine on Linux and Mac, and at least the 3D plane display works on Windows. Whether meshes work there we'll see once we fixed the `ilastiktools` segfault.

PR #1642, opened 2017-11-09
fix error in counting workflow on box deletion

### LINKED ISSUE #1637: Counting: deleting a box prints error message
Deleting a box in Object counting will print an error message. Ilastik still seems to continue unaffected.

```
ERROR 2017-11-06 09:53:49,510 excepthooks 29576 140167730824960 Unhandled exception in thread: 'MainThread'
ERROR 2017-11-06 09:53:49,511 excepthooks 29576 140167730824960 TypeError: invalid result from BoxListModel.flags(), NoneType cannot be converted to PyQt5.QtCore.ItemFlags in this context
```

Seems to be related to PyQt4 -> PyQt5 transition...

k-dominik:
Summary ;):
* `BoxListModel.flags` returns `Qt.NoItemFlags` per default

I have also looked around all ilastik and befriended repos: we should be fine. Everywhere else there are default cases that return some kind of `Qt.ItemFlags` thing.

Fixes #1637

### DIFF
--- a/ilastik/widgets/boxListModel.py
+++ b/ilastik/widgets/boxListModel.py
@@ -219,6 +219,8 @@ class BoxListModel(ListModel):
                 return Qt.ItemIsEnabled | Qt.ItemIsSelectable
             else:
                 return Qt.NoItemFlags
+        else:
+            return Qt.NoItemFlags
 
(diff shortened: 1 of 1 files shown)


chaubold:
Thanks for the fix and thanks for checking the rest of the codebase!

Issue #1571, opened 2017-08-04
SerialPickledValueSlot vs. SerialPickleableSlot


ilastikdev:
We have these two classes living side by side in the base/appletSerializer.py
Any hidden reasons why we have two?

PR #1650, opened 2017-11-22
`SerialPickledValueSlot` replaced with `SerialPickleableSlot`

k-dominik:
In `appletSerializer.py`, we had two classes that were doing basically the same thing: `SerialPickledValueSlot` and `SerialPickleableSlot`. Both were used in the code-base without an apparent reason. Since `SerialPickleableSlot` was already a bit nicer, I have ditched `SerialPickledValueSlot`. Furthermore, I changed all occurrences of `SerialPickledValueSlot` to `SerialPickleableSlot`.

I have also improved the error handling in `SerialPickleableSlot` which often failed unnecessarily.

### DIFF
--- a/ilastik/applets/base/appletSerializer.py
+++ b/ilastik/applets/base/appletSerializer.py
@@ -842,7 +842,7 @@ class SerialClassifierFactorySlot(SerialSlot):
 
 
 class SerialPickleableSlot(SerialSlot):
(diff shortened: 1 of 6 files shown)


chaubold:
Adresses #1571

k-dominik:
thank you for referring to the corresponding issue. Your memory is beyond comparison. And also thank you for the review!

PR #1651, opened 2017-11-22
Fix: load projects saved with python 3 version

k-dominik:
Cherry picked important commits from https://github.com/ilastik/ilastik/pull/1650

I've tested this on linux with Pixel Classification, Tracking, Carving, Object Classification, Autocontext. Didn't run into any issues - tests also pass.

Addresses #1571

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -14,7 +14,7 @@ install:
   - git clone https://github.com/ilastik/ilastik-meta c:\ilastik\ilastik-meta
   - cd ilastik\ilastik-meta
   - git submodule update --init --recursive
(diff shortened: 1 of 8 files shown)


k-dominik:
Ah, caveat: Carving loading of objects does not work, but this is unrelated, fixed in python 3. Will add this to `python-2-hotfixes`...

k-dominik:
hang on, I was wrong about the tests...

k-dominik:
this PR now also contains fixes for our CI to work with the python-2 stuff again.

k-dominik:
in view of time, I'll merge it in, as its similar to #1650

Issue #1657, opened 2017-11-23
Carving: importing a carving project fails to import objects


k-dominik:
This issue should affect all versions of ilastik

importing a carving project (via `import project`) will always loose all objects that are saved in the source project file.

Reason being that `CarvingSerializer._serializeToHdf5` tries to be smart about saving only "new" objects:

https://github.com/ilastik/ilastik/blob/58d6ec7b6957797dda644cfe5a42f8fb098980a1/ilastik/workflows/carving/carvingSerializer.py#L39-L43

`opCarving._dirtyObjects` is empty, however, after loading a project (first step of import), so no object is saved in the new project file (second step of import)

k-dominik:
k, consider it fixed. I'll open PRs tomorrow after testing it a bit more.

k-dominik:
fixed

PR #1658, opened 2017-11-24
fix carving import

k-dominik:
fixes issue #1657

cherry-picking commit 4aa1269. (from python-2-legacy)

Before serializing the objects, it is first checked, which objects are already present in the h5 file.

All objects in `opCarving._dirtyObjects` are still processed (old behaviour),
but additionally all objects not present in the file are also saved.


chaubold:
Looks good to me, nice and comprehensible with the set logic and comments!

k-dominik:
thx!

PR #1661, opened 2017-11-28
check availability of nifty_with_cplex/gurobi with subprocess

FynnBe:
also check for WITH_CPLEX/WITH_GUROBI flags after import


FynnBe:
thanks for the review and the good suggestion! I guess it was a bit rushed yesterday.

k-dominik:
hey @FynnBe, next time, could you please commit your style cleanups in a separate commit to make it clear what the fix is actually about.

FynnBe:
yes of course, thanks for the hint!

PR #1675, opened 2017-12-06
force allowed axis orders for counting workflow

FynnBe:
allow for higher dimensional input to counting, if higher dimensions have length 1

PR #1676, opened 2017-12-06
CountingSrv: remove regressors that depend on gurobiwrapper

chaubold:
and make RF the default.

@FynnBe already tested this locally and it works on Windows, too


FynnBe:
works fine for me
and btw, the other counting fix is independent of this (https://github.com/ilastik/ilastik/pull/1675)

PR #1680, opened 2017-12-19
silence subprocess output in test imports of nifty

FynnBe:


chaubold:
Thanks for the PR, looks good to me, but I'd like to try on my windows VM later before we merge.

PR #1686, opened 2018-01-10
Disable 3D widget by default. Enable 3D widget for Carving GUI when the data is 3D

wolny:
We need to pass the workflow class name to the VolumeEditor in order to be able to correctly initialise the 3D widget, i.e. we want the 3D widget to be visible only for workflows for certain workflows (currently only the carving workflow (?))

### DIFF
--- a/ilastik/applets/labeling/labelingGui.py
+++ b/ilastik/applets/labeling/labelingGui.py
@@ -133,7 +133,8 @@ class LabelingGui(LayerViewerGui):
             # Slot that gives a list of label names
             self.labelNames = None # labelNames.value
 
-    def __init__(self, parentApplet, labelingSlots, topLevelOperatorView, drawerUiPath=None, rawInputSlot=None, crosshair=True):
+    def __init__(self, parentApplet, labelingSlots, topLevelOperatorView, drawerUiPath=None, rawInputSlot=None,
+                 crosshair=True, is_3d_widget_visible=False):
         """
         Constructor.
 
@@ -176,7 +177,8 @@ class LabelingGui(LayerViewerGui):
         super(LabelingGui, self).__init__(parentApplet,
                                           topLevelOperatorView,
                                           [labelingSlots.labelInput, labelingSlots.labelOutput],
-                                          crosshair=crosshair)
+                                          crosshair=crosshair,
+                                          is_3d_widget_visible=is_3d_widget_visible)
 
         self.__initShortcuts()
         self._labelingSlots.labelEraserValue.setValue(self.editor.brushingModel.erasingNumber)
--- a/ilastik/applets/layerViewer/layerViewerGui.py
+++ b/ilastik/applets/layerViewer/layerViewerGui.py
@@ -76,12 +76,12 @@ class LayerViewerGui(with_metaclass(LayerViewerGuiMetaclass, QWidget)):
     """
     Implements an applet GUI whose central widget is a VolumeEditor
     and whose layer controls simply contains a layer list widget.
-    Intended to be used as a subclass for applet GUI objects.
+    Intended to be used as a superclass for applet GUI objects.
 
     Provides: Central widget (viewer), View Menu, and Layer controls
     Provides an EMPTY applet drawer widget.  Subclasses should replace it with their own applet drawer.
     """
-    
+
     ###########################################
     ### AppletGuiInterface Concrete Methods ###
     ###########################################
@@ -115,7 +115,8 @@ class LayerViewerGui(with_metaclass(LayerViewerGuiMetaclass, QWidget)):
     ###########################################
     ###########################################
 
-    def __init__(self, parentApplet, topLevelOperatorView, additionalMonitoredSlots=[], centralWidgetOnly=False, crosshair=True):
+    def __init__(self, parentApplet, topLevelOperatorView, additionalMonitoredSlots=[], centralWidgetOnly=False,
+                 crosshair=True, is_3d_widget_visible=False):
         """
         Constructor.  **All** slots of the provided *topLevelOperatorView* will be monitored for changes.
         Changes include slot resize events, and slot ready/unready status changes.
@@ -173,7 +174,8 @@ class LayerViewerGui(with_metaclass(LayerViewerGuiMetaclass, QWidget)):
         self.saved_layer_visibilities = None
 
         self._initCentralUic()
-        self._initEditor(crosshair=crosshair)
+
+        self._initEditor(crosshair=crosshair, is_3d_widget_visible=is_3d_widget_visible)
         self.__viewerControlWidget = None
         if not centralWidgetOnly:
             self.initViewerControlUi() # Might be overridden in a subclass. Default implementation loads a standard layer widget.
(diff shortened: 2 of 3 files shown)


wolny:
BTW, the tests are failing now cause they depend on the `VolumeEditor` changes I've made here: https://github.com/ilastik/volumina/pull/198

akreshuk:
I would prefer to keep ilastik and volumina as separated as possible. Volumina shouldn't really know anything about ilastik workflows. I would add an argument like "add_3d_widget" to the volumeEditor that the workflow could pass when creating it, default to "off".

wolny:
Thanks for the comment @akreshuk. Totally makes sense, as passing the workflow name to the `VolumeEditor` (Volumina) breaches the boundaries between the two project. I'll pass just the `add_3d_widget` flag to the `VolumeEditor` as you suggest, and do all the necessary computation of the flag in `ilastik`.

wolny:
BTW I've made the necessary changes in `carving/preprocessingGui.py` and `carving/carvingGui.py` (those are the two entry points which initialise the `VolumeEditor`). 

I've added the `is_3d()` helper function to both `carving/opCarving` and `carving/opPreprocessing` for, since I didn't want to refactor them too much (Lazyflow's `Operator` is their common class their inherit from, but it doesn't seem like a good place to put the `is_3d()` function in there. We might as well extract `is_3d()` to some utility class/module in ilastik if you have any suggestions).

ilastik/applets/layerViewer/layerViewerGui.py
```diff
@@ -76,12 +76,12 @@ class LayerViewerGui(with_metaclass(LayerViewerGuiMetaclass, QWidget)):
     """
     Implements an applet GUI whose central widget is a VolumeEditor
     and whose layer controls simply contains a layer list widget.
-    Intended to be used as a subclass for applet GUI objects.
+    Intended to be used as a superclass for applet GUI objects.
```
k-dominik: nice catch :)

ilastik/applets/labeling/labelingGui.py
```diff
@@ -133,7 +133,8 @@ def __init__(self):
             # Slot that gives a list of label names
             self.labelNames = None # labelNames.value
 
-    def __init__(self, parentApplet, labelingSlots, topLevelOperatorView, drawerUiPath=None, rawInputSlot=None, crosshair=True):
+    def __init__(self, parentApplet, labelingSlots, topLevelOperatorView, drawerUiPath=None, rawInputSlot=None,
+                 crosshair=True, add_3d_widget=False):
```
k-dominik: with `add_3d_widget` I have the same suggestion as in volumina - as th ewidget is added anyhow and only visibility is altered. `3d_widget_visible` ?

ilastik/applets/layerViewer/layerViewerGui.py
```diff
@@ -115,7 +115,8 @@ def stopAndCleanUp(self):
     ###########################################
     ###########################################
 
-    def __init__(self, parentApplet, topLevelOperatorView, additionalMonitoredSlots=[], centralWidgetOnly=False, crosshair=True):
+    def __init__(self, parentApplet, topLevelOperatorView, additionalMonitoredSlots=[], centralWidgetOnly=False,
+                 crosshair=True, add_3d_widget=False):
```
k-dominik: `3d_widget_visible` ?

ilastik/applets/layerViewer/layerViewerGui.py
```diff
@@ -173,7 +174,8 @@ def __init__(self, parentApplet, topLevelOperatorView, additionalMonitoredSlots=
         self.saved_layer_visibilities = None
 
         self._initCentralUic()
-        self._initEditor(crosshair=crosshair)
+
+        self._initEditor(crosshair=crosshair, add_3d_widget=add_3d_widget)
```
k-dominik: `3d_widget_visible` ?

ilastik/applets/layerViewer/layerViewerGui.py
```diff
@@ -615,11 +617,11 @@ def _initCentralUic(self):
         localDir = os.path.split(__file__)[0]
         uic.loadUi(localDir+"/centralWidget.ui", self)
 
-    def _initEditor(self, crosshair):
+    def _initEditor(self, crosshair, add_3d_widget):
```
k-dominik: `3d_widget_visible` ?

ilastik/applets/layerViewer/layerViewerGui.py
```diff
@@ -615,11 +617,11 @@ def _initCentralUic(self):
...
         """
         Initialize the Volume Editor GUI.
         """
-        self.editor = VolumeEditor(self.layerstack, parent=self, crosshair=crosshair)
+        self.editor = VolumeEditor(self.layerstack, parent=self, crosshair=crosshair, add_3d_widget=add_3d_widget)
```
k-dominik: `3d_widget_visible` ?

ilastik/workflows/carving/carvingGui.py
```diff
@@ -252,11 +252,17 @@ def makeColortable():
...
+        tagged_shape.update(self.topLevelOperatorView.InputData.meta.getTaggedShape())
+        is_3d = tagged_shape['x'] > 1 and tagged_shape['y'] > 1 and tagged_shape['z'] > 1
+        return is_3d
+
     def _after_init(self):
         super(CarvingGui, self)._after_init()
-        if self.render:self._toggleSegmentation3D()
-        
+        if self.render:
+            self._toggleSegmentation3D()
```
k-dominik: ! really nice one. Didn't understand the old line at first glance :D

ilastik/workflows/carving/opPreprocessing.py
```diff
@@ -492,5 +492,3 @@ def reset(self):
         '''reset sigma and filter to values of last preprocess'''
         self.applet._gui.setSigma(self.initialSigma)
         self.applet._gui.setFilter(self.initialFilter)
-
```
k-dominik: Seems a bit unrelated :)

ilastik/workflows/carving/carvingGui.py
wolny: 😆

ilastik/applets/labeling/labelingGui.py
wolny: yeah, that name makes more sense. I'll change the name here and in Volumina, so it's consistent everywhere. Thanks!

wolny: changed the name to `is_3d_widget_visible`, cause var names cannot start with a number :) ✅

ilastik/applets/layerViewer/layerViewerGui.py
wolny: ✅

ilastik/applets/layerViewer/layerViewerGui.py
wolny: ✅

ilastik/applets/layerViewer/layerViewerGui.py
wolny: ✅

ilastik/applets/layerViewer/layerViewerGui.py
wolny: ✅

ilastik/workflows/carving/opPreprocessing.py
wolny: ✅

wolny:
Thanks for the review @k-dominik. I've changed the naming according to your suggestions, and removed the noisy import reformatting done by PyCharm.

k-dominik:
Awesome, thank you for this very quick fix - already checked it out and it works :)

PR #1687, opened 2018-01-17
Disabled testObjectCountingMutiImageGui

k-dominik:
this test fails sometimes and we do not really know why or how
to trigger these fails. Therefore, it is disabled for the time
being.


k-dominik:
I have confirmed that the test is indeed skipped -> going to merge it.

Issue #1635, opened 2017-11-02
Tracking with Learning: Manual annotations unusably slow


chaubold:
I just tried to annotate a few tracks to learn the weights for tracking, on a `2D+t+c` dataset of 1000x1000 pixels and 3 color channels, 360 time frames, where cells are detected in a prediction map (not binary segmentation).

**Whenever I click an object, ilastik _thinks_ for ~5secs before it becomes responsive again** (and then automatically goes to the next time frame etc as expected). But I do not understand **why** it takes so long. If I right click an object and run the semi-automatic tracking it can do 50 frames within those 5 seconds (where it needs to compare segmentation masks between frames and if the overlap is big enough and unambiguous, it continues the track). Hence it should not be the thresholding of every frame that slows it down. Inserting a label should not take long and if it can do it 50 times within that timeframe it shouldn't affect _interactivity_ either. 

Is there something associated with a "click" that's slowing it down? The right click menu shows up instantly, though.

Lastly, could it be the visualization? Zooming in does not make it faster either. 

The data was originally coming from tiffs, so I used ilastik's `DataConversion` workflow to create a HDF5 file, which has the following chunking settings:
```
>$ h5dump -p 1_Input.h5
HDF5 "1_Input.h5" {
GROUP "/" {
   DATASET "exported_data" {
      DATATYPE  H5T_STD_U8LE
      DATASPACE  SIMPLE { ( 361, 1002, 1004, 3 ) / ( 361, 1002, 1004, 3 ) }
      STORAGE_LAYOUT {
         CHUNKED ( 1, 716, 715, 1 )
         SIZE 2217724080
      }
      FILTERS {
         NONE
      }
      FILLVALUE {
         FILL_TIME H5D_FILL_TIME_ALLOC
         VALUE  0
      }
      ALLOCATION_TIME {
         H5D_ALLOC_TIME_INCR
      }
}
```

k-dominik:
have you tried the same image in our python 2 version?

chaubold:
Good point, no I didn't try that yet

ilastikdev:
I have run the latest code on two big data sets (Fratini and Mueller) and have not experienced any delays in tracking training. I will try to reproduce the problem,

Janez

ilastikdev:
I can not reproduce the problem on Linux.
Are you on a MAC?

chaubold:
No, that was on Linux. I'll share the project file with you.

ilastikdev:
[...]

PR #1688, opened 2018-01-17
remove redundant line

ilastikdev:
fixes 
https://github.com/ilastik/ilastik/issues/1635


akreshuk:
It seems like the redundant line was something we forgot to remove during debugging. Out with it.

Issue #1692, opened 2018-01-17
Go to the next unlabeled division leads nowhere


akreshuk:
Pressing this button takes me to a field of view with no divisions and even no objects in the center. Could the object in question be highlighted somehow?

ilastikdev:
fixed in 
https://github.com/ilastik/ilastik/pull/1704

ilastikdev:
PR 
https://github.com/ilastik/ilastik/pull/1704

Issue #1691, opened 2018-01-17
Labeling divisions is stuck in tracking with learning


akreshuk:
When labeling divisions, an error appears along the lines of "Parent id must be the active track". It's impossible to make it go away or correct it. It is, however, not printed in the logs.

ilastikdev:
Fixed in
https://github.com/ilastik/ilastik/pull/1704

Issue #1690, opened 2018-01-17
Filter should be called Field of View in the tracking applet 


akreshuk:
At the bottom, below the parameter list.

ilastikdev:
fixed in 
https://github.com/ilastik/ilastik/issues/1690

ilastikdev:
PR 
https://github.com/ilastik/ilastik/pull/1704

Issue #1693, opened 2018-01-17
Parameter names cut off in tracking with learning


akreshuk:
![tracking_with_learning_track](https://user-images.githubusercontent.com/901064/35058320-d94d9402-fbb7-11e7-8f63-3faef239582c.png)

See the parameter table on the left - parameter names are cut off, even though they look fine in the regular tracking workflow. Could it be changed to use the same widgets?

ilastikdev:
this behavior is not specific to a workflow - it relates to the width of the user controls in the last ilastik project closed.
(one can make conservation tracking pop up with cut off parameter names also)

ilastikdev:
fixed in 5e6bc83

ilastikdev:
PR https://github.com/ilastik/ilastik/pull/1704

ilastikdev:
@k-dominik 
please check this again

k-dominik:
at least on macos it looks fine.

PR #1704, opened 2018-01-31
new functionality in tracking and a few fixes

ilastikdev:


ilastikdev:
fixes 
https://github.com/ilastik/ilastik/issues/1690

ilastikdev:
fixes #1691

k-dominik:
Hello @ilastikdev 

I've tried the PR and found something odd - at least to me it's odd.
So i've been annotating tracks in the "Training" applet and the behaviour that puzzled me occurred when annotating divisions:

Expected behaviour:
* I annotate a track, find a division
* I go back one time frame, hit "Division Event"
* Some buttons on the left get greyed out, "Division Event" looks active
* I click the parent (jumps to next frame)
* I click once on each child - the tracks appear.
* Buttons go back to normal

Observed behaviour:
* I annotate a track, find a division
* I go back one time frame, hit "Division Event"
* Some buttons on the left get greyed out, ~"Division Event" looks active~ _the button says "Stop Division Event"_
* I click the parent (jumps to next frame)
* I click once on each child - the tracks appear.
* Buttons go back to normal, except _the button still says "Stop Division Event"_
* even if I start a new track, _the button says "Stop Division Event"_
* i have to click on "stop division event" to make it go away

is this the expected behaviour?

k-dominik:
In the Tracking Applet, the drawer layout seems to deviate a bit from what we see in other applets.

with a normal width, it looks like this:

tracking1

this is the minimum width to see it all:
tracking2

is this change intended (in 1.3.0 one could have it way slimmer)? Can you confirm, that you see the same issue on your computer? It could as well be an issue on my machine somehow? This is how it looks for me in 1.3.0:
tracking3

ilastikdev:
Minimum width was increased so that the longest string "Transition Neighbourhood" gets displayed without cutting off the words.

There may be a global solution to this by increasing the user controls width all together.

On my machine with three different resolution screens, all three screens show correct behavior.
However, this may still depend on pixels per font and/or point ratio. 
I can not test this because I can not reproduce the error.
You can play with the minimum width in the ui file in .../ilastik/applets/tracking/structured
and see if this will change the appearance on your machine.

k-dominik:
Hi @ilastikdev , @FynnBe big changes are merged in now and we could expect the tests to pass after rebasing this branch onto the current master. Would you mind doing that?

ilastikdev:
fixes #1692  #1693

k-dominik:
tests pass on my machine (against ilastik/lazyflow@11a13f566ec6748b6cdda736f2888ead76d20c67 )

PR #1707, opened 2018-02-06
Improved carving [Work In Progress]

DerThorsten:
Hi All,
Here are the following changes
* naive block-wise parallel watersheds (Yes there will be hard block borders in the supervoxels)
* agglomerative clustering will reduce the supervoxels. Users can specify by which factor they want to reduce the number of supervoxels (this will more or less remove the  hard block borders )
* new marchingcubes works like a charm (no changes where needed besides the bugfixes in the marching cubes repo)
* This needs a very new version of nifty >= 0.24 (@constantinpape  might be willing to make a new build)

The gui and the code in `carvingTools.py` still needs a bit of polishing and maybe someone besides me should test this.
Besides that, it is ready


constantinpape:
I have merged your recent branch into my branch.
I will need to test it a bit more. If all works out, I can build a new package in the coming days.

k-dominik:
Thanks @DerThorsten for this great PR and for taking the time to walk me through it a bit this morning. We arrived at the following ToDos before we merge it:

- [ ] clean up `carvingTools.py`
- [ ] rework applet gui a bit, reduce the width
- [ ] check whether it works with 2D data
- [ ] config parallelism with lazyflow threads setting
- [ ] Add current processing step output to progress-reporter
- [ ] Investigate why the "completed segments" layer does not work atm -> fix it

akreshuk:
@wolny , completed segments not showing up looks like a good place to start

akreshuk:
I got rid of the other done slot and now both "completed Objects" layers pull from the same output slot and display the same strange caching behavior.

wolny:
Thanks for the info. Good that it's consistent between the layers now :smile:

k-dominik:
ah well... the message in abbda1f should say 'Re-enabled multicut'

PR #1720, opened 2018-03-13
Extend color table list dynamically to accommodate more than 16 labels

wolny:
Quick and dirty fix for the following: https://github.com/ilastik/ilastik/issues/933.
A proper fix/refactoring will follow in a subsequent PR.


akreshuk:
This is indeed a quick and dirty fix which would only work if the labels are added one by one. Otherwise, you'll add several labels of the same color. Still, I think it's the only way we are using it, so if it works, let's put it in, @stuarteberg ?

wolny:
Yeah currently it works by adding labels one by one during deserlialization, so this change works. Another quick (and a bit less dirty) fix we could use, is to initialize `_colorTable16` with the proper size (and random colors) in the LabelingGui...

PR #1722, opened 2018-03-20
fixed #1662

### LINKED ISSUE #1662: 2D data from h5 files
When trying to read data from `h5` files, we only look for datasets with `3 <= dimenisonality <= 5` (see snippet from `DataSelectionGui`below).
Sure this is easily altered, but I was wondering if we had a reason exclude data with `dimensionality == 2`.

https://github.com/ilastik/ilastik/blob/d23d3e5e708fe2a72aa83b9ae566e05f14d894df/ilastik/applets/dataSelection/dataSelectionGui.py#L699-L700

for further reference, there is a function with similar functionality in `lazyflow.utility.pathHelpers.lsHdf5`

hafizfahad143:


k-dominik:
The code change is fine, but could you please, for the future, give the pull request a descriptive name e.g.
> Allow 2D images to be loaded from H5 files

In the body of the PR you should also describe your changes a bit. Here, of course it is trivial, but one could still write something like:

> Loading data from `.h5` files was limited to data with 3 to 5 dimensions. In this PR I have changed the lower limit to 2 dimensions. Side effect: Also non-image data (e.g. tables) might now show up in the imge selection dialog when loading from `.h5` files with multiple datasets.

k-dominik:
Hello Hafiz, could you please commit the new changes (the ones for multicut) to a different branch and open a separate PR?

PR #1723, opened 2018-03-20
Colortable16 changes

akreshuk:
make all gui classes use volumina colortable
make that colortable colorblind-friendly

### DIFF
--- a/ilastik/applets/blockwiseObjectClassification/blockwiseObjectClassificationGui.py
+++ b/ilastik/applets/blockwiseObjectClassification/blockwiseObjectClassificationGui.py
@@ -30,6 +30,7 @@ from PyQt5.QtGui import QColor
 
 from ilastik.utility.gui import threadRouted
 from volumina.api import LazyflowSource, ColortableLayer
+from volumina import colortables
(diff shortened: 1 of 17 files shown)


akreshuk:
For reference, still not working:

- [ ] when loading a prob map into object classification, preview is shown with red/green
- [ ] deleting labels in object classification completely messed up: colors are not properly recycled, displayed pmaps are not made dirty
- [ ] when switching to export applet, "Predictions - exported" shows thresholding with default threshold value instead of user selected one. This is before any export is done at all. "Predictions Preview"  is fine.

akreshuk:
I fixed everything I found to be wrong before. Someone else needs to test it on a different machine now, with a new eye.

k-dominik:
Looked at it and found some minor stuff which I fixed right away.

One thing is still open:

when showing multi-channel images in the `DataSelectionApplet`, we try to be smart in order to show RGB(A) images. Probability maps are then shown as RGB-ish images starting from red, to green... etc. Before the colors matched up (by coincidence, or design) with our default colors, this is not the case anymore. Do we want to resolve it before we merge this one @akreshuk ?

PR #1728, opened 2018-04-04
 AssertionError during creation of MultiCut project

hafizfahad143:
fixes#1653 
in a code i add an other condition to activate the DT Watershed Applet that raw data as well as probability data is ready to activate this applet

### DIFF
--- a/ilastik/workflows/edgeTrainingWithMulticut/edgeTrainingWithMulticutWorkflow.py
+++ b/ilastik/workflows/edgeTrainingWithMulticut/edgeTrainingWithMulticutWorkflow.py
@@ -353,7 +353,7 @@ class EdgeTrainingWithMulticutWorkflow(Workflow):
         batch_processing_busy = self.batchProcessingApplet.busy
 
         self._shell.setAppletEnabled( self.dataSelectionApplet,             not batch_processing_busy )
-        self._shell.setAppletEnabled( self.wsdtApplet,                      not batch_processing_busy and input_ready and not superpixels_available_from_file )
+        self._shell.setAppletEnabled( self.wsdtApplet,                      not batch_processing_busy and input_ready and opDataSelection.ImageGroup[0][1].ready() and not superpixels_available_from_file )
         self._shell.setAppletEnabled( self.edgeTrainingWithMulticutApplet,  not batch_processing_busy and input_ready and superpixels_ready )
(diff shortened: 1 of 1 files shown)


ilastik/workflows/edgeTrainingWithMulticut/edgeTrainingWithMulticutWorkflow.py
```diff
@@ -353,7 +353,7 @@ def handleAppletStateUpdateRequested(self):
         batch_processing_busy = self.batchProcessingApplet.busy
 
         self._shell.setAppletEnabled( self.dataSelectionApplet,             not batch_processing_busy )
-        self._shell.setAppletEnabled( self.wsdtApplet,                      not batch_processing_busy and input_ready and not superpixels_available_from_file )
+        self._shell.setAppletEnabled( self.wsdtApplet,                      not batch_processing_busy and input_ready and opDataSelection.ImageGroup[0][1].ready() and not superpixels_available_from_file )
```
k-dominik: While your code works for a single lane, it would be nice to have it a bit more generic. Other workflows implement a `_inputReady` method, e.g. in Object Classification:

https://github.com/ilastik/ilastik/blob/428b5180e63f8656b36865d43db0eef62e053314/ilastik/workflows/objectClassification/objectClassificationWorkflow.py#L423-L433

Albeit this is a bit more code, this is more generic and also works for more than a single image lane.

You can find examples of its usage here:

https://github.com/ilastik/ilastik/blob/428b5180e63f8656b36865d43db0eef62e053314/ilastik/workflows/objectClassification/objectClassificationWorkflow.py#L630

ilastik/workflows/edgeTrainingWithMulticut/edgeTrainingWithMulticutWorkflow.py
```diff
@@ -353,7 +353,7 @@ def handleAppletStateUpdateRequested(self):
...
         self._shell.setAppletEnabled( self.edgeTrainingWithMulticutApplet,  not batch_processing_busy and input_ready and superpixels_ready )
```
k-dominik: While you are at it, it would be nice if you found a condition that would not enable the training applet before superpixels have been computed (as it is the case right now, `superpixels_ready`

PR #1733, opened 2018-04-11
Feature user axis order

FynnBe:
Separate internal axis order (aka axisorder) from outside/user axis order (aka original axisorder)
Requiers https://github.com/ilastik/volumina/pull/200


k-dominik:
So I let the test run on linux and they passed (yay!). However I get tons of warnings like this:

```
appletSerializer 24308 140311762884352 PickleableSlot version mismatch detected. Trying to load slot value.
```

is this expected?

FynnBe:
I think there is something wrong when serializing Structured Tracking projects. The 'loaded_version' is None. Maybe it is simply because I created them with my regular installed ilastik, but even in projects created in my dev env, I can't find the relatively newly added VERSION attribute

FynnBe:
The "unused command line args" is to be expected, because they are not used by the ilastik shell, but forwarded to the workflow. The check for used args is done before that happens... I don't see an easy way around that without changing the args checking logic.

FynnBe:
Somehow absolute paths were introduced again. Nice reminder to fix that bug...

k-dominik:
cool, iss this ready to be merged?

FynnBe:
Yes, the corresponding lazyflow and volumina PRs are merged already

k-dominik:
Okay, I've added what I hope is a fix for the circle-ci mess. Let's see

k-dominik:
look at this! All green :)

PR #1760, opened 2018-05-16
Update opStructuredTracking.py

### LINKED ISSUE #1756: Tracking with learning: var referenced before assignment
Happens, while computing the weights for the events.
Logs:
```
  File "/opt/ilastik-1.3.1b1-Linux/ilastik-meta/ilastik/ilastik/applets/tracking/structured/opStructuredTracking.py", line 302, in _runStructuredLearning
    if type == None:
```
UI error:
trackingproblem

varunotelli:
fixed #1756


ilastikdev:
Hi,

Thanks for the suggestion,
but the new line type = None should be placed  at line 248 (before the for loop).

Before we do that we should investigate why _type did not properly return a value or why the for loop was not entered, that is, why the trackSet is empty.

PR #1769, opened 2018-05-24
#-1758 Object classification: 'Live Update' should be a button not a checkbox for consistency with the other workflows

Amir22010:
Object classification: 'Live Update' should be a button not a checkbox for consistency with the other workflows. let me know your view
liveupdate


Amir22010:
closed contains commits of master branch too, will resolve it and give u only feature branch pr

Issue #1758, opened 2018-05-15
Object classification: 'Live Update' should be a button not a checkbox for consistency with the other workflows.


wolny:
![objclassliveupdate](https://user-images.githubusercontent.com/706781/40059232-de818e1c-5853-11e8-9d46-4243f715a5c1.png)

PR #1777, opened 2018-06-04
Change Live Update QCheckBox to QPushButton

Radi85:
Hey @wolny, 
Can you check this out? #1758


k-dominik:
Hey hey,

i just glanced over you proposed changes! Thank you very much for the contribution.
We tend to not commit automatic pep8 changes for now, in order to preserve history better and not to obfuscate the real changes so much.
So could you please only commit the relevant changes?

PR #1779, opened 2018-06-06
Fix show invisible test fail

k-dominik:
This is a bit of a bummer:

`tests/test_applets/pixelClassification/testPixelClassificationHeadless.py` has been failing silently for a while now, but it went unnoticed, because of the way the test has been invoked. This commit makes sure that a test failure is reflected in the return value.

Summary:

basically just changed `nose.run` to `nose.main`. The only difference between those functions is that `nose.run` returns `True` if all tests pass and `False` if tests fail. One would have to handle the return value explicitly. `nose.main` will exit after test execution with a proper return value indicating successful test runs with `0` and failures with `1`

About the failing test:

@FynnBe one of your commits (8080c81c6a99122f6754aa8b0548d1554d80fc45) brought that failure to light (did not cause it as far as I can tell, but now some old code runs, that hasn't run before). Would you mind looking into that?

### DIFF
--- a/tests/test_applets/blockwiseObjectClassification/testOpBlockwiseObjectClassification.py
+++ b/tests/test_applets/blockwiseObjectClassification/testOpBlockwiseObjectClassification.py
@@ -332,4 +332,4 @@ if __name__ == "__main__":
     import nose
     sys.argv.append("--nocapture")    # Don't steal stdout.  Show it on the console as usual.
     sys.argv.append("--nologcapture") # Don't set the logging level to DEBUG.  Leave it alone.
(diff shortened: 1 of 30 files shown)


FynnBe:
Further reducing the data size (to speed up test) failed for xy = (20, 20)
It works for (100, 100), but already fails for (50, 50).
This is very confusing, as the failing assert statement checks the dtype of meta data 'drange'...?

k-dominik:
Just for completeness, the problem arose from the cutout subregion being to large if the data shape was small(ish).
This should probably raise an Exception or something.

PR #1789, opened 2018-06-15
Add a minimum of 2 Labels in each workflow

### LINKED ISSUE #1743: Automatically add initial number of labels
Training in Pixel Classification and Object classification doesn't make much sense for less than two labels. We should add those automatically, saves two clicks.

paulhfu:
fixes #1743
in object classification and pixel classification there are now two labels initially, just like in tracking


ilastik/applets/objectClassification/objectClassificationGui.py
```diff
@@ -175,7 +175,14 @@ def __init__(self, parentApplet, op):
         self.labelingDrawerUi.checkShowPredictions.toggled.connect(
             self.handleShowPredictionsClicked)
 
-        #select all the features in the beginning
+        # Add two permanent labels because it makes no sense to have less here
+        self._addNewLabel()
+        self._addNewLabel()
+        self.labelingDrawerUi.labelListModel.makeRowPermanent(0)
+        self.labelingDrawerUi.labelListModel.makeRowPermanent(1)
+        self.labelingDrawerUi.labelListView.shrinkToMinimum()
```
k-dominik: I think this line kind of breaks the layout for me. Same holds true for Object Classification.

* with this PR
pc-new
* old state
pc-old

could you please make it behave it like in the lower screenshot?

paulhfu:
Hi @k-dominik, i applied your suggested changes. Could you review it again?

k-dominik:
hmmm, i think the problem is, that the `test_7_EraseCompleteLabel` in `testPixelClassificationGui` somehow isn't terminated properly. If I run it locally, it just halts...

k-dominik:
i pushed some minimal fixes, if the tests pass now on the CI (they do locally) it should be fine

k-dominik:
Hey @paulhfu ,

I went over the PR again, and, unfortunately found something that needs to be fixed before we can merge it in:

- [ ] In tracking (not manual) in the object count classifier, behavior is sort of weird. 2 initial labels are correctly added. Then, if I add two new ones (two objects, three objects) I can only delete the last one. Could you please fix that?

k-dominik:
thanks @paulhfu thank you for the fix. I'll play with it a  bit more, and merge it if I don't find anything :)

paulhfu:
@k-dominik , don't merge it yet. In obj. classification, by start labeling with the second label, an exception is thrown. I will take care of it.

paulhfu:
@k-dominik , should be fine now. Not sure if there is a better fix than the last commit.

ilastik/applets/labeling/labelingGui.py
```diff
@@ -636,10 +637,19 @@ def _addNewLabel(self):
...
+        if self._allowDeleteLastLabelOnly and self._defTwoInitialLabels:
+            # make previous label unremovable when we have at least three labels
             if newRow > 2:
                 self._labelControlUi.labelListModel.makeRowPermanent(newRow - 1)
+        elif self._allowDeleteLastLabelOnly:
+            # make previous label unremovable
+            if newRow > 0:
+                self._labelControlUi.labelListModel.makeRowPermanent(newRow - 1)
+        elif self._defTwoInitialLabels:
+            # if there are only two labels remaining make the unremovable
```
k-dominik: the comment is a bit misleading. shouldn't it say something like "if a third label is added, allow all to be deleted"...

ilastik/applets/labeling/labelingGui.py
```diff
@@ -752,10 +762,19 @@ def _onLabelRemoved(self, parent, start, end):
         oldcount = self._labelControlUi.labelListModel.rowCount() + 1
         logger.debug("removing label {} out of {}".format( row, oldcount ))
 
-        if self._allowDeleteLastLabelOnly:
+        if self._defTwoInitialLabels and self._allowDeleteLastLabelOnly:
```
k-dominik: hmmm, just conceptually (and i am aware that it might have been different before), would it make sense to delete the label first, and then add all this logic? Just in case something goes wrong on label delete, the state will kind of be more consistent? What do you think?

ilastik/applets/labeling/labelingGui.py
```diff
@@ -871,3 +890,13 @@ def allowDeleteLastLabelOnly(self, enabled):
         assume that all previous cell counts are given.
         """
         self._allowDeleteLastLabelOnly = enabled
+
+    def defTwoInitialLabels(self, enabled):
+        # in some workflows wen need always at least two labels to function correctly.
+        # setting enabled to True restricts having less than two labels
```
k-dominik: you mean it forces to have at least two labels?

ilastik/applets/labeling/labelingGui.py
```diff
@@ -871,3 +890,13 @@ def allowDeleteLastLabelOnly(self, enabled):
...
+        self._addNewLabel()
+        self._addNewLabel()
+        self.labelingDrawerUi.labelListModel.makeRowPermanent(0)
+        self.labelingDrawerUi.labelListModel.makeRowPermanent(1)
+
+        self._defTwoInitialLabels = enabled
```
k-dominik: if above comment is true, this could have a more catchy name, like: `forceAtLeastTwoLabels`

ilastik/applets/labeling/labelingGui.py
paulhfu: Isn't the label already deleted in listModel.py/removeRow() which is called before _onLabelRemoved? In _onLabelRemoved are just the label selection and the colors updated. Don't know what difference it would make if i move it to the end of the function.

paulhfu: Okay mabe it makes sense, just to be sure that the whole delete-process was properly executed to that point

paulhfu: I will do the same in _addNewLabel

ilastik/applets/labeling/labelingGui.py
paulhfu: Yes! I'll change the comment

ilastik/applets/objectClassification/objectClassificationApplet.py
```diff
@@ -22,6 +22,7 @@
 from ilastik.applets.base.standardApplet import StandardApplet
 
 from .opObjectClassification import OpObjectClassification
+from .objectClassificationGui import ObjectClassificationGui
```
k-dominik: same as above

ilastik/applets/objectClassification/objectClassificationGui.py
```diff
@@ -163,20 +164,24 @@ def __init__(self, parentApplet, op):
         self._colorTable16_forpmaps[15] = QColor(Qt.black).rgba() #for objects with NaNs in features
         
         # button handlers
-        self._interactiveMode = False
+        self.interactiveMode = False
```
k-dominik: uhm, was this a bug?? the setters are also still there, and the logic in them isn't needed anymore?

ilastik/applets/pixelClassification/pixelClassificationApplet.py
```diff
@@ -21,6 +21,7 @@
 ###############################################################################
 from ilastik.applets.base.standardApplet import StandardApplet
 from .opPixelClassification import OpPixelClassification
+from .pixelClassificationGui import PixelClassificationGui
```
k-dominik: please move back to function

ilastik/workflows/carving/carvingApplet.py
```diff
@@ -23,6 +23,7 @@
 
 from ilastik.utility import OpMultiLaneWrapper
 from .opCarving import OpCarving
+from .carvingGui import CarvingGui
```
k-dominik: pls move back to function

Issue #1441, opened 2017-03-27
No access to raw data from project in headless


akreshuk:
If the classifier has been trained already, running the headless mode with such a project should not require access to the raw data anymore.

stuarteberg:
**Background:**

The thing that makes this tricky is that the entire graph is connected and initialized when the project file is deserialized.  If the project has, say, 5 input files, then there are 5 slots feeding into the classifier training operator.  The classifier slot is therefore "not ready" unless all 5 of those slots are "ready" in turn.  But it doesn't make sense to declare a data input slot "ready" if we don't know anything about it (e.g. shape, axis order, dtype).

**Possible solution:**

Add a flag to the headless interface, `--dont-load-training-files` (or, ideally, something less awkward and more general).  When that flag is detected, the `DataSelection` applet won't deserialize *any* lanes into the workflow.  But the other applets, such as `FeatureSelection` and `PixelClassification`, will still deserialize their settings as usual, including the classifier cache.  I *think* deserialization and batch mode processing will still proceed normally, despite the fact that the workflow will have zero lanes.

Possible "gotchas": Some applets, such as thresholding store cached image data for each lane in the project file.  We have to make sure that the corresponding deserializers don't attempt to load those caches into the workflow operators -- there will be zero lanes, and therefore no caches in which they can inject the restored data.

chaubold:
I might be totally wrong here, but if we run pixel classification in batch mode and have a trained random forest, then we should be able to simply ignore all parts of the graph that led to the Random Forest, meaning we could chop their connections when running in batch (probably already when setting up the workflow). That way the Random Forest serves as *input* to the graph and is ready because we can load it from the project. But I bet there's a good reason why this is not so easy, as always :smile:

hno2:
This is fixed now, right?

k-dominik:
true, thanks for the reminder!

PR #1790, opened 2018-06-16
No need for raw data to run a project in headless

wolny:
Motivation:
unfortunately in order to run workflows in headless mode the raw data that the models were trained on has to be available. this has been quite uncomfortable for the users especially when the raw data has significant size (GBs) and has to be carry around in order to run tasks in headless.

Current solution:
- if we're in headless and the raw data exists the flow is the same as before
- if we're in headless and the raw data does not exists, tell the `opDataSelection` to fake the raw data and we are able to proceed with the batch
- if we're in headless, raw data does not exists and `--retrain` was passed -> we inform the user that retraining is not possible

Faking the raw data is possible, due to additional attributes to the lane metadata (`DatasetInfo`) serialized to the project file. The serialization is backward compatible, however because the 'old' project files lack the necessary metadata headless mode without raw data won't work (it obviously still work if the raw data is present, see 1st point above).

TODO:
- [x] Ensure backward compatibility
- [x] Extensive manual tests (different workflows, different dimensionality of raw data, multiple image lanes)
- [x] Write unit tests 
- [x] Write integration testes

Original issue: https://github.com/ilastik/ilastik/issues/1441

### DIFF
--- a/ilastik/applets/dataSelection/dataSelectionSerializer.py
+++ b/ilastik/applets/dataSelection/dataSelectionSerializer.py
@@ -366,6 +366,7 @@ class DataSelectionSerializer( AppletSerializer ):
         # Unready datasets are represented with an empty group.
         if len( infoGroup ) == 0:
             return None, False
+
         datasetInfo = DatasetInfo()
 
         # Make a reverse-lookup of the location storage strings
@@ -373,7 +374,11 @@ class DataSelectionSerializer( AppletSerializer ):
         datasetInfo.location = LocationLookup[ infoGroup['location'].value.decode('utf-8') ]
         
         # Write to the 'private' members to avoid resetting the dataset id
-        datasetInfo._filePath = infoGroup['filePath'].value.decode('utf-8')
+        if headless:
+            datasetInfo._filePath = ''
+        else:
+            datasetInfo._filePath = infoGroup['filePath'].value.decode('utf-8')
+
         datasetInfo._datasetId = infoGroup['datasetId'].value.decode('utf-8')
 
         try:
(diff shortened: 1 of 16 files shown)


ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -203,19 +209,6 @@ def filePath(self, newPath):
     def datasetId(self):
         return self._datasetId
 
-    DatasetInfoSchema = \
```
wolny: It doesn't seems to be used anywhere, so I just deleted it as obsolete

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -343,13 +348,23 @@ def setupOutputs(self):
...
+                        opReader.SubVolumeRoi.setValue(datasetInfo.subvolume_roi)
+                    opReader.WorkingDirectory.setValue(self.WorkingDirectory.value)
+                    opReader.SequenceAxis.setValue(datasetInfo.sequenceAxis)
+                    opReader.FilePath.setValue(datasetInfo.filePath)
+                else:
+                    # Use fake reader: allows to run the project in a headless
+                    # mode without the raw data
+                    opReader = OpZeroDefault(parent=self)
+                    opReader.MetaInput.meta = MetaDict(shape=datasetInfo.laneShape,
+                                             dtype=datasetInfo.laneDtype,
```
FynnBe: 👎 'manual linting bot alert'

ilastik/applets/dataSelection/dataSelectionSerializer.py
```diff
@@ -493,6 +521,16 @@ def unload(self):
             This way we can avoid invalid state due to a partially loaded project. """ 
         self.topLevelOperator.DatasetGroup.resize( 0 )
 
+    def _shouldRetrain(self):
```
FynnBe: maybe add a beautiful '@property' here?

ilastik/applets/dataSelection/opDataSelection.py
wolny: :crying_cat_face: :broken_heart:

wolny: done :white_check_mark:

ilastik/applets/dataSelection/dataSelectionSerializer.py
wolny: this method should not be a part of  `dataSelectionSerializer` public API, so I'd rather keep it protected

FynnBe: I don't see how '@property' would change anything about that. (It's always public anyway, the underscore can stay to indicate that it's meant to be private.)
Anyhow, it's fine as is 👍

wolny: true! on the 2nd thought, I think the `@property` would be nice in here :)

wolny: done :white_check_mark:

wolny:
Unit and integration tests are in place now, so this is good to go from my side. Would be great to have it in in the next release, cause users have been waiting for this feature for a long time.

wolny:
Thanks for the review @FynnBe!

PR #1793, opened 2018-06-19
Always instantiate the CacheMemoryManager on start

clbarnes:
Previously, the manager would only be instantiated on startup in
specific circumstances (when a refresh interval is explicitly set).
Therefore, for some headless workflows, the cache purging thread was
never started and the caches would grow uncontrollably.

Now it is always instantiated with the default interval (which can then
be changed later).

### DIFF
--- a/ilastik_main.py
+++ b/ilastik_main.py
@@ -97,6 +97,7 @@ def main(parsed_args, workflow_cmdline_args=[], init_logging=True):
     preinit_funcs = []
     # Must be first (or at least before vigra).
     preinit_funcs.append(_import_opengm)
+    preinit_funcs.append(_instantiate_cache_manager)
 
     lazyflow_config_fn = _prepare_lazyflow_config(parsed_args)
     if lazyflow_config_fn:
@@ -287,6 +288,14 @@ def _import_opengm():
         import opengm # noqa
     except ImportError:
         pass
+
+
+def _instantiate_cache_manager():
+    """Ensure that the CacheMemoryManager thread is started during setup.
+    """
+    import lazyflow
+    from lazyflow.operators.cacheMemoryManager import CacheMemoryManager
+    CacheMemoryManager()
 
 
 def _prepare_lazyflow_config(parsed_args):
--- a/tests/test_workflows/test_all/testAllHeadless.py
+++ b/tests/test_workflows/test_all/testAllHeadless.py
@@ -120,7 +120,16 @@ class TestHeadlessWorkflowStartupProjectCreation(object):
         parsed_args, workflow_cmdline_args = ilastik_main.parser.parse_known_args()
         shell = ilastik_main.main(
             parsed_args=parsed_args, workflow_cmdline_args=workflow_cmdline_args, init_logging=False)
+        
+        self.check_cache_manager_started()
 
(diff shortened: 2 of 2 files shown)


clbarnes:
I thought it must be in there somewhere - I should have rebased the test first to check.

However, I suspect it is still possible to start a workflow without kicking off the memory manager thread, because it explains the behaviour I see in my https://github.com/clbarnes/skeleton_synapses/blob/master/debug_scripts/minimal_prediction.py : with this script, the RAM usage increases linearly with the number of ROIs processed regardless of the LAZYFLOW_TOTAL_RAM_MB. Querying `Memory` shows that the limit has been set, but the RSS blows through the limit with no signs of slowing down (blue is RSS, orange is limit).

linear

If I instantiate `CacheMemoryManager()` in the script, the limitation comes into effect and the RSS plateaus out below the limit; the same is true if I set the refresh interval (which instantiates the cache memory manager in the `pre_init`).

k-dominik:
wew, that plot looks interesting. The yellow line looks too perfect though, don't you think?

I have pushed a branch that modifies your `minimal_prediction.py` script to setup the cleanup interval and also to configure the memory: https://github.com/k-dominik/skeleton_synapses/tree/add-log

Also, I added some logging that makes it easier to see the internal cleanups

k-dominik:
sorry, closed it by mistake

clbarnes:
Thanks for your changes!

Sorry if I wasn't clear - the orange line is meant to be perfect as it's just there for reference, to show what I'd set the RAM limit to. Forgot to `ax.legend()`.

Without explicitly doing `Memory.setAvailableRAM`, that gets picked up fine from the environment variable. And with your edits, the limit will be correctly enforced because the `CacheMemoryManager` has been instantiated manually and therefore started (this is pretty much what I'd done in my testing). The same is true if the environment variable `LAZYFLOW_STATUS_MONITOR_SECONDS` is set, because the `CacheMemoryManager` is then instantiated as part of `_prepare_lazyflow_config`: the RAM limit is then enforced. Without either of these things, the RAM usage grows linearly as above.

My suspicion is that somehow I'm sidestepping the first instantiation of `CacheMemoryManager`. I'll do some more testing to try to prove this - I'll try some logging too (my first attempt I just got swamped with messages, so I'll do some more in-depth handler control).

k-dominik:
ahye, okay that explaines the line. I still don't think that there is a need to instantiate the `CacheMemoryManager`. Tiy cab verify that by deleting the `CacheManager().setRefreshInterval(status_interval_secs)` line.
then it will clean up the caches less often and I would suspect, that it overshoots the 2GBs then sometimes.

clbarnes:
OK, I can't replicate the results I thought I was getting yesterday so I must have been doing something wrong. I suspect it's that generally when I'm running my major script, I'm `source`ing one thing which `source`s another which `export`s the variables, and then `exec`s the python shell, which then spawns a `Process` which spawns a `Process` which actually uses the ilastik shell - the environment variables must be getting lost along the way.

I think the source of error in my testing is that environment variables added with `os.putenv` don't actually get picked up by later calls to `os.environ.__getitem__` or `os.getenv`, so they get ignored.

PR #1803, opened 2018-07-29
Add window-leveling to object classification, labeling and counting

### LINKED ISSUE #1754: Add window-leveling to object classification
The window-levelling button is only present in pixel classification, it should be made a generic property of the labelingGui

paulhfu:
fixes #1754
Window-Leveling should be in every labeling-gui available.


k-dominik:
Very nice, checked it in object classification, tracking, counting!

Since all the Gui classes you edited here derive from `LabelingGui`, I still wonder whether we could somehow get away with having something like a `_setupLayers` setting up the window-leveling bit for the raw data in `LabelingGui` and just call it from the respective child classes `setupLayers`; what do you think @paulhfu

paulhfu:
Thanks for the comment @k-dominik !
Not sure if I get this right. But, LabelingGui already has a setupLayers which is called in the respective child's setupLayers. It also already contains all setup for the window-leveling. This setup is just never executed since on instantiation of  LabelingGui no rawInputSlot object is given to LabelingGui which is required for the setup. 
It would certainly make more sense to set everything in the base class. I will do that with the next commit.

k-dominik:
hey @paulhfu ,

the commits look a bit odd (duplicate commit messages), maybe we should try to resolve this together.

ilastik/applets/counting/countingGui.py
```diff
@@ -647,10 +651,19 @@ def setupLayers(self):
...
         if inputDataSlot.ready():
-            inputLayer = self.createStandardLayerFromSlot( inputDataSlot )
-            inputLayer.name = "Input Data"
-            inputLayer.visible = True
-            inputLayer.opacity = 1.0
+            inputLayer = None
+            for i in range(len(layers)):
+                if layers[i].name == "Raw Input":
+                    inputLayer = layers[i]
+                    layers[i], layers[-1] = layers[-1], layers[i]
```
k-dominik: Hey, for things like these, where you swap stuff, it would be really helpful to add a line of comment on why (not obvious to me)

ilastik/applets/labeling/labelingGui.py
```diff
@@ -48,6 +48,8 @@
 
 from ilastik.applets.labeling.labelingImport import import_labeling_layer
 
+from volumina.api import \
```
k-dominik: I don't know if you committed some auto changes here in this file in general (by some linter)... But our current agreement is not to do such changes because it's easier to navigate file history.
In any case I would also _always_ opt for parentheses for multi-line imports such as
```python
from volumina.api import (
    GrayscaleLayer, ColortableLayer, LazyflowSinkSource
)
```

paulhfu:
Hi @k-dominik,
Good that you noticed that. When i moved the init of the raw layer to the baseclass, i changed the order of layers in the layerstack. In counting there where obvious problems which is why i reordered it there. But the update problem in the other workflows also result from that. So, either i apply the reordering which i did in counting also to the other workflows or i find out why the layer stack has to be in this static order and change it. We can just talk Monday about that.

k-dominik:
@paulhfu, so should I give this another review? Really want to merge it soon :)

paulhfu:
@k-dominik, sure go ahead.

ilastik/applets/objectClassification/labelingDrawer.ui
```diff
@@ -153,7 +153,21 @@
...
        </item>
+	   <item>
+		<widget class="QToolButton" name="thresToolButton">
+		 <property name="text">
+		  <string>...</string>
+		 </property>
+		</widget>
+	   </item>
        <item>
+        <widget class="QToolButton" name="thresToolButton">
```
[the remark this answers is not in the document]
paulhfu: You're totally right. I removed it.

paulhfu:
Hey @k-dominik, rebase is done

PR #1818, opened 2018-08-22
fixed loading of multiple datasets

k-dominik:
A user made me aware that selecting multiple (n) datasets in dataselection applet (add new -> add separate image(s)) resulted in a single added image and (n-1) empty added lanes.
I fixed this by not returning directly when a dataset is successfully added. This is the "old", working behaviour.
Furthermore, I made sure, that each path returns the correct value.

@FynnBe , since this could potentially mess with your new filter resizing dialog thing, is there something I should test other than the following (which I have already confirmed works):

* loaded a 3D dataset (big), selcted all the filters, then added another one with `10` `z`-slices only. The dialog appears and lets me modify the selected filters appropriately.

### DIFF
--- a/ilastik/applets/dataSelection/dataSelectionGui.py
+++ b/ilastik/applets/dataSelection/dataSelectionGui.py
@@ -597,20 +597,18 @@ class DataSelectionGui(QWidget):
         # Resize the slot if necessary            
         if len( opTop.DatasetGroup ) < endingLane+1:
             opTop.DatasetGroup.resize( endingLane+1 )
-        
-        ret_val = False
+
         # Configure each subslot
         for laneIndex, info in zip(list(range(startingLane, endingLane+1)), infos):
             try:
                 self.topLevelOperator.DatasetGroup[laneIndex][roleIndex].setValue( info )
-                ret_val = True
             except DatasetConstraintError as ex:
                 return_val = [False]
                 # Give the user a chance to fix the problem
                 self.handleDatasetConstraintError(info, info.filePath, ex, roleIndex, laneIndex, return_val)
                 if return_val[0]:
                     # Successfully repaired graph.
-                    ret_val = True
+                    continue
                 else:
                     # Not successfully repaired.  Roll back the changes
                     opTop.DatasetGroup.resize(originalSize)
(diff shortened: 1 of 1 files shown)


ilastik/applets/dataSelection/dataSelectionGui.py
```diff
@@ -599,31 +598,35 @@ def _configureOpWithInfos(self, roleIndex, startingLane, endingLane, infos):
...
+                    return False
             except Exception as ex:
                 msg = "Wasn't able to load your dataset into the workflow.  See error log for details."
                 log_exception( logger, msg )
                 QMessageBox.critical( self, "Dataset Load Error", msg )
                 opTop.DatasetGroup.resize( originalSize )
+                return False
 
-        return False
+        return ret_val
```
FynnBe: Maybe it is clearer, to simply return True here and remove 'ret_val', as 'ret_val' will always be True, if the interpreter gets down here and no 'return False' is hit before...

FynnBe:
Maybe add two 'too' small datasets at the same time and check that the it's not popping up again if the second one is in line with the corrections done for the firstly added and the dialog is still popping up, if the second one is even more restrictive.
I would expect this to function properly as is, but those are 'famous last words'...

k-dominik:
I did the following:

* added two additional datasets, the second additional one even smaller than the first additional one.

After resolving the first one, the following popup only contained the "bad" scale for the second one. I think this also demonstrates that it would work with two smaller datasets of the same size without having to modify the selected features for each one.

ilastik/applets/dataSelection/dataSelectionGui.py
```diff
@@ -598,32 +597,34 @@ def _configureOpWithInfos(self, roleIndex, startingLane, endingLane, infos):
...
                 self.topLevelOperator.DatasetGroup[laneIndex][roleIndex].setValue( info )
-                return True
             except DatasetConstraintError as ex:
                 return_val = [False]
                 # Give the user a chance to fix the problem
                 self.handleDatasetConstraintError(info, info.filePath, ex, roleIndex, laneIndex, return_val)
                 if return_val[0]:
                     # Successfully repaired graph.
-                    return True
+                    continue
```
FynnBe: I don't see the purpose of this 'continue', but it doesn't hurt either.

k-dominik: it's verbose. I guess that was kind of forgotten the last time, that it should continue here, not return something

FynnBe: Yes. It is at the end of the for loop though, and thus redundant, but emphasising this explicitly actually makes sense. Sorry for the hiccup

PR #1856, opened 2018-10-08
Save sigma in counting workflow

### LINKED ISSUE #1753: Counting: sigma turns to 2.5 after deserialization
A user is observing that the value for sigma is reverting back to the default 2.5 after he deserializes his project.

emilmelnikov:
* Include `OpTrainCounter.Sigma` in `CountingSerializer`
* Refactor `__init__()` formatting and `isDirty()` logic in `CountingSerializer`

Close #1753.

### DIFF
--- a/tests/test_applets/objectCounting/testObjectCountingGui.py
+++ b/tests/test_applets/objectCounting/testObjectCountingGui.py
@@ -147,7 +147,7 @@ class TestObjectCountingGui(ShellGuiTestCaseBase):
             gui = countingClassApplet.getMultiLaneGui()
             opCount = countingClassApplet.topLevelOperator
 
-            opCount.opTrain.Sigma.setValue(type(self).COUNTING_SIGMA)
+            opCount.opTrain.Sigma.setValue(self.COUNTING_SIGMA)
  
             # Select the labeling drawer
             self.shell.setSelectedAppletDrawer(COUNTING_APPLET_INDEX)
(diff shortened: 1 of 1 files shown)


ilastik/applets/counting/countingSerializer.py
```diff
@@ -212,42 +212,65 @@ def _deserialize(self, group, slot):
...
-
-
-        super(CountingSerializer, self).__init__(projectFileGroupName,
-                                                            slots=slots)
+        self.predictionSlot = SerialPredictionSlot(
+            operator.PredictionProbabilities,
+            operator,
+            name='Predictions',
+            subname='predictions{:04d}',
+        )
```
FynnBe: Why are you changing this? this is even pep8 correct, or am I missing smth?

ilastik/applets/counting/countingSerializer.py
```diff
@@ -212,42 +212,65 @@ def _deserialize(self, group, slot):
...
-                               subname="rois{:04d}"),
-                 SerialBoxSlot(operator.opTrain.BoxConstraintValues,operator.opTrain,
-                              name="Values",
-                               subname="values{:04d}"),
-                 SerialBoxSlot(operator.boxViewer.rois, operator.boxViewer,
-                              name="ViewRois",
-                              subname="viewrois{:04d}"),
-                 SerialCountingSlot(operator.Classifier,
-                                      operator.classifier_cache,
-                                      name="CountingWrappers")
```
FynnBe: ... I see some indentation are indeed messed up.. ok then^^

ilastik/applets/counting/countingSerializer.py
emilmelnikov: This is basically for the consistency with the `slots`, and `slots` has multiple constructor calls with long parameter lists, so I thought that hanging indents would be more maintainable.

k-dominik:
@emilmelnikov do you think you could add a test for this as well?

tests/test_applets/objectCounting/testObjectCountingGui.py
```diff
@@ -50,6 +48,10 @@
 logger.setLevel(logging.DEBUG)
 
 
+# Sample Sigma value for OpCounting.opTrain (OpTrainCounter).
+COUNTING_SIGMA = 4.2
```
k-dominik: Not a huge fan of global variables, since it is used in one class only, you can make it a class variable in `setupClass`

emilmelnikov: Got it, although I don't see any harm in a module-level constant used in a test module.

PR #1876, opened 2018-10-30
Fix halloween

k-dominik:
extract to a temporary directory rather than the installation directory, otherwise admin permissions are needed on win


k-dominik:
okay, I confirmed that it will work, off to `rc2`

PR #1881, opened 2018-11-05
Fix typo in conda package name

m-novikov:


k-dominik:
thx, have you gone through the whole thing already? Otherwise I'd keep it open in case you find something else :)

k-dominik:
very good :)

Issue #1408, opened 2017-01-18
Automatic object naming in carving


akreshuk:
Especially if there
are many objects that one would like to "carve out" of an image volume
it is rather annoying that you have to save each object manually by
clicking the "save segmented object" button and then typing in a label.
It would be very helpful if the user would have the option to automate
this process to some extent so that each object is saved automatically
under a certain name (which the user could define at the beginning)
followed by an automatically incremented number.

PR #1882, opened 2018-11-05
Generate default name for objects in carving. Closes: #1408

m-novikov:


ilastik/workflows/carving/carvingGui.py
```diff
@@ -245,33 +246,46 @@ def onSegmentButton(self):
...
+
+    def getObjectNames(self):
+        return self.topLevelOperatorView.AllObjectNames[:].wait()
+
     def saveAsDialog(self, name=""):
         '''special functionality: reject names given to other objects'''
+        namesInUse = self.getObjectNames()
+
+        def generateObjectName():
+            return "{} {}".format(self.objectPrefix, len(namesInUse) + 1)
```
k-dominik: just a matter of taste, but I prefer the [F-string syntax](https://www.python.org/dev/peps/pep-0498/)...
```suggestion
            return f"{self.objectPrefix} {len(namesInUse) + 1}"
```

m-novikov:
> Once I reach 10 objects, the sorting gets off (as it's alphanumerical) object 10 will be shown before object 2. That is also not really desirable.
I suppose we can add 0 padding then. If someone gets more than 99 elements with same prefix sorting will be minor problem.

k-dominik:
> > Once I reach 10 objects, the sorting gets off (as it's alphanumerical) object 10 will be shown before object 2. That is also not really desirable.
> 
> I suppose we can add 0 padding then. If someone gets more than 99 elements with same prefix sorting will be minor problem.

Come to think of it, right now we do know that the number is added rightmost in the string. So padding would not really be necessary - however, I think it is more user-friendly to pad for readability with other tools. Do you think something with a `{placeholder}`, like we do in the export dialog, for the user to specify the number format could make sense?

k-dominik:
This is very nice. I guess the cherry for the users would be to have an explanatory tooltip

PR #1893, opened 2018-11-22
Pytest runner

m-novikov:


tests/conftest.py
```diff
@@ -0,0 +1,228 @@
...
+
+    bucket = [None]
+
+    def testfunc():
+        try:
+            return pyfuncitem.obj()
+        except Exception:
+            bucket[0] = sys.exc_info
+
+    thread = threading.Thread(target=testfunc)
```
emilmelnikov: Is it possible to use futures and thread pools from [`concurrent.futures`](https://docs.python.org/3/library/concurrent.futures.html) here?

tests/conftest.py
```diff
@@ -0,0 +1,228 @@
...
+            parent = item.parent
+            parent._failed = item
+            if call.excinfo.errisinstance(TimeoutException):
+                rep.wasxfail = "reason: Timeout"
+                rep.outcome = "skipped"
+
+
+def pytest_runtest_setup(item):
+    """
+    Handle chain fails for GUI tests
```
k-dominik: it would be nice to add a few words here, what is so special about gui test chains, and why they need handling.

tests/conftest.py
```diff
@@ -0,0 +1,228 @@
...
+    """
+    Handle reports and chain fails for GUI tests
+    """
+    outcome = yield
+
+    if is_gui_test(item):
+        rep = outcome.get_result()
+
+        if rep.failed and call.excinfo is not None:
+            parent = item.parent
```
k-dominik: please comment here, why this has to be set up explicitly

tests/conftest.py
```diff
@@ -0,0 +1,228 @@
...
+    Defines protocol for legacy GUI test execution
+    """
+    if not is_gui_test(pyfuncitem):
+        return
+
+    bucket = [None]
+
+    def testfunc():
+        try:
+            return pyfuncitem.obj()
```
k-dominik: what does is the `.obj()` in this case?

ilastik/utility/itertools.py
```diff
@@ -0,0 +1,31 @@
+import itertools
+
+
+not_set = object()
+
+
+def pairwise(iterable, tail=not_set):
```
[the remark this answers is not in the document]
m-novikov: more_itertools has `pairwise` function but I need one that provides me with a trailing tail.
See examples in the tests

pytest.ini
```diff
@@ -1,2 +1,4 @@
 [pytest]
 python_files = test*.py
```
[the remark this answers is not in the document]
m-novikov: It would work, but `cd tests` is workaround to another problem with hydra library. We need to update ilastik deps with this one.

tests/conftest.py
m-novikov: Calls an underlying test function. Added a comment.

tests/conftest.py
m-novikov: Added clarification [comment](https://github.com/ilastik/ilastik/pull/1893/commits/ea14a2b0cc18532e9411c25373a73411f6ae8885)

tests/conftest.py
```diff
@@ -0,0 +1,228 @@
...
+class GuiTestSuite:
+    """
+    GUI test suite runner
+    poll function should be scheduled from timer for periodic updates
+    """
+    def __init__(self, queue: queue.Queue, shell):
+        self._queue = queue
+        self._shell = shell
+
+        self._current = None
```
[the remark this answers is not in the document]
m-novikov: Renamed

tests/conftest.py
m-novikov: Switched to concurrent futures for timeouts, thanks.

ilastik/utility/itertools.py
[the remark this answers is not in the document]
emilmelnikov: There is a `stagger` function that does approximately the same:
```python
>>> list(more_itertools.stagger(range(3), offsets=(0, 1), longest=True))                                                                                                                                                                                                       
[(0, 1), (1, 2), (2, None)]
```

Actually, I am not sure whether introducing a dependency just for this case makes sense.
However, `more_itertools` seems quite usable in the other parts of the project as well.

k-dominik:
:pray:

PR #1900, opened 2018-11-29
Applet Bar improvement

m-novikov:
This is an implementation of numbers and open/closed icons in applet bar.
The current implementation of AppletDrawerToolBox had some problems, so I was forced to split it into AppletBarMngr that is responsible for tracking applet and AppletBarToolBox that contains presentation logic.
What's left:
[ ] Write tests for AppletBarMngr
[ ] Write tests for AppletDrawerToolBox
[ ] Maybe change icons to something better


akreshuk:
could you post a screenshot of what it looks like now?

m-novikov:
ilastik_applet_bar_numbers_collapsible

m-novikov:
It is possible to do it without icons, only using UTF-8 symbols. But the added benefit of icons is that it is the only way to make tab area bigger. Currently, it is quite narrow for a header.

akreshuk:
Looks very nice!

ilastik/shell/gui/ilastikShell.py
```diff
@@ -1064,9 +1066,8 @@ def handleAppletBarItemExpanded(self, modelIndex):
         """
         The user wants to view a different applet bar item.
         """
-        drawerIndex = modelIndex
-        if drawerIndex != -1:
-            self.setSelectedAppletDrawer(drawerIndex)
+        print("Switch to", modelIndex)
```
k-dominik: should be removed before merge

PR #1916, opened 2018-12-13
preserve index consistency in Gui dataset tables

### LINKED ISSUE #1857: Error loading file in multicut when removing and adding lanes
Hi!

The error was kinda weird because the message box that popped up was empty except for the title. See attached screenshot.

error loading file

What triggered the error was: 

* I opened an existing multicut project
* Removed some of the lanes completely
* Added several new images in the Raw Data tab
* Added a single probabilities image to one of the new lanes in the Probabilities tab

After I just confirmed the error, however, I could just add the image in a second try and everything worked fine afterwards.

This was found in the log:

```
ERROR 2018-10-08 13:47:03,137 log_exception 84285 140327745288000 Traceback (most recent call last):
  File "/g/kreshuk/snajder/ilastik-1.3.0-Linux/ilastik-meta/ilastik/ilastik/applets/dataSelection/dataSelectionGui.py", line 399, in addFiles
    self.addFileNames(fileNames, roleIndex, startingLane)
  File "/g/kreshuk/snajder/ilastik-1.3.0-Linux/ilastik-meta/ilastik/ilastik/applets/dataSelection/dataSelectionGui.py", line 455, in addFileNames
    startingLane, endingLane = self._determineLaneRange(fileNames, roleIndex, startingLane)
  File "/g/kreshuk/snajder/ilastik-1.3.0-Linux/ilastik-meta/ilastik/ilastik/applets/dataSelection/dataSelectionGui.py", line 503, in _determineLaneRange
    assert startingLane < len(self.topLevelOperator.DatasetGroup)
AssertionError
```

paulhfu:
store and update current index of buttons in Gui dataset table within the button object.
This fixes #1857

### DIFF
--- a/ilastik/applets/dataSelection/addFileButton.py
+++ b/ilastik/applets/dataSelection/addFileButton.py
@@ -63,7 +63,7 @@ class AddFileButton(QPushButton):
             "/../../shell/gui/icons/16x16/actions/list-add.png"),
             "Add..." if new == False else "Add New...", parent)
 
-        self.index = index  # This is a call to the setter function  self.index
+        self._index = index
         # drop down menu for different add options
         menu = QMenu(parent=self)
         menu.addAction("Add separate Image(s)...").triggered.\
--- a/ilastik/applets/dataSelection/addFileButton.py
+++ b/ilastik/applets/dataSelection/addFileButton.py
@@ -53,11 +53,11 @@ class AddFileButton(QPushButton):
 
     def __init__(self, parent, *, index=None, new=False):
         """
-        -- ``new`` - boolean parameter to indicate if this button is used to
-            add new lanes or files to new roles corresponding to an
-            existing lane (such as prediction maps)
-        -- ``index`` -  QModelIndex parameter, index of the
-            gui dataset table cell to which this button is added
+        Args:
+            parent (QWidget): Parent widget
+            index (QModelIndex): Index of the gui dataset table cell to which this button is added
+            new (bool): Indicating if this button is used to add new lanes or files to new roles
+            corresponding to an existing lane (such as prediction maps)
         """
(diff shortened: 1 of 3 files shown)


ilastik/applets/dataSelection/addFileButton.py
```diff
@@ -61,6 +61,7 @@ def __init__(self, parent, new=False):
             "/../../shell/gui/icons/16x16/actions/list-add.png"),
             "Add..." if new == False else "Add New...", parent)
 
+        self.index = index
```
k-dominik: We have discussed this line a few weeks ago if I recall and I'm opting to quote the zen of python here:

```
readability counts
```

Please don't use properties in the constructor. For me properties are to interact with the object from the outside. It's kind of weird to do it from within.

also, with the default argument `None` this produces an invalid object (accessing `index` afterwards will fail). What are you trying to do achieve with the type check for `QModelIndex`? Setting it to a different type will leave the Object in incomplete state (`__index` is not set).

ilastik/applets/dataSelection/addFileButton.py
```diff
@@ -77,3 +78,12 @@ def __init__(self, parent, new=False):
                     self.addRemoteVolumeRequested.emit)
 
         self.setMenu( menu )
+
+    @property
+    def index(self):
+        return self.__index
```
k-dominik: the double undescore `__` is bad practice imo. Single underscore is enough to indicate one should not modify the variable from the outside...

ilastik/applets/dataSelection/addFileButton.py
paulhfu: I dont know if the next commit resolves this. Do you suggest to make _index a 'public' attribute and get rid of getter/setter?

PR #1926, opened 2019-01-07
Refactor startup script and command-line parsing

emilmelnikov:
* Refactor `--clean_paths` option implementation
* Modify GUI launch command-line handling
* Remove direct access to argument parser instance
* Remove comments about modifying command-line args

### DIFF
--- a/bin/train_headless.py
+++ b/bin/train_headless.py
@@ -132,7 +132,7 @@ def generate_trained_project_file( new_project_path,
     
     # Manually configure the arguments to ilastik, as if they were parsed from the command line.
     # (Start with empty args and fill in below.)
-    ilastik_args = ilastik_main.parser.parse_args([])
+    ilastik_args = ilastik_main.parse_args([])
     ilastik_args.new_project = new_project_path
     ilastik_args.headless = True
     ilastik_args.workflow = 'Pixel Classification'
--- a/examples/example_python_client.py
+++ b/examples/example_python_client.py
@@ -20,7 +20,7 @@ os.environ["LAZYFLOW_TOTAL_RAM_MB"] = "2000"
 
 # Programmatically set the command-line arguments directly into the argparse.Namespace object
 # Provide your project file, and don't forget to specify headless.
-args = ilastik_main.parser.parse_args([])
+args = ilastik_main.parse_args([])
 args.headless = True
 args.project = '/Users/bergs/MyProject.ilp' # REPLACE WITH YOUR PROJECT FILE
 
--- a/ilastik.py
+++ b/ilastik.py
@@ -18,13 +18,20 @@
 #
 # See the LICENSE file for details. License information is also available
 # on the ilastik web site at:
-#		   http://ilastik.org/license.html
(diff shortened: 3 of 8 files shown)


emilmelnikov:
Users shouldn't read the source code to find out how to use the app, and the reference can be obtained with the usual `python ilastik.py --help`, which is always up to date.

If there is a need for invocation examples, then it would be better to include them into README, Sphinx docs or even as an `--help` epilogue, because they will be more visible. However, I am not sure whether we need this, because the majority of our users never use the command line.

k-dominik:
the comments were meant for devs

emilmelnikov:
Do developers really need these comments? It seems trivial to find out where the `ArgumentParser` is defined and how it works. If you think the examples are valuable, I suggest to move them to Sphinx docs, or to the bottom of the help message (that is, they would be printed together with the list of options when invoking `--help`).

stuarteberg:
The comments aren't there for reference, per se, but to make it convenient to test ilastik from an interactive IDE, (e.g. Eclipse PyDev, PyCharm). In an IDE, it can be a minor hassle to configure command-line options, since it involves editing the project preferences, etc.  Usually, it's just easier to edit the "main" file.  And since it's something I used to do frequently, it was nice to have the common options handy.

emilmelnikov:
@stuarteberg Interesting, I haven't thought about this type of workflow before. Please look at db551f478de948f9503213c1ccf8bb65cc36cec1 that adds support for argument files. Will this work for your case? If specifying `@args.txt` manually is inconvenient too, we could add it to args by default if the file `args.txt` exists.

stuarteberg:
This approach is a little less straightforward, but it does make `ilastik.py` a little cleaner, which is nice.  If the other ilastik devs are satisfied, then it's fine by me.

k-dominik:
@emilmelnikov while this approach works, I think it introduces a "feature" we want to implement differently at some point. I already see users using it and that doesn't feel right. At some point we should allow for something like `json`, or `toml` to be used.

I would opt for reverting the last changes and just leaving the comments inside. Otherwise this PR is fine.

emilmelnikov:
The `@argfile` is not a config file per se, it's mainly used for
1. passing a large number of arguments without exceeding the shell argument list length limit (I've seen this in some popular unix utility)
2. specifying arguments in environments where it is not convenient to invoke the app with the arguments directly (cf. JVM-based apps like IDEA with their `java.vmoptions` file).

However, I agree that the current state of CLI interface does not look good. If you still think that we should wait until a good CLI substitute arrives, then we can revert that "feature" and create a feature request.

k-dominik:
> ```python
> # The name makes it obvious that it is not intended for end-users.
> p = pathlib.Path("internal-startup-options")
> if p.exists():
>     sys.argv[1:1] = shlex.split(p.read_text()), comments=True)
> ```

this is a sound proposition I think. With that we are free to use it and once we feel like we have something better that should be exposed to users we can silently remove it.

k-dominik:
I thought this one was merged ages ago...

PR #1928, opened 2019-01-09
Create a feature request template for GitHub

emilmelnikov:
A feature request template, similar to a bug report template,
improves GitHub issue submission UI.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,32 @@
+---
+name: Feature request
+about: Let us know about the missing features
+---
+
+## Description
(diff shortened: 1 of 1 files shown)


emilmelnikov:
1. _Feature Request_ is a more or less established term, in the spirit of pull/merge requests (cf. [RStudio Feature Requst Guide](https://github.com/rstudio/rstudio/wiki/Writing-Good-Feature-Requests)).
2. My initial intention was to motivate the author to think about the cost of the feature they are proposing, because not every feature worth implementing. However, you are right that the users shouldn't care about that.

Do we also need a "disclaimer" note, something like "Many thanks for your input, but we don't promise that this request will be implemented"?

PR #1945, opened 2019-01-29
Use dot as decimal separator to be consistent

m-novikov:


ilastik/workflows/carving/preprocessingDrawer.ui
```diff
@@ -26,7 +26,7 @@
    <string>PreprocessDrawer</string>
   </property>
   <property name="locale">
-   <locale language="German" country="Germany"/>
+   <locale language="English" country="UnitedStates"/>
```
k-dominik: I don't know if it wouldn't be better to remove the locale tag alltogether, since we don't use it anywhere else and just let the system decide how to show stuff.

LICENSE
```diff
@@ -1,6 +1,6 @@
   ilastik: interactive learning and segmentation toolkit
 
-      Copyright (C) 2011-2014, the ilastik developers
```
emilmelnikov: Do we really need date ranges in copyright notices?

Some people [suggest](https://softwareengineering.stackexchange.com/questions/130478/what-copyright-date-for-an-update-to-an-open-source-project-from-last-year) just leaving the original publication date.

m-novikov: Propose update for now and discuss it on ilastik meeting. It should be consistent across licenses in ilastik files

k-dominik:
just a minor thing: could you also please make the labels consistent as described in #1950 while you are at it :)

PR #1951, opened 2019-01-30
Fix carving mem

### LINKED ISSUE #1852: Carving, MemoryError should be handled more gracefully, or even better: avoided alltogether
**Describe the bug**

Preprocessing in carving will throw a `MemoryError` when data is too big for the RAM

**Expected behavior**

At least give a clear error message on what is happening, with maybe some suggestion on how to resolve it (use a subset of the data, get a machine with more RAM)

**To Reproduce**
Within the Carving workflow
1. Add some 3d data in the data selection applet, I have used good old `sample_A_20160501.hdf`
2. Go to the preprocessing applet, choose "dark lines" and go with the defaults, hit _run_

**Error message/traceback**
please add tracebacks with formatting, such as

```pytb
# Some longish stuff left out, should be easy to reproduce yourself
  File "/.../ilastik-meta/lazyflow/lazyflow/slot.py", line 887, in __call__
    result_op = self.operator.execute(self.slot, (), self.roi, destination)
  File "/.../ilastik-meta/ilastik/ilastik/workflows/carving/opPreprocessing.py", line 237, in execute
    reduce_to=self.ReduceTo.value)
  File "/.../ilastik-meta/ilastik/ilastik/workflows/carving/carvingTools.py", line 140, in simple_parallel_ws
    res = bincount = numpy.bincount(overseg.ravel().astype('int64'))
MemoryError
```

**Screenshots**
error

**Desktop (please complete the following information):**
 - ilastik version: `1.3.2b3`
 - OS: `Ubuntu 18.04

constantinpape:
I think this fixes #1852. I don't have a dataset to reproduce the first `MemoryError` that @k-dominik reported though.
Also haven't properly tested all the changes yet.


constantinpape:
I have made quite some changes to speed up and clean up carvingTools.py.
There are still some issues / possible improvements I see in the carving workflow:
- The nifty.rag functions are not exposed for the right combinations of datatypes, that's why I had to resolve to [this construction](https://github.com/ilastik/ilastik/blob/fix_carving_mem/ilastik/workflows/carving/carvingTools.py#L154). If we rebuild nifty for ilastik this should be fixed.
- An additional size filter for watershed fragments could speed up preprocessing quite a bit.
- In [OpFilter](https://github.com/ilastik/ilastik/blob/fix_carving_mem/ilastik/workflows/carving/opPreprocessing.py#L50) only some of the filters are computed block-wise and they all use vigra and not fastfilters.
- As far as I can see [OpSimpleWatershed](https://github.com/ilastik/ilastik/blob/fix_carving_mem/ilastik/workflows/carving/opPreprocessing.py#L175) is just there doing nothing.

constantinpape:
Ok, I addressed some of the points from above.
What still remains:
- datatypes for the nifty rag: we can resolve this once re-building nifty
- size filtering for initial watershed: Don't know if this a good idea, because we would need to expose this to the user and probably wan't to avoid having another parameter.

This is good to merge from my side. Let me know what you think @k-dominik

constantinpape:
@k-dominik I realized that there is actually value in what vigra does for the hessianOfGaussian (implementing versions that only return one of the eigenvalues), because it saves memory
if this is discarded in the block-wise computation already.
Was really easy to implement this in python though.

k-dominik:
this is pending a dependency rebuild on our side

PR #1963, opened 2019-02-22
Move tracking export lugins to trackingExportApplet. 

m-novikov:
Plugins export shouldn't be available for manual tracking since all plugins assume the existence of hypothesisGraph which is absent in manual tracking.

### DIFF
--- a/ilastik/applets/tracking/base/trackingBaseDataExportApplet.py
+++ b/ilastik/applets/tracking/base/trackingBaseDataExportApplet.py
@@ -52,7 +52,7 @@ class TrackingBaseDataExportApplet( DataExportApplet ):
         title,
         is_batch: bool = False,
         default_export_filename: str = '',
-        doPluginExport: Optional[PluginExportCallable] = None
+        pluginExportFunc: Optional[PluginExportCallable] = None
     ):
         self.export_op = None
         self._default_export_filename = default_export_filename
@@ -66,7 +66,7 @@ class TrackingBaseDataExportApplet( DataExportApplet ):
             SerialDictSlot(self.topLevelOperator.AdditionalPluginArguments)
         ]
         self._serializers = [DataExportSerializer(self.topLevelOperator, title, extra_serial_slots)]
-        self._doPluginExport = doPluginExport
+        self._pluginExportFunc = pluginExportFunc
 
         super(TrackingBaseDataExportApplet, self).__init__(workflow, title, isBatch=is_batch)
 
@@ -79,7 +79,7 @@ class TrackingBaseDataExportApplet( DataExportApplet ):
 
     @property
     def supports_plugins(self):
-        return self._doPluginExport is not None
+        return self._pluginExportFunc is not None
 
     def getMultiLaneGui(self):
         if self._gui is None:
@@ -243,9 +243,10 @@ class TrackingBaseDataExportApplet( DataExportApplet ):
 
         self.progressSignal(-1)
 
-        status = self._doPluginExport(lane_index, filename, plugin, checkOverwriteFiles, argsSlot)
-
-        self.progressSignal(100)
+        try:
+            status = self._pluginExportFunc(lane_index, filename, plugin, checkOverwriteFiles, argsSlot)
+        finally:
+            self.progressSignal(100)
 
         if not status:
             return False
--- a/ilastik/applets/tracking/base/trackingBaseDataExportApplet.py
+++ b/ilastik/applets/tracking/base/trackingBaseDataExportApplet.py
@@ -22,17 +22,23 @@ from __future__ import absolute_import
 import os
 import logging
 
-from typing import Callable
+from typing import Callable, Optional
 
 from ilastik.applets.base.appletSerializer import SerialSlot, SerialDictSlot
 from ilastik.applets.dataExport.dataExportApplet import DataExportApplet
 from ilastik.applets.dataExport.dataExportSerializer import DataExportSerializer
 from ilastik.applets.tracking.base.opTrackingBaseDataExport import OpTrackingBaseDataExport
-from ilastik.plugins import pluginManager
+from ilastik.plugins import pluginManager, TrackingExportFormatPlugin
 from ilastik.utility import OpMultiLaneWrapper
+from lazyflow.slot import InputSlot
 from lazyflow.utility import format_known_keys, PathComponents, getPathVariants, make_absolute
 
 logger = logging.getLogger(__name__)
+
+PluginExportCallable = Callable[
+    [int, str, TrackingExportFormatPlugin, bool, InputSlot],
+    None
+]
 
 
 class TrackingBaseDataExportApplet( DataExportApplet ):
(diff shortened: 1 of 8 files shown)


k-dominik:
sorry about the force-push, but I figured you haven't pulled yet

ilastik/applets/tracking/base/trackingBaseDataExportApplet.py
```diff
@@ -185,4 +203,79 @@ def _configure_operator_with_parsed_args(cls, parsed_args, opTrackingDataExport)
         # configure super operator
         DataExportApplet._configure_operator_with_parsed_args(parsed_args, opTrackingDataExport)
 
+    @property
+    def op(self):
```
k-dominik: any compelling reason not to use the `self.toplevelOperator` property, implemented in `98` of this file?

ilastik/applets/tracking/base/trackingBaseDataExportApplet.py
```diff
@@ -185,4 +203,79 @@ def _configure_operator_with_parsed_args(cls, parsed_args, opTrackingDataExport)
...
+        logger.info("Export source is: %s", op.SelectedExportSource.value)
+        plugin_export_selected = op.SelectedExportSource.value == OpTrackingBaseDataExport.PluginOnlyName
+
+        if plugin_export_selected:
+            logger.info("Export source plugin selected!")
+            return self._export_with_plugin(lane_index, checkOverwriteFiles, op.SelectedPlugin.value)
+
+        return True
+
+    def getPartiallyFormattedName(self, lane_index: int, path_format_string: str) -> str:
```
k-dominik: this doesn't use `lane_index`, any reason to keep it?

ilastik/applets/tracking/base/trackingBaseDataExportApplet.py
```diff
@@ -19,19 +19,28 @@
...
+logger = logging.getLogger(__name__)
+
 
 class TrackingBaseDataExportApplet( DataExportApplet ):
     """
     This a specialization of the generic data export applet that
     provides a special viewer for tracking output.
     """
-    def __init__(self, workflow, title, is_batch=False, default_export_filename=''):
+    def __init__(self, workflow, title, is_batch=False, default_export_filename='', doPluginExport=None):
```
k-dominik: I assume you imported `Callable` to annotate this...

ilastik/applets/tracking/base/trackingBaseDataExportApplet.py
m-novikov: Fixed use lane index to fetch RawDatasetInfo. Not sure if it has any impact because manual tracking limits you to 1 dataset.

ilastik/applets/tracking/base/trackingBaseDataExportApplet.py
```diff
@@ -185,4 +216,75 @@ def _configure_operator_with_parsed_args(cls, parsed_args, opTrackingDataExport)
...
+            logger.error("Could not find selected plugin %s", pluginName)
+            return False
+
+        plugin = pluginInfo.plugin_object
+        logger.info("Exporting tracking result using %s", pluginName)
+
+        name_format = self.topLevelOperator.getLane(lane_index).OutputFilenameFormat.value
+        partially_formatted_name = self.getPartiallyFormattedName(lane_index, name_format)
+
+        if plugin.exportsToFile:
```
emilmelnikov:
```python
filename = partially_formatted_name
if plugin.exportsToFile and not os.path.basename(filename):
    filename = os.path.join(filename, 'pluginExport.txt')
```

OTOH, do we want to use the default file name if we get a directory name? If so, then we'd better to check whether `os.path.isdir(filename)` instead of hoping that the name we get ends with `/`.

ilastik/applets/tracking/base/trackingBaseDataExportApplet.py
```diff
@@ -185,4 +216,75 @@ def _configure_operator_with_parsed_args(cls, parsed_args, opTrackingDataExport)
...
+        ''' Takes the format string for the output file, fills in the most important placeholders, and returns it '''
+
+        raw_dataset_info = self.topLevelOperator.RawDatasetInfo[lane_index].value
+        project_path = self.topLevelOperator.WorkingDirectory.value
+        dataset_dir = PathComponents(raw_dataset_info.filePath).externalDirectory
+        abs_dataset_dir = make_absolute(dataset_dir, cwd=project_path)
+
+        nickname = raw_dataset_info.nickname.replace('*', '')
+        if os.path.pathsep in nickname:
+            nickname = PathComponents(nickname.split(os.path.pathsep)[0]).fileNameBase
```
emilmelnikov: I am not sure, but:
```python
nickname = pathlib.Path(nickname).stem
```

ilastik/applets/tracking/base/trackingBaseDataExportGui.py
```diff
@@ -156,10 +158,11 @@ def _chooseSettings(self):
     def _initAppletDrawerUic(self):
         # first check whether "Plugins" should be made available
         availableExportPlugins = self._getAvailablePlugins()
-        if len(availableExportPlugins) > 0:
+        if self._enable_plugins and len(availableExportPlugins) > 0:
```
emilmelnikov:
```suggestion
        if self._enable_plugins and availableExportPlugins:
```

ilastik/workflows/tracking/conservation/conservationTrackingWorkflow.py
```diff
@@ -107,9 +107,12 @@ def __init__( self, shell, headless, workflow_cmdline_args, project_creation_arg
...
 
         self.default_export_filename = '{dataset_dir}/{nickname}-exported_data.csv'
-        self.dataExportApplet = TrackingBaseDataExportApplet(self, 
-                                                             "Tracking Result Export", 
-                                                             default_export_filename=self.default_export_filename)
+        self.dataExportApplet = TrackingBaseDataExportApplet(
+            self,
+            "Tracking Result Export",
+            default_export_filename=self.default_export_filename,
+            doPluginExport=self._doPluginExport,
```
emilmelnikov: Since `_doPluginExport` is not used anywhere else, doesn't use the local state and is very simple, maybe define right it here as a local function? Ditto for the other definitions.

k-dominik:
thx @emilmelnikov for the comments. You raise good points, however, this actually has to be merged today (noon-ish). So if @m-novikov does not find the time to respond I'm still fine with this pr.

ilastik/workflows/tracking/conservation/conservationTrackingWorkflow.py
m-novikov: I prefer not to define functions inside constructors. Hard to find usages and extract in the future(prone to transform into closures).

ilastik/applets/tracking/base/trackingBaseDataExportApplet.py
m-novikov: Regarding `os.path.isdir`
This needs a bit more work, not only in this code path. We need to handle user provided paths consistently.

ilastik/applets/tracking/base/trackingBaseDataExportApplet.py
m-novikov: Not in this case.
If we have `dataset.hdf5/raw` as an input `stem` function will return `raw` instead of `dataset.hdf5`.
I think in the future we should split external/internal paths

k-dominik:
thank you both @m-novikov and @emilmelnikov for the work on this one, looking really good

PR #1973, opened 2019-03-06
Implementing mask atlas

Tomaz-Vieira:
Supersedes and fixes all observations from https://github.com/ilastik/ilastik/pull/1960
Depends on https://github.com/ilastik/lazyflow/pull/293 (new constructor for OpAxesReorder)
Depends on https://github.com/ilastik/lazyflow/pull/290 (cleanup and small fixes to pixel mask usage)

# Implements Masks for pixel classification and Atlases for Object Classification

## Refactoring along the way
Cleans up some of the code in PixelClassificationWorkflow and Object Classification Workflow by introducing SlotNameEnum, removing a lot of code replication and making better use of inheritance in the children of Object Classification Workflow. The SlotNameEnum removes some of the magical constants used throughout the code and ties the names of the slots to its indices. 

As noted in the comments from https://github.com/ilastik/ilastik/pull/1960, this is still not the best solution, and we should probably move over to something like `myOp.myOutput` and `myOp.myOtherOutput` rather than `myOp.Output[myenum.MY_OUTPUT]` and `myOp.Output[myenum.MY_OTHER_OUTPUT]`, but there is a lot of fixing required before we get to that. For the time being, it is still a bit more readable.

### Pixel Classification
Enables the use of masks, which avoids computations on ROIs completely covered by 0-valued pixels and discards predictions in the masked pixels of the blocks that actually were picked for computations.

#### Gui Usage:
Just load a mask using the Prediction Mask tab and go through the normal workflow. 0-valued pixels will be masked out.

#### Headless Usage:
```
time python3 ilastik/ilastik.py \
    --headless --project ~/testProjects/trainedFakeProject.ilp \
    --export_source="Probabilities"  --output_filename_format /tmp/headlessTest.h5 \
    --prediction_mask ~/SampleData/fakeMaskData/fakeMask.png  \
    --raw_data ~/SampleData/fakeMaskData/fakeData.png
```

### Object Classification
Enables the use of Atlases when doing object classification. Selecting an image atlas will cause the workflow to map object centers to the color on the same coordinates as the selected Atlas. When using Pixel + Object Classification, the areas of the atlas with 0 value will also be masked out of the pixel classification, just like in the standalone masked Pixel Classification Workflow.

If the user has selected an Atlas image for the workflow, then the exported object table will contain columns labeled `AtlasMapping_<channel index>`, which contain the values of all the channels in the atlas image in the center point of the object being described in that table line. The AtlasMapping values are implemented as part of the `Standard Object Features` and are therefore automatically written to the object table whenever an Atlas is given as input.

#### Gui Usage:
Just load an atlas using the Atlas tab and go through the normal workflow. 0-valued pixels will be masked out in the Pixel  Object Classification Workflow and object centers will be mapped to the colors in the atlas.

#### Headless Usage:
```
./ilastik.py --headless --project ~/myProj.ilp \
                --prediction_map ~/myPredictions.h5 \
                --raw_data ~/myRawData.png \
                --atlas ~/myAtlas.png \
                --table_filename=/tmp/myobjtable.csv  
```

### DIFF
--- a/ilastik/applets/objectExtraction/opObjectExtraction.py
+++ b/ilastik/applets/objectExtraction/opObjectExtraction.py
@@ -595,13 +595,14 @@ class OpRegionFeatures(Operator):
         atlas_mapping = numpy.zeros(num_center_points * num_channels_in_atlas)
         atlas_mapping = atlas_mapping.reshape(num_center_points, num_channels_in_atlas)
         for objIdx, center_coords in enumerate(region_centers):
+            rounded_coords = center_coords.round().astype(numpy.uint64)
             slicing = [0] * len(axes)
             slicing[axes['c']] = slice(None)
 
-            slicing[axes['x']] = round(center_coords[0])
-            slicing[axes['y']] = round(center_coords[1])
+            slicing[axes['x']] = rounded_coords[0]
+            slicing[axes['y']] = rounded_coords[1]
             if len(center_coords) == 3:
-                slicing[axes['z']] = round(center_coords[2])
+                slicing[axes['z']] = rounded_coords[2]
 
             atlas_value = atlasImage[slicing]
             atlas_mapping[objIdx] = atlas_value
--- a/ilastik/applets/objectExtraction/opObjectExtraction.py
+++ b/ilastik/applets/objectExtraction/opObjectExtraction.py
@@ -595,14 +595,15 @@ class OpRegionFeatures(Operator):
         atlas_mapping = numpy.zeros(num_center_points * num_channels_in_atlas)
         atlas_mapping = atlas_mapping.reshape(num_center_points, num_channels_in_atlas)
         for objIdx, center_coords in enumerate(region_centers):
-            slicing = numpy.zeros(len(axes)).astype(numpy.uint32)
+            slicing = [0] * len(axes)
+            slicing[axes['c']] = slice(None)
 
-            slicing[axes['x']] = center_coords[0]
-            slicing[axes['y']] = center_coords[1]
+            slicing[axes['x']] = round(center_coords[0])
+            slicing[axes['y']] = round(center_coords[1])
             if len(center_coords) == 3:
-                slicing[axes['z']] = center_coords[2]
+                slicing[axes['z']] = round(center_coords[2])
 
-            atlas_value = atlasImage[tuple(slicing)]
+            atlas_value = atlasImage[slicing]
             atlas_mapping[objIdx] = atlas_value
         return atlas_mapping
 
--- a/ilastik/applets/objectExtraction/opObjectExtraction.py
+++ b/ilastik/applets/objectExtraction/opObjectExtraction.py
@@ -114,6 +114,7 @@ def make_bboxes(binary_bbox, margin):
 class OpCachedRegionFeatures(Operator):
     """Caches the region features computed by OpRegionFeatures."""
     RawImage = InputSlot()
+    Atlas = InputSlot(optional=True)
     LabelImage = InputSlot()
     CacheInput = InputSlot(optional=True)
     Features = InputSlot(rtype=List, stype=Opaque)
@@ -135,6 +136,7 @@ class OpCachedRegionFeatures(Operator):
         # Hook up the actual region features operator
         self._opRegionFeatures = OpRegionFeatures(parent=self)
         self._opRegionFeatures.RawVolume.connect(self.RawImage)
+        self._opRegionFeatures.Atlas.connect(self.Atlas)
         self._opRegionFeatures.LabelVolume.connect(self.LabelImage)
         self._opRegionFeatures.Features.connect(self.Features)
 
@@ -304,6 +306,7 @@ class OpObjectExtraction(Operator):
 
     RawImage = InputSlot()
     BinaryImage = InputSlot()
+    Atlas = InputSlot(optional=True)
     BackgroundLabels = InputSlot(optional=True)
 
     # Bypass cache (for headless mode)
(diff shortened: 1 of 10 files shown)


ilastik/applets/objectExtraction/opObjectExtraction.py
```diff
@@ -576,7 +588,27 @@ def _augmentFeatureNames(self, features):
 
         return feature_names_with_default
 
-    def _extract(self, image, labels):
+    def _createAtlasMapping(self, region_centers:numpy.ndarray, atlasImage:numpy.ndarray):
+        axes = {tag.key:idx for idx, tag in  enumerate(atlasImage.axistags)}
+        num_center_points = len(region_centers)
+        num_channels_in_atlas = atlasImage.shape[axes['c']] if 'c' in axes else 1
+        atlas_mapping = numpy.zeros(num_center_points * num_channels_in_atlas)
```
k-dominik:
```suggestion
        atlas_mapping = numpy.zeros((num_center_points, num_channels_in_atlas))
```
will save you a line (generates the array in the right shape)

ilastik/applets/objectExtraction/opObjectExtraction.py
```diff
@@ -576,7 +588,27 @@ def _augmentFeatureNames(self, features):
...
+        atlas_mapping = atlas_mapping.reshape(num_center_points, num_channels_in_atlas)
```
k-dominik:
```suggestion

```

ilastik/workflows/objectClassification/objectClassificationWorkflow.py
```diff
@@ -420,17 +427,15 @@ def handleAppletStateUpdateRequested(self, upstream_ready=False):
...
-                    all([sub[i].ready() for i in range(nRoles)])
-        else:
-            input_ready = False
-
-        return input_ready
+    def _inputReady(self):
+        image_group_slot = self.dataSelectionApplet.topLevelOperator.ImageGroup
+        for input_lane_slot in image_group_slot:
+            for role in self.InputImageRoles:
+                if role == self.InputImageRoles.ATLAS:
```
k-dominik: could be made generic if we had a concept of "optionalness" for the image roles

ilastik/workflows/objectClassification/objectClassificationWorkflow.py
```diff
@@ -551,17 +556,64 @@ class ObjectClassificationWorkflowPixel(ObjectClassificationWorkflow):
...
-                                                        "Input Data", 
-                                                        batchDataGui=False,
-                                                        forceAxisOrder=None, 
-                                                        instructionText=data_instructions )
-        opData = self.dataSelectionApplet.topLevelOperator
-        opData.DatasetRoles.setValue(['Raw Data'])
+    @property
+    def ExportNames(self):
+        class ExtraExportNames(SlotNameEnum):
+            PIXEL_PROBABILITIES = super(self.__class__, self).ExportNames.getNext()
```
k-dominik: This reorders the "tabs" in the input data applet gui. With "optional" roles we could probably automatically order those in a meaningful way.

ilastik/utility/slot_name_enum.py
```diff
@@ -0,0 +1,53 @@
+import enum
+from typing import List, Tuple
+from itertools import chain
+
+class SlotNameEnum(enum.IntEnum):
```
k-dominik: I don't know if this class should be extended, or one should think of something else, but in principle there are two types of data roles: 1) required (`raw_data`...), optional (`mask`...). Would be cool to represent this somehow

ilastik/workflows/objectClassification/objectClassificationWorkflow.py
```diff
@@ -670,38 +708,28 @@ class ObjectClassificationWorkflowBinary(ObjectClassificationWorkflow):
     workflowName = "Object Classification (from binary image)"
```
k-dominik: the `ObjectClassificationWorkflowBinary` lacks an implementation of `createInputApplets`, I assume that's some leftover stuff from previous comments (updating it to `abstactmethod`)

ilastik/utility/slot_name_enum.py
```diff
@@ -0,0 +1,53 @@
...
+    the human readable version that you get from instances of this enum:
+
+    e.g a value defined like so:
+        MY_DATA_SLOT = 123
+    will have its slotName rendered as
+        'My Data Slot'
+    """
+
+    @property
+    def slotName(self) -> str:
```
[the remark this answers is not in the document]
emilmelnikov: I would also add that implicitly deriving GUI labels from constant names is probably not a good idea, maybe it would be better to explicitly specify the "display name"? Something like `MY_PRECIOUS_SLOT = SlotDef(42, "My Precious Slot")`.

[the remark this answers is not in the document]
Tomaz-Vieira: There is so much wrong with these enums that there is almost no point investing more time into making it "better", as it is already fundamentally wrong. They are just a less messy implementation of current logic.

What we should eventually do is have slots be defined in the data selection operator  as proper slots, so that they can even be marked as `optional` instead of having them be subslots in the generic Input slot =/

ilastik/workflows/objectClassification/objectClassificationWorkflow.py
Tomaz-Vieira: Oh well... falling back to straighforward IntEnums which dictate the order of the tabs -.-

ilastik/utility/slot_name_enum.py
Tomaz-Vieira: We can eventually push that stuff into the opDataSelection, but that'll have to wait for the next release =/

ilastik/workflows/objectClassification/objectClassificationWorkflow.py
Tomaz-Vieira: No idea how i missed that, but yeah. Fixed.

ilastik/applets/objectExtraction/opObjectExtraction.py
Tomaz-Vieira: So  a few lines below there is this: `atlas_mapping[obj_idx] = atlas_value`; `atlas_value` is simply a sampling of the Atlas image, so it might come back as a multi-channel point if the user is using a multi-channel image. That's why I have to reshape the atlas_mapping, so that every item in it has as many channels as the sampling from the atlas

PR #1974, opened 2019-03-07
Prevent Qt imports using import hooks

emilmelnikov:
Instead of modifying the import path and using a dummy module,
add a special finder that blacklists PyQt5 in the headless mode.


m-novikov:
Great :)
Would be nice to move this to separate file (e.g. `ilastik.utility.imp`)
And add some tests, from top of my head something like this:
```python
import importlib

import pytest

from ilastik.utility import imp


def test_blacklist():
    import PyQt5 # Shoud not raise

    blacklist = imp.ImportBlacklist('PyQt5')
    blacklist.install()

    with pytest.raises(ImportError):
        importlib.reload(PyQt5)

    blacklist.uninstall()

    importlib.reload(PyQt5)
```

emilmelnikov:
Now it breaks a lot of tests because these failing tests import some workflows (that, it turn, import PyQt5), before running `ilastik_main.main`...

ilastik_main.py
```diff
@@ -57,19 +60,72 @@
...
+    enabled: bool
+
+    def __init__(self, *blacklist: str):
+        """Create a new blacklist."""
+        self.blacklist = blacklist
+        self.enabled = True
+        self._caller = None
+
+    def find_spec(self, fullname, _path, _target=None):
+        if (not self.enabled) or (fullname not in self.blacklist):
```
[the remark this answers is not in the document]
emilmelnikov: This attr was supposed to be used later: I started to work on the ilastik Python API, and wanted to have a switch for different contexts (that is, you should be able to create multiple contexts and use them independently), but now I understand that it's much trickier with module imports. So, it's better to remove it, at least for now.

emilmelnikov:
> Would be nice to move this to separate file (e.g. `ilastik.utility.imp`)

Yes, but I am a bit worried about the code that gets executed by importing `ilastik.utility.imp` itself (currently, just importing `ilastik` changes your `sys.path`).

m-novikov:
> > Would be nice to move this to a separate file (e.g. `ilastik.utility.imp`)
> 
> Yes, but I am a bit worried about the code that gets executed by importing `ilastik.utility.imp` itself (currently, just importing `ilastik` changes your `sys.path`).

But we can check it with a test

emilmelnikov:
OK, I've moved things to the separate module, because `ilastik_main` imports `ilastik` anyway. What should we do with the failing tests though?

m-novikov:
> OK, I've moved things to the separate module, because `ilastik_main` imports `ilastik` anyway. What should we do with the failing tests though?

Maybe we can create `pytest.fixture` that unloads PyQt5 and annotate headless tests with it.

k-dominik:
do we ever plan on shipping ilastik without `Qt`? I am honestly not sure how bad it is to import qt in headless mode, other than aesthetic considerations.

emilmelnikov:
> do we ever plan on shipping ilastik without `Qt`? I am honestly not sure how bad it is to import qt in headless mode, other than aesthetic considerations.

I thought we are banning Qt from the headless mode because we don't want windows to suddenly pop up during the execution... However, I agree that messing with the module system and/or placing dummy modules instead of `PyQt5` is too much magic in this case.

> But on the other hand might be nice to know that we can just unplugg it when ever switching the gui.

It seems that Qt is too deeply integrated into ilastik right now, replacing it would require a lot of work, and also very unlikely to happen (there is no real alternatives to Qt for desktop, widget-based apps).

Do you think we should remove this altogether?

akreshuk:
I vaguely recall there were some licensing considerations involved, as well as the possibility to run on clusters with no X installed at all. Not sure any of that is still relevant though.

tests/conftest.py
```diff
@@ -237,3 +236,28 @@ def _keyfunc(obj):
         return cls.__module__, cls.__name__, obj.name
 
     return sorted(iterable, key=_keyfunc)
+
+
+@pytest.fixture
+def noguimods():
```
k-dominik: so this is a bit awkward, since now two hard-coded lists of ignored modules exist in the code and here in the fixture. The `ImportBlacklist` class could potentially come in handy for other cases in the future, and I, in general really like to have this mechanism.
I don't think it it's easily possible to ~~hack~~ adapt the code, such that one would only have a single point of definition of blacklisted modules (especially given that they are only blacklisted on certain execution conditions...)

emilmelnikov: Now when I look at this PR I have to admit it looks over-engineered to me. Aside from testing, I also don't see any viable use cases for changing import blacklists more than once per process run. In other words, it seems cleaner to blacklist the needed modules just once as early as possible, and also don't test that at all, because additional loading/unloading complexity required for these tests will be used only for testing.

To summarize, I think we should leave just `ImportBlacklist.find_spec` (because it is needed for the blacklisting to work) and `if args.headless:blacklist_modules("PyQt")`, while removing context managers, exceptions and tests.

emilmelnikov: Alternatively, one can still test this thing by creating a tiny test script that blacklists something, and then running it as a separate process:

```python
def test_module_blacklist():
    assert subprocess.run(["python", "script_that_blacklists_modules.py"]).returncode == 0
```

```python
# script_that_blacklists_modules.py
from ilastik.utility import imp
imp.ModuleBlacklist("spam")
import spam
```

emilmelnikov:
I am closing this in favor of [simpler solution](https://github.com/ilastik/ilastik/pull/2093) described above

PR #1975, opened 2019-03-10
EdgeTrainingGui: Fix filtering step when updating edge labels

stuarteberg:
Fixes a bug in `EdgeTrainingGui._handle_edge_label_clicked()` that slightly affects performance.

It isn't supposed to add 0-labeled edges to the slot value -- any omitted edges are implicitly presumed to be zero anyway.  But due to a bug, 0-labels weren't filtered out correctly.

### DIFF
--- a/ilastik/applets/edgeTraining/edgeTrainingGui.py
+++ b/ilastik/applets/edgeTraining/edgeTrainingGui.py
@@ -25,7 +25,7 @@ import numpy as np
 
 from PyQt5.QtCore import Qt
 from PyQt5.QtGui import QColor, QPen, QIcon
-from PyQt5.QtWidgets import QWidget, QLabel, QDoubleSpinBox, QVBoxLayout, QHBoxLayout, \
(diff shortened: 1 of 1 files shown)


ilastik/applets/edgeTraining/edgeTrainingGui.py
```diff
@@ -231,7 +231,7 @@ def _handle_edge_label_clicked(self, updated_edge_labels):
 
         new_labels = dict( edge_labels )
         new_labels.update( updated_edge_labels )
-        for sp_id_pair, new_label in new_labels:
+        for sp_id_pair, new_label in dict(new_labels).items():
```
emilmelnikov:
```suggestion
        for sp_id_pair, new_label in new_labels.items():
```

Or maybe even:

```python
new_labels = {**edge_labels, **updated_edge_labels}
new_labels = {key: label for key, label in new_labels.items() if label}
```

k-dominik: :+1:

PR #1977, opened 2019-03-18
Validation mask for visualization, Work in Progress

Beinabih:
added a layer that can visualize 1, 2 or 4 blocks as validation set

### DIFF
--- a/ilastik/applets/networkClassification/nnClassGui.py
+++ b/ilastik/applets/networkClassification/nnClassGui.py
@@ -403,7 +403,7 @@ class NNClassGui(LabelingGui):
             self.labelingDrawerUi.comboBox.addItems(self.classifiers)
 
             self.topLevelOperator.ModelPath.setValue(self.classifiers)
-            self.model = TikTorchLazyflowClassifierFactory(None, self.tiktorch_path, 0, self.batch_size)
+            self.model = TikTorchLazyflowClassifierFactory(self.tiktorch_path)
 
             self.set_BlockShape()
 
--- a/ilastik/applets/networkClassification/nnClassGui.py
+++ b/ilastik/applets/networkClassification/nnClassGui.py
@@ -132,7 +132,7 @@ class NNClassGui(LabelingGui):
         """
         menus = super(NNClassGui, self).menus()
 
-        advanced_menu = QMenu("Advanced", parent=self)
+        advanced_menu = QMenu("TikTorch", parent=self)
 
         def settingParameter():
             """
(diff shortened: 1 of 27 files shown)


FynnBe:
The size of a tile is not fixed and not meaningful from a theoretical point of view. On top it is hardware dependent. I would suggest we avoid specifying validation regions via number of tiles and instead give a user the opportunity to specify as many validation blocks as they want. These validation regions should be specified in pixel space (detail suggestion: the user should specify such a region in all dimensions of their dataset, whereas we should internally work with the fixed five dimensional tczyx representation)

Beinabih:
It would be more intuitive for the user thats true. However, the average user has no idea but the blockshape size, so if the image has only a small number of blocks he/she could unintentionally select more of the blocks as validation blocks as wanted. These edge cases will be avoided when directly specify the number of blocks. correct me if i am wrong.

ilastik/applets/networkClassification/opNNclass.py
```diff
@@ -340,6 +343,30 @@ def mergeLabels(self, from_label, into_label):
     def clearLabel(self, label_value):
         for laneIndex in range(len(self.InputImages)):
             self.getLane(laneIndex).opLabelPipeline.opLabelArray.clearLabel(label_value)
+
+    def get_val_layer(self, parameters):
```
emilmelnikov: * Functions named `get*` usually return something, but this function assigns and doesn't return anything.
* `val` abbreviation usually means "value" and probably should be expanded here.

```suggestion
    def set_validation_mask(self, parameters):
```

ilastik/applets/networkClassification/nnClassGui.py
```diff
@@ -157,10 +159,9 @@ def __init__(self, parent):
         super(QDialog, self).__init__(parent=parent)
 
         self.validation_size = QComboBox(self)
-        self.validation_size.addItem('10%')
-        self.validation_size.addItem('20%')
-        self.validation_size.addItem('30%')
-        self.validation_size.addItem('40%')
+        self.validation_size.addItem('1')
```
emilmelnikov: Maybe use `QSpinBox` and allow users to set any value they want (within a "reasonable" range)?

FynnBe:
> It would be more intuitive for the user thats true. However, the average user has no idea but the blockshape size, so if the image has only a small number of blocks he/she could unintentionally select more of the blocks as validation blocks as wanted. 

I don't expect the user to think in 'blocks' at all. They are however aware of the pixel space of their data...

>These edge cases will be avoided when directly specify the number of blocks. correct me if i am wrong.
As 'blocks' is a somewhat internal concept, we should not rely on the awareness of the user. At the end of the day, the validation dataset size will be judged by the number of pixels and maybe it will be taken into consideration how widely these labeled validation pixels are spread. 

The user is actually aware of the block shape in a very hidden way, as the validation block shape is the same as the training block shape (without the batch dimension). This is purely for technical reasons though and could even change in the future.
Furthermore, if the user labels whole blocks for validation these need not to align with our internal training-block-shape, which is model and to some extent also GPU dependent. We do not want the user to be required to relabel, because they change the model, etc...

FynnBe:
continued at https://github.com/ilastik/ilastik/pull/2033

PR #1978, opened 2019-03-21
Update splash screen and about dialog

k-dominik:
Updated the splash screen to this one:

splash

font-wise I used the one we have on our webpage (istok-web) which I find more appealing than this bertha druq...

I have also added a dedicated about dialog:

about


ilastik/shell/gui/aboutDialog.py
```diff
@@ -0,0 +1,97 @@
...
+
+        main_layout.addLayout(content_layout)
+        main_layout.addWidget(btn_box)
+
+        self.setLayout(main_layout)
+        self.setStyleSheet("background-color: white;")
+        self.setFixedSize(splash_pixmap.width() * 2.5, splash_pixmap.height() * 1.05)
+
+
+def showAboutDialog(parent=None):
```
emilmelnikov: Maybe make this a classmethod of AboutDialog?

ilastik/shell/gui/aboutDialog.py
```diff
@@ -0,0 +1,97 @@
...
+)
+from PyQt5.QtCore import Qt
+from functools import partial
+
+
+import ilastik
+
+
+about_text = f"""
+    ilastik version {ilastik.__version__} <br /><br />
```
emilmelnikov: A very minor point, but one could use `<p>` instead of `<br>` (not sure if that's relevant outside the Web).

k-dominik:
Hey @emilmelnikov, thank you very much for taking the time to look into this! Your input is highly appreciated and I followed both of your suggestions.  The cluncky name for the `showAboutDialog` still stays, since shorter alternatives like `show`, or `create` are already taken.

akreshuk:
The new font is a real pleasure to look at after the old one. Do you use the same one for the main version and the subversion below? The 3s look a bit different. Otherwise, great!

k-dominik:
> The new font is a real pleasure to look at after the old one. Do you use the same one for the main version and the subversion below? The 3s look a bit different. Otherwise, great!

Great eyes :). It's a non-system font. It is rendered to the image in inkscape. The subversion string is generated dynamically. We could ship the custom font with ilastik if we wanted to though (it's available with the open font license.

emilmelnikov:
> The cluncky name for the `showAboutDialog` still stays, since shorter alternatives like `show`, or `create` are already taken.

Maybe `createAndShowModal` or something similar would be better? Now it's a bit little repetitive...

k-dominik:
thx, that was a good one @emilmelnikov . changed it accordingly

PR #1982, opened 2019-04-04
Update CONTRIBUTING.md and .gitignore

BioinfoTongLI:
- updated unittest cmd with pytest; 
- minor update of .gitignore to ignore vim user tmp files and pytest output project

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,6 +72,3 @@ tests/recorded_test_cases/ObjectClassification/cube_objects_raw.npy
 # vim undo/swp file
 *.un~
 *.swp
-
-# pytest output
-tests/test_tracking_project.ilp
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -69,4 +69,9 @@ tests/recorded_test_cases/ObjectClassification/cube_objects_raw.npy
 .idea/
 
 
+# vim undo/swp file
+*.un~
+*.swp
 
+# pytest output
+tests/test_tracking_project.ilp
```
k-dominik: I would rather change the test in such a way that the file is created outside (`tempfile`...) than adding this exclusion here.
But you don't have to do it, I can add this when I have some time.

BioinfoTongLI: OK, I will remove it.

k-dominik:
great, thanks!

PR #1987, opened 2019-04-09
force remove ilastik-meta

FynnBe:


FynnBe:
This is the output if the given channels are not sufficient:
```
>>> create-dev-env.bat testtesttest
         new environment name: testtesttest
               using Channels:
          installing Packages: ilastik-dependencies-no-solvers
linking to existing directory: ILASTIK-META_LOCAL_SOURCE_PATH not provided
WARNING: The conda.compat module is deprecated and will be removed in a future release.
Collecting package metadata: done
Solving environment: failed

PackagesNotFoundError: The following packages are not available from current channels:

  - ilastik-dependencies-no-solvers

Current channels:

  - https://repo.anaconda.com/pkgs/main/win-64
  - https://repo.anaconda.com/pkgs/main/noarch
  - https://repo.anaconda.com/pkgs/free/win-64
  - https://repo.anaconda.com/pkgs/free/noarch
  - https://repo.anaconda.com/pkgs/r/win-64
  - https://repo.anaconda.com/pkgs/r/noarch
  - https://repo.anaconda.com/pkgs/msys2/win-64
  - https://repo.anaconda.com/pkgs/msys2/noarch

To search for alternate channels that may provide the conda package you're
looking for, navigate to

    https://anaconda.org

and use the search bar at the top of the page.
```
I'd say this output is sufficiently clear. After every step the bat sript checks the errorlevel 'if %errorlevel% neq 0 goto :end' and exists as appropriate.

emilmelnikov:
This is somewhat irrelevant to the issue, but do we have any reasons to use bash and cmd instead of Python? We assume that conda is already installed anyway, so it should be easier to maintain one Python script instead of two platform-dependent ones.

FynnBe:
I like the idea @emilmelnikov . Just out of curiosity, I looked up how one could make a symlink with python and this is indeed possible for Unix and Windows (os.symlink). I don't see other possible obstacles. Let's add a python script and deprecate these scripts.

FynnBe:
Let's open a new PR for a python version of these dev scripts...

PR #1990, opened 2019-04-15
Replace platform-specific development environment scripts with Python

emilmelnikov:
As mentioned in #1987, it might be a good idea to replace Windows cmd and Unix bash scripts with Python, because the latter is portable and easier to maintain.


CONTRIBUTING.md
```diff
@@ -151,7 +151,7 @@ source ~/.bash_profile  # MAC
 
 **Note:** the following steps guide you through the manual process of creating a development environment.
 For convenience, there is a script that achieves the same effect located in `ilastik-meta/ilastik/scripts`.
-You can create a development environment (given miniconda is installed) by invoking `./create-devenv.sh -c ilastik-forge -c conda-forge idev ../../` from within `ilastik-meta/ilastik/scripts`.
+You can create a development environment (given anaconda or miniconda is installed) by invoking `python ilastik/scripts/devenv.py -e idev create` from within `ilastik-meta`.
```
k-dominik: shouldn't you indicate the version WITH linking to the existing ilastik-meta (ie specifying a `--location`)?

scripts/devenv.py
```diff
@@ -0,0 +1,260 @@
...
+        self._site_packages_path = None
+
+    def __repr__(self):
+        return f"{__class__.__name__}({self.name!r})"
+
+    def exists(self) -> bool:
+        """Does this environment exist?"""
+        if self._exists is not None:
+            return self._exists
+        for line in run_stdout("conda", "env", "list"):
```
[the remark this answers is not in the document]
emilmelnikov: OK good, then, I'll try to move queries to JSON where possible.

scripts/devenv.py
```diff
@@ -0,0 +1,260 @@
...
+        return self._exists
+
+    def create(self, *, packages: Iterable[str], channels: Iterable[str] = ()) -> None:
+        """Create this environment.
+
+        Args:
+            packages: Package specs to install (package names and versions).
+            channels: Additional channels to use.
+        """
+        chan_args = []
```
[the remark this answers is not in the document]
emilmelnikov: conda subprocess will bail anyway when `run` is executed, I considered it's better to check this upstream in `command_create`.

scripts/devenv.py
```diff
@@ -0,0 +1,260 @@
...
+
+class CondaEnv:
+    """Environment in the conda package manager.
+
+    Attributes:
+        name: Name of the target conda environment.
+    """
+
+    @staticmethod
+    def installed() -> bool:
```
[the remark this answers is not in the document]
emilmelnikov: True, now I think we should just remove this: even `conda_installed()` doesn't really belong to `CondaEnv`.

CONTRIBUTING.md
emilmelnikov: By default, location is the current working directory, you can change it with `-l|--location`.

k-dominik: then you should at least note that you have to do it from `ilastik-meta` (which is, of course implicitly given in your command).

k-dominik:
LGTM

CONTRIBUTING.md
emilmelnikov: It is mentioned in the doc:
> ... from within `ilastik-meta`

I'll reword it anyway to make this read crystal clear.

k-dominik: uh sorry - my bad :/

emilmelnikov:
> Now I'm getting
> 
> ```
> (base) C:\repos\ilastik-meta\ilastik\scripts>python devenv.py -n test create -l /repos/ilastik-meta
> Traceback (most recent call last):
>   File "devenv.py", line 259, in <module>
>     main()
>   File "devenv.py", line 207, in main
>     env = CondaEnv(args.env)
> AttributeError: 'Namespace' object has no attribute 'env'
> ```

Sorry, I've changes `-e` to `-n` but forgot to change the argparse attributes as well.

> and maybe, if we are already at it, we could check the input package for any ilastik pacakge? if one wants to add packages, one might forget to specify 'ilastik-dependencies-no-solvers' like I just did...

It already includes `ilastik-dependencies-no-solvers` as a default package.

FynnBe:
> It already includes `ilastik-dependencies-no-solvers` as a default package.

I know, but when you add your own packages, it overwrite the default ilastik package, which is fine. We should maybe make sure, though that there is at least one ilastik package specified. If one installs some other packages, ilastik-meta does not even need to be there and then removing it fails... no one wants an ilastik dev env without any ilastik package, anyway...

emilmelnikov:
> > It already includes `ilastik-dependencies-no-solvers` as a default package.
> 
> I know, but when you add your own packages, it overwrite the default ilastik package, which is fine. We should maybe make sure, though that there is at least one ilastik package specified. If one installs some other packages, ilastik-meta does not even need to be there and then removing it fails... no one wants an ilastik dev env without any ilastik package, anyway...

Then we need to keep some whitelist of required packages, which seems to be an overkill: we may introduce some other meta-packages in the future, and people using this this with non-standard packages are likely know what they are doing anyway.

Nevertheless, it might be a good idea to add a note in the option help message, something like this:
```
-p, --package   conda packages to install; make sure to include one of the ilastik packages
```

PR #1998, opened 2019-04-25
fix conftest

k-dominik:
* fixed `sys.exec_info` being a function
* fixed last gui test not being executed

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -55,20 +55,19 @@ def pytest_pyfunc_call(pyfuncitem):
     if not is_gui_test(pyfuncitem):
         return
 
-    bucket = [None]
+    exc_info = None
 
     def testfunc():
         try:
             # Call actual test function
             return pyfuncitem.obj()
         except Exception:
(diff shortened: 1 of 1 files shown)


emilmelnikov:
Just a minor suggestion: `bucket = [None]` and `bucket[0] = ...` could be replaced with `exc_info = None`, `nonlocal exc_info` in the closure definition, and `exc_info = ...` in the same place.

tests/conftest.py
```diff
@@ -62,7 +62,7 @@ def testfunc():
             # Call actual test function
             return pyfuncitem.obj()
         except Exception:
-            bucket[0] = sys.exc_info
+            bucket[0] = sys.exc_info()
```
emilmelnikov: Just a minor suggestion: `bucket = [None]` and `bucket[0] = sys.exc_info()` could be replaced with `exc_info = None`, `nonlocal exc_info` in the closure definition, and `exc_info = sys.exc_info()` in the same place.

k-dominik: reads less weird! thx

PR #2013, opened 2019-05-09
Use uint 32 in carvingTools

constantinpape:
Use uint32 throughout the watershed function in `carvingTools.py`. This will use less memory and was not possible earlier, because nifty did not expose the correct dtypes (should be fixed now).
Also, use `projectScalarNodeDataToPixels` to go from superpixel labels to volumetric segmentation by default.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,11 +24,11 @@ jobs:
     - restore_cache:
         keys:
         # This branch if available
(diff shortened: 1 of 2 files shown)


constantinpape:
This does not work yet in 2d, because the corresponding function in nifty is not exposed to python.
I will fix that in nifty and make a new release.

k-dominik:
pending `nifty` rebuild on our side

PR #2014, opened 2019-05-09
remove old splash, just show fist!

k-dominik:
gif
continuing on the minimalist tip for our splash screen, it will only show a small version of the fist image now with transparency.


emilmelnikov:
Not sure if that's an issue, but maybe move `splashScreen.py` contents to `startShellGui` and get rid of the global variable?

ilastik/shell/gui/startShellGui.py
```diff
@@ -73,6 +76,13 @@ def _applyStyleSheet(app):
         app.setStyleSheet(styleSheetText)
 
 
+def getSplashScreen():
+    splash_path = os.path.join(os.path.split(ilastik.__file__)[0], 'ilastik-splash.png')
```
emilmelnikov: This is probably bikeshedding, but
```suggestion
    splash_path = os.path.join(os.path.dirname(ilastik.__file__), 'ilastik-splash.png')
```

PR #2016, opened 2019-05-09
Fixes "Object Classification Workflow" Data Selection readiness when using multiple lanes

Tomaz-Vieira:
The `Object Classification Workflow`s were not correctly enabling applets when running with multiple lanes.

This PR ensures that the `_inputReady` function correctly reports its state as a boolean, which can then be correctly interpreted in operations such as `&=`.

## Reproducing the bugs this PR fixes:
- Create any "Object Classification Workflow";
- Create two lanes;
- Verify that the applet  after `Data Selection` is not enabled.

### DIFF
--- a/ilastik/workflows/objectClassification/objectClassificationWorkflow.py
+++ b/ilastik/workflows/objectClassification/objectClassificationWorkflow.py
@@ -430,7 +430,7 @@ class ObjectClassificationWorkflow(Workflow):
                     continue
                 if not input_lane_slot[role].ready():
                     return False
-        return True and len(image_group_slot)
+        return True and len(image_group_slot) > 0
 
     def postprocessClusterSubResult(self, roi, result, blockwise_fileset):
         """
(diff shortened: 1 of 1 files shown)


ilastik/workflows/objectClassification/objectClassificationWorkflow.py
```diff
@@ -430,7 +430,7 @@ def _inputReady(self):
                     continue
                 if not input_lane_slot[role].ready():
                     return False
-        return True and len(image_group_slot)
+        return True and len(image_group_slot) > 0
```
emilmelnikov:
```suggestion
        return bool(len(image_group_slot))
```

Or at least just `len(image_group_slot) > 0`...

Tomaz-Vieira: Oops ^^
`return True` is just leftover from previous versions where `return`ing `True` was all it did =)

emilmelnikov: Maybe even go for `bool(image_group_slot)`? If there is no `__bool__` dunder, then the `__len__` is used anyway.

Issue #2004, opened 2019-05-03
Batchprocessing, drag-and-drop, multi-selection not working on windows


k-dominik:
## Describe the bug
In batchprocessing, drag-and-drop, multi-selection not working on windows

## Expected behavior
Please add a clear and concise description of what you expected to happen.

## To Reproduce
Within the Pixel Classification workflow
1. Process any kind of data until the batch processing applet
2. In the batchprocessing applet, try to drag and drop data into the list, or click on _Select Raw Data Files_. The former will not work, the latter will only allow to select a single file.

## ToDo:

- [x] check whether drag and drop works on the other platforms, other applets

## Desktop (please complete the following information):
 - ilastik version: 1.3.3.b1
 - OS: all

PR #2027, opened 2019-05-21
Fix: allow to select multiple files in FileDialog

k-dominik:
allow to select multiple files in batch file selection dialog

This is part1 to fixing #2004


ilastik/applets/dataSelection/dataSelectionGui.py
```diff
@@ -419,12 +419,10 @@ def getImageFileNamesToOpen(cls, parent_window, defaultDirectory):
         
         file_dialog = QFileDialog(
             parent_window, caption="Select Images", directory=defaultDirectory, filter=filt_all_str)
+        file_dialog.setFileMode(QFileDialog.ExistingFiles)
```
emilmelnikov: https://doc.qt.io/qt-5/qfiledialog.html#options-prop:
> Options (particularly the DontUseNativeDialogs option) should be set before changing dialog properties or showing the dialog.

k-dominik: ah, I see
cheers

PR #2035, opened 2019-06-25
Add drag-n-drop support in  batch processing applet

### LINKED ISSUE #2004: Batchprocessing, drag-and-drop, multi-selection not working on windows

k-dominik:
Summay:

* added drag and drop support
* factored out BatchRoleWidget to handle data
* added custom exception (as opposed to asserts)
* removed "feature" where one would supply less number of "secondary" images
  in batch processing
* improved docstring for `run_export`
* removed some legacy stuff

closes #2004

### DIFF
--- a/ilastik/applets/batchProcessing/batchProcessingApplet.py
+++ b/ilastik/applets/batchProcessing/batchProcessingApplet.py
@@ -1,10 +1,8 @@
-from __future__ import division
-from __future__ import absolute_import
-from builtins import range
 import copy
 import weakref
 from collections import OrderedDict
 import logging
+import typing
 logger = logging.getLogger(__name__)  # noqa
 
 import numpy
@@ -65,32 +63,45 @@ class BatchProcessingApplet(Applet):
             parsed_args, role_names)
         return self.run_export(role_path_dict, parsed_args.input_axes, sequence_axis=parsed_args.stack_along)
 
-    def run_export(self, role_data_dict, input_axes=None, export_to_array=False, sequence_axis=None):
-        """
-        Run the export for each dataset listed in role_data_dict,
-        which must be a dict of {role_index : path-list} OR {role_index : DatasetInfo-list}
-
-        As shown above, you may pass either filepaths OR preconfigured DatasetInfo objects.
-        The latter is useful if you are batch processing data that already exists in memory as a numpy array.
-        (See DatasetInfo.preloaded_array for how to provide a numpy array instead of a filepath.)
+    def run_export(
+        self,
+        role_data_dict: typing.Dict[
+            typing.Hashable, typing.List[typing.Union[str, DatasetInfo]]
+        ],
+        input_axes: typing.Optional[str] = None,
+        export_to_array: bool = False,
+        sequence_axis: typing.Optional[str] = None,
+    ) -> typing.List[typing.Union[str, numpy.array]]:
+        """Run the export for each dataset listed in role_data_dict
 
         For each dataset:
-            1. Append a lane to the workflow
-            2. Configure the new lane's DataSelection inputs with the new file (or files, if there is more than one
-               role).
-            3. Export the results from the new lane
-            4. Remove the lane from the workflow.
+                1. Append a lane to the workflow
+                2. Configure the new lane's DataSelection inputs with the new file (or files, if there is more than one
+                   role).
+                3. Export the results from the new lane
+                4. Remove the lane from the workflow.
 
-        By appending/removing the batch lane for EACH dataset we process, we trigger the workflow's usual
-        prepareForNewLane() and connectLane() logic, which ensures that we get a fresh new lane that's
-        ready to process data.
+            By appending/removing the batch lane for EACH dataset we process, we trigger the workflow's usual
+            prepareForNewLane() and connectLane() logic, which ensures that we get a fresh new lane that's
+            ready to process data.
 
-        After each lane is processed, the given post-processing callback will be executed.
-        signature: lane_postprocessing_callback(batch_lane_index)
+            After each lane is processed, the given post-processing callback will be executed.
+            signature: lane_postprocessing_callback(batch_lane_index)
 
-        export_to_array: If True do NOT export to disk as usual.
-                         Instead, export the results to a list of arrays, which is returned.
-                         If False, return a list of the filenames we produced to.
+        Args:
+            role_data_dict: dict with role_name: list(paths) of data that should be processed.
+              You may pass either filepaths OR preconfigured DatasetInfo objects.
+              The latter is useful if you are batch processing data that already exists in memory as a numpy array.
+              (See DatasetInfo.preloaded_array for how to provide a numpy array instead of a filepath.)
+            input_axes: axis description to override from the default role
+            export_to_array: If True do NOT export to disk as usual.
+              Instead, export the results to a list of arrays, which is returned.
+              If False, return a list of the filenames we produced to.
+            sequence_axis: stack along this axis, overrides setting from default role
+
+        Returns:
+            list containing either strings of paths to exported files,
+              or numpy.arrays (depending on export_to_array)
         """
         results = []
 
--- a/ilastik/applets/batchProcessing/batchProcessingApplet.py
+++ b/ilastik/applets/batchProcessing/batchProcessingApplet.py
@@ -1,17 +1,19 @@
-import copy
+import logging
 import weakref
 from collections import OrderedDict
-import logging
-import typing
-logger = logging.getLogger(__name__)  # noqa
+from typing import Dict, Hashable, List, Optional, Union, Mapping, Iterable
 
 import numpy
 import vigra
 from lazyflow.request import Request
-from ilastik.utility import log_exception
+
 from ilastik.applets.base.applet import Applet
 from ilastik.applets.dataSelection import DataSelectionApplet
-from ilastik.applets.dataSelection.opDataSelection import DatasetInfo, OpMultiLaneDataSelectionGroup
+from ilastik.applets.dataSelection.opDataSelection import (
+    DatasetInfo, OpMultiLaneDataSelectionGroup)
+from ilastik.utility import log_exception
+
+logger = logging.getLogger(__name__)  # noqa
 
 
 class BatchProcessingApplet(Applet):
@@ -65,13 +67,13 @@ class BatchProcessingApplet(Applet):
 
     def run_export(
         self,
-        role_data_dict: typing.Dict[
-            typing.Hashable, typing.List[typing.Union[str, DatasetInfo]]
+        role_data_dict: Mapping[
+            Hashable, Iterable[Union[str, DatasetInfo]]
         ],
-        input_axes: typing.Optional[str] = None,
+        input_axes: Optional[str] = None,
         export_to_array: bool = False,
-        sequence_axis: typing.Optional[str] = None,
-    ) -> typing.List[typing.Union[str, numpy.array]]:
+        sequence_axis: Optional[str] = None,
+    ) -> Union[List[str], List[numpy.array]]:
         """Run the export for each dataset listed in role_data_dict
 
         For each dataset:
@@ -92,7 +94,7 @@ class BatchProcessingApplet(Applet):
             role_data_dict: dict with role_name: list(paths) of data that should be processed.
               You may pass either filepaths OR preconfigured DatasetInfo objects.
               The latter is useful if you are batch processing data that already exists in memory as a numpy array.
-              (See DatasetInfo.preloaded_array for how to provide a numpy array instead of a filepath.)
+              (See :meth: `DatasetInfo.preloaded_array` for how to provide a numpy array instead of a filepath.)
             input_axes: axis description to override from the default role
             export_to_array: If True do NOT export to disk as usual.
               Instead, export the results to a list of arrays, which is returned.
--- a/ilastik/applets/batchProcessing/batchProcessingGui.py
+++ b/ilastik/applets/batchProcessing/batchProcessingGui.py
@@ -214,6 +214,9 @@ class BatchProcessingGui( QTabWidget ):
         dominant_role_name = role_names[0]
         num_paths = len(role_path_dict[dominant_role_name])
 
+        if num_paths == 0:
+            return
+
         for role_name in role_names[1:]:
             paths = role_path_dict[role_name]
             if len(paths) == 0:
--- a/ilastik/applets/batchProcessing/batchProcessingGui.py
+++ b/ilastik/applets/batchProcessing/batchProcessingGui.py
@@ -18,22 +18,117 @@
 # on the ilastik web site at:
 #           http://ilastik.org/license.html
 ###############################################################################
-from builtins import range
 import os
 import logging
+import typing
 from collections import OrderedDict
 from functools import partial
-logger = logging.getLogger(__name__)
 
-from PyQt5.QtCore import QTimer, Qt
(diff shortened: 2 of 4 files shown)


ilastik/applets/batchProcessing/batchProcessingGui.py
```diff
@@ -18,22 +18,117 @@
...
+        main_layout.setContentsMargins(0, 0, 0, 0)
+        main_layout.addLayout(button_layout)
+        main_layout.addWidget(self.list_widget)
+
+        self.clear_button.clicked.connect(self.clear)
+        self.select_button.clicked.connect(self.select_files)
+
+        self.setLayout(main_layout)
+
+    def get_all_item_strings(self) -> typing.List[str]:
```
emilmelnikov: Maybe use shorter and more specific name (and turn it into a property or generator)?

```suggestion
    @property
    def filepaths(self) -> typing.List[str]:
```

ilastik/applets/batchProcessing/batchProcessingApplet.py
```diff
@@ -1,10 +1,8 @@
-from __future__ import division
-from __future__ import absolute_import
-from builtins import range
 import copy
 import weakref
 from collections import OrderedDict
 import logging
+import typing
```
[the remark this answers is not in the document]
k-dominik: good point, and thanks for mentioning the black isort feature!

ilastik/applets/batchProcessing/batchProcessingGui.py
```diff
@@ -18,22 +18,117 @@
...
+class FileListWidget(QListWidget):
+    """QListWidget with custom drag-n-drop for file paths
+    """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    # Drag-and-drop
+    def dropEvent(self, dropEvent):
+        urls = dropEvent.mimeData().urls()
+        filepaths = list(map(QUrl.toLocalFile, urls))
```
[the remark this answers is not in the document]
k-dominik: ah :) good catch, I think I used it for some debug output and only removed the `print`s... :)

ilastik/applets/batchProcessing/batchProcessingGui.py
```diff
@@ -18,22 +18,117 @@
...
+        """
+        all_item_strings = []
+        for row in range(self.list_widget.count()):
+            all_item_strings.append(self.list_widget.item(row).text())
+        return all_item_strings
+
+    def select_files(self):
+        preference_name = 'recent-dir-role-{}'.format(self._role_name)
+        recent_processing_directory = PreferencesManager().get(
+            'BatchProcessing', preference_name, default=os.path.normpath('~'))
```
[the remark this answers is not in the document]
k-dominik: no reason, really. Just moved this code without questioning it :)

ilastik/applets/batchProcessing/batchProcessingGui.py
```diff
@@ -18,22 +18,117 @@
...
+class BatchRoleWidget(QWidget):
+    """Container Widget for Batch File list and buttons
+    """
+    def __init__(self, role_name: str, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self._role_name = role_name
+        self._init_ui()
+
+    def _init_ui(self):
```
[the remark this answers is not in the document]
k-dominik: I see your point, of course. I just like to separate any UI initialization from the rest of the `__init__`. In this case it's not so long so we could include it in the init. But even there I would make a comment that everything below is just UI/layout stuff.

ilastik/applets/batchProcessing/batchProcessingApplet.py
```diff
@@ -65,32 +63,45 @@ def run_export_from_parsed_args(self, parsed_args):
...
-        (See DatasetInfo.preloaded_array for how to provide a numpy array instead of a filepath.)
+    def run_export(
+        self,
+        role_data_dict: typing.Dict[
+            typing.Hashable, typing.List[typing.Union[str, DatasetInfo]]
+        ],
+        input_axes: typing.Optional[str] = None,
+        export_to_array: bool = False,
+        sequence_axis: typing.Optional[str] = None,
+    ) -> typing.List[typing.Union[str, numpy.array]]:
```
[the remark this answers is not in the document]
k-dominik: :+1: !

ilastik/applets/batchProcessing/batchProcessingGui.py
```diff
@@ -133,37 +206,26 @@ def initAppletDrawerUi(self):
...
-        self.list_widgets[role_index].clear()
-    
     def run_export(self):
         role_names = self.parentApplet.dataSelectionApplet.topLevelOperator.DatasetRoles.value
 
         # Prepare file lists in an OrderedDict
-        role_path_dict = OrderedDict()
-        role_path_dict[0] = BatchProcessingGui.get_all_item_strings(self.list_widgets[0])
-        num_datasets = len(role_path_dict[0])
+        role_path_dict = OrderedDict(
```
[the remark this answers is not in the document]
k-dominik: I went for being consistent with the namings, but didn't shorten, as I found the "long" version slightly more readable.

tests/test_applets/batchProcessing/test_batchProcessingGui.py
```diff
@@ -0,0 +1,203 @@
...
+    _instances = []
+
+    def __new__(cls, *args, **kwargs):
+        instance = mock.Mock()
+        cls._instances.append(instance)
+        return instance
+
+
+def assert_buttons_in_init_state(batch_processing_gui):
+    assert batch_processing_gui.run_button.isEnabled() is True
```
emilmelnikov:
```python
gui = batch_processing_gui
assert gui.run_button.isEnabled() and not gui.cancel_button.isVisible()
```

Ditto for other assert.

Also, maybe rename them to `assert_gui_in_init_state` and `assert_gui_in_running_state`? We shouldn't care about buttons per se.

tests/test_applets/batchProcessing/test_batchProcessingGui.py
```diff
@@ -0,0 +1,203 @@
...
+from ilastik.applets.batchProcessing.batchProcessingGui import (
+    BatchProcessingDataConstraintException,
+    BatchProcessingGui,
+    BatchRoleWidget,
+    FileListWidget,
+)
+from PyQt5.QtCore import Qt, QUrl
+
+
+def construct_widget(qtbot, widget_class, *args, **kwargs):
```
emilmelnikov: This could be made marginally simpler if callers create widgets themselves: `construct_widget(qtbot, BatchRoleWidget("TestRoleName"))`. Also, maybe `prepare_widget` would be more descriptive?

tests/test_applets/batchProcessing/test_batchProcessingGui.py
```diff
@@ -0,0 +1,203 @@
...
+
+    file_list_widget.dropEvent(drag_n_drop_event)
+    assert file_list_widget.count() == len(filenames)
+    for url in filenames:
+        items = file_list_widget.findItems(url, Qt.MatchFixedString)
+        assert len(items) == 1
+
+
+def test_BatchRoleWidget(qtbot, batch_role_widget, filenames):
+    assert batch_role_widget.select_button.text() == f"Select TestRoleName Files..."
```
emilmelnikov: I wouldn't check the contents of UI strings in tests...

Does this assert check some internal widget state? If so, then (ideally) we need a non-UI state field in widget or something similar.

tests/test_applets/batchProcessing/test_batchProcessingGui.py
```diff
@@ -0,0 +1,203 @@
...
+    assert batch_processing_gui.run_button.isEnabled() is False
+    assert batch_processing_gui.cancel_button.isVisible() is True
+    assert batch_processing_gui.cancel_button.isEnabled() is True
+
+
+def test_BatchProcessingGui_execution(qtbot, batch_processing_gui, role_names, filenames):
+    assert batch_processing_gui.count() == len(role_names)
+    assert_buttons_in_init_state(batch_processing_gui)
+
+    with mock.patch("ilastik.applets.batchProcessing.batchProcessingGui.Request", new=RequestMock):
```
emilmelnikov: This could cause potential problems because `_instances` is not cleared after the function exit.
We could use a locally created callable that accumulates instances:

```python
instances = []

def RequestMock():
    instances.append(mock.Mock())
    return instances[-1]

with ...:
    ...
    assert len(instances) == 0
```

tests/test_applets/batchProcessing/test_batchProcessingGui.py
```diff
@@ -0,0 +1,203 @@
...
+@pytest.mark.parametrize(
+    "secondary_files",
+    [tuple(), ("sec1",), ("sec1", "sec2"), ("sec1", "sec2", "sec3"), ("sec1", "sec2", "sec3", "sec4")],
+)
+def test_BatchProcessingGui_nonmatching_raises(qtbot, batch_processing_gui, role_names, filenames, secondary_files):
+    batch_processing_gui._data_role_widgets[role_names[0]].list_widget.addItems(filenames)
+    secondary_data_role = batch_processing_gui._data_role_widgets[role_names[1]]
+    with mock.patch("ilastik.applets.batchProcessing.batchProcessingGui.Request", new=RequestMock):
+        secondary_data_role.list_widget.addItems(secondary_files)
+        if len(secondary_files) > 0 and (len(secondary_files) != len(filenames)):
```
emilmelnikov:
```suggestion
        if len(secondary_files) in range(1, len(filenames)):
```

tests/test_applets/batchProcessing/test_batchProcessingGui.py
```diff
@@ -0,0 +1,203 @@
...
+            with qtbot.capture_exceptions() as exceptions:
+                qtbot.mouseClick(batch_processing_gui.run_button, Qt.LeftButton, delay=1)
+            assert len(exceptions) == 1
```
emilmelnikov: Move this and the next line under the context manager?

tests/test_applets/batchProcessing/test_batchProcessingGui.py
k-dominik: > I wouldn't check the contents of UI strings in tests...

why not? In this case I want to make sure that the gui is communicating the right string...

tests/test_applets/batchProcessing/test_batchProcessingGui.py
k-dominik: hm, actually I want to run this code path whenever there are one or more `secondary_files` except if the length of `secondary_files` _and_ `filenames` are equal. With the range check (which I really like and will definitely add to my active Python vocabulary), but in this case it doesn't help me.

I'll do it like

```python
if len(secondary_files) not in (0, len(filenames)):
...
```

tests/test_applets/batchProcessing/test_batchProcessingGui.py
k-dominik: i think i had a name like `assert_running_state` in the beginning and changed it to be more explicit once I noticed that I am only testing the buttons, really...

k-dominik:
@emilmelnikov, I've finally sorted the CI - will do a bit of rebasing after you had a look at the latest changes in `test_batchProcessingGui.py`

tests/test_applets/batchProcessing/test_batchProcessingGui.py
emilmelnikov: This mixes GUI logic testing (how clicks and presses affect the GUI state) and GUI presentation testing (whether a certain widget has a certain color or text). If you test presentation too, then after each presentation change in code you need to change exactly the same string in a test. (If the presentation logic is indeed complex, then it probably should be moved out from presentation altogether).

tests/test_applets/batchProcessing/test_batchProcessingGui.py
emilmelnikov: In that case, maybe list the failing examples explicitly in parameters?

```python
from contextlib import nullcontext as does_not_raise

from pytest import raises


@pytest.mark.parametrize(
    "secondary_files,expectation",
    [
        ((), does_not_raise()),
        (("sec1",), raises(BatchProcessingDataConstraintException)),
        ...,
    ]
def test_BatchProcessingGui_nonmatching_raises(..., secondary_files, expectation):
    ...
    with expectation:
        qtbot.mouseClick(batch_processing_gui.run_button, Qt.LeftButton, delay=1)
```

http://doc.pytest.org/en/latest/example/parametrize.html#parametrizing-conditional-raising

k-dominik: I would give that two thumbs up :)

tests/test_applets/batchProcessing/test_batchProcessingGui.py
k-dominik: sure, I see this problem too, but in a sense I don't care if that string is set internally somewhere correctly - I really want to see that it will take the role name. As a compromise I can offer to test whether role name is in the presentation string - not the exact wording. But that has to be tested - otherwise the users will not know where to add what data.

emilmelnikov: `assert roleName in button.text()` should be OK.

tests/test_applets/batchProcessing/test_batchProcessingGui.py
```diff
@@ -186,20 +188,39 @@ def test_BatchProcessingGui_clear(qtbot, batch_processing_gui, role_names, filen
         assert current_data_role_widget.list_widget.count() == 0
 
 
+@contextmanager
+def does_not_raise():
```
emilmelnikov: Any reason not to use `pytest.raises` and `contextlib.nullcontext`?

LGTM otherwise.

k-dominik: In short, `pytest.raises` doesn't work ;) and I didn't know about `nullcontext`.

afaik `PyQt5` won't backpropagate exceptions to callers. So qtbot installs its own excepthook and informs you after the test ran through if any exceptions occur (by setting the test to a failure).

tests/test_applets/batchProcessing/test_batchProcessingGui.py
```diff
@@ -0,0 +1,224 @@
...
+does_not_raise = nullcontext()
+
+
+@contextmanager
+def assert_raises(exception_type):
+    with capture_exceptions() as exp:
+        yield
+        if not exp:
+            raise AssertionError("No exception raised")
+        elif exp[0][0] == exception_type:
```
emilmelnikov:
```suggestion
        elif issubclass(exp[0][0], exception_type):
```

k-dominik: i had that first, but it actually returns the class, not the instance...

emilmelnikov: Yes, since this is consistent with LSP and `pytest.raises`.

k-dominik: allright ;)

PR #2047, opened 2019-06-28
Fix subslot indexing

m-novikov:
Companion to https://github.com/ilastik/lazyflow/pull/310

### DIFF
--- a/ilastik/applets/dataSelection/dataLaneSummaryTableModel.py
+++ b/ilastik/applets/dataSelection/dataLaneSummaryTableModel.py
@@ -88,8 +88,8 @@ class DataLaneSummaryTableModel(QAbstractItemModel):
 
             def handleDatasetInfoChanged(slot):
                 # Get the row of this slot
-                laneSlot = slot.operator
-                laneIndex = laneSlot.operator.index( laneSlot )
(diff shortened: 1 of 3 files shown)


ilastik/applets/dataSelection/dataLaneSummaryTableModel.py
```diff
@@ -88,8 +88,8 @@ def handleNewLane( multislot, laneIndex):
 
             def handleDatasetInfoChanged(slot):
                 # Get the row of this slot
-                laneSlot = slot.operator
-                laneIndex = laneSlot.operator.index( laneSlot )
+                assert slot.subindex, f"BUG: Excpected nested slot {slot}"
```
emilmelnikov: Maybe move asserts into Slot?

```python
@property
def subindex(self):
    if not self.nested:
        raise AttributeError("expected nested slot")
```

m-novikov: The correct approach probably would be refactoring is to replace `Slot.index` method with property, but we have level=1 slot wrapped with `OperatorWrapper` which introduces a lot of weirdness. I would stick with asserts for now. Also `subindex = ()` is valid subindex for toplevel slots.

PR #2054, opened 2019-07-03
Refactor serializer

emilmelnikov:


m-novikov:
By the way, an interesting fact I discovered yesterday writing serializers for tiktorch. Strings containing `"\0"` or bytes with `b"\x00"` is impossible to save in the hdf5. One should use opaque `numpy.void` datatype explicitly for byte encoded strings to avoid errors.

emilmelnikov:
I am closing this since this branch quite old and not too significant.

Issue #2057, opened 2019-07-09
Tracking with learning, TestStructuredLearningTrackingHeadless.testStructuredLearningTrackingHeadless fails


k-dominik:
## Describe the bug
One of our tests, `tests/test_applets/structuredLearningTracking/testStructuredLearningTrackingHytraWithDivisions.py`, fails.

## Expected behavior
Should pass

## To Reproduce
```bash
pytest tests/test_applets/structuredLearningTracking/testStructuredLearningTrackingHytraWithDivisions.py
```

## Error message/traceback

If applicable, please add tracebacks with formatting, such as

```pytb
ERROR 2019-07-09 10:30:43,388 log_exception 12903 140162677741376 Traceback (most recent call last):
  File "/.../ilastik-meta/ilastik/ilastik/shell/projectManager.py", line 441, in _loadProject
    self.workflow.onProjectLoaded( self )
  File "/.../ilastik-meta/ilastik/ilastik/workflows/tracking/structured/structuredTrackingWorkflow.py", line 483, in onProjectLoaded
    self.batchProcessingApplet.run_export_from_parsed_args(self._batch_input_args)
  File "/.../ilastik-meta/ilastik/ilastik/applets/batchProcessing/batchProcessingApplet.py", line 66, in run_export_from_parsed_args
    return self.run_export(role_path_dict, parsed_args.input_axes, sequence_axis=parsed_args.stack_along)
  File "/.../ilastik-meta/ilastik/ilastik/applets/batchProcessing/batchProcessingApplet.py", line 131, in run_export
    export_to_array=export_to_array)
  File "/.../ilastik-meta/ilastik/ilastik/applets/batchProcessing/batchProcessingApplet.py", line 252, in _run_export_with_empty_batch_lane
    self.dataExportApplet.prepare_lane_for_export(batch_lane_index)
  File "/.../ilastik-meta/ilastik/ilastik/workflows/tracking/structured/structuredTrackingWorkflow.py", line 403, in prepare_lane_for_export
    self.result = runLearningAndTracking(withMergerResolution=False)
  File "/.../ilastik-meta/ilastik/ilastik/workflows/tracking/structured/structuredTrackingWorkflow.py", line 363, in runLearningAndTracking
    withBatchProcessing=True)
  File "/.../ilastik-meta/ilastik/ilastik/applets/tracking/structured/opStructuredTracking.py", line 372, in _runStructuredLearning
[...]

Issue #2056, opened 2019-07-09
Tracking with Learning: learning parameters fails, annotations incomplete


k-dominik:
## Describe the bug
In tracking with learning, I annotated some tracks. When running the learn tracking weights, ilastik throws an exception.

## Expected behavior
Please add a clear and concise description of what you expected to happen.

## To Reproduce
* Do the usual tracking stuff (I had mitocheck), train all classifiers and annotate a track.
* In the "Tracking Structured Learning" applet, click _Calculate Tracking weights_

## Error message/traceback

If applicable, please add tracebacks with formatting, such as

```pytb
  File "/.../ilastik-meta/ilastik/ilastik/applets/tracking/structured/structuredTrackingGui.py", line 466, in _learn
    progressVisitor=self.progressVisitor
  File "/.../ilastik-meta/ilastik/ilastik/applets/tracking/structured/opStructuredTracking.py", line 331, in _runStructuredLearning
    children = [int(self.getLabelTT(time+1, division[0][i])) for i in [0, 1]]
  File "/.../ilastik-meta/ilastik/ilastik/applets/tracking/structured/opStructuredTracking.py", line 331, in <listcomp>
    children = [int(self.getLabelTT(time+1, division[0][i])) for i in [0, 1]]
  File "/.../ilastik-meta/ilastik/ilastik/applets/tracking/structured/opStructuredTracking.py", line 483, in getLabelTT
    labels = self.Annotations.value["labels"][time]
```

## Desktop (please complete the following information):
 - ilastik version: `1.3.3b2`
 - OS: Ubuntu 18.04

Issue #2005, opened 2019-05-03
Tracking: AttributeError: 'DiGraph' object has no attribute '_node'


k-dominik:
## Describe the bug
[per user report](https://forum.image.sc/t/ilastik-1-3-3-first-beta/25097/4?u=k-dominik) this error pops up in tracking with learning. The workflow still seems to work.

## Error message/traceback

If applicable, please add tracebacks with formatting, such as

```pytb
ERROR 2019-05-03 09:43:26,542 excepthooks 17782 140070798771968 Unhandled exception in thread: 'Worker #3'
ERROR 2019-05-03 09:43:26,543 excepthooks 17782 140070798771968 Traceback (most recent call last):
  File ".../ilastik-1.3.3b1-Linux/ilastik-meta/volumina/volumina/tiling.py", line 782, in _fetch_tile_layer
    img = ims_req.wait()
  File ".../ilastik-1.3.3b1-Linux/ilastik-meta/volumina/volumina/pixelpipeline/imagesources.py", line 422, in wait
    return self.toImage()
  File ".../ilastik-1.3.3b1-Linux/ilastik-meta/volumina/volumina/pixelpipeline/imagesources.py", line 428, in toImage
    self._arrayreq.wait()
  File ".../ilastik-1.3.3b1-Linux/ilastik-meta/volumina/volumina/pixelpipeline/slicesources.py", line 48, in wait
    return self._sp(self._ar.wait())
  File ".../ilastik-1.3.3b1-Linux/ilastik-meta/volumina/volumina/pixelpipeline/datasources.py", line 504, in wait
    rawData = self._rawRequest.wait()
  File ".../ilastik-1.3.3b1-Linux/ilastik-meta/volumina/volumina/pixelpipeline/datasources.py", line 219, in wrapper
    return func(*args, **kwargs)
  File ".../ilastik-1.3.3b1-Linux/ilastik-meta/volumina/volumina/pixelpipeline/datasources.py", line 243, in wait
    a = self._req.wait()
  File ".../ilastik-1.3.3b1-Linux/ilastik-meta/lazyflow/lazyflow/request/request.py", line 531, in wait
    return self._wait(timeout)
  File ".../ilastik-1.3.3b1-Linux/ilastik-meta/lazyflow/lazyflow/request/request.py", line 559, in _wait
    self._wait_within_request(current_request)
[...]

PR #2058, opened 2019-07-10
Fix tracking learning

### LINKED ISSUE #2052: Tracking with learning: some leftover networkx 1.x code
we'll have to ensure no networkx 1.x syntax (e.g. `graph.node[key]`) is used anymore. Should not be closed without a test that runs when solvers are installed.

k-dominik:
Addresses a few issues related to learning tracking weights in Tracking.

Summary:

* there was some incomplete refactor from `networkx 1.x` to `2.x`
* added tests

Todo:
* [x] refactor `insertAnnotationsToHypothesesGraph`
* [x] rebuild hytra (needed, see https://github.com/ilastik/hytra/pull/8)

fixes #2052, #2056, #2057, #2005

### DIFF
--- a/ilastik/applets/tracking/structured/opStructuredTracking.py
+++ b/ilastik/applets/tracking/structured/opStructuredTracking.py
@@ -582,12 +582,12 @@ class OpStructuredTracking(OpConservationTracking):
                             if (misdetectionLabel not in labels[t+1][dest]):
                                 intersectSet = labels[t][source].intersection(labels[t+1][dest])
                                 lenIntersectSet = len(intersectSet)
-                                assert ((t,source) in list(traxelgraph._graph.edges.keys()) and (t+1,dest) in list(traxelgraph._graph.edges[(t,source)].keys()),
+                                if lenIntersectSet > 0:
+                                    assert ((t, source), (t+1, dest)) in traxelgraph._graph.edges.keys(), (
                                         "Annotated arc that you are setting 'value' of is NOT in the hypotheses graph. " + \
                                         "Your two objects have either very dissimilar features or they are spatially distant. " + \
                                         "Increase maxNearestNeighbors in your project or force the addition of this arc by changing the code here :)" + \
                                         "source ---- dest "+str(source)+"--->"+str(dest)+"       : "+str(lenIntersectSet)+" , "+str(intersectSet))
-                                if lenIntersectSet > 0:
                                     traxelgraph._graph.edges[((t,source), (t+1,dest))]['value'] = lenIntersectSet
 
         for parentTrack in list(divisions.keys()):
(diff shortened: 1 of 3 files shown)


ilastik/applets/tracking/structured/opStructuredTracking.py
```diff
@@ -565,36 +572,49 @@ def insertAnnotationsToHypothesesGraph(self, traxelgraph, annotations,misdetecti
...
+                for destination_object_id in labels[t+1].keys():
+                    # skip if misdetection inside
+                    if (misdetectionLabel in labels[t+1][destination_object_id]):
+                        continue
+
+                    intersectSet = labels[t][source_object_id].intersection(labels[t+1][destination_object_id])
+                    lenIntersectSet = len(intersectSet)
+                    if lenIntersectSet > 0:
+                        if ((t, source_object_id), (t+1, destination_object_id)) not in traxelgraph._graph.edges.keys():
+                            raise AnnotationHypothesisgraphMismatchException(
```
emilmelnikov: Consider moving the message string into `__init__`:

```python
class SpamException:
    def __init__(self, src_obj_id, track_ids):
        msg = f"... {src_obj_id} ... {track_ids} ..."
        super().__init__(msg)
````

File: ilastik/applets/tracking/structured/opStructuredTracking.py [hunk 2]
```diff
@@ -531,19 +535,22 @@ def _type(self, time, track):
             elif type != None:
                 return [type]
 
-    def getLabel(self, time, track, labels):
+    @staticmethod
+    def getLabel(time, track, labels):
         for label in list(labels[time].keys()):
```
emilmelnikov: If we are here anyway, maybe refactor this and `getLabelT` along the way?

```python
return next(label for label, v in labels[time].items() if len(v) == 1 and track in v, False)
```

tests/test_applets/structuredLearningTracking/testOpStructuredTracking.py
```diff
@@ -0,0 +1,204 @@
...
+            return False
+        if self.Timestep == other.Timestep:
+            if self.Id == other.Id:
+                return True
+
+        return False
+
+
+@pytest.fixture
+def expected():
```
emilmelnikov: Do we need to keep "nodes" and "edges" structures together? We could split them into 2 different fixtures.

k-dominik: you're absolutely right. I kept them together as I was just understanding the tracking code, and there annotations are kept together in a dict. But here there is no benefit, other than mirroring the data structures used in the code.

k-dominik:
thx @emilmelnikov for the suggestions! I moved the exception message out but will keep the data structure mirroring the code.

tests/test_applets/structuredLearningTracking/testOpStructuredTracking.py
emilmelnikov: Oh, I missed that, sorry. If it mirrors a structure in SUT, then definitely keep it as-is!

PR #2066, opened 2019-07-25
Use QRect instead of ROI or x/y/w/h

### LINKED ISSUE #2049: Counting, problem serializing/deserializing boxes to project file
## Describe the bug

Boxes are not properly serialized/deserialized from the project. It looks like interpretation of the axisorder is off somehow, since the height of the box is intact.

## Expected behavior
Loading a project should preserve boxes.

## Screenshots
**before loading**
before

**after loading**
after

 - ilastik version: 1.3.2post1, 1.3.3b2
 - OS: Ubuntu 18.04

emilmelnikov:
This also fixes invalid boxes when opening an existing project.
`BoxController.addNewBox` was called from a different source with
different axis order.

Close #2049.


ilastik/utility/gui/roi.py
```diff
@@ -0,0 +1,66 @@
...
+        >>> roi2rect(["spam", "y", "x"], [0, 2, 3], [0, 12, 103])
+        PyQt5.QtCore.QRect(3, 2, 100, 10)
+        >>> roi2rect(["spam", "y", "x"], [0, 12, 103], [0, 2, 3])
+        PyQt5.QtCore.QRect(3, 2, 100, 10)
+        >>> roi2rect(["spam", "y"], [0, 2, 3], [0, 12, 103])  # doctest: +IGNORE_EXCEPTION_DETAIL
+        Traceback (most recent call last):
+          ...
+        ValueError
+    """
+    try:
```
k-dominik: I don't know if one should enforce `axes`, `start` and `stop` to all be of the same length

ilastik/utility/gui/roi.py
```diff
@@ -0,0 +1,66 @@
+###############################################################################
+#   ilastik: interactive learning and segmentation toolkit
+#
+#       Copyright (C) 2011-2014, the ilastik developers
```
k-dominik: sorry to be so pedantic, but one could update the year here...

emilmelnikov: Good catch.

However, we may not need to update these years at all [[1]] [[2]] [[3]]. As far as I understood, it's better to add a year each time you "modify" the code, that is, when you do either an initial release or copy a part of code from some other project, you add a date.

Finally, it would be great to have some sort of license checker as a part of our CI pipeline.

[1]: https://stackoverflow.com/a/2391555
[2]: https://opensource.stackexchange.com/a/1265
[3]: http://www.gnu.org/licenses/gpl-howto.html

k-dominik: the date has been bugging me ever since, but since we have it and you added a new license... In any case we have two options:

Maybe we should discuss getting rid of the date, or even what our switch to MIT/BSD License is doing ;)

ilastik/applets/counting/countingGuiBoxesInterface.py
```diff
@@ -895,12 +861,10 @@ def addNewBox(self, pos: QRect) -> None:
         if any(isinstance(elem, ResizeHandle) for elem in self.scene.selectedItems()):
             return
 
-        # Avoid accidental click-drags.
-        # FIXME: Exploratory testing showed that a single no-drag click makes a 4x4 rectangle, which is probably wrong.
-        if pos.width() <= 4 and pos.height() <= 4:
+        if pos.isEmpty():
```
k-dominik: :+1:

PR #2069, opened 2019-07-26
Suggest features, use time.time instead of os.times

k-dominik:
previously the code used `os.times()[4]` (elapsed time since some point
in the past) to set start and stop times in order to report timing.
This doesn't work on windows

from the Python docs

_"See the Unix manual page times(2) or the corresponding Windows Platform
API documentation. On Windows, only user and system are known; the other
attributes are zero."_

with ~~`time.time()`~~ `time.perf_counter` this is portable.

Also switched all occurrences of `time.time` to `time.perf_counter` (see @emilmelnikov's comment)


emilmelnikov:
It seems like `time.time()` is used just for the time delta, in that case, `time.perf_counter()` is more appropriate (it doesn't account for leap seconds, for example) and (in theory) more accurate.

PR #2073, opened 2019-08-01
Fail on GUI test timeout

m-novikov:


tests/conftest.py
```diff
@@ -55,22 +55,21 @@ def pytest_pyfunc_call(pyfuncitem):
...
+            fut.set_result(result)
+        except Exception as e:
+            fut.set_exception(e)
 
-    with futures.ThreadPoolExecutor(max_workers=1) as executor:
-        fut = executor.submit(testfunc)
-        fut.result(timeout=GUI_TEST_TIMEOUT)
+    runner = threading.Thread(target=testfunc)
+    runner.daemon = True
+    runner.start()
```
k-dominik: For me it's not obvious when will this thread be killed...

m-novikov: This tread will be killed either when the test function completes or if it hangs it will be killed on process exit (this is why it's a daemon thread).

PR #2075, opened 2019-08-02
Run test withing QThread

m-novikov:


m-novikov:
randomly fails with segfaults during normal test run

PR #2079, opened 2019-08-07
Counting: Fix box export and add box import

### LINKED ISSUE #2045: Counting, export list of boxes doesn't work
## Describe the bug
In counting it is possible to export a list of boxes. This fails

## Expected behavior
Should produce a file with list of boxes

## To Reproduce
Within the Counting workflow
1. Add some 2ddata in the data selection applet
2. Select features
3. Draw a box and right click on the list of boxes, select export..

## Error message/traceback

```pytb
  File "/.../ilastik-1.3.3b2-Linux/ilastik-meta/ilastik/ilastik/applets/counting/countingGuiBoxesInterface.py", line 1064, in saveBoxesToCSV
    fh.write(" , ".join(header) +"\n")
TypeError: a bytes-like object is required, not 'str'
```
 - ilastik version: [e.g. 1.3.3b2]
 - OS: Ubuntu 18.04

### LINKED ISSUE #2046: Counting, import list of boxes
## Description

The counting workflow allows for export of boxes. Might be really useful to import box locations in other projects.

kind of depends on fixing #2045 first

## Scenario

User would use this feature in order to compare the counts at the same locations in different images (channels).

## Rationale

https://forum.image.sc/t/cell-density-counting/27082

emilmelnikov:
Closes #2045.
Closes #2046.

### DIFF
--- a/ilastik/applets/counting/countingGuiBoxesInterface.py
+++ b/ilastik/applets/counting/countingGuiBoxesInterface.py
@@ -20,9 +20,11 @@
 ###############################################################################
 
 import colorsys
+import csv
 import logging
 import time
 import warnings
+from typing import Iterable, TextIO, Tuple
 
 import numpy as np
 
@@ -204,6 +206,7 @@ class QGraphicsResizableRect(QGraphicsRectItem):
 
         QGraphicsRectItem.__init__(self, 0, 0, w, h, parent=parent)
         self.Signaller = QGraphicsResizableRectSignaller(parent=parent)
+
 
         scene.addItem(self)
 
@@ -819,7 +822,6 @@ class BoxInterpreter(QObject):
 
 
 class BoxController(QObject):
-
     fixedBoxesChanged = pyqtSignal(dict)
     viewBoxesChanged = pyqtSignal(dict)
 
@@ -848,7 +850,6 @@ class BoxController(QObject):
         self.scene.selectionChanged.connect(self.handleSelectionChange)
 
         boxListModel.boxRemoved.connect(self.deleteItem)
-        boxListModel.signalSaveAllBoxesToCSV.connect(self.saveBoxesToCSV)
 
     def getCurrentActiveBox(self):
         pass
@@ -937,6 +938,19 @@ class BoxController(QObject):
         # super(type(self.boxListModel),self.boxListModel).removeRow(index)
         el.release()
 
+    def deleteAll(self) -> None:
+        """Delete all boxes."""
+        for i in reversed(range(len(self._currentBoxesList))):
+            self.deleteItem(i)
+
+    def rois(self) -> Iterable[Tuple[int, int, int, int]]:
+        """Start X, start Y, end X and end Y box coordinates (start inclusive, end exclusive)."""
+        for box in self._currentBoxesList:
+            item = box.getRectItem()
+            left, top = item.topLeftDataPos()
+            right, bottom = item.bottomRightDataPos()
+            yield left, top, right, bottom
+
     def selectBoxItem(self, index):
         [el._rectItem.setSelected(False) for el in self._currentBoxesList]  # deselect the others
         self._currentBoxesList[index]._rectItem.setSelected(True)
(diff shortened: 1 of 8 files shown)


emilmelnikov:
Column names were taken from the existing code. However, having column names with spaces seems wrong. Should we leave them as-is or change them to snake case or something similar?

Current column names:
```python
["ID", "StartX", "StartY", "StopX", "StopY", "Count", "Average density", "Std density"]
```

ilastik/widgets/boxListView.py
```diff
@@ -231,35 +227,40 @@ def allowFixIcon(self):
...
-        menu.addAction(export)
+        import_ = QAction("Import...", self)
+        import_.triggered.connect(self._import_csv)
+        menu.addAction(import_)
 
         menu.exec_(self.mapToGlobal(event.pos()))
 
+    def _export_csv(self):
+        filename, _filter = QtWidgets.QFileDialog.getSaveFileName(
+            self, "Export Boxes", os.path.expanduser("~"), "CSV (*.csv)", "CSV (*.csv)"
```
Tomaz-Vieira: In GUI mode, we usually default to exporting stuff to the same directory where the project file resides, which is usually the value inside the `WorkingDirectory` slots. Maybe we should be pointing to that directory rather than `~` ?

ilastik/applets/counting/countingGuiBoxesInterface.py
```diff
@@ -970,41 +973,39 @@ def _getNextBoxColor(self):
...
-                        "%5.2f" % averagedens,
-                        "%5.2f" % stddensity,
-                    ]
-                    logger.debug("line " + ",".join(line))
-                    fh.write(",".join(line) + "\n")
+        try:
+            with open(filename, newline="") as csvfile:
+                for row in csv.DictReader(csvfile):
+                    cols = [int(row[name]) for name in colnames]
+                    self.addNewBox(roi2rect(cols[:2], cols[2:]))
```
Tomaz-Vieira: I guess a bit nitpicky, but it's possible that we throw an exception after having added a few boxes. It might be safer to first collect and validate all boxes, then add all of them at once (or rollback, if you're feeling adventurous) so that we can't reach weird states of semi-imported files.

k-dominik:
> Column names were taken from the existing code. However, having column names with spaces seems wrong. Should we leave them as-is or change them to snake case or something similar?
> 
> Current column names:
> 
> ```python
> ["ID", "StartX", "StartY", "StopX", "StopY", "Count", "Average density", "Std density"]
> ```

While all csv reading software can probably handle spaces in column names, I'd be in favor of having the names nice from the start as this means less cleanups for packages that rely on that feature for functionality beyond reading. So :+1: for (and I cannot believe that I say it) camelCase (since it means only changing two headings)

ilastik/applets/counting/countingGuiBoxesInterface.py
```diff
@@ -970,41 +973,39 @@ def _getNextBoxColor(self):
...
-            with open(filename, "wb") as fh:
+            with open(filename, "w", newline="") as csvfile:
+                writer = csv.writer(csvfile, quoting=csv.QUOTE_MINIMAL)
+                writer.writerow(["ID", "StartX", "StartY", "StopX", "StopY", "Count", "Average density", "Std density"])
+
+                for row_id, box in enumerate(self._currentBoxesList):
+                    start = [str(p) for p in box.getRectItem().topLeftDataPos()]
+                    stop = [str(p + 1) for p in box.getRectItem().bottomRightDataPos()]
+                    region = box.getSubRegion()
+                    count = format(np.sum(region), "0.8g")
```
k-dominik: just out of curiosity, why `format` and not `f"{np.sum(region):0.8g}"`?

emilmelnikov: I had f-strings before, but `format` is marginally better because all these variables should be just numbers in a string form, not arbitrary strings. Besides, I didn't knew about `format` before :)

emilmelnikov:
it should work now.
I've rebased this because some changes were overlapping with previous commits.

ilastik/applets/counting/countingGuiBoxesInterface.py
emilmelnikov: Now we collect all boxes from CSV first, and only then actually add them into view/controller.

For export, we just remove the file if something went wrong.

emilmelnikov:
Also, I dropped the `ID` column because it's value was just a row number, not an actual ID.

tests/helpers/shellGuiTestCaseBase.py
```diff
@@ -159,52 +160,46 @@ def moveMouseFromCenter(self, imgView, coords, modifier=Qt.NoModifier):
     def strokeMouse(
         self,
         imgView: QAbstractScrollArea,
-        start: Union[QPoint, Iterable[int]],
-        end: Union[QPoint, Iterable[int]],
+        topLeft: Union[QPointF, QPoint, Iterable[numbers.Real]],
```
k-dominik: naming `topLeft` and `bottomRight` are nice choices for drawing boxes. This function is, however, also used in other contexts. Then I find `start` and `stop` more fitting.
same for `strokeMouseFromCenter`

emilmelnikov: The problem with `start` and `end` is that `end` suggests that the range is start-inclusive end-exclusive (as in Python ranges), whereas `bottomRight` indicates the opposite (start-inclusive end-inclusive). At least that's how `strokeMouse` worked. Besides, top-left and bottom-right are just points in a 2D image view, which is what we want anyway.

Initially, such naming made me think that `strokeMouseFromCenter` was end-exclusive, so I adjusted my test coordinates accordingly, which of course failed the tests, and I spent some time double-checking the coordinates and other code paths.

If you insist on old names, then I would suggest to at least change `strokeMouse` and all test data to end-exclusive (maybe that's a good idea anyway if we're dealing mostly with ROIs).

k-dominik: it's funny, to me it never occurred that when stroking the mouse one of the two points would not be reached (as in an "exclusive" end)... `topLeft` and `bottomRight` imply an ordering in space which is confusing here

emilmelnikov: > `topLeft` and `bottomRight` imply an ordering in space which is confusing here

Don't we always move mouse and draw on a 2D image view?

k-dominik: think of it as drawing, would you ever expect, given two points, let's call them `start_point` and `end_point`, that any of those points would not be included in the drawn line. The more I think about it the more I am in favor of renaming again, away from `topLeft` and friends

k-dominik: > There is an ordering in implementation (curr_point = start_point + point_increment), although I'm not sure whether it's right or not...

if you had a _start-_ and an _end_point this is exactly what you would expect

emilmelnikov: Yes, you are totally right about the ordering! Press-drag-release sequence could be in any of the 8 possible directions (this should be pretty obvious from the beginning, sorry).

> think of it as drawing, would you ever expect, given two points, let's call them `start_point` and `end_point`, that any of those points would _not_ be included in the drawn line.

Maybe... Again, because start-inclusive end-exclusive is a very common way to define an interval. This may be less true while dealing with graphics, but still could be confusing (my guess is this is the reason why Qt moved from start-end to offset-size format).

Slightly off-topic: there is an interesting essay by Djikstra on a very similar issue: https://www.cs.utexas.edu/users/EWD/transcriptions/EWD08xx/EWD831.html

> The more I think about it the more I am in favor of renaming again, away from `topLeft` and friends

Agreed, `topLeft` and friends definitely need to go.

So, maybe rename these parameters to `startPoint` and `endPoint`? I was thinking about literally the same names this night.

P.S. Sorry to bikeshedding here.

PR #2080, opened 2019-08-13
use conda develop instead of symlinks

k-dominik:
**Rationale:**
 with _recent_ changes to conda, it will try to repair your env when installing new packages. This means with our current approach of force-removing `ilastik-meta` and symlinking you practically cannot install new packages into your current environment... Creating a new environment every time you want to install something drove me nuts!

In preparation for this I changed `ilastik-dependencies` package not to include `ilastik-meta` in the first place. (https://github.com/ilastik/ilastik-conda-recipes/pull/66)

**Summary:**
 * Instead of force-remove - symlink, use `conda develop` to add local
 `ilastik-meta` to packages discoverable by current python environment.

You can try it out right now with the update `ilastik-dependencies-no-solvers-package` with

```bash
# from your ilastik-meta folder
python ilastik/scripts/devenv.py creaste -n idev -c kdominik ilastik-forge conda-forge defaults
```

_edit_: it's expected that the tests fail right now - the updated `ilastik-dependencies-no-solvers` package is not yet in `ilastik-forge`

### DIFF
--- a/scripts/devenv.py
+++ b/scripts/devenv.py
@@ -89,6 +89,9 @@ class CondaEnv:
         """Remove a package in this environment."""
         run("conda", "remove", "--yes", "--force" if force else None, "--name", self.name, package)
 
+    def develop(self, path: pathlib.Path):
+        run("conda", "develop", "--name", self.name, str(path))
+
     @property
     def _info(self):
         if self._info_cache is not None:
@@ -186,7 +189,7 @@ def main():
     create_ap.add_argument(
         "-l",
         "--location",
-        help="local ilastik-meta directory; skip DIR to keep the original ilastik-meta package",
+        help="local ilastik-meta directory",
         metavar="DIR",
         nargs="?",
         const=None,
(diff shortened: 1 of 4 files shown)


scripts/devenv.py
```diff
@@ -234,18 +237,11 @@ def command_create(args: argparse.Namespace, env: CondaEnv) -> None:
...
-    env_repo_path.symlink_to(location)
-
-    pth_path = env.site_packages_path / "ilastik-meta.pth"
     packages = "lazyflow", "volumina", "ilastik"
-    pth_content = "\n".join(f"{env_repo_path / pkg}" for pkg in packages)
-    logging.info(f"create {pth_path}")
-    pth_path.write_text(pth_content)
+    package_locations = [location / pkg for pkg in packages]
+
+    for package_location in package_locations:
```
emilmelnikov: Maybe simplify to

```python
for pkg in "lazyflow", "volumina", "ilastik":
    env.develop(location / pkg)
```

scripts/devenv.py
```diff
@@ -89,6 +89,9 @@ def remove_package(self, package: str, *, force=False) -> None:
         """Remove a package in this environment."""
         run("conda", "remove", "--yes", "--force" if force else None, "--name", self.name, package)
 
+    def develop(self, path: pathlib.Path):
```
emilmelnikov:
```suggestion
    def develop(self, path: pathlib.Path) -> None:
    """Install a package in development mode."""
```

Also, maybe rename to `install_package_dev` or something similar?

Issue #2082, opened 2019-08-15
Use different watershed implementation in multicut


constantinpape:
Use my implementation of distance transform watersheds in the multicut workflow.
It provides better results than the implementation that is currently used.
It is implemented [here](https://github.com/constantinpape/elf/blob/master/elf/segmentation/watershed.py#L45) (currently this is just an API mock-up, but I will add the implementation soon).

Note that this also implies changing some of the selector boxes in the GUI, because my function takes different arguments than the one that is currently used.
We need to decide which ones to expose to the user.

Side note: I am in the process of refactoring general purpose functionality into [elf](https://github.com/constantinpape/elf). For development purposes, I would suggest we use the implementations in there. Once we decide to merge the update watershed functionality into ilastik, we can discuss whether we want to add elf as a dependency to ilastik or copy the implementations into ilastik.

PR #2086, opened 2019-08-21
[wip] Improved watershed and test

boykovdn:
closing #2082, also adds a test file/


.circleci/config.yml
```diff
@@ -55,8 +55,8 @@ jobs:
         conda activate base &&
         cd ${ILASTIK_ROOT}/ilastik-meta &&
         python ilastik/scripts/devenv.py create -n ${TEST_ENV_NAME}
-        -p pytest-qt ilastik-dependencies-no-solvers
-        -c ilastik-forge conda-forge defaults
+        -p pytest-qt ilastik-dependencies-no-solvers elf
+        -c ilastik-forge conda-forge cpape defaults
```
constantinpape: We should decide how to handle the elf functionality. [elf](https://github.com/constantinpape/elf) is a small python only library I recently wrote to bundle functionality that I need in several downstream dependencies.
It's fairly small, so we could consider adding it to ilastik dependencies (and pushing it on ilastik-forge).
If you don't want to add another dependency, we could also copy the relevant code somewhere to ilastk.

ilastik/applets/wsdt/wsdtGui.py
```diff
@@ -124,50 +124,55 @@ def populate_channel_menu(*args):
...
+        configure_update_handlers( sigma_weights_box.valueChanged, op.SigmaSeeds )
+        sigma_weights_box.setToolTip("Smooth the watershed weight map with this sigma")
+        drawer_layout.addLayout( control_layout( "Weight Smooth", sigma_weights_box ) )
+        self.sigma_weights_box = sigma_weights_box
+
+        apply_nonmax_suppression_box = QCheckBox()
+        configure_update_handlers( apply_nonmax_suppression_box.toggled, op.ApplyNonmaxSuppression )
+        apply_nonmax_suppression_box.setToolTip("Filter out seeds. Use that option when some of your foreground objects have long and thin parts.")
+        drawer_layout.addLayout( control_layout( "Nonmax Suppression", apply_nonmax_suppression_box ) )
+        self.apply_nonmax_suppression_box = apply_nonmax_suppression_box
```
constantinpape: I don't think we need to expose all parameters to the user here.
`sigma_weights` and `sigma_seeds` can be unified to `sigma`.
And I don't think that we need to expose `apply_nonmax_suppression`, it can just be enabled by default.
The only downside to using non-max suppression for the seeds is a bit longer run-time + nifty dependency (which is not an issue here.)

boykovdn:
We still have to resolve the problem of outdated nifty version, as well as the dependency on elf. 
Applying nonmax suppression will not work with the current version, because the import fails due to outdated nifty. Channel cpape is for development only and should be removed when rebuilding, as it only supports python 3.7

tests/test_applets/multicutWsdt/test_Consistency.py
```diff
@@ -0,0 +1,99 @@
...
+						WS_PARAMS['sigma'],
+						WS_PARAMS['sigma'],
+						WS_PARAMS['min_size'],
+						WS_PARAMS['alpha'],
+						WS_PARAMS['pixel_pitch'],
+						WS_PARAMS['apply_nonmax_suppression'])	
+
+	return ws, max_id
+
+def get_result_operator(input_data):
```
k-dominik: you could make this and `get_result_function` fixtures, too...

tests/test_applets/multicutWsdt/test_Consistency.py
```diff
@@ -0,0 +1,99 @@
...
+
+	graph = Graph()
+	with Pipeline(graph=graph) as get_wsdt:
+		get_wsdt.add(OpArrayPiper, Input=input_data)
+		get_wsdt.add(OpCachedWsdt, FreezeCache=False)
+		wsdt_result = get_wsdt[-1].outputs['Superpixels'][:].wait()
+
+	return wsdt_result
+
+def test_Consistency(input_data):
```
k-dominik:
```suggestion
def test_consistency(input_data):
```

.circleci/config.yml
constantinpape: Yes, I want to bring it on conda forge as soon as we have nifty there.

boykovdn:
Note: In the future before merging these changes, the user documentation on the ilastik website (https://www.ilastik.org/documentation/multicut/multicut) should be updated to reflect the changes in the GUI.

boykovdn:
- [] Remove cpape channel dependency in appveryor.yml (once part of ilastik-forge)

ilastik/applets/wsdt/wsdtApplet.py
```diff
@@ -40,13 +40,10 @@ def singleLaneOperatorClass(self):
...
-                 'MinMembraneSize',
-                 'MinSegmentSize',
-                 'SigmaMinima',
-                 'SigmaWeights',
-                 'GroupSeeds',
-                 'PreserveMembranePmaps' ]
+                 'Threshold',
+                 'MinSize',
+                 'Sigma',
+                 'Alpha']
```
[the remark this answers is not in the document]
boykovdn: Sure, I can't think of a reason why not to broadcast them. ApplyNonmaxSuppression will probably be on by default in the future.

boykovdn:
- [ ] Implement regression test as suggested by @k-dominik in earlier comment

.circleci/config.yml
```diff
@@ -55,8 +55,8 @@ jobs:
         conda activate base &&
         cd ${ILASTIK_ROOT}/ilastik-meta &&
         python ilastik/scripts/devenv.py create -n ${TEST_ENV_NAME}
-        -p pytest-qt ilastik-dependencies-no-solvers black
-        -c ilastik-forge conda-forge defaults
+        -p pytest-qt ilastik-dependencies-no-solvers black elf
```
boykovdn: @k-dominik I cannot figure out why CircleCI cannot find elf (hence the tests fail) once a test tried to import it. Any clue? I thought adding it as a dependency here would make it readily available to the environment.

boykovdn:
on line 131 of testPixelClassificationGui.py a segfault happens for Linux, not sure why. @Tomaz-Vieira I saw that you had done some work on this test, do you have any quick ideas?

tests/test_applets/multicutWsdt/test_Consistency.py
```diff
@@ -0,0 +1,101 @@
...
+
+	graph = Graph()
+	with Pipeline(graph=graph) as get_wsdt:
+		get_wsdt.add(OpArrayPiper, Input=input_data)
+		get_wsdt.add(OpCachedWsdt, FreezeCache=False)
+		wsdt_result = get_wsdt[-1].outputs['Superpixels'][:].wait()
+
+	return wsdt_result
+
+def test_consistency(input_data, get_result_function, get_result_operator):
```
wolny: minor: could you rename it to something more descriptive, e.g. `test_results_equal` and change the name of the main test function

wolny:
one minor comment, LGTM overall :+1:

k-dominik:
merging this despite #2184

Thank you @boykovdn

PR #2090, opened 2019-08-29
Add StrictABC and use it for ShellABC

emilmelnikov:
`ShellABC` now uses `StrictABCMeta` metaclass and registers it's virtual
subclasses after their definitions.

PyQt5 uses a custom SIP metaclass for QObject and it's subclasses,
making it impossible to use plain `ABC` or `ABCMeta` for `ShellABC`
since `IlastikShell` inherits from `QMainWindow`.

It is possible to define a custom metaclass
`QObjectABCMeta(QObject.__class__, ABCMeta)`  and use it for `ShellABC`,
but in this case we always depend on `QObject`, even though some
subclasses should not use Qt at all (e.g. `HeadlessShell`).

We could use `ABC.register` to "virtually" subclass `ShellABC`. However,
because such subclass is not a real ABC subclass, it is possible to
create an instance even if some abstract methods are not implemented.

Therefore, we perform all necessary checks during the `register` call
in `StrictABCMeta`. This solution automatically provides proper
`__subclasshook__` and `__instancehook__`, together with more meaningful
error message.

### DIFF
--- a/ilastik/shell/gui/ilastikShell.py
+++ b/ilastik/shell/gui/ilastikShell.py
@@ -174,7 +174,7 @@ class ProgressDisplayManager(QObject):
         self.requestTimer.start()
 
         self.statusBar.addPermanentWidget(self.requestStatus)
-        
+
         self.memoryWidget = MemoryWidget()
         self.memoryWidget.showDialogButton.clicked.connect(self.parent().showMemUsageDialog)
         self.statusBar.addPermanentWidget(self.memoryWidget)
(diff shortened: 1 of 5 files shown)


m-novikov:
An alternative solution that was described here I just created QABC for use with qt based classes and added signal support for these classes. 
https://github.com/ilastik/volumina/pull/250/commits/0d5907260cc312ad5d2b4ec1f09697a4f76c1885

emilmelnikov:
Reworded the commit message and replaced fancy quotes with `"` in docstring (copied them from a web page by mistake).

emilmelnikov:
> An alternative solution that was described here I just created QABC for use with qt based classes and added signal support for these classes.
> [ilastik/volumina@0d59072](https://github.com/ilastik/volumina/commit/0d5907260cc312ad5d2b4ec1f09697a4f76c1885)

I tried to do exactly the same initially, but it's preferable that `HeadlessShell` does not import anything from `PyQt5`.

m-novikov:
What about returning passed class from `register` method then it can be used as decorator. And one can be sure that register and class won't get split during refactoring.
```
@MyABC.register
class MyImpl:
     pass
```

emilmelnikov:
> What about returning passed class from `register` method then it can be used as decorator. And one can be sure that register and class won't get split during refactoring.
> 
> ```
> @MyABC.register
> class MyImpl:
>      pass
> ```

This is a great point! I've missed the fact that the default implementation already does that:

> Changed in version 3.3: Returns the registered subclass, to allow usage as a class decorator.

PR #2093, opened 2019-08-30
Prevent Qt imports using import hooks

emilmelnikov:
Use import hooks instead of a dummy module in order to prevent importing
PyQt5 in the headless mode by mistake.

Supersedes https://github.com/ilastik/ilastik/pull/1974.


k-dominik:
all headless tests are failing. I assume that something import `matplotlib` which will import `PyQt5` somewhere. Just a hunch.

emilmelnikov:
When I run tests locally on 8cc74e95cd3958da8e0e93d5c020c9dddf105837 (current master), some windows are briefly popping up from time to time.

k-dominik:
did you limit the test execution somehow, which tests where run? was it maybe expected??

emilmelnikov:
> all headless tests are failing. I assume that something import `matplotlib` which will import `PyQt5` somewhere. Just a hunch.

Yeah... `ilastik.utility` package indirectly imports PyQt, I suspect vigra which unconditionally imports pylab for some reason.

emilmelnikov:
> did you limit the test execution somehow, which tests where run? was it maybe expected??

It seems like these are all the new tests driven by pytest-qt and not marked as GUI tests (test_batchProcessingGui and others).

k-dominik:
btw. this will also be triggered by something like:

```python
_qt_available = True
try:
    import PyQt5
except ImportError:
    _qt_available = False
```
which is not absolutely uncommon. And code like this should also be allowed in headless mode imo.

emilmelnikov:
> btw. will this also be triggered by something like?
> 
> ```python
> _qt_available = True
> try:
>     import PyQt5
> except ImportError:
>     _qt_available = False
> ```
> 
> which is not absolutely uncommon. And code like this should also be allowed in headless mode imo.

That's fine because the blacklist raises ImportError, as if the module doesn't exist. The only task here is to install the blacklist as soon as possible.

k-dominik:
the only question that remains is now, if such a thing is possible at all. I saw that the previous dummy module was probably installed after `PyQt5` was imported already (after `import ilastik` anyway).

emilmelnikov:
> the only question that remains is now, if such a thing is possible at all. I saw that the previous dummy module was probably installed after `PyQt5` was imported already (after `import ilastik` anyway).

It should be possible: PyQt is not an essential Python module, so blocking it should be feasible.

emilmelnikov:
Close in favor of https://github.com/ilastik/ilastik/pull/2099.

PR #2095, opened 2019-09-02
Fix TrackingBaseDataExportApplet tests

emilmelnikov:
When launching tests with pytest, "tearDown" has not been called, and
stale test project file has not been removed. Now the file is created
in a temporary directory, which is automatically removed after a test.

### DIFF
--- a/tests/test_applets/conservationTracking/testTrackingBaseDataExportAppletSerialization.py
+++ b/tests/test_applets/conservationTracking/testTrackingBaseDataExportAppletSerialization.py
@@ -1,4 +1,4 @@
-from typing import Mapping, Any
+from typing import Any, Mapping
 
 import h5py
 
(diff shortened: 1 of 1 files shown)


tests/test_applets/conservationTracking/testTrackingBaseDataExportAppletSerialization.py
```diff
@@ -1,71 +1,49 @@
...
-            dataExportSerializer.serializeToHdf5(testProject, self.TEST_PROJECT_FILE)
+    path = tmp_path / "test_tracking_project.ilp"
 
-        # check serialized values
-        with h5py.File(self.TEST_PROJECT_FILE) as testProject:
-            assert testProject["Tracking Result Export/SelectedPlugin"].value.decode() == 'Fiji-MaMuT'
-            assert testProject["Tracking Result Export/SelectedExportSource"].value.decode() == 'Plugin'
-            assert testProject["Tracking Result Export/AdditionalPluginArguments/bdvFilePath"].value.decode() == '/tmp/bdv.xml'
+    with h5py.File(path) as project:
+        dataExportSerializer.serializeToHdf5(project, path)
```
k-dominik: i see the convenience of this fixture, already writing the output file.
But since this test file is about testing the serialization, I think it would be preferable to not have a `ERROR at setup ...` when something goes wrong in `.serializeToHdf5`

PR #2096, opened 2019-09-03
Fix live update consistency in pixel classification

### LINKED ISSUE #2092: Pixel Classification: live update not synchronized among lanes
## Describe the bug
In pixel classification, when working with multiple images, _live update_ used to be synchronized along lanes. Right now the lanes behave weird and independently. One can also arrive at inconsistent states where _live update_ appears to be pressed but one can still go back to feature selection (see screenshot)

## Expected behavior
Pressing _live update_ in __any__ lane should update it globally.

## To Reproduce
Within the Pixel Classification workflow
1. Add multiple images
2. Select some features in the _Feature Selection Applet_
3. In training, do some annotations, click _live update_ mode
4. Switch a lane, see that _live update_ is disabled

## Screenshots
blah

 - ilastik version: this problem was introduced in `1.3.3b2`

the cause is most likely somewhere in these two commits:
* 52bd155bc7524d2a7dd3378388db44d896a1f7bd
* f2539169463212a45f59f75bfde0e5d2757b20bf

k-dominik:
Synchronizes the status of live update button and the associated prediction cache across lanes.

Also added a test to define this behavior.

fixes #2092

### DIFF
--- a/ilastik/applets/pixelClassification/pixelClassificationGui.py
+++ b/ilastik/applets/pixelClassification/pixelClassificationGui.py
@@ -25,6 +25,7 @@ from __future__ import division
 # Built-in
 from past.utils import old_div
 import os
+from typing import Optional
 import logging
 from collections import OrderedDict
 from functools import partial
@@ -493,7 +494,10 @@ class PixelClassificationGui(LabelingGui):
             except:
                 self.render = False
 
-        self.setLiveUpdateEnabled(False)
+        # listen to freezePrediction changes
+        unsub_callback = self.topLevelOperatorView.FreezePredictions.notifyDirty(lambda *args: self.setLiveUpdateEnabled())
+        self.__cleanup_fns.append(unsub_callback)
+        self.setLiveUpdateEnabled()
 
     def initFeatSelDlg(self):
         if self.topLevelOperatorView.name == "OpPixelClassification":
@@ -782,16 +786,14 @@ class PixelClassificationGui(LabelingGui):
         return feature_images_slot.ready() and feature_images_slot.meta.shape is not None
 
     def isLiveUpdateEnabled(self):
-        return self.labelingDrawerUi.liveUpdateButton.isChecked()
+        return not self.topLevelOperatorView.FreezePredictions.value
 
-    def setLiveUpdateEnabled(self, checked: bool):
+    def setLiveUpdateEnabled(self, checked: Optional[bool] = None):
+        checked = checked if checked is not None else self.isLiveUpdateEnabled()
+        self.labelingDrawerUi.liveUpdateButton.blockSignals(True)
+        self.labelingDrawerUi.liveUpdateButton.setChecked(checked)
+        self.labelingDrawerUi.liveUpdateButton.blockSignals(False)
         if checked:
-            if not self.hasFeatures():
-                self.labelingDrawerUi.liveUpdateButton.blockSignals(True)
-                self.labelingDrawerUi.liveUpdateButton.setChecked(False)
-                self.labelingDrawerUi.liveUpdateButton.blockSignals(False)
-                QMessageBox.warning(self, "Error", "No features selected on previous step.")
-                return
             self._viewerControlUi.checkShowPredictions.setChecked(True)
             self.handleShowPredictionsClicked()
 
--- a/tests/test_applets/pixelClassification/testPixelClassificationMultiImageGui.py
+++ b/tests/test_applets/pixelClassification/testPixelClassificationMultiImageGui.py
@@ -228,7 +228,7 @@ class TestPixelClassificationGuiMultiImage(ShellGuiTestCaseBase):
 
     def test_5_InteractiveMode(self):
         """
-        Click the "interactive mode" checkbox and see if any errors occur.
+        Click the "interactive mode" button and see if any errors occur.
         """
 
         def impl():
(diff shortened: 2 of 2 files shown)


ilastik/applets/pixelClassification/pixelClassificationGui.py
```diff
@@ -784,28 +800,53 @@ def hasFeatures(self) -> bool:
...
+        self.labelingDrawerUi.liveUpdateButton.setChecked(enable)
+        self.labelingDrawerUi.liveUpdateButton.blockSignals(False)
+
+        self._update_aux_buttons(enable)
+        # Notify the workflow that some applets may have changed state now.
+        # (For example, the downstream pixel classification applet can
+        #  be used now that there are features selected)
+        self.parentApplet.appletStateUpdateRequested()
+
+    def _update_freeze_status(self, freeze: bool):
```
emilmelnikov: Could inline the argument since `freeze == not self.isLiveUpdateEnabled()`.

ilastik/applets/pixelClassification/pixelClassificationGui.py
```diff
@@ -784,28 +800,53 @@ def hasFeatures(self) -> bool:
     def isLiveUpdateEnabled(self):
         return self.labelingDrawerUi.liveUpdateButton.isChecked()
 
-    def setLiveUpdateEnabled(self, checked: bool):
-        if checked:
+    def _update_live_update_button(self, enable: bool):
```
emilmelnikov: Could inline argument since `enable == not self.topLevelOperatorView.FreezePredictions.value`.

ilastik/applets/pixelClassification/pixelClassificationGui.py
```diff
@@ -493,7 +491,25 @@ def __init__(self, parentApplet, topLevelOperatorView, labelingDrawerUiPath=None
...
+        # need to initialize the aux buttons once in case _update_live_update_button did not 
+        # result in an update
+        self._update_aux_buttons(not self.topLevelOperatorView.FreezePredictions.value)
+
+        def sync_live_update_with_freeze_value():
+            self._update_live_update_button(not self.topLevelOperatorView.FreezePredictions.value)
+
+        # listen to freezePrediction changes
+        self.topLevelOperatorView.FreezePredictions.notifyDirty(bind(sync_live_update_with_freeze_value))
+        self.__cleanup_fns.append(partial(
```
emilmelnikov: `notifyDirty` already returns an unsubscribe callback, so one could use it directly.

ilastik/applets/pixelClassification/pixelClassificationGui.py
```diff
@@ -784,28 +800,53 @@ def hasFeatures(self) -> bool:
...
-    def setLiveUpdateEnabled(self, checked: bool):
-        if checked:
+    def _update_live_update_button(self, enable: bool):
+        button_state = self.isLiveUpdateEnabled()
+        if button_state == enable:
+            return
+
+        if enable:
             if not self.hasFeatures():
                 self.labelingDrawerUi.liveUpdateButton.blockSignals(True)
```
emilmelnikov: These 3 lines are redundant since `enable == True` implies `button_state == False` which means an already unchecked button.

ilastik/applets/pixelClassification/pixelClassificationGui.py
emilmelnikov: Not sure whether we could do any better, but this method's name is awkward...

k-dominik: > Not sure whether we could do any better, but this method's name is awkward...

oh wow, i didn't even realize how bad the name was; thx ;)

ilastik/applets/pixelClassification/pixelClassificationGui.py
k-dominik: good catch!

k-dominik:
> I find it a bit hard to follow the changes, mainly because there are multiple `update_*` functions for FreezePredictions and liveUpdateButton, which should always be in sync (?)
> 
> Maybe we could define a single function that gets called on either FreezePredictions or liveUpdateButton change, and work out all state changes there?

Hm, I think there have to be at least two functions, one that updates the button from the `FreezePrediction` change, and one that updates `FreezePredictions` when the button is toggled. I don't think reworking this into one (as it was before) makes it simpler (didn't work properly before).

Tomaz-Vieira:
Nice that you managed to catch that, even after all the messing around I did with the code =)

So, it seems that the most important difference in this patch is that the applet is watching `FreezePredictions` for dirtyness, and updating accordingly. I naively assumed that there would be only a single applet for all lanes, so it would be ok to initialize it with "live update" disabled. But that meant that when we switched lanes, a new applet would be created, with "live update" disabled, even if it had been enabled in the first lane.

Here's a [patch](https://github.com/ilastik/ilastik/compare/master...Tomaz-Vieira:fix_2096?expand=1) that tries to merge both ideas; it keeps the new code but also watches for `FreezePredictions` dirtyness. Still, if you don't find it to be readable, feel free to just ignore it and carry own with your version.

k-dominik:
Hey @Tomaz-Vieira,

thanks a lot for checking it out! Without a doubt your solution is simpler, I like that :) So I just cherry picked :)

PR #2099, opened 2019-09-05
Remove dummy PyQt module for the headless mode

emilmelnikov:
Since headless mode should be able to work without PyQt, a special
dummy module was used instead of a real PyQt one. However, this approach
doesn't work if a real PyQt module has been already imported at this
time, because `import PyQt5` looks up `PyQt5` in `sys.modules` cache
first, bypassing `sys.path`.

Moreover, Qt could be still useful for it's QtCore and other modules.

Therefore, it has been decided to remove the module altogether to
prevent potential confusion, while preventing importing unneeded GUI
modules with code reviews and various analysis tools (e.g. flake8).

See https://github.com/ilastik/ilastik/pull/2093 and
https://github.com/ilastik/ilastik/pull/1974 for previous attempts to
solve the issue. The solutions tried there introduce excessive
complexity, compared to the problems they try to solve.


emilmelnikov:
> what's up wit the modification on `tests/data/inputdata/smallVideoConservationTracking.ilp`, otherwise LGTM

Apparently, this happens if you interrupt `tests/test_applets/conservationTracking/testConservationTrackingHeadless.py`. Not sure why... Reverted this.

PR #2100, opened 2019-09-05
Refactor box interpreter

emilmelnikov:
* Remove dead code
* Refactor BoxInterpreter.eventFilter and some of the surrounding code


ilastik/applets/counting/countingGuiBoxesInterface.py
```diff
@@ -626,51 +623,36 @@ def release(self):
...
-
-        :param navigationInterpreter:
-        :param positionModel:
-        :param BoxContr:
-        :param widget: The main widget
-
-        """
-
-        QObject.__init__(self)
+    def __init__(self, base: NavigationInterpreter, posModel: PositionModel, mainWidget: QWidget):
```
k-dominik: not sure how `mainWidget` is more descriptive than just widget, but after consulting `QRubberBand` docs I'm not wiser ;). The Qt sourcecode gives me the feeling that it should be called `parent`

emilmelnikov: I'm afraid that calling this parameter `parent` would give the wrong impression that `BoxInterpreter` itself has a notion of a "parent" widget, something that `QObject` does.
How about `centralWidget`, just to be consistent with `LayerViewerGui`?

k-dominik: okay, but maybe add a docstring making clear that it's just the parent for the rubberwidget.

emilmelnikov: I ended up renaming it to `newBoxParent` with the corresponding docstring.

PR #2111, opened 2019-09-18
pass empty dialog options to show native dialog

k-dominik:
without passing the options explicitly, the qt file dialog would be shown
native dialogs are much nicer (so all your usual favorites are available ...).

### DIFF
--- a/ilastik/widgets/ImageFileDialog.py
+++ b/ilastik/widgets/ImageFileDialog.py
@@ -6,7 +6,6 @@ from PyQt5.QtWidgets import QFileDialog
 
 from volumina.utility import PreferencesManager
 from ilastik.applets.dataSelection.opDataSelection import OpDataSelection
-from ilastik.config import cfg as ilastik_config
 
 
 class ImageFileDialog(QFileDialog):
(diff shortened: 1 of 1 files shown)


emilmelnikov:
It seems that [subclassing QFileDialog produces this behavior]( https://doc.qt.io/qt-5/qfiledialog.html#Option-enum) (see description of `QFileDialog::DontUseNativeDialog`).

I grepped usages of `ImageFileDialog` and they all immediately call `getSelectedPaths()`. Maybe it makes sense to convert this class to module-level function?

k-dominik:
Thanks for checking it out!

> I grepped usages of `ImageFileDialog` and they all immediately call `getSelectedPaths()`. Maybe it makes sense to convert this class to module-level function?

sure, we could do that. But I don't see much saving/benefit in this change. You'd have to call it something like `show_file_dialog_and_get_file_names` to make it explicit that a dialog is spun up, which I think the instantiation of a dialog communicates. But yeh, would be one sets of brackets and a dot less.

emilmelnikov:
I thought about a module-prefixed function `imageFileDialog.getOpenFileNames`. This is consistent with `QFileDialog.getOpenFileNames`.

k-dominik:
> I thought about a module-prefixed function `imageFileDialog.getOpenFileNames`. This is consistent with `QFileDialog.getOpenFileNames`. However, this is not a big issue, and could be left as-is.

fair enough.

PR #2114, opened 2019-09-18
Fix carving preprocessing on 2D input

emilmelnikov:


ilastik/workflows/carving/opPreprocessing.py
```diff
@@ -182,40 +182,36 @@ def setupOutputs(self):
...
                         size_regularizer=self.SizeRegularizer.value,
                         reduce_to=self.ReduceTo.value,
                     )
-                else:
-                    result_view[...], max_id = vigra.analysis.watershedsNew(volume_feat)
-            # 2d watersheds
+            elif input_.ndim == 3:
+                result[idx], max_id = vigra.analysis.watersheds(input_)
+            elif input_.ndim == 2:
+                result[idx], max_id = vigra.analysis.watershedsNew(input_)
```
constantinpape: I don't think we need the two branches here, because we can just use `vigra.analysis.watershedsNew`.
AFAIK `watersheds` and `watershedsNew` give the same output, but `watershedsNew` should be faster.
My guess is that someone just didn't refactor the 3d code branch when `watershedsNew` was introduced...

ilastik/workflows/carving/opPreprocessing.py
```diff
@@ -182,40 +182,36 @@ def setupOutputs(self):
...
-                if self.DoAgglo.value:
-                    result_view[...], max_id = watershed_and_agglomerate(
-                        volume_feat[:, :, 0],
-                        max_workers=max_workers,
-                        size_regularizer=self.SizeRegularizer.value,
-                        reduce_to=self.ReduceTo.value,
-                    )
-                else:
-                    result_view[...], max_id = vigra.analysis.watersheds(volume_feat[:, :, 0])
+                raise AssertionError("Invalid ndim")
```
k-dominik: I don't like the use of `AssertionError` in general. Just think about catching/handling it. `AssertionError` is normally generated by `assert` statements. You'd have to `except AssertionError` to do any handling there which is never a good idea: those might get optimized away when python is invoked with the appropriate flags. In any case the error message seems to imply, that you want to raise a `ValueError`. But that also doesn't really fit the if statement. I would subclass exception here.

ilastik/workflows/carving/opPreprocessing.py
emilmelnikov: OK, great! I assumed that the new watershed performed poorly on 3D.

ilastik/workflows/carving/opPreprocessing.py
emilmelnikov: The assertion error is there just to make the if-else chain complete, ndim should be checked before anyway. Also, this will go away because we only need one, new watershed implementation in both cases.

ilastik/workflows/carving/opPreprocessing.py
constantinpape: No, I am pretty sure it performs better.
AFAIK the only reason `watersheds` is still in `vigra` is because it supports some options not available in `watershedsNew` (turbo watershed), that are not relevant for us here.

emilmelnikov:
Probably we can get away with a simple test that just exercises the operator itself.

k-dominik:
> Probably we can get away with a simple test that just exercises the operator itself.

assuming that the downstream ops support 2d data...

PR #2116, opened 2019-09-20
[wip] Making random forest training optional

boykovdn:
The mean edge probability feature is overfitted by the RF, so making the training step optional can save the user time in most cases (when their data contains clear boundaries).


ilastik/applets/edgeTrainingWithMulticut/edgeTrainingWithMulticutGui.py
```diff
@@ -39,7 +43,57 @@ def initAppletDrawerUi(self):
...
+                        if imgView.isVisible():
+                            imgView.scene().joinRenderingAllTiles()
+                    self.topLevelOperatorView.FreezeCache.setValue(True)
+     
+            self.getLayerByName("Multicut Edges").visible = True
+            # self.getLayerByName("Multicut Segmentation").visible = True
+            th = threading.Thread(target=updateThread)
+            th.start()
+
+            op.FreezeClassifier.setValue(False)
```
boykovdn: I had to unfreeze the classifier in order to get the 'Easy Predict' button to update the view with multicut edges. I am not completely sure why this is the case when the classifier should not be used for this computation, and there might be a more elegant solution. Also if I freeze the classifier here, ilastik stops responding when 'Easy Predict' is toggled.

PR #2123, opened 2019-09-26
Inherit all interpreters from InterpreterABC

emilmelnikov:
Previously, interpreters relied on a custom `__subclasshook__` for
`isinstance` checks. Now, each interpreter is a subclass of
`InterpreterABC`, which is a proper abstract base class.
This change should make it easier to trace class relationships.


k-dominik:
requires https://github.com/ilastik/volumina/pull/258

PR #2124, opened 2019-09-26
Update test data '3d1c_Probabilities.h5' to have only one channel

boykovdn:
* The name suggests this file contains data for the boundaries under one channel,
which was not the case previously.
* Discard channel 0 which contains the inverse of boundary information (boundaries
dark, cell bodies bright)


k-dominik:
The file name might be a bit confusing, but it is consistent - those are the probabilities produced for the `3D1C.h5` image. Will leave as is.

PR #2126, opened 2019-09-26
Fail when conda-build is not installed

emilmelnikov:
Due to https://github.com/ilastik/ilastik/pull/2080, `install_package_dev` requires
`conda-build` to be installed. It this is not the case, `devenv.py create` fails only
after the lengthy installation process.

This PR explicitly checks the availability of `conda-build` before proceeding
with the script's command.

### DIFF
--- a/scripts/devenv.py
+++ b/scripts/devenv.py
@@ -2,7 +2,7 @@
 
 """Manage ilastik development environment."""
 
-import sys
+import sys  # isort:skip
 
 # Don't use new language features when checking the minimum Python version.
 if sys.version_info < (3, 6):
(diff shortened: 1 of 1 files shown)


scripts/devenv.py
```diff
@@ -210,6 +220,10 @@ def add_global_args(argparser: argparse.ArgumentParser) -> None:
     if shutil.which("conda") is None:
         logging.error("conda is not installed")
         return
+    
+    if run_returncode("conda", "build", "--version") != 0:
+        logging.error("conda-build is not installed; run 'conda install conda-build' inside the 'base' environment")
```
k-dominik:
```suggestion
        logging.error("conda-build is not installed; run 'conda install -n base -c conda-forge conda-build' to install conda-build in the 'base' environment")
```

PR #2138, opened 2019-10-04
Restores ability to use PreloadedDatasetInfo when batch processing

### LINKED ISSUE #2137: BatchProcessingExport.run_export no longer accepts preloaded arrays
The new PreloadedArrayDatasetInfo is an improvement, but if you pass a list of them to `run_export` as the value of the `role_data_dict`, it still gets interpreted as a truthy `role_input_path` by `export_dataset`, and so ilastik tries to convert it to a `FileSystemDatasetInfo` (and fails).

Is there a new way to run a given workflow on preloaded arrays?

Tomaz-Vieira:
The ability to use a `PreloadedDatasaetInfo` in batch processing (which is useful when using ilastik inside a python script) was lost during the big `DatasetInfo` overhaul. This patches restores that capability and adds a test which is very much based on `examples/example_python_client.py`

Fixes #2137

### DIFF
--- a/tests/test_applets/pixelClassification/testPixelClassificationHeadless.py
+++ b/tests/test_applets/pixelClassification/testPixelClassificationHeadless.py
@@ -116,8 +116,6 @@ class TestPixelClassificationHeadless(object):
         from ilastik.applets.dataSelection.opDataSelection import FilesystemDatasetInfo
 
         info = FilesystemDatasetInfo(filePath=dataset_path)
-        print(f"\n\n\n~~~~~~~~~~~~~~~~~~~~~>>>>info:{info.axiskeys}\n\n\n")
-        # import pydevd; pydevd.settrace()
         opDataSelection = workflow.dataSelectionApplet.topLevelOperator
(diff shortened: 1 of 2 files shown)


tests/test_applets/pixelClassification/testPixelClassificationHeadless.py
```diff
@@ -112,6 +116,8 @@ def create_new_project(cls, project_file_path, dataset_path):
         from ilastik.applets.dataSelection.opDataSelection import FilesystemDatasetInfo
 
         info = FilesystemDatasetInfo(filePath=dataset_path)
+        print(f"\n\n\n~~~~~~~~~~~~~~~~~~~~~>>>>info:{info.axiskeys}\n\n\n")
```
clbarnes: Is this intentional?

Tomaz-Vieira: Absolutely not. Fixed =)

clbarnes:
LGTM!

PR #2145, opened 2019-10-23
Quickfix: changed default export path for project-internal data

k-dominik:
default location changed from home-directory to project file location by overriding the base-class method.

`default_output_dir` will be accessed in `OpDataExport` to determine the output location for our magic value `{dataset_dir}`.


Tomaz-Vieira:
LGTM
Do we have any reason not to have everything default to the same directory as the poject file, though?

k-dominik:
> LGTM
> Do we have any reason not to have everything default to the same directory as the poject file, though?

I guess this is a different discussion, I don't think we should change de historic behavior of writing stuff to the dataset_dir just now, if that's what you mean.

PR #2156, opened 2019-11-05
Make training a random forest classifier optional

boykovdn:
* Use 'standard_edge_mean' feature on probability map to compute edge probabilities
* Add tickbox to toggle whether to train a classifier
* Re-use user selection of input channels from Watershed as probability map

### DIFF
--- a/ilastik/applets/edgeTraining/opEdgeTraining.py
+++ b/ilastik/applets/edgeTraining/opEdgeTraining.py
@@ -27,6 +27,8 @@ class OpEdgeTraining(Operator):
     DEFAULT_FEATURES = {"Grayscale": ["standard_edge_mean"]}
     FeatureNames = InputSlot(value=DEFAULT_FEATURES)
     FreezeClassifier = InputSlot(value=True)
+    TrainRandomForest = InputSlot(value=False)
+    WatershedSelectedInput = InputSlot()
 
     # Lane-wise
     EdgeLabelsDict = InputSlot(level=1, value={})
@@ -56,6 +58,8 @@ class OpEdgeTraining(Operator):
         self.opComputeEdgeFeatures.FeatureNames.connect(self.FeatureNames)
         self.opComputeEdgeFeatures.VoxelData.connect(self.VoxelData)
         self.opComputeEdgeFeatures.Rag.connect(self.opRagCache.Output)
+        self.opComputeEdgeFeatures.TrainRandomForest.connect(self.TrainRandomForest)
+        self.opComputeEdgeFeatures.WatershedSelectedInput.connect(self.WatershedSelectedInput)
 
         self.opEdgeFeaturesCache = OpMultiLaneWrapper(OpValueCache, parent=self, broadcastingSlotNames=["fixAtCurrent"])
         self.opEdgeFeaturesCache.Input.connect(self.opComputeEdgeFeatures.EdgeFeaturesDataFrame)
@@ -76,6 +80,7 @@ class OpEdgeTraining(Operator):
         )
         self.opPredictEdgeProbabilities.EdgeClassifier.connect(self.opClassifierCache.Output)
         self.opPredictEdgeProbabilities.EdgeFeaturesDataFrame.connect(self.opEdgeFeaturesCache.Output)
+        self.opPredictEdgeProbabilities.TrainRandomForest.connect(self.TrainRandomForest)
 
         self.opEdgeProbabilitiesCache = OpMultiLaneWrapper(
             OpValueCache, parent=self, broadcastingSlotNames=["fixAtCurrent"]
@@ -237,6 +242,8 @@ def decodeToStringIfBytes(s):
 
 
 class OpComputeEdgeFeatures(Operator):
+    WatershedSelectedInput = InputSlot()
+    TrainRandomForest = InputSlot(value=False)
     FeatureNames = InputSlot()
     VoxelData = InputSlot()
     Rag = InputSlot()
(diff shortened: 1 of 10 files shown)


stuarteberg:
This seems like a nice feature.  I have some questions:

1. a. Implicitly, the threshold value for an edge to be "on" is 0.5.  If the edge probabilities are computed directly from the pixel probabilities (rather than from an RF), then I think many users would want to specify exactly what threshold to use.  That value should be respected by the "naive segmentation" layer, for example.

   b. (I'm not 100% sure about this.) If you make the threshold customizable, then I *think* it might be necessary to re-scale the values above/below the threshold value, since the multicut algorithm assumes a threshold of 0.5.

   Say the user chooses a threshold of T=0.3.  Then, before passing the data to the multicut function, values from 0.0 to 0.3 should be "expanded" to the range 0.0 to 0.5, and values from 0.3 to 1.0 should be "compressed" to the range 0.5 to 1.0.  That way, multicut can treat the values as probabilities, but the user's desired threshold is respected.  (I know that multicut has the `beta` parameter, but that isn't quite the same as providing a custom "on" threshold value.)

   FWIW, the thresholding applet's "graphcut" mode also performs a re-scaling like I'm describing.  (Graphcut isn't supported in the current version of ilastik due to dependency issues, but [the code is still there if you're curious][1].)

[1]: https://github.com/ilastik/ilastik/blob/7dd81417f8a6af1a53b9674fdd42989a6f933da4/ilastik/applets/thresholdTwoLevels/opThresholdTwoLevels.py#L321-L335

2. I think you should add the new `TrainRandomForest` slot to the list in `EdgeTrainingSerializer.__init__()`, so that this setting is saved to the project file.

3. (Somewhat unrelated to this PR.)  If the user's data already has clear boundaries, but those boundaries happen to be "dark" rather than "bright", she might appreciate a checkbox for "invert probabilities" somewhere in the GUI, perhaps in the watershed applet.

ilastik/applets/multicut/multicutGui.py
```diff
@@ -305,13 +305,14 @@ def configure_operator_from_gui(self):
...
 
-    def _handle_mulicut_update_clicked(self):
+    def _handle_multicut_update_clicked(self):
         def updateThread():
             """
             Temporarily unfreeze the cache and freeze it again after the views are finished rendering.
             """
             with self.set_updating():
                 self.topLevelOperatorView.FreezeCache.setValue(False)
+                self.topLevelOperatorView.FreezeClassifier.setValue(False)
```
stuarteberg: `FreezeClassifier` is a member of `OpEdgeTrainingWithMulticut`, but not `OpMulticut`.  This gui "mixin" class is designed to work with operators that expose the same slots as `OpMulticut` (at least), but shouldn't require all of the slots in `OpEdgeTrainingWithMulticut`.

So, ideally, this `FreezeClassifier` change should happen in `EdgeTrainingWithMulticutGui` or its base class, `EdgeTrainingGui`.  But the complicated updating logic here makes that slightly inconvenient.  I suppose you could move the updating logic into a member function, and then allow `EdgeTrainingWithMulticutGui` to override it:

```python
    def _handle_mulicut_update_clicked(self):
        def updateThread():
            """
            Temporarily unfreeze the cache and freeze it again after the views are finished rendering.
            """
            with self.set_updating():
                self._update_multicut_views()

        self.getLayerByName("Multicut Edges").visible = True
        # self.getLayerByName("Multicut Segmentation").visible = True
        th = threading.Thread(target=updateThread)
        th.start()

    def _update_multicut_views(self):
        self.topLevelOperatorView.FreezeCache.setValue(False)
        self.update_button.setEnabled(False)
        self.live_multicut_button.setEnabled(False)

        # This is hacky, but for now it's the only way to do it.
        # We need to make sure the rendering thread has actually seen that the cache
        # has been updated before we ask it to wait for all views to be 100% rendered.
        # If we don't wait, it might complete too soon (with the old data).
        ndim = len(self.topLevelOperatorView.Output.meta.shape)
        self.topLevelOperatorView.Output((0,) * ndim, (1,) * ndim).wait()

        # Wait for the image to be rendered into all three image views
        for imgView in self.editor.imageViews:
            if imgView.isVisible():
                imgView.scene().joinRenderingAllTiles()
        self.topLevelOperatorView.FreezeCache.setValue(True)
        self.update_button.setEnabled(True)
        self.live_multicut_button.setEnabled(True)
```

Then, in `EdgeTrainingWithMulticutGui`:

```python
    def _update_multicut_views(self):
        self.topLevelOperatorView.FreezeClassifier.setValue(False)
        super()._update_multicut_views()
        self.topLevelOperatorView.FreezeClassifier.setValue(True)
```

I wouldn't call that elegant, but I think it will work.

constantinpape:
>     1. a. Implicitly, the threshold value for an edge to be "on" is 0.5.  If the edge probabilities are computed directly from the pixel probabilities (rather than from an RF), then I think many users would want to specify exactly what threshold to use.  That value should be respected by the "naive segmentation" layer, for example.
>        b. (I'm not 100% sure about this.) If you make the threshold customizable, then I _think_ it might be necessary to re-scale the values above/below the threshold value, since the multicut algorithm assumes a threshold of 0.5.
>        Say the user chooses a threshold of T=0.3.  Then, before passing the data to the multicut function, values from 0.0 to 0.3 should be "expanded" to the range 0.0 to 0.5, and values from 0.3 to 1.0 should be "compressed" to the range 0.5 to 1.0.  That way, multicut can treat the values as probabilities, but the user's desired threshold is respected.  (I know that multicut has the `beta` parameter, but that isn't quite the same as providing a custom "on" threshold value.)

@stuarteberg I think these are good points, just a few comments: 
- if we expose this thresholding parameter, it should work for the Simple Probabilities and the Random Forest probabilities in the same way.
I think it might still be useful to change it for the Random Forest Probabilities in the same manner and
it would be confusing to have it behave differently for Simple Probabilities
- if we introduce this probability scaling, I think we should not expose the beta parameter. They would interact in a weird manner and it would be confusing for users.

k-dominik:
> > ```
> > 1. a. Implicitly, the threshold value for an edge to be "on" is 0.5.  If the edge probabilities are computed directly from the pixel probabilities (rather than from an RF), then I think many users would want to specify exactly what threshold to use.  That value should be respected by the "naive segmentation" layer, for example.
> >    b. (I'm not 100% sure about this.) If you make the threshold customizable, then I _think_ it might be necessary to re-scale the values above/below the threshold value, since the multicut algorithm assumes a threshold of 0.5.
> >    Say the user chooses a threshold of T=0.3.  Then, before passing the data to the multicut function, values from 0.0 to 0.3 should be "expanded" to the range 0.0 to 0.5, and values from 0.3 to 1.0 should be "compressed" to the range 0.5 to 1.0.  That way, multicut can treat the values as probabilities, but the user's desired threshold is respected.  (I know that multicut has the `beta` parameter, but that isn't quite the same as providing a custom "on" threshold value.)
> > ```
> 
> @stuarteberg I think these are good points, just a few comments:
> 
>     * if we expose this thresholding parameter, it should work for the Simple Probabilities and the Random Forest probabilities in the same way.
>       I think it might still be useful to change it for the Random Forest Probabilities in the same manner and
>       it would be confusing to have it behave differently for Simple Probabilities
> 
>     * if we introduce this probability scaling, I think we should not expose the beta parameter. They would interact in a weird manner and it would be confusing for users.

+1 for getting rid of the beta parameter and just having a threshold there for both cases, "simple" probabilities, and RF predictions. Using a threshold there is more graphic and thus easier for users to grasp than some abstract beta. Also we could recolor the edges with respect to this threshold so users get a "feeling" for the result.

stuarteberg:
>if we introduce this probability scaling, I think we should not expose the beta parameter. They would interact in a weird manner and it would be confusing for users.

I agree that it would be confusing to have two parameters that do *almost* (but not quite) the same thing.  But for multicut experts (or ilastik developers) I'd be in favor of keeping a secret control, which could be shown in ilastik's special `--debug` mode, assuming it's not too much trouble.

stuarteberg:
>if we introduce this probability scaling, I think we should not expose the beta parameter. They would interact in a weird manner and it would be confusing for users.

I agree that it would be confusing to have two parameters that do *almost* (but not quite) the same thing.  But for multicut experts (or ilastik developers) I'd be in favor of keeping a secret control, which could be shown in ilastik's special `--debug` mode, assuming it's not too much trouble.

boykovdn:
Forced tests to recompute, but now CircleCI passes and Appveyor doesn't, for the same code.

wolny:
LGTM. The code is well structured and readable :+1:

ilastik/applets/edgeTraining/opEdgeTraining.py
```diff
@@ -248,41 +255,66 @@ def setupOutputs(self):
...
+
+            # Could use join() or merge() here, but we know the rows are already in the right order, and concat() should be faster.
+            all_edge_features_df = pd.DataFrame(rag.edge_ids, columns=["sp1", "sp2"])
+            all_edge_features_df = pd.concat([all_edge_features_df] + edge_feature_dfs, axis=1, copy=False)
+            result[0] = all_edge_features_df
+
+        else:
+            def normalize1(series):
+                series -= np.min(series)
+                series /= np.max(series)
```
wolny: This will fail if `series` if of integer type, e.g.:
```
a = np.random.randint(0,255, (10,10))
a /= np.max(a)
```
gives:
```
TypeError: No loop matching the specified signature and casting
```
replace with `series = series / np.max(series)`

k-dominik:
> Oh! My bad.. Should have tested more thoroughly.

No worries. There were a lot of eyes on this one. And that's something that might be very invisible with toy data!

k-dominik:
merging this despite circleci failing. for this see #2184

otherwise great PR, thanks @boykovdn

PR #2158, opened 2019-11-07
Quickfix export objectids

### LINKED ISSUE #1868: Allow to export the "Objects" layer from object classification
It is now not possible to save the segmentation resulting from thresholding other than by right-click --> export on the layer, which is not practical for multiple datasets and can't be used headless.

### LINKED ISSUE #2106: Randomize colors of object layer in object classification
## Description

Would be great to enable the "randomize colors" functionality that we already have in all the superpixel layers for object classification as well, since one could have touching objects and would want to resolve clashes.

k-dominik:
Added the "object identities" layer to the set of exportable sources.

In order to enable some workflows that take might take object identities as an input (e.g. atlas/prediction mask) it is really beneficial to have this output in object classification available not only via right click export.

fixes #2106, fixes #1868

### DIFF
--- a/ilastik/applets/objectClassification/objectClassificationDataExportGui.py
+++ b/ilastik/applets/objectClassification/objectClassificationDataExportGui.py
@@ -89,6 +89,7 @@ class ObjectClassificationResultsViewer(DataExportLayerViewerGui):
             "Object Probabilities",
             "Blockwise Object Predictions",
             "Blockwise Object Probabilities",
+            "Object Identities",
             "Pixel Probabilities",
         ]
 
@@ -113,6 +114,10 @@ class ObjectClassificationResultsViewer(DataExportLayerViewerGui):
                 layer.visible = False
                 layer.name = layer.name + "- Preview"
             layers += previewLayers
+        elif selection == "Object Identities":
+            previewSlot = self.topLevelOperatorView.ImageToExport
+            layer = self._initColortableLayer(previewSlot)
+            layers += layer
         else:
             assert False, "Unknown selection."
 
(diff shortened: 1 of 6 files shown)


ilastik/applets/objectClassification/objectClassificationDataExportGui.py
```diff
@@ -113,6 +114,10 @@ def setupLayers(self):
                 layer.visible = False
                 layer.name = layer.name + "- Preview"
             layers += previewLayers
+        elif selection == "Object Identities":
+            previewSlot = self.topLevelOperatorView.ImageToExport
+            layer = self._initColortableLayer(previewSlot)
```
Tomaz-Vieira: Seems like it'd make more semantic sense if `_initColortableLayer` would return a single layer (as the singular name implies), and then just add it to the `layers`  list with `.append` or whatever. Or perhaps you're expecting `_initColortableLayer` to eventually return more than one layer? In that case, maybe that should be pluralized?

ilastik/applets/objectClassification/objectClassificationDataExportGui.py
```diff
@@ -151,3 +156,16 @@ def _initPredictionLayers(self, predictionSlot):
...
+        objectssrc.setObjectName("LabelImage LazyflowSrc")
+        ct = colortables.create_default_16bit()
+        ct[0] = QColor(0, 0, 0, 0).rgba()  # make 0 transparent
+        layer = ColortableLayer(objectssrc, ct)
+        layer.name = "Object Identities - Preview"
+        layer.setToolTip("Segmented objects, shown in different colors")
+        layer.visible = False
+        layer.opacity = 0.5
+        layer.colortableIsRandom = True
+        return [layer]
```
Tomaz-Vieira: I really wish all these layer parameters could be set inside `ColortableLayer.__init__`. Maybe we can get some patches to volumina to do that later

In fact, this looks eerily similar to the code in `objectExtractionGui.py `. If updating `ColortableLayer` in `volumina` would be too much of a hassle, than maybe this should be a `@(static|class)method` so that we can reuse it and don't replicate this logic around.

k-dominik: totally agree - since this patch is targeted to finalize 1.3.3, I will open an Issue about this in volumina but will proceed with the changess for now

PR #2173, opened 2020-01-31
Remove filepath in h5 before updating

k-dominik:
This fixes our functionality of replacing the path of an image file after it has been moved.

Checks whether `filepath` is in the h5 group already and removes it before
updating if necessary. Also made sure to save the new filepath utf-8 encoded.

"+" some sneaky black changes courtesy of my pre-commit hook...

### DIFF
--- a/ilastik/applets/dataSelection/dataSelectionSerializer.py
+++ b/ilastik/applets/dataSelection/dataSelectionSerializer.py
@@ -98,7 +98,7 @@ class DataSelectionSerializer(AppletSerializer):
 
     @timeLogged(logger, logging.DEBUG)
     def _serializeToHdf5(self, topGroup, hdf5File, projectFilePath):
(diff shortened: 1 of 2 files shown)


Tomaz-Vieira:
Nice catch!
Did you per chance update z5py? I think the tests are blowing up because the new version of z5py expects `str`s rather than `Path`s

k-dominik:
> Nice catch!
> Did you per chance update z5py? I think the tests are blowing up because the new version of z5py expects `str`s rather than `Path`s

Thanks for having a look. It seems like there seems to be an inconsistency between the apis of h5py and z5py (I've opened an issue [here](https://github.com/constantinpape/z5/issues/160)). For now just explicitly converting the path to string for the z5 case.

k-dominik:
thanks @Tomaz-Vieira !

PR #2178, opened 2020-02-14
[WIP]Adds --distributed to ilastik headless

Tomaz-Vieira:
This PR adds the ability to run ilastik with `mpirun`, by passing in the `--distributed` flag when running in headless mode. Also,  one can use the `--distributed-block-size=<integer>` to determine the size of the tiles that are worked on by each individual worker, so e.g. `--distributed-block-size=512` will operate on tiles of shape `Shape5D(x=512, y=512, z=512, c=512, t=512)`, clamped to the maximum size of the raw data on each dimension.

- Reworks argparser creation in `DataSelectionApplet` so it's easier for the `BatchProcessingApplet` to add the new `--distributed` flag;
- Creates `--distributed` and `--distributed-block-size` command line options
- Requires https://github.com/ilastik/lazyflow/pull/335 to actually do the mpi orchestration

This also includes some changes in the way ilastik guesses axiskeys for its datasources. This was needed for the batch processing service, but might not fit perfectly with common ilastik usage.
TODO: Test new axiskeys guessing

## Running
0 - Install `mpi4py`: `conda install mpi4py` (Install via `pip` on the HPCs, though, so that it gets a chance to compile using the native implementation of mpi)
1 - Train your classifier as usual in interactive mode;
2 - Run headless like so:
```
export HDF5_USE_FILE_LOCKING=FALSE # multiple processes will open the .ilp file
NUM_PROCESSES=8
mpirun -N ${NUM_PROCESSES} python ilastik.py --headless \
    --project ~/MyProject.ilp \
    --distributed --distributed_block_size=256 \
    /path/to/your/data.png
```


k-dominik:
I guess this can be closed as https://github.com/ilastik/ilastik/pull/2241 was merged, right @Tomaz-Vieira

PR #2196, opened 2020-03-13
Fix windows dll discovery for grpc

m-novikov:


ilastik.py
```diff
@@ -63,7 +63,7 @@ def isvalidpath_win(path):
     if sys.platform.startswith("win"):
         # Empty PATH except for gurobi and CPLEX and add ilastik's installation paths.
         path = list(filter(isvalidpath_win, _env_list("PATH")))
-        path += subdirs("Qt4/bin", "Library/bin", "python", "bin")
+        path += subdirs("Qt4/bin", "Library/bin", "Library/mingw-w64/bin", "python", "bin")
```
k-dominik: `Qt4` cannot really be needed anymore. I checked and it doesn't exist in the environments... Also there is nothing from qt5 :)

```suggestion
        path += subdirs("Library/bin", "Library/mingw-w64/bin", "python", "bin")
```

PR #2210, opened 2020-03-20
Execute legacy GUI tests before anything else

m-novikov:
It seems like there is slowdown because when GUI tests run in the end which may result in timeouts.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -170,6 +170,7 @@ def pytest_runtestloop(session):
     if session.config.getoption("run_legacy_gui"):
         for tstcls, gui_test_bag in itertools.groupby(_sorted_guitests(guitests), get_guitest_cls):
             run_gui_tests(tstcls, gui_test_bag)
+            ThreadRouter.app_is_shutting_down = False
 
     elif guitests:
         warnings.warn("Skipping legacy GUI test to enable please use --run-legacy-gui option\n")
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -166,15 +166,17 @@ def pytest_runtestloop(session):
     # Modify session leaving only normal tests as session.items
     # Gui test should be run separately
     guitests, session.items = split_guitests(session.items)
-    _pytest_runtestloop(session)
 
     if session.config.getoption("run_legacy_gui"):
         for tstcls, gui_test_bag in itertools.groupby(_sorted_guitests(guitests), get_guitest_cls):
             run_gui_tests(tstcls, gui_test_bag)
+            ThreadRouter.app_is_shutting_down = False
```
k-dominik: Hmm, I feel this should be set before calling `run_gui_tests(...)` so that when exiting the loop, the `app_is_shutting_down` value is True (as it should be)?
Or do you expect any threadrouted stuff afterwards?

m-novikov: It's only set to false once the application is shutdown so in this case after each GUI test. The first test still runs in app_is_shutting_down = False because nothing happened before.

k-dominik: but after the tests are done, this should be `True`, right? With the current code it is set to `False` in after the last test, which is the "wrong" state. I know we implicitly assume that is not a problem, but well...

m-novikov: After tests are done it still should be False, because we are testing ilastik in running state. Otherwise a lot of callback marked as thread router wouln't execute.

k-dominik: after the gui tests are done, none of that should happen, right?

m-novikov: Why? There are still tests for some applets which fail if variable is not reset.

k-dominik: well, in any case - it seems like the app_is_shutting_down variable is triggered after every gui test as far as I understand the code. So all gui tests with the same class except the first one will get an invalid threadrouter, this can't be the solution.

PR #2214, opened 2020-03-24
mnt: numpy.fromstring -> numpy.frombuffer

k-dominik:
the quest of reducing number of warnings during tests...

gets rid of

`DeprecationWarning: The binary mode of fromstring is deprecated, as it behaves surprisingly on unicode inputs. Use frombuffer instead`


m-novikov:
Just to note maybe it makes sense to split it into multiple pull-requests, because some changes clearly should have no side-effect like fromstring -> frombuffer rename. This way we would see improvements faster and can contribute also.

PR #2217, opened 2020-03-25
Circle activate workflows

k-dominik:


m-novikov:
If we use `continuumio/miniconda3` image then to fix all conda init issues we would only need to set environment variable `BASH_ENV: /opt/conda/etc/profile.d/conda.sh` in yaml.

.circleci/config.yml
```diff
@@ -1,82 +1,137 @@
...
+    - save_cache:
+        key: v1-repo-{{ .Environment.CIRCLE_SHA1 }}
+        paths:
+        - ~/ilastik
+
+  configure_conda:
+    working_directory: ~/ilastik/ilastik
+    parallelism: 1
+    environment:
+      BASH_ENV: /opt/conda/etc/profile.d/conda.s
```
m-novikov: Typo in path should end with .sh

k-dominik: I've removed it completely - the miniconda3 container already seems to be correctly configured

.circleci/config.yml
```diff
@@ -1,82 +1,130 @@
...
         cd ${ILASTIK_ROOT}/ilastik-meta/ilastik &&
         mkdir test-results &&
         xvfb-run --server-args="-screen 0 1024x768x24" pytest --run-legacy-gui --junitxml=test-results/junit.xml
+    - store_test_results:
+        path: test-results
+    - save_cache:
+        name: store dependency cache
+        key: v1.4.8-dep-{{ .Branch }}-{{ epoch }}
+        paths:
+        - /opt/conda/pkg
```
m-novikov:
```suggestion
        - /opt/conda/pkgs
```

k-dominik: thx

k-dominik:
3000 force-pushs later....

PR #2222, opened 2020-03-31
do not normalize

FynnBe:
does not work like this as reorder_feature_image always is of type float32.
we need to check meta data instead

Let's discuss and do it better!


m-novikov:
What happens if input data doesn't use the full range of data type (e.g. uint8 data only uses 0 - 128)?

FynnBe:
ok, to sum up our online offline conversation: 

- no normalization in ilastik
- only batch normalization for now
- add statistics object for global normalization later

PR #2226, opened 2020-04-02
fix leaving cplex and gurobi in PATH

k-dominik:
The previous method of filtering would not leave even the default installation paths of cplex or gurobi in PATH.

This commit allows for paths like
`C:\something\arbitrary\gurobi*\**`
`C:\something\arbitrary\cplex*\**`

which is not possible with `pathlib.PurePath.match` alone.


k-dominik:
thx for the trust, I tested it manually and will write one after moving `ilastik.py`

PR #2232, opened 2020-04-06
supply loaded axistags to PreloadedArray

k-dominik:
when deserializing a project in headless mode with the training data missing,
the axistags have to be deserialized, too. Otherwise ilastik guesses them.
If this guess is wrong, the data doesn't fit the workflow anymore.

* fixes running ilastik projects in headless without the data.
* fixes memory allocation for data for missing data in headless (previously an array of the same size as the missing image would have been allocated in memory).

### DIFF
--- a/ilastik/applets/dataSelection/opDataSelection.py
+++ b/ilastik/applets/dataSelection/opDataSelection.py
@@ -47,7 +47,8 @@ from ilastik.utility import OpMultiLaneWrapper
 from lazyflow.utility.pathHelpers import splitPath, globH5N5, globNpz, PathComponents
 from lazyflow.utility.helpers import get_default_axisordering
 from lazyflow.operators.opReorderAxes import OpReorderAxes
-from lazyflow.graph import Graph, Operator
+from lazyflow.operators import OpZeroSource
+from lazyflow.graph import Graph
 
 
 def getTypeRange(numpy_type):
@@ -387,6 +388,47 @@ class PreloadedArrayDatasetInfo(DatasetInfo):
         opReader = OpArrayPiper(parent=parent)
         opReader.Input.setValue(self.preloaded_array)
         return opReader.Output
+
+
+class DummyDatasetInfo(DatasetInfo):
+    """Special DatasetInfo for datasets that can't be found in headless mode
+    """
+
+    def __init__(self, *, project_file: h5py.File = None, **info_kwargs):
+        super().__init__(**info_kwargs)
+
+    @property
+    def legacy_location(self) -> str:
+        return "Outer Space"
+
+    @property
+    def effective_path(self) -> str:
+        return "Outer Space"
+
+    @property
+    def display_string(self) -> str:
+        return "Dummy Zero Array"
+
+    def to_json_data(self) -> Dict:
+        raise NotImplemented("Dummy Slots should not be serialized!")
+
+    def get_provider_slot(self, parent: Operator) -> OutputSlot:
+        opZero = OpZeroSource(shape=self.laneShape, dtype=self.laneDtype, axistags=self.axistags, parent=parent)
+        return opZero.Output
+
+    @classmethod
+    def from_h5_group(cls, data: h5py.Group, params: Dict = None):
+        laneShape = tuple(data["shape"])
+        default_tags = get_default_axisordering(laneShape)
+
+        params = params or {}
+        if "laneShape" not in params:
+            params["laneShape"] = laneShape
+        if "default_tags" not in params:
+            params["default_tags"] = default_tags
+        if "laneDtype" not in params:
+            params["laneDtype"] = numpy.uint8
+        return super().from_h5_group(data, params)
 
 
 class UrlDatasetInfo(DatasetInfo):
--- a/lazyflow/operators/valueProviders.py
+++ b/lazyflow/operators/valueProviders.py
@@ -496,3 +496,29 @@ class OpZeroDefault(Operator):
     def propagateDirty(self, slot, subindex, roi):
         if slot == self.Input:
             self.Output.setDirty(roi)
+
+
+class OpZeroSource(Operator):
+    """Operator that serves zeros on it's Output
+
+    This operator will never set anything dirty as it cannot be changed.
+    """
+
+    Output = OutputSlot()
+
+    def __init__(self, dtype, shape, *, graph=None, parent=None, **kwargs):
+        super().__init__(graph=graph, parent=parent)
+
+        self._meta = {"dtype": dtype, "shape": shape}
+        self._meta.update(kwargs)
+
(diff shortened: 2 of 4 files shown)


Tomaz-Vieira:
Hey, thanks for catching that =)

I have no idea why I didn't do it, but the sensible thing to do would be to call `PreloadedArrayDatasetInfo.from_h5_group(infoGroup)`, and have `PreloadedArrayDatasetInfo` override the base implementation of that method slightly, to provide the empty array (or, even better, to to so with a sparse array or whatever). That way, all deserialization logic, as you suggested, would be in the specialized class instead of external to it =)

k-dominik:
>  have no idea why I didn't do it, but the sensible thing to do would be to call `PreloadedArrayDatasetInfo.from_h5_group(infoGroup)`, and have `PreloadedArrayDatasetInfo` override the base implementation of that method slightly, to provide the empty array (or, even better, to to so with a sparse array or whatever). That way, all deserialization logic, as you suggested, would be in the specialized class instead of external to it =)

I agree on the `.from_h5_group` bit, but will probably put it in a new subclass of DatasetInfo.

lazyflow/operators/valueProviders.py
```diff
@@ -496,3 +496,29 @@ def execute(self, slot, subindex, roi, result):
...
+
+    Output = OutputSlot()
+
+    def __init__(self, dtype, shape, *, graph=None, parent=None, **kwargs):
+        super().__init__(graph=graph, parent=parent)
+
+        self._meta = {"dtype": dtype, "shape": shape}
+        self._meta.update(kwargs)
+
+    def setupOutputs(self):
```
Tomaz-Vieira, line 515: Can we move this logic into the constructor? We can leave an empty `setupOutputs` method if lazyflow demands it, but since this operator has no Inputs it'll never be called anyway, and then we can move closer to a world of meaningful constructors in ilastik. As a bonus, we don't even have to keep the private `meta` dict in the operator, as `dtype` and `shape` can both be used straight from the constructor params into the `Output.meta` dict

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -389,6 +390,47 @@ def get_provider_slot(self, parent: Operator) -> OutputSlot:
...
+
+    def to_json_data(self) -> Dict:
+        raise NotImplemented("Dummy Slots should not be serialized!")
+
+    def get_provider_slot(self, parent: Operator) -> OutputSlot:
+        opZero = OpZeroSource(shape=self.laneShape, dtype=self.laneDtype, axistags=self.axistags, parent=parent)
+        return opZero.Output
+
+    @classmethod
+    def from_h5_group(cls, data: h5py.Group, params: Dict = None):
```
Tomaz-Vieira, line 420: The reason the `params` dict is accepted in the base `DatasetInfo.from_h5_group` is just so that the child classes can parse their specific fields and pass those up those up to the super method (which i'm not sure was good design anyway). If you think it might be useful to override some of the parameters, like `laneShape` or `default_tags`, then maybe it would be a good idea to have those as explicit parameters of the `from_h5_group` method

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -389,6 +390,47 @@ def get_provider_slot(self, parent: Operator) -> OutputSlot:
...
+        laneShape = tuple(data["shape"])
+        default_tags = get_default_axisordering(laneShape)
```
Tomaz-Vieira, line 422: I wonder if this can bite us later on... This `DummyDatasetInfo` is supposed to be a stand-in for missing data, and that missing data might have different default axistags depending on the file type (e.g.: .png might be yxc, precomputed chunks might be cxy). This will probably never be a problem because the axistags are always saved into the .ilp files, and therefore the default_tags would never have to be used, but it's something to keep in mind.

lazyflow/operators/valueProviders.py
k-dominik: yes, that makes sense for this operator. But just to clarify - `setupOutput` method _is_ in fact called (Otherwise the tests would also fail).

ilastik/applets/dataSelection/opDataSelection.py
k-dominik: yes, good point. The nice thing, if we were to save them, would be more straight forward deserialization (without the `get_default_axisordering` workaround, which is, as you already stated meaningless since `axisorder` is always supplied).

ilastik/applets/dataSelection/opDataSelection.py
k-dominik: do you imply I am misusing it? I see no other way to pass up the missing values that have not been serialized, but are needed for initialization...

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -389,6 +390,47 @@ def get_provider_slot(self, parent: Operator) -> OutputSlot:
...
+
+    @property
+    def display_string(self) -> str:
+        return "Dummy Zero Array"
+
+    def to_json_data(self) -> Dict:
+        raise NotImplemented("Dummy Slots should not be serialized!")
+
+    def get_provider_slot(self, parent: Operator) -> OutputSlot:
+        opZero = OpZeroSource(shape=self.laneShape, dtype=self.laneDtype, axistags=self.axistags, parent=parent)
```
Tomaz-Vieira, line 416: ah, beautiful constructors \o/

ilastik/applets/dataSelection/opDataSelection.py
Tomaz-Vieira: Ah, now I see what's happening. I myself added the `params` parameter to all of the `from_h5_group` methods, which now I think was a bad idea, because no client of those methods ever uses that parameter, and probably never should; it's a dictionary, which can take anything and ignore anything, so it would be very non-intuitive for the clients of those methods and is only justifiable in the base class, since there it is needed to support the specialized `DatasetInfo`s. If we actually wanted the specialized DatasetInfos to allow for some parameters to be overridden, those should be explicit parameters in their respective `from_h5_group` method. 

Ideally the `from_h5_group` methods should be `@staticmethod`s rater than `@classmethod`s, and the base `from_h5_group` should be renamed but still be a `@classmethod` to support the specialized DatasetInfo classes.

Anyway, I guess we can ignore this for now. I can get back to it later and then fix all of the specialized DatasetInfos.

k-dominik: :+1:

ilastik/applets/dataSelection/opDataSelection.py
k-dominik: :_)

PR #2233, opened 2020-04-08
fix import

jburel:
Problem noticed while testing the example with ilastik-meta 1.4.0b1 from conda


k-dominik:
one tiny thing though, could you maybe also resolve the missing `os` import, while you're at it?

jburel:
@k-dominik adjusted the import (one added, one remove)

PR #2238, opened 2020-04-15
open project files as readonly per default in headless

### LINKED ISSUE #1870: open projects in read-only mode per default in headless mode
right now when running multiple instances of ilastik in headless mode accessing the same project file, one has to specify the `--readonly=True` option (and even provide a value...), otherwise `h5py` will throw an exception when trying to access the project file from the second project.

The only case I can think of where opening a project file not in read-only is when retraining. And there is a flag for that, that could already be used to determine in what mode to open the project file. Otherwise I would suggest to make it the default behavior to open project files in read-only mode in headless operation.

k-dominik:
Removes a major annoyance from ilastik's headless mode: allowing for concurrent access to project files per default.

Summary:
* `--readonly` flag kept. Projects will be opened per default with read access in GUI mode, without read access in headless mode. the `--readonly` flag now takes a parameter to allow changing this behavior.

fixes #1870

### DIFF
--- a/ilastik/app.py
+++ b/ilastik/app.py
@@ -27,6 +27,7 @@ from typing import List, Optional, Sequence, Tuple
 
 import ilastik.config
 from ilastik.config import cfg as ilastik_config
+from ilastik.utility.commandLineProcessing import str2bool
 
 logger = logging.getLogger(__name__)
 
@@ -38,8 +39,14 @@ def _argparser() -> argparse.ArgumentParser:
     ap.add_argument("--project", help="A project file to open on startup.")
     ap.add_argument(
         "--readonly",
-        help="Open all projects in read-only mode, to ensure you don't " "accidentally make changes.",
-        action="store_true",
+        nargs="?",
+        default=None,
+        const="true",
+        type=str2bool,
+        help=(
+            "Open all projects in read-only mode, to ensure you don't accidentally make changes. "
+            "Per default projects are opened with read access in GUI mode, without read access in headless mode."
+        ),
     )
     ap.add_argument(
         "--new_project", help="Create a new project with the specified name. Must also specify " "--workflow."
@@ -352,6 +359,10 @@ def _prepare_auto_open_project(parsed_args):
     parsed_args.project = os.path.expanduser(parsed_args.project)
     # convert path to convenient format
     path = PathComponents(parsed_args.project).totalPath()
+
+    # readonly
+    if parsed_args.readonly is None:
+        parsed_args.readonly = parsed_args.headless
 
     def loadProject(shell):
         # This should work for both the IlastikShell and the HeadlessShell
--- a/ilastik/app.py
+++ b/ilastik/app.py
@@ -27,7 +27,7 @@ from typing import List, Optional, Sequence, Tuple
 
 import ilastik.config
 from ilastik.config import cfg as ilastik_config
-from ilastik.utility.commandLineProcessing import str2bool
+from ilastik.utility.commandLineProcessing import OptionalFlagAction
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


emilmelnikov:
Maybe change `--readonly` to accept an optional argument and ask people to use `--headless --readonly=0` instead?

```python
ap = argparse.ArgumentParser()
ap.add_argument(
    "--readonly", nargs="?", default=None, constant='true', type=_str2bool
)
args = ap.parse_args()

if args.readonly is None:
    args.readonly = args.headless

def _str2bool(s):
    t = s.strip().lower()
    if t in {"false", "0", "no"}:
        return False
    if t in {"true", "1", "yes"}:
        return True
    raise ValueError(f"{s!r} cannot be converted to bool")
```

Tomaz-Vieira:
I agree with Emil. Otherwise it will be possible to do this:
```
ilastik --headless --readonly --headless-write-access
```

which is unclear as to what should happen, from the user's perspective

k-dominik:
> Maybe change `--readonly` to accept an optional argument and ask people to use `--headless --readonly=0` instead?
> 
> ```python
> ap = argparse.ArgumentParser()
> ap.add_argument("--readonly", nargs="?", default=None, const='true', type=_str2bool)
> args = ap.parse_args()
> 
> if args.readonly is None:
>     args.readonly = args.headless
> 
> def _str2bool(s):
>     t = s.strip().lower()
>     if t in {"false", "0", "no"}:
>         return False
>     if t in {"true", "1", "yes"}:
>         return True
>     raise ValueError(f"{s!r} cannot be converted to bool")
> ```

I like it!

emilmelnikov:
> I like it!

Just a quick addition: if you're going to use `_str2bool` approach of allowing various strings, consider adding `off`/`on` for false/true flags. This is consistent with `getboolean` in `configparser`.

emilmelnikov:
I think we can improve error messages for users by implementing this as an action instead of just type conversion function:

```python
import argparse
import configparser


class OptionalFlagAction(argparse.Action):
    """Similar to store_true or store_false actions, but also allows an optional argument.

    This action is typically used with ``default`` and ``const`` in :meth:`argparse.ArgumentParser.add_argument`.
      * If the option is not supplied, value of ``default`` will be used (usually left out since it is None by default).
      * If the option is supplied without an argument, value of ``const`` will be used (usually True or False).
      * If the option is supplied with an argument, that argument will be converted to boolean.

    Valid arguments are the same as in :meth:`configparser.ConfigParser.getboolean`.
    """

    def __init__(self, nargs=argparse.OPTIONAL, **kwargs):
        if nargs != argparse.OPTIONAL:
            raise ValueError(f"action {self.__class__.__name__} can only be used with nargs={argparse.OPTIONAL!r}")

        if not isinstance(kwargs.get("const"), bool):
            raise ValueError(f"action {self.__class__.__name__} can only be used with const=True or const=False")

        super().__init__(nargs=nargs, **kwargs)

    def __call__(self, parser, namespace, values, option_string=None):
        if values is None:
            value = self.default

        elif isinstance(values, bool):
            value = values

        else:
            states = configparser.ConfigParser.BOOLEAN_STATES
            try:
                value = states[str(values).lower()]
            except KeyError:
                parser.error(f"invalid value {values!r} for {option_string}; try one of: {', '.join(states)}")
                return

        setattr(namespace, self.dest, value)


# Usage: argument_parser.add_argument("--readonly", action=OptionalFlagAction, const=True)
```

In this case, if user enters something like `--readonly=enabled`, they will get a nice error message that you'd see from argparser, instead of an ugly traceback.

k-dominik:
atm you'd get something like this (no traceback):

```
ilastik.py: error: argument --readonly: invalid str2bool value: 'blah'
```
I would argue it's already good enough.

k-dominik:
but since you have already went through the trouble of implementing it, why don't you add it to this PR (just push to the branch)...

emilmelnikov:
Oh, that's right, I should have checked how it fails beforehand...

k-dominik:
this is one beautiful error message:

```
ilastik.py: error: invalid value 'blah' for --readonly; try one of: 1, yes, true, on, 0, no, false, off
```
:heart_eyes_cat:

Issue #2170, opened 2020-01-13
Smarter axes guess


alyst:
## Description

Use the axes from the last loaded image (if it's compatible) as the initial guess for the axes of the new image.

## Scenario

User adds new images to the *Raw Data* tab, no *"Constraint of 'Pixel Classification' applet was violated"* alert box followed by the image properties dialog popup are being shown.

## Rationale

It would make adding raw images much easier.

k-dominik:
Hi @alyst,

i thought ilastik was doing that (and it did in previous versions). Seems like a change undid this. This will most likely be reverted to the original behavior ( or something maybe even smarter - will be discussed ).

Thanks bringing this up. Wasn't caught in the reviews.

PR #2241, opened 2020-04-24
Guesses axistags for new data to be the same as the last lane if sufficiently similar

Tomaz-Vieira:
Users often use images from the same dataset as different lanes. Those usually match in axistags, but the user was forced to edit the DatasetInfo for every new lane of the same dataset.

This PR modifies `dataSelectionGui` to guess that the new lane should have the same axistags as the previous lane if it matches a few similarity criteria. It also unifies data selection code from several places in the code: `DataSelectionGui`, `BatchPRocessingApplet`, `BatchProcessingGui`. Some of the simplification in the `BatchProcessingApplet` had already been done to support the `--distributed` flag, so the `ilastik_distributed` PR was merged into this one.

Additionally, this also revives the Precomputed Chunks functionality, though with the original inefficient implementation.

Finally, this introduces the `--ignore-training-axistags` flag, that makes headless not use the training axistags onto the new datasets, but rather tries to derive the axis order from data in the files, like the `axes` property in `.n5` files

Needs https://github.com/ilastik/ndstructs/tree/0.0.3dev0 or later to run with `--distributed`

closes https://github.com/ilastik/ilastik/issues/2170


ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -924,9 +945,54 @@ def addLane(self, laneIndex):
...
+        return len(self.innerOperators)
+
+    def get_lane(self, lane_idx: int) -> OpDataSelectionGroup:
+        return self.innerOperators[lane_idx]
+
+    def get_role_info(self, lane_idx: int, role: str) -> Optional[DatasetInfo]:
+        info_slot = self.DatasetGroup[lane_idx][self.role_names.index(role)]
+        return info_slot.value if info_slot.ready() else None
+
+    def get_last_axes_keys(self) -> Dict[str, Optional[AxisTags]]:
```
k-dominik, line 988: "...and naming things". Here I would rather have a more descriptive name. So either something that does not imply the implementation ("last") or that more clearly states that this method gets the axistags of the last lane. So maybe either `get_axes_keys_last_lane`. Or make this method more generic, like `get_axis_keys(self, lane_idx)` and allow a `-1` for lane-index. That way the gui implementation of a guess doesn't dripple down into the operator. Does that make sense?

ilastik/applets/dataSelection/dataSelectionGui.py
```diff
@@ -571,13 +572,37 @@ def _createDatasetInfo(self, roleIndex: int, filePath: Path, roi=None) -> Filesy
...
         )
 
+    def guess_axistags_for(self, role: Union[str, int], info: DatasetInfo) -> Optional[AxisTags]:
+        if self.topLevelOperator.num_lanes == 0:
+            return info.axistags
+        lane = self.topLevelOperator.getLane(-1)
+        previous_info = lane.get_dataset_info(role)
+        if previous_info is None or previous_info.default_tags == previous_info.axistags:
+            return info.axistags
+        if previous_info.__class__ != info.__class__ or len(previous_info.axiskeys) != len(info.axiskeys):
```
k-dominik, line 591: I would argue this comparison is too strong:

* one could have copied the last data to project internal -> won't work anymore
* even if one dataset can be resolved to `RelativeDatasetInfo` and the other one "only" to `AbsoluteDatasetInfo` it won't work.

There are arguments for and against that. I don't know what happens more often in reality. But as it stands right now, having two identical files in different folders could result in the axistags being properly applied in one case, and not in the other.

k-dominik:
Ah it's so great to have this working. Do you think it is possible without breaking a leg to make the stack import use the same code-path // have the same functionality there?

ilastik/applets/dataSelection/opDataSelection.py
Tomaz-Vieira: Makes total sense. In fact, I think I eventually reached the same conclusion as I was fixing the code some more.

ilastik/applets/dataSelection/dataSelectionGui.py
Tomaz-Vieira: Fair point. I'll remove or refine the class comparison

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -924,9 +945,54 @@ def addLane(self, laneIndex):
...
+        if self.num_lanes == 0:
+            return {role_name: None for role_name in self.role_names}
```
emilmelnikov, line 990:
```suggestion
            return dict.fromkeys(self.role_names, None)
```

lazyflow/distributed/TaskOrchestrator.py
```diff
@@ -0,0 +1,78 @@
+from mpi4py import MPI
+from threading import Thread
+import enum
+from typing import Generator, TypeVar, Generic, Callable, Tuple
+import uuid
+
+END = "END-eb23ae13-709e-4ac3-931d-99ab059ef0c2"
```
k-dominik, line 7: could you please add some comments here on the lingo used? What is a datum? What is a tag. So that ppl can make sense of this code without going through an mpi tutorial. (is 13 a special number?). Thx

lazyflow/distributed/TaskOrchestrator.py
```diff
@@ -0,0 +1,78 @@
...
+            raise Exception("Trying to orchestrate tasks with no workers!!!")
+        self.workers = {rank: _Worker(self.comm, rank) for rank in range(1, num_workers + 1)}
+
+    def get_finished_worker(self) -> _Worker[TASK_DATUM]:
+        status = MPI.Status()
+        self.comm.recv(source=MPI.ANY_SOURCE, tag=Tags.TASK_DONE, status=status)
+        return self.workers[status.Get_source()]
+
+    def orchestrate(self, task_data: Generator[TASK_DATUM, None, None]):
+        print(f"ORCHESTRATOR: Starting orchestration of {len(self.workers)}...")
```
k-dominik, line 47: do you rely on printing? or was that just for debugging? might consider switching to logging.

ilastik/applets/dataSelection/dataSelectionGui.py
```diff
@@ -403,10 +403,11 @@ def addFiles(self, roleIndex, startingLaneNum=None):
 
     def addFileNames(self, paths: List[Path], startingLaneNum: int, roleIndex: int):
         # If the user didn't cancel
-        if paths:
+        for path in paths or []:
             try:
-                new_infos = self._createDatasetInfos(roleIndex, paths)
-                self.addLanes(new_infos, roleIndex=roleIndex, startingLaneNum=startingLaneNum)
+                full_path = self._get_dataset_full_path(path, roleIndex=roleIndex)
+                info = self.instantiate_dataset_info(url=str(full_path), role=roleIndex)
```
k-dominik, line 409: again, naming with url and path is a bit unfortunate

ilastik/applets/dataSelection/dataSelectionApplet.py
```diff
@@ -105,6 +121,87 @@ def topLevelOperator(self):
...
+                    named_configured_roles = {k: v for k, v in zip(role_arg_names, role_arg_values) if v}
+                    if values:
+                        if named_configured_roles:
+                            raise ValueError(
+                                "You can only have trailing file paths if no other role was set by name. "
+                                f"You have set the following roles by name: {named_configured_roles}"
+                            )
+                        setattr(namespace, role_arg_names[0], values)
+
+            return ExtraTrailingArgumentsAction
```
k-dominik, line 140: i'd move this to commandLineProcessing

ilastik/applets/dataSelection/dataSelectionApplet.py
```diff
@@ -105,6 +121,87 @@ def topLevelOperator(self):
...
+                  "axistags defined by the training data axistags. If --ignore-training-axistags is set, dataset axistags "
+                  "will be inferred from the file conventions, e.g.: 'axes' in .n5, and not from the "
+                  "training data axistags. Defaults to using the training data axistags"),
+            required=False,
+            type=parse_input_axes,
+            default=[None] * len(role_names)
+        )
+
+        arg_parser.add_argument(
+            "--ignore-training-axistags", "--ignore_training_axistags",
```
k-dominik, line 191: I'm only writing this here, but this goes for all the command line args that you've added. I think we should stick to the very uncommon onderscore. Mixing this now, and enabling this makes things confusing.

ilastik/applets/batchProcessing/batchProcessingApplet.py
```diff
@@ -83,38 +106,32 @@ def run_export(
...
             export_to_array: If True do NOT export to disk as usual.
               Instead, export the results to a list of arrays, which is returned.
               If False, return a list of the filenames we produced to.
-            sequence_axis: stack along this axis, overrides setting from default role
 
         Returns:
             list containing either strings of paths to exported files,
               or numpy.arrays (depending on export_to_array)
         """
+        assert not (export_to_array and export_function)
```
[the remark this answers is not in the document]
Tomaz-Vieira: Agreed, I also don't like it. It's still here only for backwards compatibility with clients like `examples/example_python_client.py`

ilastik/applets/batchProcessing/batchProcessingApplet.py
```diff
@@ -123,74 +140,42 @@ def lerpProgressSignal(a, b, p):
...
-            self.dataSelectionApplet.topLevelOperator.removeLane(original_num_lanes, original_num_lanes)
-
-    @property
-    def num_lanes(self) -> int:
-        return len(self.dataSelectionApplet.topLevelOperator.DatasetGroup)
-
-    @property
-    def role_names(self) -> List[str]:
-        return self.dataSelectionApplet.topLevelOperator.DatasetRoles.value
+            self.dataSelectionApplet.popLane()
```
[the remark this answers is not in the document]
k-dominik: sorry, my bad. I misread the diff here

ilastik/applets/batchProcessing/batchProcessingGui.py
```diff
@@ -316,30 +310,17 @@ def partial_copy(name, obj):
...
-                role_path_dict[role_name] = [None] * num_paths
-
-            if len(role_path_dict[role_name]) != num_paths:
-                raise BatchProcessingDataConstraintException(
-                    f"Number of files for '{role_name!r}' does not match! " f"Exptected {num_paths} files."
-                )
-
         # Run the export in a separate thread
-        export_req = Request(partial(self.parentApplet.run_export, role_path_dict))
+        lane_configs = self.parentApplet.dataSelectionApplet.create_lane_configs(role_inputs=role_inputs)
```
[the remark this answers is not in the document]
k-dominik: hm. Maybe I'll rephrase:

This code is in the UI part. I wanted to make a distinction on what qualifies as an exception here. And for me, from a UI point of view, not having added the right number of files does not qualify as an exception (like something went wrong). I wouldn't want a user to get this shock of "ugh, I broke something" for something as mundane as adding files in the GUI. I think an information would be more in place (even if it's relaying the same sentence).

tests/test_ilastik/test_workflows/test_DatasetInfo.py
```diff
@@ -202,14 +202,3 @@ def test_h5_stack_via_star_file_glob_and_stared_internal_path(h5_stack_dir, empt
     )
 
     assert info.filePath == expected_filepath
-
-
-def test_guess_tags_for_singleton_axes(h5_1_100_200_1_1, empty_project_file):
```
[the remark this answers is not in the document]
k-dominik: thank you for the clarification!

Tomaz-Vieira:
> Quite the ride. I like the direction this is going!
> 
> In general I would really advise on adding docstrings to your functions/classes/methods. I don't think this is ever a wasted effort.
> 
> Also it seems you have not added a single test. This seems a bit odd, as there is at least a few new command line arguments, so new functionality.
> Do you think the Distributed stuff is not testable on CI?

You're totally right, I completely forgot to test the new command line params -.- I guess I was too focused on making the other tests pass... I'll work on that. The `--distributed` para can work in the CI, I guess, as long as it's fine to also install mpi4py in the test environment

tests/test_ilastik/test_workflows/test_HeadlessPixelClassificationWorkflow.py
```diff
@@ -0,0 +1,178 @@
...
+    assert project.exists()
+    assert raw_data.parent.exists()
+
+    ilastik_dot_py = Path(__file__).parent.parent.parent.parent / "ilastik.py"
+    subprocess_args = [
+        "python",
+        str(ilastik_dot_py),
+        "--headless",
+        "--project=" + str(project),
+        "--raw-data=" + str(raw_data),
```
k-dominik, line 83: interesting, wouldn't have thought this works with kebap-case for raw data.
```suggestion
        "--raw_data=" + str(raw_data),
```

Tomaz-Vieira: As I was reworking the cmd line parsing, I added this to `dataSelectionApplet.py`:
```py3
    arg_parser.add_argument(
        "--" + arg_name,
        "--" + arg_name.replace("_", "-"), #allow kebap-case
        nargs="+",
        help=f"List of input files for the {role_name} role",
    )
```

This allows for both cases and prevents some silly errors we would get before, where ilastik could think that `--raw-data` was the name of a file

lazyflow/distributed/TaskOrchestrator.py
```diff
@@ -0,0 +1,103 @@
+from mpi4py import MPI
+import enum
+from typing import Iterable, TypeVar, Generic, Callable, Tuple
+import uuid
+import logging
+
+print(f"THIS IS MY NAME: {__name__}")
```
k-dominik, line 7: leftover debug print?

lazyflow/distributed/TaskOrchestrator.py
```diff
@@ -0,0 +1,103 @@
...
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
```
k-dominik, line 9: is not set in `logging_config.json`

lazyflow/distributed/TaskOrchestrator.py
```diff
@@ -0,0 +1,103 @@
...
+
+    In order to use this class, applications must be launched with mpirun: e.g.: mpirun -N <num_workers> ilastik.py
+    """
+
+    def __init__(self, comm=None):
+        self.comm = comm or MPI.COMM_WORLD
+        self.rank = self.comm.Get_rank()
+        num_workers = self.comm.size - 1
+        if num_workers <= 0:
+            raise Exception("Trying to orchestrate tasks with no workers!!!")
```
k-dominik, line 52: could be maybe a `RuntimeError`? Just so that noone has to do bare `catch Exception` when running this. Or even better have your own `OrchestratorException` class for users...

lazyflow/operators/ioOperators/opFormattedDataExport.py
```diff
@@ -258,3 +279,36 @@ def run_export(self):
...
+            cutout = self.get_roi()
+            orchestrator.orchestrate(cutout.split(block_shape=block_shape))
+        else:
+
+            def process_tile(tile: Slice5D, rank: int):
+                self.set_roi(tile)
+                slices = tile.to_slices(output_meta.getAxisKeys())
+                with z5py.File(n5_file_path, "r+") as n5_file:
+                    dataset = n5_file[self.OutputInternalPath.value]
+                    dataset[slices] = self.ImageToExport.value
```
k-dominik, line 312: I'm a bit concerned about this approach (please correct my if I have overlooked something here). Please correct me if I am wrong, but each worker will always process a whole block at once. With the standard block size of 256, this can blow up with 3D time series. Just consider a 2-channel block: `256**4 * 2 * 32bit ~ 35GB`. In order not to blow up nodes with this configuration one would feel inclined to lower the block size, but this will be inefficient quickly due to halo.
Blocking over time should be reconsidered IMO.

ilastik/applets/dataSelection/dataSelectionApplet.py
```diff
@@ -105,6 +121,87 @@ def topLevelOperator(self):
...
+        )
+
+        def parse_input_axes(raw_input_axes: str) -> List[Optional[vigra.AxisTags]]:
+            input_axes = [axes.strip() for axes in raw_input_axes.split(",")]
+            if len(input_axes) > len(role_names):
+                raise ValueError("Specified input axes exceed number of data roles ({role_names})")
+            if len(input_axes) == 1:
+                input_axes = input_axes * len(role_names)
+            else:
+                input_axes += ["None"] * (len(role_names) - len(input_axes))
```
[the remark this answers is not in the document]
k-dominik: bump.

ilastik/applets/batchProcessing/batchProcessingGui.py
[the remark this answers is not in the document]
k-dominik: Do you disagree that getting a popup _error_ (uncaught exception) message is a bit too hysterical here?

ilastik/applets/dataSelection/dataSelectionApplet.py
```diff
@@ -105,6 +121,87 @@ def topLevelOperator(self):
...
+        arg_parser.add_argument(
+            "--ignore-training-axistags", "--ignore_training_axistags",
+            help="Do not use training data to guess input data axis on headless runs. Can still use file axis conventions",
+            action="store_true",
+        )
+
+        arg_parser.add_argument(
+            "--stack-along",
+            "--stack_along",
+            help="Axis along which stack datasets (e.g.: my_yx_slices*.tiff) are to be stacked",
```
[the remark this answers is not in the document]
k-dominik: again, I would say that we should say in this doc string that the stacked axes comes last - otherwise some parameters (subregion and so on) will be impossible to infer.

lazyflow/operators/ioOperators/ioOperators.py
```diff
@@ -641,6 +641,14 @@ def execute(self, slot, subindex, rroi, result):
 
         # Save the axistags as a dataset attribute
         self.d.attrs["axistags"] = self.Image.meta.axistags.toJSON()
+        if isinstance(self.d, h5py.Dataset):
+            for index, tag in enumerate(self.Image.meta.axistags):
+                self.d.dims[index].label = tag.key
```
k-dominik, line 646: :+1: for sneaking in proper axis labels!

lazyflow/utility/pathHelpers.py
```diff
@@ -413,3 +423,9 @@ def globNpz(path: str, globString: str):
 def globList(listOfPaths, globString):
     matches = [x for x in listOfPaths if fnmatch.fnmatch(x, globString)]
     return matches
+
+
+if __name__ == "__main__":
+    import doctest
+
+    doctest.testmod()
```
k-dominik, line 431: can be removed; we have removed those main sections in the tests (or well started to do so), so it is against that "style" to add those main functions here.

```bash
pytest --doctest-modules ...
```

tests/test_ilastik/test_workflows/test_HeadlessPixelClassificationWorkflow.py
k-dominik: :+1:, I like that!

tests/test_ilastik/test_applets/batchProcessing/test_batchProcessingGui.py
```diff
@@ -207,10 +207,10 @@ def assert_raises(exception_type):
     "secondary_files,expectation",
     [
         (tuple(), does_not_raise),
-        (("sec1",), assert_raises(BatchProcessingDataConstraintException)),
+        (("sec1",), does_not_raise),
```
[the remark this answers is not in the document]
k-dominik: could you please comment on this?

ilastik/applets/dataSelection/dataSelectionApplet.py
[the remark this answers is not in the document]
k-dominik: okay, a previous comment cleared this up for me. resolved...

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -93,33 +94,49 @@ def __init__(
...
+        return opMetadataInjector.Output
+
+    @abstractmethod
+    def get_non_transposed_provider_slot(
+        self, parent: Optional[Operator] = None, graph: Optional[Graph] = None
+    ) -> OutputSlot:
+        """Gets an OutputSlot which can be queried for the data of this DatasetInfo.
+
+        Like with operators, either parent or graph must be provided, but not both"""
+        pass
```
k-dominik, line 138: doesn't appear that this will work as expected?!

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -384,8 +403,8 @@ def to_json_data(self) -> Dict:
     def display_string(self) -> str:
         return "Preloaded Array"
 
-    def get_provider_slot(self, parent: Operator) -> OutputSlot:
-        opReader = OpArrayPiper(parent=parent)
+    def get_provider_slot(self, parent: Optional[Operator] = None, graph: Optional[Graph] = None) -> OutputSlot:
+        opReader = OpArrayPiper(parent=parent, graph=graph)
```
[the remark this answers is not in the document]
k-dominik: sure, but `get_provider_slot` returns a slot as it appears. And sure, slots come from operators, but ... I don't think it's obvious that one of `{parent, graph}` has to be supplied. both are marked optional

ilastik/applets/dataSelection/dataSelectionSerializer.py
```diff
@@ -63,11 +64,9 @@ class DataSelectionSerializer(AppletSerializer):
     """
 
     # Constants
-    LocationStrings = {"FileSystem": FilesystemDatasetInfo, "ProjectInternal": ProjectInternalDatasetInfo}
     InfoClassNames = {
-        ProjectInternalDatasetInfo.__name__: ProjectInternalDatasetInfo,
-        FilesystemDatasetInfo.__name__: FilesystemDatasetInfo,
-        RelativeFilesystemDatasetInfo.__name__: RelativeFilesystemDatasetInfo,
+        klass.__name__: klass
+        for klass in [ProjectInternalDatasetInfo, FilesystemDatasetInfo, RelativeFilesystemDatasetInfo, UrlDatasetInfo]
```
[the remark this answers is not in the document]
k-dominik: lol, sorry :)

lazyflow/operators/ioOperators/opFormattedDataExport.py
```diff
@@ -150,7 +156,22 @@ def setupOutputs(self):
             )
 
         new_start, new_stop = tuple(clipped_start), tuple(clipped_stop)
+        return new_start, new_stop
+
+    def get_cutout(self) -> Slice5D:
+        input_axiskeys = self.Input.meta.getAxisKeys()
+        cutout_start, cutout_stop = self.get_new_roi()
+        cutout_slices = tuple(slice(start, stop) for start, stop in zip(cutout_start, cutout_stop))
+        return Slice5D.zero(**{axis: slc for axis, slc in zip(input_axiskeys, cutout_slices)})
```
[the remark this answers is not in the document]
k-dominik: maybe that's it there alright, maybe only make this thing return those `slice(0, 1)`s and have a separate function to create a slice from axistags. `numpy.zeros` will also only give you an array of zeros (in the specified dtype), no matter what you put into the constructor. Also with `slice(0,1)` you will get something  non-zero from a  list (or similar). There is `slice(0)`, which I probably would expect if I get another slice-leaning object. Currently, `Slice5D.zero()` returns an object that when used in indexing will return an array of 5D shape with a single element (the origin data point), but for someone familiar with slices, a zero slice would return an array with zero elements.

tests/test_ilastik/test_workflows/test_HeadlessPixelClassificationWorkflow.py
```diff
@@ -0,0 +1,178 @@
...
+
+def test_headless_2d3c_with_same_raw_data_axis(testdir, pixel_classification_ilp_2d3c: Path, tmp_path: Path):
+    raw_100x100y3c: Path = create_h5(numpy.random.rand(100, 100, 3), axiskeys="yxc")
+    output_path = tmp_path / "out_100x100y3c.h5"
+    run_headless_pixel_classification(
+        testdir, project=pixel_classification_ilp_2d3c, raw_data=raw_100x100y3c, output_filename_format=str(output_path)
+    )
+
+
+def test_headless_2d3c_with_swizzled_raw_data_axis(testdir, pixel_classification_ilp_2d3c: Path, tmp_path: Path):
```
k-dominik, line 114: ...naming things... a term like "permuted" is maybe more common that "swizzled" (at least I had to look it up)

tests/test_ilastik/test_workflows/test_HeadlessPixelClassificationWorkflow.py
```diff
@@ -0,0 +1,178 @@
...
+    raw_3c100x100y: Path = create_h5(numpy.random.rand(3, 100, 100), axiskeys="cyx")
+    output_path = tmp_path / "out_3c100x100y.h5"
+
+    # default behavior is to try to apply training axistags to the batch data, and therefore fail because raw data's
+    # dimensions do not match that of the training data
```
k-dominik, line 119: I guess the problem here is only the position of the `c` axis... if I knew nothing but this test I'd probably think dimensions would somehow all have to match. But you wouldn't really see an issue with x-y permuations

tests/test_ilastik/test_workflows/test_HeadlessPixelClassificationWorkflow.py
```diff
@@ -0,0 +1,178 @@
...
+        testdir,
+        project=pixel_classification_ilp_2d3c,
+        raw_data=raw_3c100x100y,
+        output_filename_format=str(output_path),
+        input_axes="cyx",
+    )
+
+    # alternatively, since the generated h5 data has the axistags property, we can ignore training data and use that
+    # instead, by using the '--ignore_training_axistags' flag
+    run_headless_pixel_classification(
```
k-dominik, line 139: hang on. Do I understand correctly, that the default behavior now is:

* try to force the training data axistags (probably the ones of the last lane), even if something like proper axistags comes with the data
* if `--ignore-training-flags` is passed, only then the axistags in the file are used.

This seems wrong. Shou;dn't the logic (`DataSelectionGui.guess_axistags_for`) be the same in headless mode as in gui mode? (only apply lane axistags if those were changed?)

ilastik/applets/dataSelection/opDataSelection.py
k-dominik: ah, didn't realize it was in the abc...

lazyflow/distributed/TaskOrchestrator.py
Tomaz-Vieira: Indeed. Changed to a `ValueError`, which seems more appropriate than an `Exception`

lazyflow/operators/ioOperators/opFormattedDataExport.py
Tomaz-Vieira: True that; I overlooked 3D data and/or time series.... Well, the blocks can have arbitrary shape, so maybe defaulting to something more sensible, like

```
{x:256, y: 256, t: 1, z: 1, c:slice(None)}
```
would be a better idea? We could make t so that the user must specify a block size instead of trying to be clever about

lazyflow/utility/pathHelpers.py
Tomaz-Vieira: Moved tests out of doctest and into `tests/`

lazyflow/operators/ioOperators/opFormattedDataExport.py
k-dominik: I'd keep 3D blocks 3D, 2D blocks 2D, so something like

```
{x:256, y: 256, t: 1, z: 256, c:slice(None)}
```
this would be "clamped" in the case of 2D data anyway. But I wouldn't want to force slice-wise computation

tests/test_ilastik/test_workflows/test_HeadlessPixelClassificationWorkflow.py
Tomaz-Vieira: Hmmm I think I agree with you that (even though I don't like it) it would be more consistent to only apply the training axistags onto the new data if the training axistags were changed from the default. Still, we run into the problem that when running headless, the input data may not be available any more. In that case, I don't think there is a way to tell if the axistags have been forced or not. We could save that into the `.ilp` files, but there would be no backwards compatibility.

Also, forcing the  axistags from the last lane is the legacy behavior anyway (create project, permute axis, run headless, check output: it is also permuted), so it might be a good idea to keep that behavior as it is.

k-dominik: allright, let's keep it this way then.

ilastik/applets/batchProcessing/batchProcessingApplet.py
```diff
@@ -54,11 +56,42 @@ def parse_known_cmdline_args(self, cmdline_args):
...
-            help="The length of the side of the tiles used in distributed mode",
-            type=int,
-            default=256,
+            "--distributed_block_roi",
+            "--distributed-block-roi",
+            help=textwrap.dedent(
+                """
+                Determines the dimensions of the blocks used to split the data in distributed mode.
+                Values can be:"
+                    A literal python Dict[str, Optional[int]], with keys in 'xyztc'. Missing keys will default like so:
```
k-dominik, line 84: Nice, making this a dict frees us from all the ambiguities of passing just a sequence! Being very nitpicky I'd mention the simple value for block roi size first. This is what most users want. Typing a dictionary seems like pretty advanced usage without clear benefits. I most cases that I can think of it is more beneficial to crank up the spacial dimensions.

Tomaz-Vieira: > Being very nitpicky I'd mention the simple value for block roi size first

OK. I think I liked the dict explanation first, so that the simple value can then be explained as a special case of the dict, but I do agree that it might overcomplicate for normal users.

> Also our command line parser is very very picky about having spaces in values.

Are you referring to [this](https://github.com/ilastik/ilastik/blob/a1bb868b0a8d43ac3c89e681cc89d43be9591ff7/ilastik/utility/commandLineProcessing.py#L68)? If so, then we're safe; I'm using `ast.literal_eval` for the parsing. Other than that, there's just the responsibility of using quotes when passing parameters in the command line, and `argparse` should handle it just fine.

k-dominik: hm, I was more referring to a [bug in cpython](https://bugs.python.org/issue22433)

tests/test_ilastik/test_applets/batchProcessing/test_batchProcessingGui.py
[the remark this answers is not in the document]
Tomaz-Vieira: So for the switch from `assert_raises` to `does_not_raise`: It is perfectly legitimate to run a batch where some of the iterations of the batch do not have a matching second role; For example, in pixel classification, it is totally fine to have the first iteration of the batch to have `Raw Data` (role 1) and `Prediction Mask` (role 2), and then have the second iteration of the batch only have `Raw Data`, since `Prediction Mask` is an optional input, anyway.

Of course, what role is or isn't optional is determined by the workflow (not the Gui), and checked by the `DataSelectionApplet`; Ideally we should instantiate (rather than mock) a `DataSelectionApplet` and use that for testing, but at this time it is so entangled with everything else that it is really hard to do that.

tests/test_ilastik/test_workflows/test_HeadlessPixelClassificationWorkflow.py
Tomaz-Vieira: Indeed. I rewrote the comments to reflect that

ilastik/applets/batchProcessing/batchProcessingApplet.py
Tomaz-Vieira: Ha. I wasn't aware of that.

Anyway, AFAIU, it only happens when there's an "unknown optional" and a "known positional", which causes the "known positional" to capture the "unknown optional". Since the `BatchProcessingApplet` extends the `DataSelectionApplet` argument parser with the `--distributed-block-roi` parameter, by the time the `unspecified_input_files` positional argument is parsed, `--distributed-block-roi` will no longer be unknown, so the parsing won't break. And since the workflows themselves don't handle any other positional arguments by themselves, we shouldn't get this miss-parsing when they run their `parse_known_args` either.

It's too bad that there is now this extra bit of implicit dependency in the argument parsing. Maybe we should start deprecating the usage of the positional arguments and eventually remove it altogether. Anyway, I added another variation to the tests for the sake of paranoia.

ilastik/applets/dataSelection/dataSelectionApplet.py
```diff
@@ -124,209 +221,110 @@ def parse_known_cmdline_args(cls, cmdline_args, role_names):
...
-                )
-                logger.warning("Performance may suffer as a result.  For best performance, use a chunked HDF5 volume.")
-                logger.warning(
-                    "*******************************************************************************************"
-                )
+    def create_dataset_info(
+        self, url: Union[Path, str], axistags: Optional[vigra.AxisTags]=None, sequence_axis: str="z"
+    ) -> DatasetInfo:
+        url = str(url)
+        if isUrl(url):
```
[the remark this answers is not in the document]
Tomaz-Vieira: I agree. I think the appropriate thing to do would be to change `isUrl` to `isRemote`, since a URL can very well refer to a local path... In fact, the `UrlDatasetInfo` should probably be renamed to `RemoteDatasetInfo` too.

If you don't mind, I'd rather do this in another PR, though. It's an old function, which is used by a few other files in the project =/

ilastik/applets/dataSelection/dataSelectionApplet.py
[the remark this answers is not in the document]
Tomaz-Vieira: I've added some more help text to clarify what will happen with stacked images and the axis along which they stack

ilastik/applets/batchProcessing/batchProcessingApplet.py
k-dominik: I agree that we need to phase positional arguments out. They do more harm than they add convenience imo.

tests/test_ilastik/test_applets/batchProcessing/test_batchProcessingGui.py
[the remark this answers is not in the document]
k-dominik: okay I guess I understand this now. I still think that the functionality of the _batch processing applet_ to raise an exception whenever it determines that roles are not matching. Hence the mocking. With that code path removed I'm not sure this test makes any sense at all.

k-dominik:
okay, we'll merge this now and @Tomaz-Vieira will do a follow-up PR fixing  #2296 :)

PR #2247, opened 2020-05-12
Fix halo offshoot

m-novikov:


tests/test_lazyflow/test_operators/test_tiktorch/test_classifier.py
```diff
@@ -0,0 +1,28 @@
...
+def test_get_halo_returns_values_specified_by_tags(pb_session):
+    model_session = ModelSession(session=pb_session, factory=mock.Mock())
+    assert model_session.get_halo(axes="xy") == [256, 128]
+    assert model_session.get_halo(axes="yx") == [128, 256]
+    assert model_session.get_halo(axes="yxc") == [128, 256, 1]
+
+
+def test_get_halo_returns_0_if_value_is_unspecified(pb_session):
+    model_session = ModelSession(session=pb_session, factory=mock.Mock())
+    assert model_session.get_halo(axes="xyz") == [256, 128, 0]
```
k-dominik, line 27: philosophical question, is the size of the block along a dimension that is not there, if you're asking for that dimensions,  0, or 1?

Tomaz-Vieira:
This sounds like the kind of problem `ndstructs` was created to solve =)
Maybe we should discuss merging some common codepaths later

tests/test_lazyflow/test_operators/test_tiktorch/test_classifier.py
Tomaz-Vieira: I would say that an unmentioned dimension should be `1` if we're talking about a shape, but a `0` if we're talking about a halo

k-dominik: unmentioned? then it should not even exist imo.. if I ask for `xy` then I don't need to have the singletons (or sizes) for `z`

k-dominik: in any case I assume that the code relies on the zeros somewhere. And you are of course right that for a halo one want's 0

PR #2262, opened 2020-07-03
Change default file mode when writing hdf5/n5 image files

k-dominik:
I don't think there is an issue about this, but @constantinpape brought this up:

currently we overwrite image files with the same file name. In general this makes sense, of course, but for hdf5/n5 files, it is possible to add multiple datasets to the result file.

This PR changes the behavior; hdf5/n5 files are now opened in append mode, datasets are overwritten if already present.

I have opened #2410 in order to keep track of progress on hdf5 table export. I think we should merge this already and live with not being able to do the same for the h5 table export for now.

### DIFF
--- a/lazyflow/operators/ioOperators/opExportSlot.py
+++ b/lazyflow/operators/ioOperators/opExportSlot.py
@@ -27,6 +27,7 @@ from builtins import object
 import os
 import shutil
 import collections
+import contextlib
 from functools import partial
 
 import numpy
@@ -265,7 +266,7 @@ class OpExportSlot(Operator):
             with OpStreamingH5N5Reader.get_h5_n5_file(export_components.externalPath, mode="a") as h5N5File:
                 # Create a temporary operator to do the work for us
                 opH5N5Writer = OpH5N5WriterBigDataset(parent=self)
-                if export_components.internalPath in h5N5File:
+                with contextlib.suppress(KeyError):
                     del h5N5File[export_components.internalPath]
                 try:
                     opH5N5Writer.CompressionEnabled.setValue(compress)
--- a/lazyflow/operators/ioOperators/opExportSlot.py
+++ b/lazyflow/operators/ioOperators/opExportSlot.py
@@ -262,18 +262,11 @@ class OpExportSlot(Operator):
         # Create and open the hdf5/n5 file
         export_components = PathComponents(self.ExportPath.value)
         try:
-            if os.path.isdir(export_components.externalPath):  # externalPath leads to a n5 file
-                shutil.rmtree(export_components.externalPath)  # n5 is stored as a directory structure
-            else:
-                os.remove(export_components.externalPath)
-        except OSError as ex:
-            # It's okay if the file isn't there.
-            if ex.errno != 2:
-                raise
-        try:
-            with OpStreamingH5N5Reader.get_h5_n5_file(export_components.externalPath, "w") as h5N5File:
+            with OpStreamingH5N5Reader.get_h5_n5_file(export_components.externalPath, mode="a") as h5N5File:
                 # Create a temporary operator to do the work for us
                 opH5N5Writer = OpH5N5WriterBigDataset(parent=self)
+                if export_components.internalPath in h5N5File:
+                    del h5N5File[export_components.internalPath]
                 try:
                     opH5N5Writer.CompressionEnabled.setValue(compress)
                     opH5N5Writer.h5N5File.setValue(h5N5File)


lazyflow/operators/ioOperators/opExportSlot.py
```diff
@@ -262,18 +262,11 @@ def _export_h5n5(self, compress=False):
...
-            # It's okay if the file isn't there.
-            if ex.errno != 2:
-                raise
-        try:
-            with OpStreamingH5N5Reader.get_h5_n5_file(export_components.externalPath, "w") as h5N5File:
+            with OpStreamingH5N5Reader.get_h5_n5_file(export_components.externalPath, mode="a") as h5N5File:
                 # Create a temporary operator to do the work for us
                 opH5N5Writer = OpH5N5WriterBigDataset(parent=self)
+                if export_components.internalPath in h5N5File:
+                    del h5N5File[export_components.internalPath]
```
emilmelnikov, line 269:
```suggestion
                h5N5File.pop(export_components.internalPath, None)
```

emilmelnikov (approved):
LGTM

lazyflow/operators/ioOperators/opExportSlot.py
k-dominik: I'm afraid `z5py` does not support `pop` :/

emilmelnikov: Oh, that's unfortunate. In this case:
```python
with contextlib.suppress(KeyError):
    del h5N5File[export_components.internalPath]
```

k-dominik: do you really find that more readable than `if ... in ... :` ?

emilmelnikov: It's more "Pythonic": https://docs.python.org/3/glossary.html#term-lbyl.

I like this style overall a bit more because it does things in a single "transaction".

In this case it doesn't really matter though, either option is totally fine!

k-dominik: Now I'm torn ;)

ilastik/widgets/exportObjectInfoDialog.py
```diff
@@ -273,8 +272,9 @@ def validate_before_exit(self):
         self.accept()
 
     def is_valid_path(self, path):
-        match = path.rsplit(".", 1)
-        if len(match) == 1 or match[1] not in ALLOWED_EXTENSIONS:
+        pc = PathComponents(path)
+        ext = pc.extension
+        if ext not in ALLOWED_EXTENSIONS:
             return False
         return True
```
emilmelnikov, line 279:
```suggestion
        return PathComponents(path).extension in ALLOWED_EXTENSIONS
```

ilastik/utility/exportFile.py
```diff
@@ -447,10 +448,17 @@ def write_all(self, mode, compression=None):
         count = 0
         self.ExportProgress(0)
         if mode in ("h5", "hd5", "hdf5"):
-            with h5py.File(self.file_name, "w") as fout:
+            pc = PathComponents(self.file_name)
+
+            with h5py.File(pc.externalPath, "a") as fout:
+                if pc.internalPath not in fout:
+                    fout.create_group(pc.internalPath)
```
emilmelnikov, line 455:
```suggestion
                fout.require_group(pc.internalPath)
```

ilastik/widgets/exportObjectInfoDialog.py
```diff
@@ -256,15 +258,12 @@ def validate_before_exit(self):
...
                 title = "Warning"
                 text = "No file extension or invalid file extension ( %s )\nAllowed: %s"
-                match = path.rsplit(".", 1)
-                if len(match) == 1:
-                    ext = "<none>"
-                else:
-                    ext = match[1]
+
+                ext = path.ext or "<no file extension>"
                 text %= (ext, ", ".join(ALLOWED_EXTENSIONS))
```
emilmelnikov, line 267: This is a very... creative way to format a string. We probably should change this to an f-string.

k-dominik: pushed the wrong branch

ilastik/utility/exportFile.py
k-dominik: pushed the wrong branch

ilastik/widgets/exportObjectInfoDialog.py
k-dominik: pushed the wrong branch

Issue #2265, opened 2020-07-07
Host bioimage.io manifest in ilastik repo


constantinpape:
Wei suggested to host the manifest.yaml for our ilastik models in an ilastik rep so that it's easier for us to change it.
I think this makes sense, he explained it here: https://github.com/bioimage-io/bioimage.io/issues/58#issuecomment-655121166

Where do we want to put it? Here, in tiktorch or somewhere else?

cc @FynnBe @m-novikov

constantinpape:
I made a repo for it here:
https://github.com/ilastik/bioimage-io-models

PR #2267, opened 2020-07-08
Add bioimage.io manifest

constantinpape:
Draft for the bioimage.io manifest.
We could also make a new repo `ilastik/models` that contains the manifest and could also hold some of the models itself.
See https://github.com/deepimagej/models.

Still need to replace some of the things copied from the deepimagej template.

See also #2265

### DIFF
--- a/manifest.bioimage.io.yaml
+++ b/manifest.bioimage.io.yaml
@@ -1,0 +1,35 @@
+format_version: 0.2.0
+config:
+  id: ilastik
+  name: ilastik
+  tags:
+    - ilastik
+  logo: https://raw.githubusercontent.com/deepimagej/models/master/logos/logo.png
+  icon: https://raw.githubusercontent.com/deepimagej/models/master/logos/icon.png
+  splash_title: deepImageJ
+  splash_subtitle: "What's the ilastik blurb?"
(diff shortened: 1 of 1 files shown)


constantinpape:
I thought about this a bit more and I think it would make more sense to have a `ilastik/models` repo where we can also dump models that don't have another repo.
What do you think @k-dominik @m-novikov @FynnBe?
If you agree I can go ahead and create that repo.

k-dominik:
I agree that if we want to add models and the likes, then it probably makes more sense to separate it. 'models' sounds a tad generic for the repo...

constantinpape:
I don't mind a different repo name.
If 'models' sounds too generic we could also go with
'bioimage-io-models'
or
'deep-ilastik-models'

FynnBe:
I like 'bioimage-io-models', they don't all have to be deep. Someone could theoretically add ilastik's random forest as a bioimage-io model...

constantinpape:
> I like 'bioimage-io-models', they don't all have to be deep. Someone could theoretically add ilastik's random forest as a bioimage-io model...

Makes sense; I will create the 'bioimage-io-models' repo later and add the manifest.

PR #2276, opened 2020-07-22
Changes needed for dependency upgrade qt 5.12

k-dominik:
qt 5.12 in combination with Python >= 3.7.4 will bubble up some errors that would previously be ignored (see also https://github.com/ilastik/volumina/pull/261).

This PR includes similar changes to main ilastik.

Todo before this can be merged:

- [x] rebuild on win
- [x] rebuild on osx
  - [x] test on osx
- [x] move packages to main channel
- [x] remove the `REMOVE BEFORE MERGE` commits ;)

### DIFF
--- a/ilastik/applets/counting/countingGuiBoxesInterface.py
+++ b/ilastik/applets/counting/countingGuiBoxesInterface.py
@@ -166,8 +166,8 @@ class ResizeHandle(QGraphicsRectItem):
 
 class QGraphicsResizableRectSignaller(QObject):
     """
-     This class is used to emit signals since only QObjects can do it.
-     Multiple inheritance is not supported for qt-python classes (Qt 4.10)
+    This class is used to emit signals since only QObjects can do it.
+    Multiple inheritance is not supported for qt-python classes (Qt 4.10)
     """
 
     signalHasMoved = pyqtSignal(QPointF)  # The resizable rectangle has moved the new position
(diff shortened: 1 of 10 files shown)


ilastik/applets/dataSelection/dataSelectionGui.py
```diff
@@ -601,9 +600,9 @@ def addStack(self, roleIndex, laneIndex):
         try:
             # FIXME: do this inside a Request
             self.parentApplet.busy = True
-            inner_path = stack_info.importAsLocalDataset(project_file=self.project_file)
+            inner_path = stack_info.importAsLocalDataset(project_file=self.get_project_file())
```
m-novikov, line 603: Does new qt handle properties in a special way?

m-novikov (approved):

ilastik/applets/dataSelection/dataSelectionGui.py
k-dominik: yes

> The cause is PyQt5 accessing all class attributes via PyObject_GetAttr during initialization of the Gui Widget.
In our case, the class is not fully initialized by that time but class properties are still executed (via PyObject_GetAttr).

ilastik/applets/counting/countingGuiBoxesInterface.py
```diff
@@ -183,7 +183,7 @@ class QGraphicsResizableRect(QGraphicsRectItem):
     hoverColor = QColor(255, 0, 0)  # _hovering and selection color
 
     def __init__(self, x, y, h, w, scene=None, parent=None, editor=None):
-        """"
+        """ "
```
m-novikov, line 186: typo

k-dominik: thx, removed that!

PR #2277, opened 2020-07-23
Fixes pixel label deserialization

Tomaz-Vieira:
As [pointed out](https://forum.image.sc/t/ilastik-project-file-cannot-be-loaded-on-another-computer/40545) by user JesseDicello,  deserialization of Pixel Labels still had some problems leftover from the big 1.3.3 object classification workflow cleanup.

This PR  fixes axis reordering in Object Classification Workflows so that they are identical to 1.3.2, and so that they will be serialized in an identical manner.

It also corrects the bad assumption that all pixel labels were being saved without dropping dummy axis; labels will now be read back using the current axis keys of the label slot (except for some version-specific exceptions needed for backwards compatibility).

Finally, labels are being saved with their axistags as attributes in the H5 dataset, so that at least this part of the .ilp is no longer dependent on ilastik's internal state.

### DIFF
--- a/ilastik/__init__.py
+++ b/ilastik/__init__.py
@@ -41,7 +41,7 @@ def _format_version(t):
     return ".".join(str(i) for i in t)
 
 
-__version_info__ = (1, 4, "0b7")
+__version_info__ = (1, 4, "0b8")
 
 __version__ = _format_version(__version_info__)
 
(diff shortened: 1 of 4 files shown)


k-dominik:
Okay there is one click path that I found that might be possible to fix:

pixel classification project created in 1.3.3post3, added annotations.
Opened with this PR and changed axisorder on the input. The annotations are gone. This does not happen with projects created with this PR.

k-dominik:
merging this with #2278 addressing the issue mentioned in https://github.com/ilastik/ilastik/pull/2277#issuecomment-663403380

PR #2280, opened 2020-07-27
Fix obj export empty image

### LINKED ISSUE #2279: Failed to export: "Index 1 is out of bounds [...] "
Hi.
I'm trying out Ilastik and it looks great so far!
But I got an issue with the Export of my data in CSV. For one picture it worked perfectly, now nothing is happening (Configure Feature Table Export). When I try to export in hdf5 for example (Choose Export Image Settings..) I got this error. Maybe one of you got an idea what I am doing wrong?
It seems like an excel-file is created and deleted at the same moment.

Thank you.
M

image

k-dominik:
Tiny fix when exporting object classification to h5 table. Previously this would raise an exception when an image was empty (=no objects).

also made sure callbacks of the progress signal would be unsubscribed in any case. Previously those would've been kept in case of error (and progress would be reported n-times).

maybe fixes #2279

### DIFF
--- a/ilastik/utility/exportFile.py
+++ b/ilastik/utility/exportFile.py
@@ -249,6 +249,10 @@ def create_slicing(axistags, dimensions, margin, feature_table):
         yields also the actual object id
     """
     assert margin >= 0, "Margin muss be greater than or equal to 0"
+    if len(feature_table) == 0:
+        yield from ()
+        return
+
     time = feature_table[Default.TimeColumnName].astype(np.int32)
     minx = feature_table["Bounding Box Minimum_0"].astype(np.int32)
     maxx = feature_table["Bounding Box Maximum_0"].astype(np.int32)
@@ -443,7 +447,7 @@ class ExportFile(object):
                     for file_name in file_names:
                         zip_file.write(file_name)
         self.ExportProgress(100)
-        logger.info("exported %i tables" % count)
+        logger.info(f"exported {count} tables to {self.file_name}.")
 
     def _add_columns(self, table_name, columns):
         if table_name in iter(self.table_dict.keys()):
--- a/ilastik/utility/exportFile.py
+++ b/ilastik/utility/exportFile.py
@@ -1,4 +1,3 @@
-from builtins import range
 import collections
 import numpy as np
 import numpy.lib.recfunctions as nlr
@@ -9,6 +8,9 @@ from sys import stdout
 from zipfile import ZipFile
 from ilastik.applets.objectExtraction.opObjectExtraction import default_features_key
 import logging
+
+from typing import Iterator, List, Tuple
+
 
 logger = logging.getLogger(__name__)
 try:
@@ -243,16 +245,48 @@ def _slice_range(low, high, max_):
     return slice(start, end, None)
 
 
-def create_slicing(axistags, dimensions, margin, feature_table):
-    """
-    Returns an iterator on the slices for each object roi
-        yields also the actual object id
+def create_slicing(
+    axistags: AxisTags, dimensions: Tuple[int, int, int, int, int], margin: int, feature_table: np.ndarray
+) -> Iterator[Tuple[List[slice], int]]:
+    """Transforms bounding box in feature table to slicing
+
+    ``feature_table`` should have the following fields:
+
+        ``labelimage_oid``
+            object ids, reset per timeframe
+        ``Bounding Box Minimum_0``
+            minimum in x-direction
+        ``Bounding Box Maximum_0``
+            maximum in x-direction
+        ``Bounding Box Minimum_1``
+            minimum in y-direction
+        ``Bounding Box Maximum_1``
+            maximum in y-direction
+
+        ``feature_table`` may also have the following fields (in case of 3D data):
+
+        ``Bounding Box Minimum_2``
+            minimum in z-direction
+        ``Bounding Box Maximum_2``
+            maximum in z-direction
+
+
+    Args:
+        axistags: image axis tags
+        dimensions: 5D image dimensions in (t, x, y, z, c) -order
+        margin: how much "halo" around each object's bounding box
+        feature_table: structured feature array
+
+
+    Yields:
+        tuple with slices for each object (bounding box + margin) and the object id
     """
     assert margin >= 0, "Margin muss be greater than or equal to 0"
+    assert len(dimensions) == 5, "Dimensions must be 5D"
     if len(feature_table) == 0:
-        yield from ()
         return
 
+    oids = feature_table["labelimage_oid"].astype(np.int64)
     time = feature_table[Default.TimeColumnName].astype(np.int32)
     minx = feature_table["Bounding Box Minimum_0"].astype(np.int32)
     maxx = feature_table["Bounding Box Maximum_0"].astype(np.int32)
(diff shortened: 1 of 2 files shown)


ilastik/utility/exportFile.py
```diff
@@ -249,6 +249,9 @@ def create_slicing(axistags, dimensions, margin, feature_table):
         yields also the actual object id
     """
     assert margin >= 0, "Margin muss be greater than or equal to 0"
+    if len(feature_table) == 0:
+        return None
```
Tomaz-Vieira, line 253: This function's docstring says that it returns an iterator, but this line returns a `None`. Seems like it should return an empty iterator if that makes sense, or throw an exception if it doesn't.

k-dominik: Heya, thanks a lot for picking this up! I've addressed it so you could have another look :)

Tomaz-Vieira (approved):
LGTM =)

ilastik/utility/exportFile.py
```diff
@@ -249,35 +249,38 @@ def create_slicing(axistags, dimensions, margin, feature_table):
...
+            maxz = feature_table["Bounding Box Maximum_2"].astype(np.int32)
+        except ValueError:
+            minz = [0] * table_shape
+            maxz = [1] * table_shape
+
+        indices = list(map(axistags.index, "txyzc"))
+        excludes = indices.count(-1)
+        oid = 1
+        for i in range(table_shape):
+            if time[i] != time[i - 1]:
```
emilmelnikov, line 272: When `i = 0`, `time[i] != time[i - 1]` compares first and last time points, which is confusing. Is this the intended behviour?

ilastik/utility/exportFile.py
```diff
@@ -249,35 +249,38 @@ def create_slicing(axistags, dimensions, margin, feature_table):
...
+                oid = 1
+            # noinspection PyTypeChecker
+            slicing = [
+                slice(time[i], time[i] + 1),
+                _slice_range(minx[i] - margin, maxx[i] + margin, dimensions[1]),
+                _slice_range(miny[i] - margin, maxy[i] + margin, dimensions[2]),
+                _slice_range(minz[i] - margin, maxz[i] + margin, dimensions[3]),
+                slice(None),
+            ]
+            yield [slicing[x] for x in indices][: 5 - excludes], oid
```
emilmelnikov, line 282: This seems to be invalid if `indices` has `-1` elements somewhere in the middle or on the left side.

ilastik/utility/exportFile.py
```diff
@@ -249,35 +249,38 @@ def create_slicing(axistags, dimensions, margin, feature_table):
...
+        try:
+            minz = feature_table["Bounding Box Minimum_2"].astype(np.int32)
+            maxz = feature_table["Bounding Box Maximum_2"].astype(np.int32)
+        except ValueError:
+            minz = [0] * table_shape
+            maxz = [1] * table_shape
+
+        indices = list(map(axistags.index, "txyzc"))
+        excludes = indices.count(-1)
+        oid = 1
```
emilmelnikov, line 270: Nitpick, but maybe update `oid` at the beginning of the loop?

```python
oid = 0
for ...:
    if ...:
        oid = 0
    oid += 1
```

ilastik/utility/exportFile.py
```diff
@@ -249,35 +249,38 @@ def create_slicing(axistags, dimensions, margin, feature_table):
...
+        maxx = feature_table["Bounding Box Maximum_0"].astype(np.int32)
+        miny = feature_table["Bounding Box Minimum_1"].astype(np.int32)
+        maxy = feature_table["Bounding Box Maximum_1"].astype(np.int32)
+        table_shape = feature_table.shape[0]
+        try:
+            minz = feature_table["Bounding Box Minimum_2"].astype(np.int32)
+            maxz = feature_table["Bounding Box Maximum_2"].astype(np.int32)
+        except ValueError:
+            minz = [0] * table_shape
+            maxz = [1] * table_shape
```
emilmelnikov, line 266: For consistency with the other values:

```suggestion
            minz = np.zeros(table_shape, np.int32)
            maxz = np.ones(table_shape, np.int32)
```

ilastik/utility/exportFile.py
```diff
@@ -249,35 +249,38 @@ def create_slicing(axistags, dimensions, margin, feature_table):
...
-            slice(time[i], time[i] + 1),
-            _slice_range(minx[i] - margin, maxx[i] + margin, dimensions[1]),
-            _slice_range(miny[i] - margin, maxy[i] + margin, dimensions[2]),
-            _slice_range(minz[i] - margin, maxz[i] + margin, dimensions[3]),
-            slice(None),
-        ]
-        yield [slicing[x] for x in indices][: 5 - excludes], oid
-        oid += 1
+    if len(feature_table) == 0:
+        yield from ()
```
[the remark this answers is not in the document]
k-dominik: hmmm I had this first, and then thought for some reason that it would be more clear with the if else. But you're right, there's the danger of someone adding code after the currently indented one ...

ilastik/utility/exportFile.py
k-dominik: I guess there is again some implicit "information in here". The assumption is that data will always have at least `txy` and only `z` and `c` are optional. I'm gonna change it to something like:

```python
yield [slicing[x] for x in indices if x != -1], oid
```

ilastik/utility/exportFile.py
k-dominik: i think per definition oid 0 is the background and not included in the table. So `oid = 1` is the first object. Or do I misunderstand the point of your comment?

ilastik/utility/exportFile.py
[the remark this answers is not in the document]
emilmelnikov: I suggested this mainly as an early exit pattern, in order to reduce indentation.

ilastik/utility/exportFile.py
emilmelnikov: It's just a trick intended to keep `oid` updates in one place of the loop.

k-dominik: anyway. I removed that logic. The oid is in the table so why not use it.

k-dominik:
Hi @emilmelnikov , thanks for nudging me to actually think about what the code is doing. The function I had touch anyway could be simplified.

ilastik/utility/exportFile.py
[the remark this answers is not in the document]
emilmelnikov: Oh, and you also don't need `yield from ()` here: since the function already contains the `yield` keyword, it will be treated as a generator function.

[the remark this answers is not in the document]
k-dominik: didn't know this automagic.

ilastik/utility/exportFile.py
k-dominik: removed check completely

ilastik/utility/exportFile.py
```diff
@@ -243,12 +245,34 @@ def _slice_range(low, high, max_):
     return slice(start, end, None)
 
 
-def create_slicing(axistags, dimensions, margin, feature_table):
+def create_slicing(
+    axistags: AxisTags, dimensions: Tuple[int, int, int, int, int], margin: int, feature_table: np.ndarray
+) -> Iterator[Tuple[List[slice], int]]:
     """
-    Returns an iterator on the slices for each object roi
+    Returns an iterator on the slices for each object (bounding box + margin)
```
emilmelnikov, line 252: May as well use the `Returns:` section from Google Sphinx format.

ilastik/utility/exportFile.py
```diff
@@ -243,12 +245,34 @@ def _slice_range(low, high, max_):
...
         yields also the actual object id
+
+    Args:
+        axistags: image axis tags
+        dimensions: 5D image dimensions in (t, x, y, z, c) -order
+        margin: how much "halo" around each object's bounding box
+        feature_table: structured array with at least the following fields:
```
emilmelnikov, line 259: Maybe move schema description from parameter doc into the function doc? In this case, the description can be properly formatted with definition lists:

```
``feature_table`` should have the following fields:

``labelimage_oid``
    object ids, reset per timeframe
``Bounding Box Minimum_0``
    minimum in x-direction

``feature_table`` may also have the following fields:

``Bounding Box Minimum_2``
    minimum in z-direction
```

emilmelnikov (approved):

ilastik/utility/exportFile.py
k-dominik: google styleguide even foresees a `Yields` keyword instead of `Returns` for generators

PR #2282, opened 2020-08-03
Removes the range parameter from usage of volumina's NormalizableLayer

Tomaz-Vieira:
Adjusts usage of `NormalizableLayer` (and its descendants such as `AlphaModulatedLayer`) everywhere, removing he `range` parameter, which wasn't doing anything useful.

This PR only exists because of https://github.com/ilastik/volumina/pull/265 and therefore should be merged together with that.


Tomaz-Vieira:
> With your PR, it is only working in the range adjusted with window leveling.

I have no idea how you manage to catch those, but I'm glad that you do =)
I changed the widget logic in volumina to get the current values set by the "Window Level" button, but still allows for the full range

Issue #2283, opened 2020-08-07
Load h5 datasets as separate volumes in one go


vzinche:
## Description

Currently `Add New` -> `Add separate Image(s)` supports loading only one h5 dataset at a time, while `Add a 3D/4D volume` allows you to specify a pattern or load all the datasets at the same. Would be great to have the same for `Add separate Image(s)`

## Scenario

I have 20 separate volumes stored as 20 separate datasets in one h5 file. Would be great if to load them all I could just select all the datasets from the file at the same time (or specify a pattern).

## Rationale

I don't want to repeat the  `Add separate Image(s)` procedure 20 times to load 20 volumes.

Issue #2129, opened 2019-09-27
inner dataset path selection leaked outside projects


k-dominik:
## Describe the bug
When a file hierarchical file (e.g. h5) is added, that contains multiple paths, the first time it is added, one is asked to select an inner dataset. When one opens a new project afterwards and adds the same dataset, one has to go to dataset properties to choose a different path. I do see that this could be regarded as a feature if one has a file, where e.g. all dataset present different layers. In reality, however, datasets might be split over channels, time, z, and users might choose to work on them individually. It might then become a nuisance to always have to go to the dataset properties to change it (which is not obvious, too).

## Expected behavior
Default inner paths should not leak between projects. If I open a new project, the inner path should be left blank and should be choosen. If I add another hierarchical dataset with a similar dataset to the same project, I guess, it should be fine to add the same inner path as a default one. but _not_ across project boundaries.

Specifically the code relating to `def _add_default_inner_path(self, roleIndex: int, inner_path: str):` (and all the other default file handling needs to be changed accordingly).

I feel like https://github.com/ilastik/volumina/pull/257 was very much inspired to circumvent this leaking in tests.

PR #2285, opened 2020-08-10
Add multi mode to subvolume selection dialog

emilmelnikov:
Multiple dataset selection can be used to add multiple
datasets when opening an HDF5/N5/etc. files.

Also, clean up some surrounding code and remove
direct accesses to dialog widgets.

Closes https://github.com/ilastik/ilastik/issues/2283.

### DIFF
--- a/ilastik/applets/dataSelection/dataSelectionGui.py
+++ b/ilastik/applets/dataSelection/dataSelectionGui.py
@@ -21,10 +21,10 @@ from __future__ import absolute_import
 # 		   http://ilastik.org/license.html
 ###############################################################################
 # Python
+import contextlib
 import os
-import re
 from pathlib import Path
-from typing import Dict, List, Set
+from typing import Dict, List, Optional, Set
 import threading
 import h5py
 from functools import partial
@@ -35,9 +35,6 @@ logger = logging.getLogger(__name__)
 # PyQt
 from PyQt5 import uic
 from PyQt5.QtWidgets import QDialog, QMessageBox, QStackedWidget, QWidget
-
-# lazyflow
-from lazyflow.request import Request
 
 # volumina
 from volumina.utility import preferences
@@ -51,7 +48,6 @@ from ilastik.applets.base.applet import DatasetConstraintError
 from .opDataSelection import (
     DatasetInfo,
     RelativeFilesystemDatasetInfo,
-    FilesystemDatasetInfo,
     ProjectInternalDatasetInfo,
     UrlDatasetInfo,
 )
@@ -138,12 +134,6 @@ class DataSelectionGui(QWidget):
 
     ###########################################
     ###########################################
-
-    class UserCancelledError(Exception):
-        # This exception type is raised when the user cancels the
-        #  addition of dataset files in the middle of the process somewhere.
-        # It isn't an error -- it's used for control flow.
-        pass
 
     def __init__(
         self,
@@ -401,17 +391,17 @@ class DataSelectionGui(QWidget):
         paths = ImageFileDialog(self).getSelectedPaths()
         self.addFileNames(paths, startingLaneNum, roleIndex)
 
-    def addFileNames(self, paths: List[Path], startingLaneNum: int, roleIndex: int):
-        # If the user didn't cancel
-        if paths:
-            try:
-                new_infos = self._createDatasetInfos(roleIndex, paths)
-                self.addLanes(new_infos, roleIndex=roleIndex, startingLaneNum=startingLaneNum)
-            except DataSelectionGui.UserCancelledError:
-                pass
-            except Exception as ex:
-                log_exception(logger)
-                QMessageBox.critical(self, "Error loading file", str(ex))
+    def addFileNames(self, paths: List[Path], startingLaneNum: int, roleIndex: int) -> None:
+        try:
+            new_infos = self._dataset_infos(roleIndex, paths)
+            if not new_infos:
+                return
+
+            self.addLanes(new_infos, roleIndex=roleIndex, startingLaneNum=startingLaneNum)
+
+        except Exception as ex:
+            log_exception(logger)
+            QMessageBox.critical(self, "Error loading file", str(ex))
 
     def _findFirstEmptyLane(self, roleIndex):
         opTop = self.topLevelOperator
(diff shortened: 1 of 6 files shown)


ilastik/applets/dataSelection/dataSelectionGui.py
```diff
@@ -515,21 +505,60 @@ def _determineLaneRange(self, infos: List[DatasetInfo], startingLaneNum=None):
...
+
+        internal_paths = DatasetInfo.getPossibleInternalPathsFor(file_path)
+        if not internal_paths:
+            raise RuntimeError(f"File {file_path} has no image datasets")
+        if len(internal_paths) == 1:
+            return internal_paths
+
+        auto_internal_paths = list(self._get_previously_used_inner_paths(role_index).intersection(internal_paths))
+        if len(auto_internal_paths) == 1:
+            return auto_internal_paths
```
Tomaz-Vieira, line 558: If a user selects an .h5 file with multiple datasets, then picks one, that dataset will be recorded as an `auto_internal_path`. Say that the user then changes his mind, and removes that file from the data selection. Then, he tries to add it again, but this time he wants to add multiple datasets at once; He will not be able to, because `auto_internal_paths` will have his recorded selection from the first time he added the file, and this function will return before reaching `SubvolumeSelectionDlg`. In fact, he will never be able to use the multi-dataset selection for that file again, until he deletes the `.ilastik_preferences` file.

ilastik/widgets/hdf5SubvolumeSelectionDialog.py
```diff
@@ -18,50 +18,78 @@
...
+        label_lines = ["<p>Your file contains multiple images.</p>"]
+        if multi:
+            label_lines.append("<p>Select images that you would like to open.</p>")
+        else:
+            label_lines.append("<p>Select image that you would like to open.</p>")
+        layout.addWidget(QLabel("".join(label_lines), parent=self))
+
+        if multi:
+            self.items = QListWidget(parent=self)
+            self.items.setSelectionMode(QListWidget.MultiSelection)
```
Tomaz-Vieira, line 63: Is there any way we can get `Shift + Click` working on this? Otherwise people have to click every single one of the datasets they're trying to add =)

emilmelnikov: All the "normal" keyboard shortcuts should work, including Ctrl-A. Do you think we should mention them in a help text?

ilastik/applets/dataSelection/dataSelectionGui.py
emilmelnikov: Good catch! Could you explain why do we need to remember previously used paths anyway?

k-dominik: hmm this was a "convenience function". Let's say a user has multiple h5 files with the same internal layout (always `/raw` and `/somethingelse`). We try to be helpful and if the user has selected the `raw` one for the first image, automatically load `raw` for the next file that is added (if `raw` is present).

emilmelnikov: Maybe it would be better to add an "Apply to All" checkbox? We could make it checked by default to preserve the original behavior.

Also, how exactly do these selections persist between different actions? Do we save them in a project file?

ilastik/widgets/hdf5SubvolumeSelectionDialog.py
k-dominik: Now that you mention that, maybe it's the easiest (and most helpful) to add this information in a help-text, as you suggest. I would probably always try the standard things anyway, but I'm sure users will appreciate this info!

ilastik/widgets/hdf5SubvolumeSelectionDialog.py
```diff
@@ -18,50 +18,78 @@
...
+        if self.multi:
+            raise ValueError("selectedPath can be used only when multi=False")
+        if self.exec_() != self.Accepted:
+            return None
+        return self.combo.currentText()
+
+    def select_paths(self) -> Optional[List[str]]:
+        """Select and return item texts, or None if the dialog has not been accepted."""
+        if not self.multi:
+            raise ValueError("selectedPaths can be used only when multi=True")
```
k-dominik, line 88: not sure something that doesn't take a parameter should raise `ValueError`. From [the docs](https://docs.python.org/3/library/exceptions.html#ValueError):

> Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as IndexError.

ilastik/applets/dataSelection/dataSelectionGui.py
```diff
@@ -515,21 +505,60 @@ def _determineLaneRange(self, infos: List[DatasetInfo], startingLaneNum=None):
...
-        rois may be None, in which case it is ignored.
-        """
-        if rois is None:
-            rois = [None] * len(filePaths)
-        assert len(rois) == len(filePaths)
+    def _join_dataset_path(self, file_path: Path, internal_path: str) -> str:
+        """Return a proper path to the dataset in a file."""
+        p = file_path.absolute()
+        with contextlib.suppress(ValueError):
+            p = p.relative_to(self.topLevelOperator.WorkingDirectory.value)
```
k-dominik, line 512: probably unrelated comment: I will always stumble over this `WorkingDirectory` thing being the project file dir...

ilastik/applets/dataSelection/dataSelectionGui.py
k-dominik: there are good arguments for either the possibilities you mention. There has been an [issue about it](https://github.com/ilastik/ilastik/issues/2129) where the fix results in the current behavior:

inner paths are only "remembered" for the current session with the current project. No other persistence. I'm not quite sure what you mean by "actions".

emilmelnikov: > I'm not quite sure what you mean by "actions".
I meant a user action such as drag-and-drop, or clicking on the "Add..." button.

OK, so how users usually add files in ilastik? Do they frequently add/delete many different files that have multiple internal datasets, or load everything once in the beginning.

What if a user adds one multi-dataset file first, and then wants to add another multi-dataset file, but with different datasets? For me it would be very confusing if the system just goes and silently adds some previously selected datasets.

k-dominik: > What if a user adds one multi-dataset file first, and then wants to add another multi-dataset file, but with different datasets? For me it would be very confusing if the system just goes and silently adds some previously selected datasets.

usually all files will bear the same internal structure in experiments. So I'd say this is the more common case. At least when this behavior was not in place we got complaints - I haven't heard/read any complaints about this behavior not being there.

However, I like your suggestion best, to have a checkbox like "remember internal path(s) for all following h5 files" or sometihng like that.

k-dominik:
hey @emilmelnikov, did you abandon this because of too much discussion? I think it's on a very good path!

emilmelnikov:
> hey @emilmelnikov, did you abandon this because of too much discussion? I think it's on a very good path!

I picked this up a couple of days ago, but then considered leaving the original dialog as-is, and adding a new "Add multiple files from pattern..." dialog. That dialog could accept a glob-like (or regex-like) path string, and add all datasets from multiple files in one step.

It seems that the Valentina's use case is relatively rare, and at the same time the semantics of storing multiple previously used dataset paths is not obvious, which could lead to subtle bugs and UX problems.

emilmelnikov:
Closed in favour of https://github.com/ilastik/ilastik/pull/2367.

PR #2286, opened 2020-08-13
Add JSONSerialSlot

m-novikov:


ilastik/applets/base/jsonSerializerRegistry.py
```diff
@@ -0,0 +1,97 @@
...
+
+class _DictSerialzierRegistry:
+    _VERSION_TAG = "__serializer_version"
+
+    def __init__(self):
+        self._serializer_by_type = {}
+
+    def register_serializer(self, type_, version=1):
+
+        def _register_serializer(serializer_cls):
```
k-dominik, line 43: I wonder if this could be exploited to get some abc-like behavior, where one would enforce methods `serialize`, `deserialize` and attribute `serializer_version`. I think it should not be the decorator that specifies the version, but the serializer itself. While the serializer class gets the version passed in the dict when deserializing, it is not aware of it it's own version.

ilastik/applets/base/appletSerializer.py
```diff
@@ -1198,3 +1201,37 @@ def updateWorkingDirectory(self, newdir, olddir):
         to be updated. Child Classes should overwrite this method
         if they store relative paths."""
         pass
+
+
+class JSONSerialSlot(SerialSlot):
+    """
+    Implements the logic for serializing a json serializable object slot.
+
+    wraps value with numpy.void to avoid the following error:
```
k-dominik, line 1210: outdated? or did you forget to do it in code?

k-dominik:
I really like this idea and also it's implementation. It's especially useful to have the tests early on to document expectations on capabilities.

emilmelnikov:
I would like to discuss an alternative design that uses Python protocols.

Instead of `serialize`, the class of a value being saved should support the mapping protocol, so we can just `group.attrs[name] = json.dumps(dict(value))` in `_saveValue`.

Instead of `deserialize`, the class of a value being saved should have `fromdict` method that does the job.

This approach has an advantage of keeping object-data transformations close to the actual implementation, and doesn't require new abstractions, while being generic.

If some case need to to pre- or post-processing, we can add `transform_from` and `transform_to` parameters to `JsonSerialSlot` that are functions `Dict[str, Any] -> Dict[str, Any]`.

ilastik/applets/base/jsonSerializerRegistry.py
m-novikov: Regarding version tag, serializers should not be exposed to this tag, it was an error that it got passed to serializer with a dict.

k-dominik:
> I would like to discuss an alternative design that uses Python protocols.

When we discussed this I came with similar ideas, but @m-novikov convinced me that his approach is also very viable and has some very nice advantages.

1) Versioning (albeit not part of this PR anymore) in future this is very easily extensible to multiple  versions of serializers, always staying backwards compatible. Something like
  ```python
  @register_serializer(Foo, version=1)
  class MyV1Serializer: ...
  
  
  @register_serializer(Foo, version=2)
   class MyV2Serializer: ...
  ```
  will very nicely document the different versions. The only thing in place we would need then is developer guidelines that restrict editing of serializers to the ones that have _not_ yet been part a release version of ilastik. Otherwise one would write a new serializer and give it a new version. No clustered `if ... else ...` needed. Very clean :D

2) Straight forward to write serializers for external classes. No need to subclass those in order to add serialization methods

3) Sure, the serialization code is not close to the implementation of the objects anymore - however, with type annotations you get still most of the benefits.


Would you add something @m-novikov ?

So I stay with my original comment - I like this approach a lot.

m-novikov:
> Would you add something @m-novikov ?

Another benefit if one would want to have different serialization schema (e.g. plain object vs nested) they only need to create an additional registry.

emilmelnikov:
> Versioning (albeit not part of this PR anymore) in future this is very easily extensible to multiple versions of serializers, always staying backwards compatible.

It seems weird to have multiple pieces of code that do almost the same. We don't do this in other places when we change other code in a _compatible_ way. Of course if you need to support multiple _incompatible_ interfaces/implementations this is the way to go, but then it probably should be done somewhere upstream because it breaks the entire serialization process.

> Straight forward to write serializers for external classes. No need to subclass those in order to add serialization methods

You don't need to subclass either way:

```python
class MyRandomClassBundle:
    def __init__(self, v: MyRandomClass):
        self.v = v

    def __getitem__(self, k):
        return getattr(self.v, k)

    @classmethod
    def fromdict(self, d):
        v = MyRandomClass(...)
        v.foo = d["bar"]
        return cls(v)
```

The main benefit is to be able to work with this bundle as a usual, transparent data structure.

> Sure, the serialization code is not close to the implementation of the objects anymore - however, with type annotations you get still most of the benefits.

I don't think type annotations are the main concern here, the main problem are invariants (e.g. field1 should be non-negative if field2 is True).

> Another benefit if one would want to have different serialization schema (e.g. plain object vs nested) they only need to create an additional registry.

There are also [standard JSON encoders and decoders](https://docs.python.org/3/library/json.html#encoders-and-decoders) that we could use instead of custom serializers.

---

I think the main difference here is having a separate class with `serialize` / `deserialize` methods and registry vs. passing (optional) `serialize` / `deserialize` functions into `JSONSerialSlot`.

I  guess my main objection is having a lot of classes that don't hold any values and, in most cases, could be replaced with generic data transformations. That is, most objects that we want to serialize hold data somewhere, and this data could (and probably should!) be serialized independent of the object type.

PR #2295, opened 2020-08-27
fixed error when adding more than 15 objects in object count classifier

### LINKED ISSUE #2294: object count classifer error when adding more than 14 objects
## Describe the bug
per [user report](https://forum.image.sc/t/ilastik-object-count-classification-bug/41996?u=k-dominik) we were made aware that an error occurs in tracking when adding the label for "15 objects"

## Expected behavior
Should not result in an error of course

## To Reproduce
Within the tracking workflow
1. Add some 2d/3d/4d+t data
2. In the object count classifier start adding objects until you are at "15 objects"

## Error message/traceback

If applicable, please add tracebacks with formatting, such as

```pytb
ERROR 2020-08-25 14:39:24,313 log_exception 11247 140639160215360 Logged the above exception just in case PyQt loses it.
ERROR 2020-08-25 14:39:24,313 excepthooks 11247 140639160215360 Unhandled exception in thread: 'MainThread'
ERROR 2020-08-25 14:39:24,314 excepthooks 11247 140639160215360 Traceback (most recent call last):
  File "/.../ilastik-versions/ilastik-1.3.3post3-Linux/ilastik-meta/ilastik/ilastik/utility/bind.py", line 67, in __call__
    self.f(*(self.bound_args + args[0 : self.numUnboundArgs]))
  File "/.../ilastik-versions/ilastik-1.3.3post3-Linux/ilastik-meta/ilastik/ilastik/applets/labeling/labelingGui.py", line 721, in _addNewLabel
    self._labelingSlots.labelNames.setValue(operator_names, check_changed=False)
  File "/.../ilastik-versions/ilastik-1.3.3post3-Linux/ilastik-meta/lazyflow/lazyflow/slot.py", line 73, in call_in_setup_context
    return func(self, *args, **kwargs)
  File "/.../ilastik-versions/ilastik-1.3.3post3-Linux/ilastik-meta/lazyflow/lazyflow/graph.py", line 171, in __exit__
    sig_setup_complete()
  File "/.../ilastik-versions/ilastik-1.3.3post3-Linux/ilastik-meta/lazyflow/lazyflow/utility/orderedSignal.py", line 80, in __call__
    f(*args, **kw)
  File "/.../ilastik-versions/ilastik-1.3.3post3-Linux/ilastik-meta/ilastik/ilastik/utility/gui/threadRouter.py", line 82, in routed
    val = func(*args, **kwargs)
  File "/.../ilastik-versions/ilastik-1.3.3post3-Linux/ilastik-meta/ilastik/ilastik/applets/layerViewer/layerViewerGui.py", line 457, in updateAllLayers
    newGuiLayers = self.setupLayers()
  File "/.../ilastik-versions/ilastik-1.3.3post3-Linux/ilastik-meta/ilastik/ilastik/applets/objectClassification/objectClassificationGui.py", line 635, in setupLayers
    self._setPredictionColorTableForRow(predictLayer, row)
  File "/.../ilastik-versions/ilastik-1.3.3post3-Linux/ilastik-meta/ilastik/ilastik/applets/objectClassification/objectClassificationGui.py", line 726, in _setPredictionColorTableForRow
    oldcolor = self._colorTable16_forpmaps[row + 1]
IndexError: list index out of range
```

 - ilastik version: 1.3.3post3

k-dominik:
I guess someone buried the knowledge that it's not very beneficial for solving the tracking problem to have a lot of merge objects somehow implicitly in the assumption that no one will add more than 14 objects. If we want to discourage it, we should make it explicit somewhere and not rely on "people just not doing it".
This PR allows users to add as many merged objects as they want. We should deal with the algorithmic problems this imposes somewhere else.

* made sure new colors are added to colortable when initial size is exceeded.
* made sure naming of new labels follows the scheme `x Objects`.

fixes #2294

### DIFF
--- a/ilastik/workflows/tracking/common.py
+++ b/ilastik/workflows/tracking/common.py
@@ -1,0 +1,48 @@
+###############################################################################
+#   ilastik: interactive learning and segmentation toolkit
+#
+#       Copyright (C) 2011-2020, the ilastik developers
+#                                <team@ilastik.org>
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# In addition, as a special exception, the copyright holders of
+# ilastik give you permission to combine ilastik with applets,
+# workflows and plugins which are not covered under the GNU
+# General Public License.
+#
+# See the LICENSE file for details. License information is also available
+# on the ilastik web site at:
+#          http://ilastik.org/license.html
+###############################################################################
+from dataclasses import dataclass
+from typing import SupportsInt
+
+
+@dataclass(frozen=True)
+class PluralInflector:
+    zero: str
+    one: str
+    many: str
+
+    # arbitrarily defined maximum in case list() is called on this object
+    max_index: int = 256
+
+    def __getitem__(self, i: SupportsInt) -> str:
+        i = int(i)
+        if i < 0:
+            raise ValueError("cannot inflect negative numbers")
+        if i > self.max_index:
+            raise IndexError(f"values above {self.max_index} not supported (got {i}).")
+        if i == 0:
+            return self.zero
+        return f"{i} {self.one if i == 1 else self.many}"
+
+    def __len__(self) -> int:
+        return self.max_index
+
+
+DIVISION_CLASSIFIER_LABEL_NAMES = PluralInflector(zero="False Detection", one="Object", many="Objects")
--- a/ilastik/applets/labeling/labelingGui.py
+++ b/ilastik/applets/labeling/labelingGui.py
@@ -24,7 +24,6 @@ from builtins import filter
 import os
 import re
 import logging
-import itertools
 from functools import partial
 
 # Third-party
--- a/ilastik/applets/objectClassification/objectClassificationGui.py
+++ b/ilastik/applets/objectClassification/objectClassificationGui.py
@@ -160,8 +160,7 @@ class ObjectClassificationGui(LabelingGui):
 
         self.labelingDrawerUi.brushSizeCaption.setVisible(False)
 
-        self._colorTable16_forpmaps = list(colortables.default16_new)
-        self._colorTable16_forpmaps[15] = QColor(Qt.black).rgba()  # for objects with NaNs in features
+        self._colorTable_forpmaps = list(colortables.default16_new)
 
         self.labelingDrawerUi.subsetFeaturesButton.clicked.connect(self.handleSubsetFeaturesClicked)
         self.labelingDrawerUi.labelAssistButton.clicked.connect(self.handleLabelAssistClicked)
@@ -453,14 +452,10 @@ class ObjectClassificationGui(LabelingGui):
         return self.topLevelOperatorView.SuggestedLabelNames([]).wait()[row_idx]
 
     def getNextLabelName(self):
-        if self._labelControlUi.labelListModel.rowCount() >= len(
-            self.topLevelOperatorView.SuggestedLabelNames([]).wait()
-        ):
-            return self._getNext(
-                self.topLevelOperatorView.LabelNames, super(ObjectClassificationGui, self).getNextLabelName
-            )
-        else:
+        try:
             return self._getNext(self.topLevelOperatorView.LabelNames, self._getNextSuggestedLabelName)
+        except IndexError:
+            return self._getNext(self.topLevelOperatorView.LabelNames, super().getNextLabelName)
 
     def getNextLabelColor(self):
         return self._getNext(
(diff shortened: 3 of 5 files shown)


ilastik/workflows/tracking/common.py
```diff
@@ -0,0 +1,16 @@
...
+        assert idx >= 0
+        if idx == 0:
+            return "False Detection"
+        elif idx == 1:
+            return "1 Object"
+        else:
+            return f"{idx:d} Objects"
+
+    def __len__(self) -> int:
+        return sys.maxsize
```
emilmelnikov, line 16: This could result in `MemoryError` if someone does `list(DivisionClassifierLabelNames())`, directly or indirectly. Is it possible to remove `__len__` and modify code that does this?

ilastik/workflows/tracking/common.py
```diff
@@ -0,0 +1,16 @@
+import sys
+
+
+class DivisionClassifierLabelNames:
```
emilmelnikov, line 4: Maybe do something more generic?

```python
@dataclass(frozen=True)
class PluralInflector:
    zero: str
    one: str
    many: str

    def __getitem__(self, i: SupportsInt) -> str:
        i = int(i)
        if i < 0:
            raise ValueError("cannot inflect negative numbers")
        if i == 0:
            return self.zero
        return f"{i} {self.one if i == 1 else self.many}"

    __call__ = __getitem__


DIVISION_CLASSIFIER_LABEL_NAMES = PluralInflector(zero="False Detection", one="Object", many="Objects")
```

emilmelnikov (changes requested):
I am worried about `sys.maxsize`.

ilastik/workflows/tracking/common.py
k-dominik: True, I also saw that problem and thought, how likely is that to happen :D. I guess the calling code could `try: ... except KeyError` instead of checking length.

k-dominik: side note: so I haven't found the code for "list(...)" yet, but it seems that it only checks if __len__ gives something nonzero and then iterates over the object until it catches `KeyError`

emilmelnikov: In this case it makes sense to choose label names at the calling site. We can add slice support to `__getitem__` to fully mimic `list`, and just use something like `LABEL_NAMES[:15]`.

k-dominik: btw, even with `__len__` removed `list(..)` will dutifully run into eternity with an object like

```python

class Blah:
     def __getitem__(self, idx):
         return idx

blah = Blah()
list(blah)
```

emilmelnikov: Interesting. It seems that we forgot that the [iterator protocol](https://docs.python.org/3/library/functions.html#iter) also works on objects with just `__getitem__`, not only on objects with `__iter__`.

What if we still add `__len__`, but `raise NotImplementedError` inside? This should prevent functions that try to create a finite sequence not to do that. OTOH, one may argue that call sites shouldn't try to create such sequences in the first place.

ilastik/workflows/tracking/common.py
k-dominik: I like the solution, but I don't really see it used anywhere else, then the explicit implementation is easier to grasp imo.

ilastik/workflows/tracking/common.py
k-dominik: Okay I had the feeling that things are over-complicated here. I pushed a small update to  that class with a max-size of 1024 (which is unreasonably large for the tracking problem).

ilastik/workflows/tracking/common.py
k-dominik: I ended up mixing and matching a bit...

ilastik/workflows/tracking/common.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+@dataclass(frozen=True)
+class PluralInflector:
+    zero: str
+    one: str
+    many: str
+
+    # arbitrarily defined maximum in case list() is called on this object
+    max_index: int = 256
```
emilmelnikov, line 32: If we decided to be general, might as well generalize this too:

```python
def inflect(n, *, zero, one, many):
    if n < 0:
        raise ValueError(f"cannot inflect negative number {n}")
    if n == 0:
        return zero
    if n == 1:
        return self.one
    return self.many.format(n)


class LazySeq(collection.abc.Sequence):
    def __init__(self, func, *, size):
        self.func = func
        self.size = size

    def __getitem__(self, i):
        if not (0 <= i < self.size):
            raise IndexError(i)
        return self.func(i)

    def __len__(self):
        return self.size


DIVISION_CLASSIFIER_LABEL_NAMES = LazySeq(
    functools.partial(inflect, zero="False Detection", one="1 Object", many="{} Objects"), size=256
)

emilmelnikov (approved):

ilastik/workflows/tracking/common.py
k-dominik: okay, thanks for the thorough review on that. I'll go with the current solution.

PR #2297, opened 2020-08-27
Defaults MPI tests to use num physical cores

Tomaz-Vieira:
Depending on what flavor of MPI one is using, and on what configuration is being used, the tests involving MPI (`--headless --distributed`) might fail with the MPI implementation being unable to allocate the previously hardcoded 4 workers.

This PR defaults the number of workers to the number of physical CPUs, which is far more likely to succede .


k-dominik (approved):
LGTM :)

emilmelnikov:
This has been already merged, but there is a built-in method for getting a number of CPUs that one is allowed to use: https://docs.python.org/3/library/os.html#os.sched_getaffinity

PR #2303, opened 2020-09-17
Headless prediction python API

m-novikov:
Test script
```python
from ilastik.experimental import api
from ilastik.experimental.api import features

import imageio
import matplotlib.pyplot as plt


IMAGE_PATH = "./TrainRaw.png"
LABELS_PATH = "./TrainLabels.png"
TEST_IMAGE_PATH = "./TestRaw.png"

train_arr = imageio.imread(IMAGE_PATH)
labels_arr = imageio.imread(LABELS_PATH)
test_arr = imageio.imread(TEST_IMAGE_PATH)
print(train_arr.shape)

### PIXEL CLASS API ###
classifier = (
    api.ClassifierBuilder()
    .add_dataset(train_arr, labels_arr)
    .use_features(api.DEFAULT_FEATURES)
    # .use_features([features.Gaussian(5), features.StructureTensorEigenvalues(5)])
    .train()
)
test_res = classifier.predict(test_arr)
train_res = classifier.predict(train_arr)  # Evaluate on original training data
### PIXEL CLASS API END ###

fig = plt.figure(figsize=(2, 2))

fig.add_subplot(2, 2, 1)
plt.imshow(train_arr, cmap="gray")

fig.add_subplot(2, 2, 2)
plt.imshow(test_arr, cmap="gray")

fig.add_subplot(2, 2, 3)
plt.imshow(test_res)

fig.add_subplot(2, 2, 4)
plt.imshow(train_res)
plt.show()
```
Using following images:
* [TestRaw](https://user-images.githubusercontent.com/5163640/93466708-faaf8a80-f8ec-11ea-8416-e9e1b8555249.png)
* [TrainLabels](https://user-images.githubusercontent.com/5163640/93466719-fb482100-f8ec-11ea-9807-0f01c15d6b7a.png)
* [TrainRaw](https://user-images.githubusercontent.com/5163640/93466724-fbe0b780-f8ec-11ea-9e33-11eafb1082f3.png)


m-novikov:
Example using ilastik project file
```python
from ilastik.experimental import api
from ilastik.experimental.api import features

import imageio
import matplotlib.pyplot as plt


IMAGE_PATH = "~/TrainRaw.png"
LABELS_PATH = "~/TrainLabels.png"
TEST_IMAGE_PATH = "~/TestRaw.png"

train_arr = imageio.imread(IMAGE_PATH)
labels_arr = imageio.imread(LABELS_PATH)
test_arr = imageio.imread(TEST_IMAGE_PATH)

### PIXEL CLASS API ###
classifier = api.ClassifierBuilder.from_project_file("./MyProjectPixel.ilp")
test_res = classifier.predict(test_arr)
train_res = classifier.predict(train_arr)  # Evaluate on original training data
### PIXEL CLASS API END ###

fig = plt.figure(figsize=(2, 2))

fig.add_subplot(2, 2, 1)
plt.imshow(train_arr, cmap="gray")

fig.add_subplot(2, 2, 2)
plt.imshow(test_arr, cmap="gray")

fig.add_subplot(2, 2, 3)
plt.imshow(test_res)

fig.add_subplot(2, 2, 4)
plt.imshow(train_res)
plt.show()

```
* [MyProjectPixel.ilp.zip](https://github.com/ilastik/ilastik/files/5239431/MyProjectPixel.ilp.zip)

k-dominik:
Always nice to see how simple pixel classification can look like :)

A few thoughts:

* what/who would be the consumer of this API? would we want to use it from our gui side (interactive setting) or is it meant for programmatic use in some kind of batch (non-interactive) setting?
* this API assumes that blockwise processing will be handled "outside" somehow. Also in a blockwise setting it would instantiate the whole graph for every block, which might be slow.
* intermediate results cannot be accessed (e.g. just some feature output)
* while i like the possibility to spawn a classifier directly from a project file, the api for constructing one from scratch with the function chaining is at least a pattern I have seen seldom in python

m-novikov:
>     * what/who would be the consumer of this API? would we want to use it from our gui side (interactive setting) or is it meant for programmatic use in some kind of batch (non-interactive) setting?

This API is meant for an end user in non-interactive setting. Running batch jobs, @constantinpape provided [an example](https://github.com/constantinpape/cluster_tools/blob/master/cluster_tools/ilastik/prediction.py#L132-L147) which would be nice to replace with python code instead of subprocess call

>     * this API assumes that blockwise processing will be handled "outside" somehow. Also in a blockwise setting, it would instantiate the whole graph for every block, which might be slow.

Yes, I plan to move graph creation outside of the predict function, it just a temporary solution. For blockwise processing I spoke with Constantin and hopefully for now it's sufficient to have block handling outside. 
It's also possible to to come up with some nice API for this use case also e.g. `classifier.stream().rois([..list_of_rois..]).predict().write_to("outfile.n5")`

>     * while i like the possibility to spawn a classifier directly from a project file, the api for constructing one from scratch with the function chaining is at least a pattern I have seen seldom in python

Builder patterns seems like a nice solution here because it allows us to delay graph creation and validate user input to avoid inconsistent graph state.

constantinpape:
I think this will simplify using ilastik from other python applications, e.g. my example code for block-wise processing above. This would make the code much easier and also more efficient (overhead of loading the ilastik project every time and resaving the output).

> For blockwise processing I spoke with Constantin and hopefully for now it's sufficient to have block handling outside.

I think users would often want to apply other operations, thresholding, connected components etc. before saving the result, so I think blocking will often be handled externally.
It still wouldn't hurt to have a nice API for "simple" blockwise processing like @m-novikov proposed.

emilmelnikov:
> Builder patterns seems like a nice solution here because it allows us to delay graph creation and validate user input to avoid inconsistent graph state.

Builders are used usually used in languages without keyword arguments and/or dict literals because it's error-prone to pass multiple positional parameters to a constructor/factory. Since Python has both, one can replace `SpamBuilder().attr1(val1).attr2(val2).build()` with `Spam(attr1=val1, attr2=val2)`.

Overall, I think it would be great if we mimic [scikit-learn estimator API](https://scikit-learn.org/stable/developers/develop.html). In this case, ilastik can participate in scikit-learn pipelines for preprocessing, grid search etc.

m-novikov:
> > Builder patterns seems like a nice solution here because it allows us to delay graph creation and validate user input to avoid inconsistent graph state.
> 
> Builders are usually used in languages without keyword arguments and/or dict literals because it's error-prone to pass multiple positional parameters to a constructor/factory. Since Python has both, one can replace `SpamBuilder().attr1(val1).attr2(val2).build()` with `Spam(attr1=val1, attr2=val2)`.
> 
> Overall, I think it would be great if we mimic [scikit-learn estimator API](https://scikit-learn.org/stable/developers/develop.html). In this case, ilastik can participate in scikit-learn pipelines for preprocessing, grid search etc.

I was thinking in terms of followin use case. I want to evaluate which features are important for my model so I compute some search grid and then train multiple classifiers and evaluate.

```python
builder = ClassifierBuilder().add_dataset(data, labels)
def _compute_score(test): ...
scores = []
for feature_set in possible_feature_sets:
     classifier = builder.use_features(feature_set).train()
     score = _compute_score(classifier.predict(data), labels)
     scores.append((score, feature_set))

scores.sort()
```

Tomaz-Vieira:
> Builders are usually used in languages without keyword arguments

I absolutely agree with this. Keyword arguments is how it's being done in webilastik too; it removes the need for a separate `*Builder` class, all arguments are immediately visible in the method signature (as opposed to split into multiple methods of the builder pattern) and the parameters can be statically checked. Also, if you have a `*Builder` object anywhere in the code, it is impossible to tell if it has some or any of its parameters already set by other parts of the code.

> Overall, I think it would be great if we mimic [scikit-learn estimator API](https://scikit-learn.org/stable/developers/develop.html). In this case, ilastik can participate in scikit-learn pipelines for preprocessing, grid search etc.

I would disagree with this, though. The scikit-learn estimator spawns untrained and has a `.fit` method which has to be called in order for this object to be useful. I don't think we should _ever_ allow for a PixelClassifier to spawn untrained, for the same reasons presented above; one cannot, at a glance, tell if it's ready to predict or not. If we're trying to make an architecture with less inconsistencies, then allowing broken objects to exist (i.e.: objects in which calling their methods produce exceptions of bogus results) is probably not a good idea.

m-novikov:
> I absolutely agree with this. Keyword arguments is how it's being done in webilastik too; it removes the need for a separate `*Builder` class, all arguments are immediately visible in the method signature (as opposed to split into multiple methods of the builder pattern) and the parameters can be statically checked. Also, if you have a `*Builder` object anywhere in the code, it is impossible to tell if it has some or any of its parameters already set by other parts of the code.

What happens in the case if we need add multiple datasets? Would you expect something like
```
classifier = Classifier(datasets=[(data1, labels1), (data2, labels2)], features=[...FeatureList...])
```
Keyword arguments force certain code structure, which is not always a good thing. User can't always know beforehand what he want's to put in the objects, so in some cases lazy initialization is more convenient and communicates intent better like in mentioned case of grid search.

Tomaz-Vieira:
> I was thinking in terms of followin use case. I want to evaluate which features are important for my model so I compute some search grid and then train multiple classifiers and evaluate.
>...
> TLDR. Allow partial intialization without spawning graph

The problem I see with partial initialization is that it is hard to tell how far anything has been partially initialized, which is the main problem with the classic architecture (i.e. it creates the workflow then does a bunch of partial initializations all around the codebase). I'll keep pushing for initialization to happen inside `__init__` until the day I die =)

Your example of iterating over several feature sets would work just fine if you were simply re-instantiating your classifier at every loop iteration, but with the added bonus of never having a potentially broken builder on the loose:

```python3
scores = []
for feature_set in possible_feature_sets:
     classifier = PixelClassifer(..., features=feature_set)
     score = _compute_score(classifier.predict(data), labels)
     scores.append((score, feature_set))
```

Tomaz-Vieira:
> What happens in the case if we need add multiple datasets? Would you expect something like
> 
> ```
> classifier = Classifier(datasets=[(data1, labels1), (data2, labels2)], features=[...FeatureList...])
> ```

yes =)

> Keyword arguments force certain code structure, which is not always a good thing. User can't always know beforehand what he want's to put in the objects, (...)

By "a user" a suppose you mean "client code", as opposed to human user? The keyworded arguments mean that at some point in the client code all arguments will have to be present, which forces the client code to be written in that way; I assume that is what you mean by "forces certain code structure". I guess I agree, but I think that's a good thing, as it directs the client code to use the API correctly, with no invisible state. It's what keeps us from doing the equivalent of checking `self.FeaturesSlot.ready()`

This kind of thing will still be necessary at the Applet level, though, as the human user might be doing exactly that - trying stuff again and again with the same params, but modifying the features at every iteration. I just don't think it should seep down to the low level programatic API, since it has caused us problems before =)

emilmelnikov:
If a user wants to dynamically build an arg list, `*args/**kwargs` are always available.

emilmelnikov:
> I would disagree with this, though. The scikit-learn estimator spawns untrained and has a `.fit` method which has to be called in order for this object to be useful. I don't think we should _ever_ allow for a PixelClassifier to spawn untrained, for the same reasons presented above; one cannot, at a glance, tell if it's ready to predict or not. If we're trying to make an architecture with less inconsistencies, then allowing broken objects to exist (i.e.: objects in which calling their methods produce exceptions of bogus results) is probably not a good idea.

There will always be methods that throw exceptions. I think compatibility with scikit-learn is worth it, because of super familiar API and interop with other libraries.

emilmelnikov:
There is a paper about scikit-learn API design, I think is't an interesting and relevant read: https://arxiv.org/abs/1309.0238

m-novikov:
> Overall, I think it would be great if we mimic scikit-learn estimator API. In this case, ilastik can participate in scikit-learn pipelines for preprocessing, grid search etc.

@emilmelnikov so you would think something like?
```
IlastikEstimator(feature_list).fit(data, labels).predict(test_data)
```

emilmelnikov:
> @emilmelnikov so you would think something like?
> 
> ```
> IlastikEstimator(feature_list).fit(data, labels).predict(test_data)
> ```

Yes, something like that. Maybe`IlastikEstimator` can accept an optional `graph` kwarg if needed so that we avoid recreating the same thing repeatedly.

m-novikov:
> Yes, something like that. MaybeIlastikEstimator can accept an optional graph kwarg if needed so that we avoid recreating the same thing repeatedly.

Point of this API to hide implementation detail and be able to use it even if we migrate to dask.
There is no point in trying to optimize this stuff if we don't have any metric.

emilmelnikov:
> Point of this API to hide implementation detail and be able to use it even if we migrate to dask.
> There is no point in trying to optimize this stuff if we don't have any metric.

Yes, sure, I just wanted to add that if we'll need to sneak something in 1 or 2 places, we can do that.

> Aside from it I doubt that batch use case would effectively use caches.

Right, I forgot that we're talking about batch processing here.

Tomaz-Vieira:
> There will always be methods that throw exceptions.

This is a completely avoidable exception, though. So much so that it could be caught statically.

> I think compatibility with scikit-learn is worth it, because of super familiar API and interop with other libraries.

If people want to use a general random forest, they can go directly to vigra or scikit-learn. Ilastik is operating one level higher than that. A `PixelClassififer` can't predict on arbitrary data; it expects pixels, in a particular shape, with a particular number of channels.

> There is a paper about scikit-learn API design

I'm not convinced by their motivations:

_"(...) having two coupled instances (i.e., an estimator object, used as a factory, and a model object, produced by the estimator) indeed decreases the ease of use and is also more likely to unnecessarily confuse newcomers. (...) decoupling estimators from models also creates parallel class hierarchies and increases the overall maintenance complexity of the project."_

Not if the "estimator object" is the `PixelClassifier` class itself, `.fit()` is a static method and an _instance_ of `PixelClassifier` is the "model"; then there is only one single class. I would find it confusing, though, to have a `predict()` method that I can't call. This is the same issue we have with classic ilastik: first we create a broken workflow, then we fix it, and only then can we use it.

m-novikov:
Removed training for now as it obviosly reqiures more discussion. But extracted part of ilp parsing into sub package.

k-dominik:
I wonder If one could add a regression/integration test case here that shows how the API is intended to be used.

m-novikov:
As agreed API surface is 2 functions:
```python
from ilastik.experimental.api import from_project_file
pipeline = from_project_file("./MyProjectPixel.ilp")
test_res = pipeline.predict(test_arr)
```

ilastik/experimental/features.py
```diff
@@ -0,0 +1,62 @@
+class _Feature:
```
k-dominik, line 1: this does not seem to handle 2D/3D features (3D data does not necessarily imply 3D features)
This means that the corresponding tests need some brushing up, too.

ilastik/experimental/api/_base.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def _guess_channel_axis_idx(shape):
+    for idx, val in enumerate(shape):
+        if val < 4:
+            return idx
+    return None
+
+
+def _guess_axistags(shape):
```
k-dominik, line 71: there is `get_default_axisordering` in `lazyflow.helpers`. I'd feel more comfortable not to introduce new code for axis-guessing, especially as this PR already implements a separate code path for pixel classification

ilastik/experimental/api/_base.py
```diff
@@ -0,0 +1,96 @@
...
+
+            self._predict_op = OpClassifierPredict(graph=graph)
+            self._predict_op.Classifier.setValue(classifer.instance)
+            self._predict_op.Classifier.meta.classifier_factory = classifer.factory
+            self._predict_op.Image.connect(self._feature_sel_op.OutputImage)
+            self._predict_op.LabelsCount.setValue(classifer.label_count)
+
+        def predict(self, data):
+            data = _make_vigra_with_cannel_axis(data)
+            num_channels_in_data = data.shape[data.axistags.index("c")]
```
k-dominik, line 44:
```suggestion
            num_channels_in_data = data.channels
```

ilastik/experimental/api/_base.py
m-novikov: I added a docstring to this function, the problem with `get_default_axisordering` that it doesn't try to guess the channel axis position. So if I had shape `(1, 128, 128)` then `get_default_axisordering` would guess `zyx` while this function would guess `cyx`

k-dominik: okay, how do you know it's `cyx` in that case and not in fact a single `z`-slice. Reading this, and I am very sorry to sort of open a discussion on this, it looks like we should _not_ allow such ambiguity and force users to always supply axistags if they want to predict with bare numpy arrays. I think this will solve a lot of issues in advance and make usage way more clear.

so something like
```python
test_res = classifier.predict(test_arr, axisorder="cyx")
```
I don't think users will hate us for this. After all we do not handle arbitrary matrices, but we have images, that always come with dimensional meanings.

m-novikov: Good point, I will use default axis tag guessing and if the array provided has the capability to store tags (e.g. vigra) then use metadata provided. I use singledispatch so it would be easy to add functionality for xarray, ndstructs

k-dominik:
Okay - So to conclude the axis guessing discussion I see that, if data is supplied as

1) `numpy.ndarray` users are somewhat expected to know our conventions for guessing - no correction possible.
2) `vigra.array` axistags can be supplied.

with the prospect of future support of other classes that support tagged axes.
I'm absolutely down with 2). However 1) introduces something that we night break in the future (e.g. by removing it), because it introduces too many issues. I see that it adds a lot of convenience! But wouldn't it be better right now to require users to always construct a `taggedView` with the appropriate axistags (`vigra` is a dependency anyway).

Tomaz-Vieira:
I'd say allowing untagged arrays is asking for trouble; axes have very well defined meanings within ilastik, and we already have trouble with axis order as it stands.

Also, I know that internally ilastik can handle vigra arrays, and I think it's good that they not exposed by the API, but  it would probably be a good idea to move away from them. I'm not sure how well they're being maintained, and vigra is often incompatible with numpy; And even some vigra functions (like the filters) expect numpy arrays instead of vigra arrays (so it's kinda internally incompatible as well =/ )

k-dominik (approved):
LGTM :rocket:

I feel piece of mind now, having only tagged arrays as inputs :)

constantinpape:
I am looking forward to test this :).

PR #2307, opened 2020-09-23
Fix tests in master (grpc import & cleaned up layer writes)

m-novikov:


k-dominik (approved):
LGTM, nice catch of the update layers problem :+1:

PR #2310, opened 2020-09-24
Add menu action for opening a config folder

emilmelnikov:
See also https://github.com/ilastik/volumina/pull/267


ilastik/shell/gui/ilastikShell.py
```diff
@@ -937,7 +937,26 @@ def _createSettingsMenu(self):
...
+                msg = f"<p>Program <tt>{e.name}</tt> is not installed</p>"
+            except Exception as e:
+                msg = f"<p>Failed to open the default file manager:<br><tt>{e}</tt></p>"
+
+            if msg:
+                msg += f"<p>Open the following config folder manually:<br><tt>{path}</tt></p>"
+                QMessageBox.critical(self, "Config Folder Error", msg)
+
         menu.addAction("&Keyboard Shortcuts").triggered.connect(editShortcuts)
+        menu.addAction("Open &Config Folder...").triggered.connect(openConfigFolder)
```
k-dominik, line 959: to be honest, I don't think that it is that important to have it in the gui.
However, now that you have added the niceness to open file locations in native file browsers, we can finally easily add the "navigate to log file" button, which will be very useful :)

ilastik/app.py
```diff
@@ -282,6 +284,12 @@ def logged_start(self):
         threading.Thread.start = logged_start
 
 
+def _init_preferences():
+    from volumina.utility.preferences import migrate
```
k-dominik, line 288: i think it's fine to trigger it here. Could you maybe add a comment so that it's clear that it is some legacy pickle->json migration that is going on here.

k-dominik (approved):
very good :) can't believe I will never see the pickled preferences file in my editor again :dagger:

k-dominik:
I'll rebuild the deps package in the meantime to allow tests to pass...

ilastik/shell/gui/ilastikShell.py
emilmelnikov: What do you think about moving our log files under the corresponding `appdirs` directory too? `ilastik_log.txt` in the home dir is slightly annoying for me, but on the other hand I see that the log file is much more important for users to locate.

ilastik/app.py
emilmelnikov: Maybe add the describing docstring to `migrate` instead?

k-dominik: okay, let's do it this way

ilastik/shell/gui/ilastikShell.py
k-dominik: sure for our users this is not of much importance, as they (hopefully) never have to look at it, I think for devs it would be nice to be able to have this location configurable.

ilastik/utility/filemanager.py
```diff
@@ -0,0 +1,36 @@
+import os
+import shutil
+import subprocess
+import sys
+from typing import Union
+
+
+def open(path: Union[str, bytes, os.PathLike]):
```
m-novikov, line 8: Another possibility is to use `QDesktopServices`
I don't know how reliable it in qt
```python
from PyQt5.QtCore import QUrl
from PyQt5.QtGui import QDesktopServices

QDesktopServices.openUrl(QUrl(f"file://{path}", QUrl.TolerantMode))
```

emilmelnikov: That's awesome, thanks!

ilastik/shell/gui/ilastikShell.py
```diff
@@ -937,7 +937,22 @@ def _createSettingsMenu(self):
         def editShortcuts():
             mgrDlg = ShortcutManagerDlg(self)
 
+        def openConfigFolder():
+            from volumina.utility import preferences
+
+            path = preferences.get_path().parent
+
+            try:
+                QDesktopServices.openUrl(QUrl(path.absolute().as_uri()))
```
m-novikov, line 946: This returns boolean success flag.

m-novikov (approved):

ilastik/app.py
```diff
@@ -140,6 +140,8 @@ def main(parsed_args, workflow_cmdline_args=[], init_logging=True):
 
     _init_threading_logging_monkeypatch()
 
+    _init_preferences()
```
Tomaz-Vieira, line 143: I really like the idea of using default directories for storing our files, but I guess I'd find it confusing if ilastik silently moved my files around; I could come looking for `~/.ilastikrc` and just not find it. Maybe we shouldn't do this automatically, or at least ask the user if he's OK with us moving it elsewhere.

Alternatively, as we were discussing, maybe ilastik could just look in multiple places until it finds a config file. It would then log which config file is being used so that no one is surprised that their configs aren't taking effect when there are 2 or more config files in the system

ilastik/config.py
```diff
@@ -87,6 +89,12 @@ def init_ilastik_config(userConfig=None):
     if userConfig is not None and not os.path.exists(userConfig):
         raise Exception("ilastik config file does not exist: {}".format(userConfig))
 
+    # Move default config to the new location.
+    oldConfigPath = os.path.expanduser("~/.ilastikrc")
+    if userConfig is None and os.path.exists(oldConfigPath) and not os.path.exists(CONFIG_PATH):
+        os.makedirs(os.path.dirname(CONFIG_PATH), exist_ok=True)
+        shutil.move(oldConfigPath, CONFIG_PATH)
+
```
Tomaz-Vieira, line 97: Since we're here anyway, what do you think about moving the call to `init_ilastik_config()` into `app.py`? I find it unnerving that it just happens silently on module `import`

emilmelnikov: Probably a good idea, but I'm afraid this will break other code that depends on module initialisation.

k-dominik: i agree that this is a good idea, so it should get an issue and be tackled in a separate PR

k-dominik (approved):
cool! could you please rebase this on master once your volumina PR is merged?

k-dominik:
merging - failure is unrelated and addressed in #2319

Issue #2311, opened 2020-09-24
Failing test TestPixelClassificationGui


m-novikov:
We have a brittle test in `tests/test_ilastik/test_applets/pixelClassification/testPixelClassificationGui.py::TestPixelClassificationGui::test_1_NewProject`
Can spontaneously fail with following exception:
```python
File "ilastik/ilastik/applets/pixelClassification/opPixelClassification.py", line 330, in _checkConstraints
    thisLaneTaggedShape["c"], validShape["c"]
ilastik.applets.base.applet.DatasetConstraintError: Constraint of 'Pixel Classification' applet was violated: All input images must have the same number of channels.  Your new image has 1 channel(s), but your other images have 3 channel(s).
```
It seems UI doesn't manage to react in time to dataset remove instruction

PR #2312, opened 2020-09-24
Wait for events in pixelClassificationGui tests. Closes: #2311

m-novikov:


k-dominik (approved):
LGTM,

it's good that we have something that we can add if tests with signal/slots are too brittle.

PR #2324, opened 2020-10-12
Fix dialog accept called in foreign thread

m-novikov:


k-dominik (approved):
cool, great that it works now!

PR #2333, opened 2020-10-22
Add popup msg

k-dominik:


emilmelnikov:
> The main idea: Workflows can have a WELCOME_MESSAGE variable that points to a file with message contents and also contains an id for the message

NN Workflow seems to be a special case that doesn't warrant another project-wide abstraction.

k-dominik:
Abandoned in favor of having this information _mainly_ on the web and only having pointers in the __applet drawer__.

PR #2340, opened 2020-11-03
API changes for tifffile upgrade

k-dominik:
With this PR we finally switch from `tifffile 0.4.post2` which we have been building ourselves. `tifffile` even changed there versioning scheme in the meantime :o . I tried to keep the changes minimal as possible (e.g. still using vigra to read the pages), but ended up having to modify `OpExportMultipageTiff` quite a bit.

TODO:
- [x] rebuild deps package on OSX
- [x] move latest deps version to main
- [x] remove `tifffile` from ilastik-forge

### DIFF
--- a/lazyflow/operators/ioOperators/opExportMultipageTiff.py
+++ b/lazyflow/operators/ioOperators/opExportMultipageTiff.py
@@ -79,7 +79,7 @@ class OpExportMultipageTiff(Operator):
         if isinstance(dtype, type):
             dtype = dtype().dtype
 
-        page_buf = RoiRequestBufferIter(self._opReorderAxes.Output, self._batch_size)
+        page_buf = RoiRequestBufferIter(self._opReorderAxes.Output, self._batch_size, iterate_axes="tzc")
         page_buf.progress_signal.subscribe(self.progressSignal)
 
         with tifffile.TiffWriter(self.Filepath.value, byteorder="<", ome=True) as writer:
--- a/lazyflow/operators/ioOperators/opExportMultipageTiff.py
+++ b/lazyflow/operators/ioOperators/opExportMultipageTiff.py
@@ -85,7 +85,7 @@ class OpExportMultipageTiff(Operator):
         with tifffile.TiffWriter(self.Filepath.value, byteorder="<", ome=True) as writer:
             writer.write(
                 data=page_buf,
-                shape=self._opReorderAxes.Output.meta.shape,
+                shape=self._opReorderAxes.Output,
                 dtype=dtype,
                 software="ilastik",
                 metadata={"axes": "".join(self._opReorderAxes.Output.meta.getAxisKeys())},
--- a/lazyflow/operators/ioOperators/opExportMultipageTiff.py
+++ b/lazyflow/operators/ioOperators/opExportMultipageTiff.py
@@ -29,7 +29,7 @@ import tifffile
 
 from lazyflow.graph import InputSlot, Operator
 from lazyflow.operators.opReorderAxes import OpReorderAxes
-from lazyflow.utility import OrderedSignal, RoiRequestBuffer
+from lazyflow.utility import OrderedSignal, RoiRequestBufferIter
 
 
 logger = logging.getLogger(__name__)
@@ -79,13 +79,13 @@ class OpExportMultipageTiff(Operator):
         if isinstance(dtype, type):
             dtype = dtype().dtype
 
-        page_buf = RoiRequestBuffer(self._opReorderAxes.Output, self._batch_size)
+        page_buf = RoiRequestBufferIter(self._opReorderAxes.Output, self._batch_size)
         page_buf.progress_signal.subscribe(self.progressSignal)
 
         with tifffile.TiffWriter(self.Filepath.value, byteorder="<", ome=True) as writer:
             writer.write(
                 data=page_buf,
-                shape=self._opReorderAxes.Output,
+                shape=self._opReorderAxes.Output.meta.shape,
                 dtype=dtype,
                 software="ilastik",
                 metadata={"axes": "".join(self._opReorderAxes.Output.meta.getAxisKeys())},
--- a/lazyflow/operators/ioOperators/opExportMultipageTiff.py
+++ b/lazyflow/operators/ioOperators/opExportMultipageTiff.py
@@ -22,19 +22,15 @@
 
 import logging
 import pathlib
-import textwrap
-import uuid
-import xml.etree.ElementTree as ET
-from typing import Any, Iterable, Sequence
 
 import numpy as np
 import psutil
 import tifffile
-import vigra
 
 from lazyflow.graph import InputSlot, Operator
 from lazyflow.operators.opReorderAxes import OpReorderAxes
-from lazyflow.utility import OrderedSignal, RoiRequestBatch
+from lazyflow.utility import OrderedSignal, RoiRequestBuffer
+
 
 logger = logging.getLogger(__name__)
 
@@ -64,8 +60,6 @@ class OpExportMultipageTiff(Operator):
         self._opReorderAxes.Input.connect(self.Input)
         self._opReorderAxes.AxisOrder.setValue(self._EXPORT_AXES)
 
-        self._page_buf = None
-
     def setupOutputs(self):
         pass
 
@@ -81,54 +75,21 @@ class OpExportMultipageTiff(Operator):
         if path.exists():
             path.unlink()
 
-        self._page_buf = _NdBuf(self._opReorderAxes.Output.meta.shape[:-2])
-
-        batch = RoiRequestBatch(
-            outputSlot=self._opReorderAxes.Output,
-            roiIterator=_page_rois(*self._opReorderAxes.Output.meta.shape),
-            totalVolume=np.prod(self._opReorderAxes.Output.meta.shape),
-            batchSize=self._batch_size,
-        )
-        batch.progressSignal.subscribe(self.progressSignal)
-        batch.resultSignal.subscribe(self._write_buffered_pages)
-        batch.execute()
-
-    def _write_buffered_pages(self, roi, page) -> None:
-        """Store a new page in the buffer and write all in-order buffered pages to the file."""
-        page_axes = self._EXPORT_AXES[-2:]
-        page = vigra.taggedView(page, self._EXPORT_AXES).withAxes(page_axes)
-
-        page_idx = roi[0][:-2]
-        self._page_buf[page_idx] = page
-
-        for i, page in self._page_buf:
-            if not i:
-                self._write_first_page(page)
-            else:
-                vigra.impex.writeImage(page, self.Filepath.value, dtype="", compression="NONE", mode="a")
-
-    def _write_first_page(self, page) -> None:
-        """Write the first page differently: it needs the OME-XML `ImageDescription` field."""
         dtype = self._opReorderAxes.Output.meta.dtype
         if isinstance(dtype, type):
             dtype = dtype().dtype
 
-        desc = _image_desc_xml(
-            dtype=dtype,
-            axes=self._opReorderAxes.Output.meta.getAxisKeys(),
-            shape=self._opReorderAxes.Output.meta.shape,
-            file_uuid=str(uuid.uuid1()),
-            file_name=pathlib.Path(self.Filepath.value).name,
-        )
+        page_buf = RoiRequestBuffer(self._opReorderAxes.Output, self._batch_size)
+        page_buf.progress_signal.subscribe(self.progressSignal)
 
-        if logger.isEnabledFor(logging.DEBUG):
-            import xml.dom.minidom
-
-            pretty_desc = xml.dom.minidom.parseString(desc).toprettyxml()
-            logger.debug(f"Generated OME-TIFF metadata:\n{pretty_desc}")
-
-        with tifffile.TiffWriter(self.Filepath.value, software="ilastik", byteorder="<") as writer:
-            writer.save(page, description=desc, planarconfig="planar")
+        with tifffile.TiffWriter(self.Filepath.value, byteorder="<", ome=True) as writer:
+            writer.write(
+                data=page_buf,
+                shape=self._opReorderAxes.Output.meta.shape,
+                dtype=dtype,
+                software="ilastik",
+                metadata={"axes": "".join(self._opReorderAxes.Output.meta.getAxisKeys())},
+            )
 
     @property
     def _batch_size(self) -> int:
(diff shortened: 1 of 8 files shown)


lazyflow/operators/ioOperators/opExportMultipageTiff.py
```diff
@@ -81,54 +84,32 @@ def run_export(self) -> None:
         if path.exists():
             path.unlink()
 
-        self._page_buf = _NdBuf(self._opReorderAxes.Output.meta.shape[:-2])
+        dtype = self._opReorderAxes.Output.meta.dtype
+        if isinstance(dtype, type):
```
emilmelnikov, line 88: I've seen this occurring in many places. We probably should have an utility func:

```python
def as_dtype(obj):
    """Like :func:`numpy.dtype`, but also works with NumPy scalar types."""
    if isinstance(obj, type):
        if issubclass(obj, numpy.generic):
            return obj().dtype
        raise ValueError(f"{obj} is not a NumPy scalar type")
    return numpy.dtype(obj)
```

But I'm not sure whether it's justified.

lazyflow/operators/ioOperators/opExportMultipageTiff.py
```diff
@@ -81,54 +84,32 @@ def run_export(self) -> None:
...
+
+        output_shape = self._opReorderAxes.Output.meta.shape
+
+        _page_buf = _ReqSyncBuf(self._opReorderAxes.Output.meta.shape, np.prod(output_shape[:-2:]))
 
         batch = RoiRequestBatch(
             outputSlot=self._opReorderAxes.Output,
-            roiIterator=_page_rois(*self._opReorderAxes.Output.meta.shape),
+            roiIterator=_page_rois(*output_shape),
             totalVolume=np.prod(self._opReorderAxes.Output.meta.shape),
```
emilmelnikov, line 98:
```suggestion
            totalVolume=np.prod(output_shape),
```

lazyflow/operators/ioOperators/opExportMultipageTiff.py
```diff
@@ -81,54 +84,32 @@ def run_export(self) -> None:
...
-
-            pretty_desc = xml.dom.minidom.parseString(desc).toprettyxml()
-            logger.debug(f"Generated OME-TIFF metadata:\n{pretty_desc}")
-
-        with tifffile.TiffWriter(self.Filepath.value, software="ilastik", byteorder="<") as writer:
-            writer.save(page, description=desc, planarconfig="planar")
+        with tifffile.TiffWriter(self.Filepath.value, byteorder="<", ome=True) as writer:
+            writer.write(
+                data=_page_buf,
+                shape=self._opReorderAxes.Output.meta.shape,
```
emilmelnikov, line 108:
```suggestion
                shape= output_shape,
```

lazyflow/operators/ioOperators/opTiffReader.py
```diff
@@ -125,11 +121,11 @@ def setupOutputs(self):
                     )
 
             if "q" in axes:
```
emilmelnikov, line 123: Is this really possible?..

lazyflow/operators/ioOperators/opExportMultipageTiff.py
```diff
@@ -149,24 +130,43 @@ def _batch_size(self) -> int:
         return min(batch_size, npages)
 
 
-class _NdBuf:
-    """Store ND-indexed items and give them back in the strict ND-ascending order."""
+class _ReqSyncBuf:
+    """Interate over items in ND-ascending order, that are set in arbitrary order
```
emilmelnikov, line 134:
```suggestion
    """Iterate over items in ND-ascending order, that are set in arbitrary order.
```

lazyflow/operators/ioOperators/opExportMultipageTiff.py
```diff
@@ -149,24 +130,43 @@ def _batch_size(self) -> int:
...
+
 
-    def __init__(self, shape: Iterable[int]):
+    TODO: add tests
+    TODO: include _page_rois, as this class relies on using the same logic for
+          keys
+    """
+
+    def __init__(self, shape, n_rois):
+        self.sig = threading.Event()
```
emilmelnikov, line 143: I don't like this solution. It should be possible to redesign this using a blocking queue.

emilmelnikov (changes requested):
I am not sure about `threading.Event`.

lazyflow/operators/ioOperators/opExportMultipageTiff.py
```diff
@@ -149,24 +130,43 @@ def _batch_size(self) -> int:
...
 
 
-class _NdBuf:
-    """Store ND-indexed items and give them back in the strict ND-ascending order."""
+class _ReqSyncBuf:
+    """Interate over items in ND-ascending order, that are set in arbitrary order
+
 
-    def __init__(self, shape: Iterable[int]):
+    TODO: add tests
```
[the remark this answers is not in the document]
k-dominik: thank you emil, this is a [wip] commit. The todo is just for me to remember what I want to do before finishing this PR

tests/test_lazyflow/test_operators/test_ioOperators/testOpTiffReader.py
```diff
@@ -73,30 +78,15 @@ def test_unknown_axes_tags(self):
...
-                    tifffile.imsave(tiff_path, data)
-                else:
-                    tifffile.imsave(tiff_path, data, metadata={"axes": "QQQ"})
-                op = OpTiffReader(graph=Graph())
-                op.Filepath.setValue(tiff_path)
-                assert op.Output.ready()
-                assert_array_equal(data, op.Output[:].wait())
-                assert op.Output.meta.axistags == vigra.defaultAxistags(test_axes)
+            data = numpy.random.randint(0, 256, test_shape).astype(numpy.uint8)
+            tiff_path = "{}/myfile_{}.tiff".format(d, axisorder)
```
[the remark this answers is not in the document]
k-dominik: out of interest, on what operating system will it not work?

lazyflow/operators/ioOperators/opExportMultipageTiff.py
k-dominik: something like that would be clearly useful

lazyflow/operators/ioOperators/opTiffReader.py
k-dominik: yes, absolutely.

lazyflow/operators/ioOperators/opExportMultipageTiff.py
k-dominik: I have to admit that I felt a bit weary when writing this concept is the main reason I wanted a review at this point. I have canvassed the stdlib before going here. So more concrete suggestions are highly welcome.

tests/test_lazyflow/test_operators/test_ioOperators/testOpTiffReader.py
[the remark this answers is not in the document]
emilmelnikov: This will probably work on Windows, but I think it's better to avoid explicit `/`s unless a library specifically says that it converts paths automatically. I doubt `tifffile` does that.

lazyflow/operators/ioOperators/opTiffReader.py
emilmelnikov: What does it mean?.. I thought we only deal with "tzyxc".

lazyflow/operators/ioOperators/opExportMultipageTiff.py
k-dominik: Also this will most definitely block if there is an uncaught exception in some worker thread. :/

emilmelnikov: Maybe something like this:

```python
import queue
from typing import Any, Iterable, MutableMapping

import numpy


class _NdBuf:
    """Store ND-indexed items and give them back in the strict ND-ascending order."""

    def __init__(self, shape: Iterable[int]):
        self._items = {}
        self._shape = tuple(shape)

    def __setitem__(self, key: Iterable[int], value: Any):
        i = numpy.ravel_multi_index(key, self._shape)
        self._items[i] = value

    def __iter__(self):
        return _NdBufIter(self._items)


class _NdBufIter:
    def __init__(self, items: MutableMapping[int, Any]):
        self._items = items
        self._i = 0

    def __iter__(self):
        return self

    def __next__(self):
        try:
            item = self._items.pop(self._i)
        except KeyError:
            raise StopIteration from None
        self._i += 1
        return item


def as_completed(buf: _NdBuf, q: queue.Queue):
    it = iter(buf)

    while True:
        kv = q.get()
        if kv is None:
            q.task_done()
            return

        buf[kv[0]] = kv[1]
        for item in it:
            q.task_done()
            yield item


def example():
    import threading
    import time
    
    buf = _NdBuf((2, 3))
    q = queue.Queue()
    
    def worker():
        # This is some order in which our blocks will be arriving.
        # Node that the first 3 items are out of order.
        for k in (0, 1), (0, 2), (0, 0), (1, 0), (1, 1), (1, 2):
            # Just some int value to send; we will be using an array here.
            v = 42 + numpy.ravel_multi_index(k, (2, 3))
            # Simulate work.
            time.sleep(0.5)
            print(f"processing {k} -> {v}")
            time.sleep(0.5)
            q.put((k, v))
        # Tell the other end to stop.
        q.put(None)
    
    threading.Thread(target=worker).start()
    
    for item in as_completed(buf, q):
        print(f"received {item}")


example()
```

Output:

```
processing (0, 1) -> 43
processing (0, 2) -> 44
processing (0, 0) -> 42
received 42
received 43
received 44
processing (1, 0) -> 45
received 45
processing (1, 1) -> 46
received 46
processing (1, 2) -> 47
received 47
```

k-dominik: interesting. I think I'll explore a bit modifying `RoiRequestBatch`. Using it right now in this ordered manner seems like a workaround.

m-novikov: Maybe it makes sense to add `join` method to `RoiRequestsBatch` to await the completion of all pending requests.

k-dominik: while I can see that a `join` method can be useful in some cases, I don't think this is want we want to use here - data can be potentially be bigger than ram, so we'd want to write/discard data as quickly as possible.

k-dominik:
okay, the current state is heavily based on the "first version" that used `Threading.Event`, only difference is the usage of Queues and some tests....

emilmelnikov:
> In the end I decided not to touch `RoiRequestBatch` as it works decently good and secure with respect to exceptions and the likes...

Why `RoiRequestBatch` even exists?.. It says that it throttles the number of active requests, but requests are not threads: they are executed in a thread pool of a fixed size. And the ability of creating and running a (nearly) unlimited number of requests is the main reason why they are needed in the first place.

lazyflow/utility/roiRequestBuffer.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def to_index(self, roi: ROI_TUPLE) -> int:
+        """Convert ROI_TUPLE to index
+
+        first ROI_TUPLE -> index = 0
+        last ROI_TUPLE -> index = len(self) - 1
+        """
+        return numpy.ravel_multi_index(roi[0][:-2], self._shape[:-2])
+
+    def __len__(self) -> int:
```
Tomaz-Vieira, line 29: I guess this assumes shapes are in `xyztc` order like the rest of the codebase... Do you think we should move into something with less magical constants in this PR? I'm super biased here, but I'd strongly argue for using `Shape5D` and `Slice5D`everywhere instead of just `Tuple[int, int, int, int, int]`. And there's even the `Slot.getShape5D()` method to help with that. In fact, we could use `Slice5D.split()` to iterate over tiles of any shape

lazyflow/utility/roiRequestBuffer.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def __iter__(self) -> Iterator[numpy.ndarray]:
+        self._roi_request_batch.execute()
+        slot_shape = self._slot.meta.shape
+        while True:
+            if self._index not in self._items:
+                k, v = self._q.get()
+                idx = self._roi_iter.to_index(k)
+                self._items[idx] = v
+                continue
```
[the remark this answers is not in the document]
k-dominik: absolutely, yes. But they have to be written out in order. so tough luck. But it's hard to construct such a case tbh.

lazyflow/utility/roiRequestBuffer.py
```diff
@@ -0,0 +1,98 @@
...
+
+        self._items: Dict[int, numpy.ndarray] = {}
+        self._q: Queue[Tuple[ROI_TUPLE, numpy.ndarray]] = Queue()
+        self._index = 0
+        self._roi_iter = roi_iter_cls(self._slot.meta.shape)
+        self._max = len(self._roi_iter)
+
+        self._roi_request_batch = self._config_batch_request()
+
+    def _config_batch_request(self) -> RoiRequestBatch:
```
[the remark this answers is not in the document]
k-dominik: hmmm readability? :D

This stems from the times I thought that it could make sense to make this thing iterate multiple times over the data. So before I had it in `__iter__`... but yeah, one could probably shove it into `__init__`

lazyflow/utility/roiRequestBuffer.py
Tomaz-Vieira: Since `ndstructs`' API is still not ready for prime time, maybe you could use a dict here, which is consistent with `MetaDict.getTaggedShape()`? It'd be easier to understand that `roi[0][:-2]` means "just the spacial axes" if it were using the label names instead of their indices

lazyflow/utility/roiRequestBuffer.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def _put(self, *val):
+        self._q.put(val)
+
+    @property
+    def progress_signal(self):
+        return self._roi_request_batch.progressSignal
+
+    def __iter__(self) -> Iterator[numpy.ndarray]:
+        self._roi_request_batch.execute()
```
Tomaz-Vieira, line 83: I have a feeling that a lot of the `__init__` initialization could be moved into this method, and that most instance variables could be just local variables inside this method. For example. `self._q` isn't used anywhere else and could just be a local `q = Queue(...)`. Same for `self._roi_request_batch`, `self._index` and possibly a few others. As a bonus, this would mean that it would be safe to call `iter(my_instance_of_RoiRequestBuffer)` multiple times, because `__iter__` would reinitialize the relevant variables to start the iteration again. As an extra bonus, caling `__iter__` would also be thread safe if all of those were local variables =)

lazyflow/utility/roiRequestBuffer.py
k-dominik: it's actually `ctz`.... proves your point I guess :)

lazyflow/operators/ioOperators/opTiffReader.py
k-dominik: it means "unknown". Happens if you, e.g. add a signleton in `tifffile`...

lazyflow/utility/roiRequestBuffer.py
k-dominik: I do agree a bit :) a bit, only because this class, by design, with the progress update and all is really only supposed to be in action once. So I'll clean it up a little and add a guard that `__iter__` is in fact only active once.

lazyflow/utility/roiRequestBuffer.py
```diff
@@ -0,0 +1,98 @@
...
+
+class RoiRequestBuffer:
+    """Iterate over first 3 dims of a 5D slot in nd-ascending order."""
+
+    def __init__(self, slot: Slot, batchsize: int, roi_iter_cls: Type[_RoiIter] = _RoiIter):
+        """
+        Args:
+            slot: slot to request data from
+            batchsize: maximum number of requests to launch in parallel
+            roi_iter_cls: mainly exposed for testing, class that produces iterators\
```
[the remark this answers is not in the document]
k-dominik: okay, removed in favor of monkeypatch

lazyflow/utility/roiRequestBuffer.py
k-dominik: I'm expecting this code path to be gone soon. The RoiIter thing will now raise if axisorder is not as expected `tzcyx`.

Tomaz-Vieira (approved):

k-dominik:
Thx to everyone for the reviews! Happy we can get rid of our own tifffile build now :)

PR #2342, opened 2020-11-10
Fix failing windows ci

k-dominik:
specify channels for package installs


k-dominik:
seems like  the cause was a glitch in conda, which seems resolved...

PR #2348, opened 2020-11-19
Update server configuration

m-novikov:


ilastik/applets/serverConfiguration/configStorage.py
```diff
@@ -80,58 +77,13 @@ def get_server(self, id_) -> typing.Optional[types.ServerConfig]:
...
         for srv in servers:
-            self._write_server_entry(srv)
+            servers_data.append(attr.asdict(srv))
 
-        if isinstance(self._dst, str):
-            with open(self._dst, 'w+') as out:
-                self._config.write(out)
-        else:
-            self._config.write(self._dst)
+        preferences.set(SERVER_CONFIG_GROUP, SERVER_CONFIG_LIST_KEY, servers_data)
```
k-dominik, line 86: the cool thing is, that this fixes #2355 :+1:  (which the previously installed write to .ilastikrc it originally caused :/)

tests/test_ilastik/test_applets/test_serverConfiguration/test_configStore.py
```diff
@@ -1,162 +1,68 @@
...
+        self.config = defaultdict(dict)
 
-        return _parse
+    def set_server_configs(self, data):
+        self.set(SERVER_CONFIG_GROUP, SERVER_CONFIG_LIST_KEY, data)
 
-    def test_parsing_server(self, parse):
-        servers = parse(self.SINGLE_SERVER)
+    def get_server_configs(self):
+        print(self.config)
```
k-dominik, line 22: maybe remove the print

tests/test_ilastik/test_applets/test_serverConfiguration/test_configStore.py
```diff
@@ -1,162 +1,68 @@
...
+    def test_parsing_server(self, prefs_stub, config_storage):
+        prefs_stub.set_server_configs(
+            [
+                {
+                    "id": "myid1",
+                    "name": "MyServer1",
+                    "address": "127.0.0.1:4412",
+                    "devices": [
+                        {"id": "cpu0", "name": "CPU0", "enabled": False},
+                        {"id": "gpu1", "name": "MyCoolGPU", "enabled": True},
```
k-dominik, line 52: lol @ name

tests/test_ilastik/test_applets/test_serverConfiguration/test_configStore.py
```diff
@@ -1,162 +1,68 @@
...
+                    ],
+                }
+            ]
+        )
+        servers = config_storage.get_servers()
+        assert 1 == len(servers)
```
k-dominik, line 58: i know of course order here doesn't matter, but for me, logically reading from left to right `len(servers)` should be first... but you seem to prefer it the other way round, which is also fine.

ilastik/applets/serverConfiguration/configStorage.py
```diff
@@ -5,73 +5,34 @@
...
-        return res
+class ServerConfigStorage:
+    def __init__(self, preferences=preferences) -> None:
+        self.__preferences = preferences
+
+    def get_servers(self):
+        servers_data = self.__preferences.get(SERVER_CONFIG_GROUP, SERVER_CONFIG_LIST_KEY, [])
+        result = []
+        for server_dict in servers_data:
+            server_dict["devices"] = [_from_dict(types.Device, d) for d in server_dict.get("devices", [])]
```
k-dominik, line 33: does it actually make sense to save all devices? What happens if server configuration changes? (sorry for not finding out in code) or do you foresee usage of multiple devices and this key holds the selected one(s)?

k-dominik (commented):
Ahh :) So much code gone and a way nicer UI (always struggled explaining everything to workshop participants, as most of it was not needed, but still cluttered the UI).

One thing from a usage perspective:
* The edit button seems to only affect (make editable again) the address field, maybe it would be good to have it next to it?
* What does the save button actually do? I assume it saves the config to file, right? Does it have to be a button, would it hurt to save every time a device is selected? It's always hard to explain why users even have to click there (other than to enable the next applet).

k-dominik:
merging this to enable documenting that state properly (and users having this improved state early). We can also iterate on this with a bit of practical experience.

PR #2356, opened 2020-12-01
Uses new Slice5D.to_tuple from ndstructs

Tomaz-Vieira:
This updates ilastik to use the new Slice5D.to_tuple implementation, which returns

`Tuple[Tuple[int|None], ...], [Tuple[int|None], ...]]`

instead of

`Tuple[np.ndsarray, np.ndarray]`.

Requires `ndstructs 0.0.5dev0`


k-dominik (approved):
I guess my only complaint would be that the tests didn't fail with `ndstructs 0.0.4...`. I merged #2297 (no clue why this has not been merged before) and tested this PR rebased on that manually. works ;)

PR #2365, opened 2020-12-11
Live update shortcut available in all labelingGuis

akreshuk:
move live update shortcut into labelingGui


k-dominik (approved):
cool, I checked manually that it is working as expected.

PR #2367, opened 2020-12-11
Add from patterns dialog

emilmelnikov:
The new dialog allows a user to enter multiple file path and dataset name patterns.

image

Supersedes https://github.com/ilastik/ilastik/pull/2285.
Closes https://github.com/ilastik/ilastik/issues/2283.

### DIFF
--- a/ilastik/applets/dataSelection/addFileButton.py
+++ b/ilastik/applets/dataSelection/addFileButton.py
@@ -38,6 +38,7 @@ class AddFileButton(QPushButton):
     It presents a drop down menu with the following options:
 
     - Add separate image(s)
+    - Add multiple image(s)
     - Add 3D/4D volume from sequence
     - Add DVID volume
     - Add precomputed chunked volume
@@ -47,6 +48,7 @@ class AddFileButton(QPushButton):
     """
 
     addFilesRequested = pyqtSignal()
+    addFromPatternsRequested = pyqtSignal()
     addStackRequested = pyqtSignal()
     addRemoteVolumeRequested = pyqtSignal()
     addPrecomputedVolumeRequested = pyqtSignal()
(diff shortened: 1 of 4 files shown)


k-dominik (commented):
ah, with the examples it is very nice.

Disclaimer, haven't looked at the code, yet.

Some observations:
* ilastik does not play well with many image lanes. We should maybe warn users before really opening all, let's say >100 images.
* Opening many image causes ilastik to freeze (during creation of lanes). Maybe a progress bar or something would be nice to communicate to users that something is going on.
* When it's so easy to add _many_ images, it would be nice to have the possibility to remove many at once, too.
* a preview would be nice (as e.g. present in the stack import, that also  works also with h5 files and internal paths, have you looked at it at all?).

in future we might want to marry at least the stack import and this multiple file import.
Ultimately it would be cool to reduce it to a _single_ dialog somehow where users can add anything from a single image to importing a stack.

ilastik/applets/dataSelection/addFileButton.py
```diff
@@ -62,32 +61,19 @@ def __init__(self, parent, *, index=None, new=False):
...
+        super(AddFileButton, self).__init__(QIcon(str(_ICON_PATH)), "Add New..." if new else "Add...", parent)
 
-        self._index = index
+        self.index = index
         # drop down menu for different add options
         menu = QMenu(parent=self)
-        menu.addAction("Add separate Image(s)...").triggered.connect(self.addFilesRequested.emit)
-        menu.addAction("Add a single 3D/4D Volume from Sequence...").triggered.connect(self.addStackRequested.emit)
+        menu.addAction("Add separate Image(s)...").triggered.connect(self.addFilesRequested)
+        menu.addAction("Add multiple Image(s)...").triggered.connect(self.addFromPatternsRequested)
```
k-dominik, line 70: A more descriptive name would be cool. The button above already enables adding of multiple images...

emilmelnikov: What name do you suggest?

PR #2375, opened 2021-01-13
add possibility to silently force cache values

k-dominik:
probably not elegant, and possibly dangerous, open to suggestions here. But at least it works.

Currently, autocontext will start retraining in batch processing (starting with the second image), [user report](https://forum.image.sc/t/ilastik-autocontext-stage2-probability-output-range-incorrect-in-headless-mode/46603). This costs time, and even worse - will retrain without the data...

This only pops up in autocontext (the same is done in pixel classification, too) because one classifier output is input for the next stage :/

cache value can now be forced, without setting it's output to dirty.

### DIFF
--- a/ilastik/workflows/newAutocontext/newAutocontextWorkflow.py
+++ b/ilastik/workflows/newAutocontext/newAutocontextWorkflow.py
@@ -242,7 +242,7 @@ class NewAutocontextWorkflowBase(Workflow):
         # Restore classifier we saved in prepareForNewLane() (if any)
         if self.stored_classifers:
             for pcApplet, classifier in zip(self.pcApplets, self.stored_classifers):
-                pcApplet.topLevelOperator.classifier_cache.forceValue(classifier)
+                pcApplet.topLevelOperator.classifier_cache.forceValue(classifier, set_dirty=False)
 
             # Release references
             self.stored_classifers = []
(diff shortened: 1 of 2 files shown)


ilastik/workflows/newAutocontext/newAutocontextWorkflow.py
```diff
@@ -242,7 +242,7 @@ def handleNewLanesAdded(self):
         # Restore classifier we saved in prepareForNewLane() (if any)
         if self.stored_classifers:
             for pcApplet, classifier in zip(self.pcApplets, self.stored_classifers):
-                pcApplet.topLevelOperator.classifier_cache.forceValue(classifier)
+                pcApplet.topLevelOperator.classifier_cache.forceValue(classifier, set_dirty=False)
```
Tomaz-Vieira, line 245: Maybe you could clarify something for me since you've already dove deep into the issue... Is it just the second pixel classification that is retraining? And if so, is it because we're calling `forceValue` on the upstream classifier first, which propagates dirty into the second classifier?

Could this be solved by restoring the classifier cache on the downstream classifier operator first and then on the upstream one?

k-dominik: It's the second classifier that is retraining. right now upstream is set first, then downstream. I'd expect that it in fact should only work in this order. The other way round, if you set upstream last, it will propagate dirty downstream completely.

I tried this for fun anyway, and it still retrains :/

PR #2378, opened 2021-01-19
Cleanup/Fix of Multicut Gui Code

k-dominik:
Reorganized the code a bit after finding a good old `RuntimeError: wrapped C/C++ object of type QPushButton has been deleted` error when _closing_ the multicut workflow at the training stage.

Made inheritance from `LabelViewerGui` explicit by converting `EdgeTrainingGui` to a Mixin. Also removed some of the explicit `ParentClass.method` calls and let this be handled with `super`.

### DIFF
--- a/ilastik/applets/multicut/multicutGui.py
+++ b/ilastik/applets/multicut/multicutGui.py
@@ -56,7 +56,7 @@ logger = logging.getLogger(__name__)
 
 # This is a mixin that can be added to any LayerViewerGui subclass
 # See MulticutGui (bottom of this file) for the standalone version.
-class MulticutGuiMixin(object):
+class MulticutGuiMixin:
 
     ###########################################
     ### AppletGuiInterface Concrete Methods ###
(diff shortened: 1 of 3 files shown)


k-dominik:
This seems to solve some issues with the gui. IMO this can be merged as is. I'm furthermore looking into making multi-lane operational again, but will open a separate PR for that.

emilmelnikov (commented):
I really don't like that those mixins access lots of instance attributes, and have their own `__init__`, but it seems there is nothing we can do about this right now.

I'd suggest at least making sure that those mixins are used with the appropriate base class:
```python
class MyMixin:
    def __init_subclass__(cls, **kwargs):
        parent = LayerViewerGui
        if parent not in cls.mro():
            raise TypeError(
                f"{self.__class__.__name__} should be used "
                f"only with {parent.__class__.__name__!r}"
            )
        super().__init_subclass__(**kwargs)
```

k-dominik:
cool pattern, thx @emilmelnikov

ilastik/applets/multicut/multicutGui.py
```diff
@@ -72,6 +72,14 @@ def stopAndCleanUp(self):
     ###########################################
     ###########################################
 
+    def __init_subclass__(cls, **kwargs):
+        """Make sure Mixin can only be used with LayerViewerGui"""
+        parent = LayerViewerGui
+        if parent not in cls.mro():
+            breakpoint()
```
emilmelnikov, line 79: Stray `breakpoint()`?

Also, after some time I realized that the original suggestion could be simplified just to `assert issubclass(cls, LayerViewerGui)` in `__init_subclass__` (it's just a sanity check after all, users won't see that anyway).

k-dominik: ah, thank you for being so thorough :D I have almost merged it ... And I agree that an assert is good enough here.

PR #2379, opened 2021-01-21
Local remote nn

m-novikov:
_description update by @k-dominik_

This PR adds a "local" NN workflow, that will take care of spinning up/down the tiktorch server.

* In order to try this the current tiktorch (not built yet) is needed


ilastik/workflows/neuralNetwork/_localLauncher.py
```diff
@@ -0,0 +1,54 @@
...
+class LocalServerLauncher:
+    _executable_path: str
+    _process: subprocess.Popen
+
+    def __init__(self, executable_path: str) -> None:
+        self._executable_path = executable_path
+        self._process = None
+
+    def start(self):
+        if self._process:
```
Tomaz-Vieira, line 34: Any reason not to start the server in `__init__`? It would make it impossible for the server to be `.start`ed multiple times and it would remove the need for this check. It also makes it so that it's always legal to call `.stop`. And as a final bonus, it makes `_process` be correctly typed as `subprocess.Popen` instead of `Optional[subprocess.Popen]`

It's a shame one can't prevent a server from being `.stop`ped multiple times without ownership semantics =/

emilmelnikov: This class can be used as a contextmanager, so start/stop could turn into `__enter__` and `__exit__`.

m-novikov: > This class can be used as a contextmanager, so start/stop could turn into __enter__ and __exit__.

This class can't be used as context manager as it doesn't have clear scope located withing one function.

> Any reason not to start the server in __init__?

Starting the separate process in constructor seems like a unexpected behavior. Explicit verb in this case communicates intent better imo

emilmelnikov: > This class can't be used as context manager as it doesn't have clear scope located withing one function.

Why not? It has only init, start and stop. Seems reasonable to acquire/release resources with contextmanager, or at least define a separate function that does it.

m-novikov: >  Why not? It has only init, start and stop. Seems reasonable to acquire/release resources with contextmanager, or at least define a separate function that does it.

lifecycle of this class connected to a workflow class, so stop should be called from `cleanUp` method

ilastik/workflows/neuralNetwork/_remoteWorkflow.py
```diff
@@ -0,0 +1,249 @@
...
+        opDataSelection = self.dataSelectionApplet.topLevelOperator
+        input_ready = len(opDataSelection.ImageGroup) > 0 and not self.dataSelectionApplet.busy
+
+        opNNClassification = self.nnClassificationApplet.topLevelOperator
+        serverConfig_finished = self.serverConfigApplet.topLevelOperator.ServerConfig.ready()
+
+        opDataExport = self.dataExportApplet.topLevelOperator
+
+        predictions_ready = input_ready and len(opDataExport.Inputs) > 0
+        # opDataExport.Inputs[0][0].ready()
```
k-dominik, line 151: leftover stuff?

ilastik/workflows/neuralNetwork/_remoteWorkflow.py
```diff
@@ -0,0 +1,249 @@
...
+    def __init__(self, shell, headless, workflow_cmdline_args, project_creation_args, *args, **kwargs):
+        graph = Graph()
+        super().__init__(shell, headless, workflow_cmdline_args, project_creation_args, graph=graph, *args, **kwargs)
+        self._applets = []
+        self._workflow_cmdline_args = workflow_cmdline_args
+
+        # Parse workflow-specific command-line args
+        parser = argparse.ArgumentParser()
+        parser.add_argument("--batch-size", help="choose the preferred batch size", type=int)
+        parser.add_argument("--halo-size", help="choose the preferred halo size", type=int)
```
k-dominik, line 56: isn't halo size network dependent?

ilastik/workflows/neuralNetwork/_localWorkflow.py
```diff
@@ -0,0 +1,248 @@
...
+
+from lazyflow.operators import tiktorch
+
+from lazyflow.graph import Graph
+
+
+logger = logging.getLogger(__name__)
+
+
+class LocalWorkflow(Workflow):
```
k-dominik, line 22: With the local workflow being more or less the same than the remote one, minus the serverconfig applet, shouldn't it make sense to subclass the remote from the local one, or at least having some common base?

tests/test_ilastik/conftest.py
```diff
@@ -7,3 +7,16 @@
...
     return conf_path / "data"
+
+
+def pytest_addoption(parser):
+    parser.addoption("--tiktorch_executable", help="tiktorch executable to use for integration tests")
+
+
+@pytest.fixture
+def tiktorch_executable_path(request):
+    tiktorch_exe_path = request.config.getoption("--tiktorch_executable", None)
```
k-dominik, line 18: I'd also probably move this to top level conftest (where also `run-legacy-gui` is defined). Or at least have both in the same place.

tests/test_ilastik/conftest.py
```diff
@@ -7,3 +7,16 @@
 def data_path():
     conf_path = pathlib.Path(__file__).parent
     return conf_path / "data"
+
+
+def pytest_addoption(parser):
+    parser.addoption("--tiktorch_executable", help="tiktorch executable to use for integration tests")
```
k-dominik, line 13:
```suggestion
    parser.addoption("--tiktorch-executable", help="tiktorch executable to use for integration tests")
```
I know, it's then not consistent with argument of ilastik executable, but I think everyone from the devs expects this weird convention only from ilastik.

ilastik/workflows/neuralNetwork/_localWorkflow.py
```diff
@@ -0,0 +1,248 @@
...
+    """
+    This class provides workflow for a remote tiktorch server
+    It has special server configuration applets allowing user to
+    connect to remotely running tiktorch server managed by user
+    """
+
+    auto_register = False
```
k-dominik, line 29: why is it disabled?

k-dominik (changes requested):
Hey hey. This needs some massaging.

I really like the prospect of having an easy to use local workflow!

In general, I wanted to test this, popped in the latest tiktorch binary and of course it didn't work. This can be expected I guess. It failed pretty ungraceful though, throwing an exception...

It would be nice to add development instructions to CONTRIBUTING.

ilastik/app.py
```diff
@@ -256,6 +260,27 @@ def _update_hbp_mode(parsed_args):
...
+        root_path = Path(root_path)
+        tiktorch_script = "run_tiktorch.sh" if platform.system() != "Windows" else "run_tiktorch.bat"
+        bundled_tiktorch_executable = root_path / "tiktorch" / tiktorch_script
+
+        if bundled_tiktorch_executable.exists():
+            tiktorch_executable = bundled_tiktorch_executable
+
+    if tiktorch_executable and tiktorch_executable.exists():
+        tiktorch_msg = "Using tiktorch executable: %s" % tiktorch_executable
+        print(tiktorch_msg)
```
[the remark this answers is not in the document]
m-novikov: Nope, this is intentional because logger may be uninitialized and we want to have this output. Same with the message "Starting ilastik from ..." in this file

ilastik/workflows/neuralNetwork/_localWorkflow.py
m-novikov: Because registration is conditional dependent on local tiktorch availability

ilastik/workflows/__init__.py
```diff
@@ -134,9 +134,13 @@
 
 # network classification, check whether required modules are available:
 try:
-    from .nnClassification import NNClassificationWorkflow
+    from . import neuralNetwork
+
+    WORKFLOW_CLASSES += [neuralNetwork.RemoteWorkflow]
+    tiktorch_exe_path = ilastik.config.cfg.get("ilastik", "tiktorch_executable", fallback=None)
```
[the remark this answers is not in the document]
m-novikov: hmm, this is done similar to https://github.com/ilastik/ilastik/blob/master/ilastik/app.py#L249 https://github.com/ilastik/ilastik/blob/master/ilastik/app.py#L256
we don't have transient config

ilastik/workflows/neuralNetwork/_localWorkflow.py
k-dominik: well, in any case, it doesn't work currently :) at least not for me if I supply tiktorch-server exec via cmd (verified that it's indeed picked up, didn't debug into it further).

ilastik/workflows/__init__.py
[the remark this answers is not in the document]
k-dominik: I'd say this is already bad for the cases you listed there, but it will not break anything. But I guess this would be out of scope for this PR then.

ilastik/app.py
```diff
@@ -256,6 +260,27 @@ def _update_hbp_mode(parsed_args):
...
+        logger.info(tiktorch_msg)
+        ilastik_config.set("ilastik", "tiktorch_executable", str(tiktorch_executable))
```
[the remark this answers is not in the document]
k-dominik: maybe it's not a good idea to safe this in config at all. It should be determined at runtime - either via the argument or via detecting it in the distro (or from ilastikrc) but I wouldn't write there. Or do you think that the benefits outweigh the potential issues?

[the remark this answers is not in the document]
m-novikov: I think we could introduce a singleton object for this kind of options, only reason it's in config right now is because it looked similar to hbp and debug flags.

k-dominik (approved):

k-dominik:
hello @m-novikov, somehow I cannot request a review from you (I assume it's because it's your PR :D). In any case, would you mind having a look at the current state?

Also, I've two questions :)

1) is the `--halo-size` not network dependent? (Batch size _might_ be, too?)
2) There is some stuff, like the model wizard, and the hyper parameter setting thing that only make sense in context of training, right, so currently those should not be exposed.

tests/test_ilastik/test_workflows/test_workflow_listing.py
```diff
@@ -0,0 +1,60 @@
...
+            ),
+        ),
+        (
+            [Mock(workflowName="Test", __name__="Test", workflowDisplayName=None)],
+            [Mock(workflowName="Test", __name__="Test", workflowDisplayName=None)],
+            (("Test", "Test"),),
+        ),
+    ],
+)
+def test_listing(monkeypatch, workflowclasses, subclasses, expected):
```
m-novikov, line 51: Awesome :)
Maybe enclose test cases with `# fmt: off` because it's hard to parse otherwise

lazyflow/slot.py
```diff
@@ -1374,7 +1374,7 @@ def _configureOperator(self, slot, oldSize=0, newSize=0, notify=True):
             self.maybe_call_within_transaction(self.operator._setupOutputs)
 
     def _setupOutputs(self):
-        """"""
+        """ """
```
m-novikov, line 1377: unrelated

ilastik/workflows/neuralNetwork/_nnWorkflowBase.py
```diff
@@ -68,19 +62,16 @@ def imageNameListSlot(self):
...
+        super().__init__(shell, headless, workflow_cmdline_args, project_creation_args, graph=graph, *args, **kwargs)
+
         self._applets = []
         self._workflow_cmdline_args = workflow_cmdline_args
 
         # Parse workflow-specific command-line args
         parser = argparse.ArgumentParser()
-        parser.add_argument("--batch-size", help="choose the prefered batch size", type=int)
-        parser.add_argument("--halo-size", help="choose the prefered halo size", type=int)
+        parser.add_argument("--batch-size", help="choose the preferred batch size", type=int)
```
m-novikov, line 73: At least `halo-size` should be dropped I think

ilastik/workflow.py
```diff
@@ -260,37 +262,49 @@ def getAvailableWorkflows():
 
     from . import workflows
 
-    for W in workflows.WORKFLOW_CLASSES + all_subclasses(Workflow):
+    def _makeWorkflowTuple(workflow_cls):
```
m-novikov, line 265: Adding the description of this tuple to parent function docstring would be nice

tests/test_ilastik/test_workflows/test_workflow_listing.py
k-dominik: hmm, what do you mean by hard to parse? Wold you prefer a single line per test case?

Issue #2385, opened 2021-01-27
ilastik fails on servers without gui libs


k-dominik:
https://github.com/ilastik/ilastik/blob/19d89a88817169699d61ae7979ce858b2c754772/ilastik/app.py#L286

something innocent as migrating preferences causes ilastik to fail in headless mode on servers that don't have gui libs installed :/

Any good ideas for workarounds?

<details>

```pytb
Traceback (most recent call last):
  File "/ilastik-release/ilastik-meta/ilastik/ilastik.py", line 122, in <module>
    main()
  File "/ilastik-release/ilastik-meta/ilastik/ilastik.py", line 118, in main
    main()
  File "/ilastik-release/ilastik-meta/ilastik/ilastik/__main__.py", line 30, in main
    hShell = app.main(parsed_args, workflow_cmdline_args)
  File "/ilastik-release/ilastik-meta/ilastik/ilastik/app.py", line 141, in main
    _init_preferences()
  File "/ilastik-release/ilastik-meta/ilastik/ilastik/app.py", line 287, in _init_preferences
    from volumina.utility.preferences import migrate
  File "/ilastik-release/ilastik-meta/volumina/volumina/__init__.py", line 47, in <module>
    from . import api
  File "/ilastik-release/ilastik-meta/volumina/volumina/api.py", line 25, in <module>
    from .pixelpipeline.imagepump import ImagePump
  File "/ilastik-release/ilastik-meta/volumina/volumina/pixelpipeline/imagepump.py", line 30, in <module>
    from volumina.pixelpipeline.slicesources import PlanarSliceSource, SyncedSliceSources
  File "/ilastik-release/ilastik-meta/volumina/volumina/pixelpipeline/slicesources.py", line 28, in <module>
    from .interface import DataSourceABC, PlanarSliceSourceABC, RequestABC
  File "/ilastik-release/ilastik-meta/volumina/volumina/pixelpipeline/interface.py", line 29, in <module>
    from volumina.utility.qabc import QABC, abstractsignal
  File "/ilastik-release/ilastik-meta/volumina/volumina/utility/__init__.py", line 25, in <module>
    from .getMainWindow import getMainWindow
  File "/ilastik-release/ilastik-meta/volumina/volumina/utility/getMainWindow.py", line 1, in <module>
    from PyQt5.QtWidgets import QApplication, QWidget, QMainWindow
ImportError: libGL.so.1: cannot open shared object file: No such file or directory
```

</details>

[...]

PR #2386, opened 2021-01-27
Do not migrate preferences in the headless mode

emilmelnikov:
Running ilastik in the headless mode caused it to crash because
attempting to migrate preferences indirectly caused import of the
volumina package, which, in turn, imports Qt GUI libraries that could
be unavailable in the headless mode.

Closes https://github.com/ilastik/ilastik/issues/2385.


k-dominik (approved):
LGTM!

PR #2390, opened 2021-02-08
Add futures like add_done_callback to requests

m-novikov:
The initial step in the direction of a bit more transparent concurrency control in volumina pixelpipeline

### DIFF
--- a/lazyflow/request/request.py
+++ b/lazyflow/request/request.py
@@ -723,7 +723,6 @@ class Request(object):
             if not complete:
                 # Call when we eventually finish
                 self._sig_execution_complete.subscribe(lambda: callback(self))
-                return
 
         if complete:
             callback(self)
--- a/lazyflow/request/request.py
+++ b/lazyflow/request/request.py
@@ -714,11 +714,16 @@ class Request(object):
         Call callback when request is finished regardless of the state
         """
         assert not self._cleaned, "This request has been cleaned() already."
+
+        # Request either still running in this case take a lock so state transition doesn't
+        # happen while we are subscribing or immidiatelly fire the callback in case it
+        # has already happened
         with self._lock:
             complete = self.execution_complete
             if not complete:
                 # Call when we eventually finish
                 self._sig_execution_complete.subscribe(lambda: callback(self))
+                return
 
         if complete:
             callback(self)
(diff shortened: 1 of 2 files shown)


k-dominik (approved):
lgtm :) The tests are explanatory and appreciated!

lazyflow/request/request.py
```diff
@@ -708,6 +708,21 @@ def _handle_finished_request(self, request, *args):
...
                 self._wake_up()
 
+    def add_done_callback(self, callback: Callable[["Request"], None]) -> None:
+        """
+        Mimic concurrent.Future interface
+        Call callback when request is finished regardless of the state
+        """
+        assert not self._cleaned, "This request has been cleaned() already."
+        with self._lock:
+            complete = self.execution_complete
```
emilmelnikov, line 718: Control flow is not obvious here. It would be great to either add a comment or maybe use an early return:

```python
with self._lock:
    if not self.execution_complete:
        self._sig_execution_complete.subscribe(lambda: callback(self))
        return

callback(self)

m-novikov: Updated :) not sure if added comment covers what you mean

k-dominik: since it was copy-pasted from the one of the `notify_...` functions, it would make sense to have only a single pattern of control flow.

emilmelnikov: OK, maybe it's not really necessary then. I was slightly confused by double-testing the `complete` variable within _and_ outside the lock, so I thought maybe it'd be better to have an early return.

k-dominik: sure, I think this looks better, too. So we could change it in the whole file, no?

m-novikov: > sure, I think this looks better, too. So we could change it in the whole file, no?

I think we should use the same pattern like in other notify functions, early return is not implementable for notifications where two flags are used https://github.com/ilastik/ilastik/pull/2390/files#diff-4801f2aba46cdd35eeca59573dbc9df361125acef1aad3e7a240928838a4f8f9R763

Tomaz-Vieira (approved):

lazyflow/request/request.py
k-dominik: okay, then let's leave it at that.

PR #2395, opened 2021-02-14
Make sure to use int datatype for potentially large volumes

k-dominik:
In `numpy` default `int` type is `long`. [On windows, default `long` is 32 bits](https://numpy.org/devdocs/reference/arrays.scalars.html#integer-types).

added a tiny helper function that should be used when reducing shapes to their volume in order not to fall into this issue

ref: [user report on image.sc](https://forum.image.sc/t/overflow-warning-in-ilastik/48790?u=k-dominik)

### DIFF
--- a/lazyflow/utility/helpers.py
+++ b/lazyflow/utility/helpers.py
@@ -20,7 +20,10 @@
 # 		   http://ilastik.org/license/
 ###############################################################################
 
-from typing import Tuple
+from functools import reduce
+from operator import mul
+from typing import Tuple, Iterable
+import numbers
 
 
 def itersubclasses(cls, _seen=None):
@@ -115,3 +118,12 @@ def get_default_axisordering(shape: Tuple[int, ...]) -> str:
         return axisorders[ndim]
     except KeyError:
         raise ValueError(f"cannot infer axis order for {ndim}-dimensional shape")
+
+
+def bigintprod(nums: Iterable[numbers.Integral]) -> int:
+    """Safe way to get the product of an iterable of integer numbers
+
+    avoids defaulting to C long (like numpy), which can result in overflows for
+    array shapes seen in practice on some operating systems (windows: long -> 32 bit).
+    """
+    return reduce(mul, map(int, nums))
--- a/lazyflow/utility/roiRequestBatch.py
+++ b/lazyflow/utility/roiRequestBatch.py
@@ -31,6 +31,7 @@ import numpy
 
 import lazyflow.stype
 from lazyflow.utility import OrderedSignal
+from lazyflow.utility.helpers import bigintprod
 from lazyflow.request import Request, SimpleRequestCondition, log_exception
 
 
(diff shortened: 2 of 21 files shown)


Tomaz-Vieira (approved):
LGTM

lazyflow/utility/roiRequestBatch.py
```diff
@@ -247,7 +250,7 @@ def _handleCompletedRequest(self, roi, result):
                 # Report progress (if possible)
                 if self._totalVolume is not None:
                     self._processedVolume += numpy.prod(numpy.subtract(roi[1], roi[0]))
```
emilmelnikov, line 252: Shouldn't the same problem happen here with `numpy.prod`?

k-dominik: good catch - that actually means that I should rethink the fix, as this only enforces `_totalVolume` to be `int` once set in this class. It is, however, calculated itself outside this class with `numpy.prod`...

k-dominik: you got me paranoid ;)

lazyflow/utility/helpers.py
```diff
@@ -115,3 +118,12 @@ def get_default_axisordering(shape: Tuple[int, ...]) -> str:
         return axisorders[ndim]
     except KeyError:
         raise ValueError(f"cannot infer axis order for {ndim}-dimensional shape")
+
+
+def shapeprod(shape: Iterable[numbers.Integral]) -> int:
```
emilmelnikov, line 123: Python 3.8 and later has [`math.prod`](https://docs.python.org/3/library/math.html#math.prod), so maybe it'd be better to add `lazyflow.future.math`, import it as `from lazyflow.future import math`, and `math.prod(shape)`? Later, we can replace `from lazyflow.future import math` with just `import math`.

emilmelnikov (commented):
LGTM, but could be make compatible with the future Python stdlib.

lazyflow/utility/helpers.py
k-dominik: I agree that it would be nice, however, we'd still need to convert to int: 
Something like `shape = numpy.array((10**31, 100), dtype="uint32")` will resort in an overflow error...

k-dominik: instead of `reduce(mul,...)` we could use `itertools.prod`...

emilmelnikov: Yes, you are correct, I erroneously assumed that `math.prod` would convert elements to int...

emilmelnikov: In this case, how about renaming `shapeprod` to `approd`, `bignumprod`, or something similar?

k-dominik: hm good point. `shapeprod` is not very informative... how about `bigintprod` as yet another alternative @emilmelnikov? :)

emilmelnikov: `bigintprod` is a perfect name!

PR #2396, opened 2021-02-16
expoerimental/API added roi parameter to pixel classification pipeline

k-dominik:
In order to be more useful for larger data, this PR adds a `roi` parameter to the `predict` method of the pixel classification api.
Right now I went for `xarray` dict-like specification, which looks very sensible.

Example:

```python
import h5py
import xarray
from ilastik.experimental.api import from_project_file

pipeline = from_project_file("my_pretrained_project.ilp")
f = h5py.File("my-big-image.h5", "r")
ds = xarray(f["data"], dims="zyx")
block_prediction = pipeline.predict(ds, {"x": slice(0, 100), "y": slice(0, 200)}  # all "z"
```

### DIFF
--- a/ilastik/experimental/api/_base.py
+++ b/ilastik/experimental/api/_base.py
@@ -43,7 +43,7 @@ def from_project_file(path) -> Pipeline:
             self._predict_op.Image.connect(self._feature_sel_op.OutputImage)
             self._predict_op.LabelsCount.setValue(classifer.label_count)
 
-        def predict(self, data):
+        def predict(self, data, xarray_roi=dict()):
             data = convert_to_vigra(data)
             num_channels_in_data = data.channels
             if num_channels_in_data != num_channels:
@@ -59,11 +59,26 @@ def from_project_file(path) -> Pipeline:
                 )
 
             self._reorder_op.Input.setValue(data)
-
-            data = self._predict_op.PMaps.value[...]
+            roi = xarray_roi_to_slicing(xarray_roi, self._predict_op.PMaps.meta.axistags.keys())
+            data = self._predict_op.PMaps[roi].wait()
             return xarray.DataArray(data, dims=tuple(self._predict_op.PMaps.meta.axistags.keys()))
 
     return _PipelineImpl()
+
+
+def xarray_roi_to_slicing(xarray_roi, vigra_axistags):
+    if any(k not in vigra_axistags for k in xarray_roi.keys()):
+        raise ValueError(f"xarray_roi contains keys ({xarray_roi.keys()} not in axistags ({vigra_axistags})")
+    if not xarray_roi:
+        return Ellipsis
+
+    out_slicing = []
+    for axis in vigra_axistags:
+        if axis in xarray_roi:
+            out_slicing.append(xarray_roi[axis])
+        else:
+            out_slicing.append(slice(None, None, None))
+    return tuple(out_slicing)
 
 
 @singledispatch
--- a/ilastik/experimental/api/__init__.py
+++ b/ilastik/experimental/api/__init__.py
@@ -1,2 +1,2 @@
-from ._base import from_project_file
+from ._base import from_project_file, xarray_roi_to_slicing
 from .types import Pipeline
--- a/tests/test_ilastik/test_experimental/test_api/test_base.py
+++ b/tests/test_ilastik/test_experimental/test_api/test_base.py
@@ -7,7 +7,7 @@ import vigra
 import xarray
 from imageio import imread
 
-from ilastik.experimental.api import from_project_file
+from ilastik.experimental.api import from_project_file, xarray_roi_to_slicing
 
 from ..types import TestData, TestProjects, ApiTestDataLookup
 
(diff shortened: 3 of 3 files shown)


ilastik/experimental/api/_base.py
```diff
@@ -43,7 +43,7 @@ def __init__(self):
             self._predict_op.Image.connect(self._feature_sel_op.OutputImage)
             self._predict_op.LabelsCount.setValue(classifer.label_count)
 
-        def predict(self, data):
+        def predict(self, data, xarray_roi=dict()):
```
emilmelnikov, line 46:
```python
def predict(self, data, xarray_roi=None):
    if xarray_roi is None:
        xarray_roi = {}
```

ilastik/experimental/api/_base.py
```diff
@@ -59,13 +59,28 @@ def predict(self, data):
...
-            data = self._predict_op.PMaps.value[...]
+            roi = xarray_roi_to_slicing(xarray_roi, self._predict_op.PMaps.meta.axistags.keys())
+            data = self._predict_op.PMaps[roi].wait()
             return xarray.DataArray(data, dims=tuple(self._predict_op.PMaps.meta.axistags.keys()))
 
     return _PipelineImpl()
 
 
+def xarray_roi_to_slicing(xarray_roi, vigra_axistags):
+    if any(k not in vigra_axistags for k in xarray_roi.keys()):
```
emilmelnikov, line 70:
```suggestion
    if set(xarray_roi).difference(vigra_axistags):
```

ilastik/experimental/api/_base.py
```diff
@@ -59,13 +59,28 @@ def predict(self, data):
...
+    if not xarray_roi:
+        return Ellipsis
+
+    out_slicing = []
+    for axis in vigra_axistags:
+        if axis in xarray_roi:
+            out_slicing.append(xarray_roi[axis])
+        else:
+            out_slicing.append(slice(None, None, None))
+    return tuple(out_slicing)
```
emilmelnikov, line 81:
```suggestion
    return tuple(xarray_roi.get(axis, slice(None)) for axis in vigra_axistags)
```

ilastik/experimental/api/_base.py
```diff
@@ -59,13 +59,28 @@ def predict(self, data):
...
+            data = self._predict_op.PMaps[roi].wait()
             return xarray.DataArray(data, dims=tuple(self._predict_op.PMaps.meta.axistags.keys()))
 
     return _PipelineImpl()
 
 
+def xarray_roi_to_slicing(xarray_roi, vigra_axistags):
+    if any(k not in vigra_axistags for k in xarray_roi.keys()):
+        raise ValueError(f"xarray_roi contains keys ({xarray_roi.keys()} not in axistags ({vigra_axistags})")
+    if not xarray_roi:
```
emilmelnikov, line 72: Do we really need this special case? That is, does `a[...] == a[:, :, :, :, :]` always holds in volumina?

Tomaz-Vieira:
Some objections against dictionaries are:

 - the set of keys is not obvious (nor is their case sensitiveness),
 - The meaning of missing keys might also not be obvious,
 - Static type checking will not catch

Tomaz-Vieira:
Some objections against dictionaries are:

 - the set of keys is not obvious (nor is their case sensitiveness);
 - The meaning of missing keys might also not be obvious;
 - Extra or mistyped keys are not immediately detected (if at all);  
 - Static type checking can't catch any errors in the dict;

m-novikov:
Another idea to return intermediate objects when we call predict and only materialize it on direct query.
```python
block_prediction = pipeline.predict(ds).isel(x=slice(0,10), y=slice(0, 17)).result()
# this kind of block interaction is nicer to read imo
predictor = pipeline.Predictor(ds)
computed = []
for block in tile(ds.shape, 256):
     tile_prediction = predictor.isel(**block).result()
     computed.append(tile_prediction)
```
it would also play nicer if we have multi-value returns

ilastik/experimental/api/_base.py
k-dominik: what is the objection against the default value?

ilastik/experimental/api/_base.py
k-dominik: cheers, I was actually hoping to get this from xarrray and didn't look into their docs yet... :) But I'll use your suggestion if there is nothing to be found.

ilastik/experimental/api/_base.py
k-dominik: it depends on the data, unfortunately. I wanted to send a clear message for what happens, when no roi is provided.

ilastik/experimental/api/_base.py
emilmelnikov: Because default function values are evaluated and assigned on function definition, _not_ evaluation.This doesn't play well with mutable values: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments.

k-dominik:
> Some objections against dictionaries are:
> 
>     * the set of keys is not obvious (nor is their case sensitiveness);
> 
>     * The meaning of missing keys might also not be obvious;
> 
>     * Extra or mistyped keys are not immediately detected (if at all);
> 
>     * Static type checking can't catch any errors in the dict;

I agree, there are some nice pro's tho:

* builtin type, anyone knows how to interact with them (I would say this is the strongest argument _for_ dicts, also that xarray implements indexing with dicts is a plus). 
* once you're in a notebook, no static type checking will help you (well, there might be tooling though that I don't know, can you mypy live in notebooks these days?)
* There is `TypedDict` (we'd have to upgrade Python though, or add yet another dependency)

> Another idea to return intermediate objects when we call predict and only materialize it on direct query.

What I liked about  the first iteration of the api is that was a single call to get a prediction. That is staying with the original `predict` interface. I could imagine something like this as part of a `blockwise_predict` function? With this `isel` method, that would return a future...

ilastik/experimental/api/_base.py
k-dominik: doesn't happen with `dict()` though...

Tomaz-Vieira:
> builtin type, anyone knows how to interact with them

this is only half-true, though. `dict` is indeed a builtin, but we don't just take ANY builtin `dict`; We take a very particular type of `dict`, which the user will have to learn to build. And he can do that either by reading English ("roi is a Dict[str, Tuple[in, int]] where the keys are in "xyztc" "), or by reading python:

```python3
```

> once you're in a notebook, no static type checking will help you

True that, but basic python functionality will still work. This produces no errors on the call site:

```python3
predict(roi={"X": (100, 200), "y": (300, 400)})
```

While this:

```python3
predict(roi=CustomRoi(X=(100, 200), y=(300,400)))
```
immediately throws a "TypeError: Got an unexpected keyword argument: X" , right at the call site.

Tomaz-Vieira:
> builtin type, anyone knows how to interact with them

this is only half-true, though. `dict` is indeed a builtin, but we don't just take ANY builtin `dict`; We take a very particular type of `dict`, which the user will have to learn to build. And he can do that either by reading English ("roi is a Dict[str, Tuple[in, int]] where the keys are in "xyztc" "), or by reading python:

```python3
def __init__(self, x: Tuple[int, int], y: Tuple[int, int], z: Tuple[int, int], c: Tuple[int, int], t: Tuple[int, int])
```

which comes with the added boni that the second version can have explicit default values and does not silently accept bad keys

> once you're in a notebook, no static type checking will help you

True that, but basic python functionality will still work. This produces no errors on the call site:

```python3
predict(roi={"X": (100, 200), "y": (300, 400)})
```

While this:

```python3
predict(roi=CustomRoi(X=(100, 200), y=(300,400)))
```
immediately throws a "TypeError: Got an unexpected keyword argument: X" , right at the call site.

> There is TypedDict
I guess that could also work; the syntax is indeed more ergonomic at the cost of runtime validation. But if ever we need to add some validation in the constructor or add some methods to the rois (enlarge, split, get_tiles, etc), we're back to a custom Roi class

k-dominik:
How about we hide that from users and offer mypy compatibility:


```Python
RoiDict = TypedDict('RoiDict', {"x": ..., "y", ....}, total=False)

class Classifier:
    ...
    def predict(roi: RoiDict):
        croi = CustomRoi(**roi)
```

Wouldn't this be way nicer for users?

Issue #2299, opened 2020-08-31
classifier retrains from all zeros data if raw data not accessible anymore


k-dominik:
right now (`1.4.0b7`) if a trained classifier hasn't been saved (say pixel classification), ilastik will try to retrain in headless mode. If the data is not available, it will be connected to "pseudo" sources with all zeros. It shouldn't retrain in that case but rather fail with helpful error message.

PR #2398, opened 2021-02-16
Make sure ilastik errors when trying to access missing data in headless

k-dominik:
When running ilastik in headless, we do not require training data to be present. However, ilastik will try to train its classifier, if it has not been saved to the project file. This can lead to training from all zeros when data is inaccessible.
This is of course not acceptable.

fixes https://github.com/ilastik/ilastik/issues/2299

### DIFF
--- a/lazyflow/operators/valueProviders.py
+++ b/lazyflow/operators/valueProviders.py
@@ -500,8 +500,15 @@ class OpZeroDefault(Operator):
             self.Output.setDirty(roi)
 
 
-class OpZeroSource(Operator):
-    """Operator that serves zeros on it's Output
+class MissingDataAccessError(Exception):
+    pass
+
+
+class OpRaisingSource(Operator):
+    """Operator that raises when data is accessed
+
+    Provides valid metadata, but does not allow to get any data.
+    Needed to provide metadata for "missing" datasets in trained projects.
 
     This operator will never set anything dirty as it cannot be changed.
     """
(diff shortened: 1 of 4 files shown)


lazyflow/operators/valueProviders.py
```diff
@@ -522,8 +529,10 @@ def setupOutputs(self):
         pass
 
     def execute(self, slot, subindex, roi, result):
-        result[...] = 0
-        return result
+        raise MissingDataAccessError(
```
emilmelnikov, line 532: The error message could be moved into `MissingDataAccessError.__init__()`.

Also, the operator name is slightly more general than the exception message. Maybe rename it to something like `OpMissingDataSource`?

emilmelnikov (approved):
The error message could be moved into `MissingDataAccessError.__init__()`.

Also, the operator name is slightly more general than the exception message. Maybe rename it to something like `OpMissingDataSource`?

emilmelnikov:
Alternatively, we could generalize the op by lifting the error message to the op constructor.

Tomaz-Vieira:
I feel rather unconfortable having a DataSource that just throws an exception... It feels like us admitting defeat.

Maybe we can do this kind of stuff right at the start of the workflow, as it is parsing the command line parameters and determining whether it is going to train or not?

k-dominik:
> I feel rather unconfortable having a DataSource that just throws an exception... It feels like us admitting defeat.
> 
> Maybe we can do this kind of stuff right at the start of the workflow, as it is parsing the command line parameters and determining whether it is going to train or not?

I agree that (in the current framework) we might want to add, maybe a method to the exporting operator that is overridden by every workflow and that can do certain checks and produce a meaningful message. Nevertheless - if all goes wrong there should still be a point of failure, so in the current architecture we need this.

emilmelnikov (approved):

Issue #1205, opened 2016-04-05
Make sure we compress all data to reduce the size of an ilastik project file


chaubold:
... which could be useful when only the trained classifiers and settings are needed for batch processing. If the project needs to be distributed to several compute nodes so they can parallelize e.g. Object Classification of a new dataset over blocks of data, only the classifiers e.t.c. are needed, but not the cached label images of the dataset originally used for training.

akreshuk:
If the label images are in a  compressed cache, they shouldn't take any space at all. And there is something positive about being able to retrain if needed. But, of course, if it takes too much space in your experience, we should get rid of it. And we shouldn't store it separately for each object classifier, if it's happening now.

chaubold:
After some more investigation I found out that the problem is mainly in tracking, where the label images displaying the resulting tracks are cached, together with a new label image after merger resolving. This quickly adds up to a lot of memory.

- [ ] check whether the data type of the cached label images is the right one and whether they are stored compressed
- [ ] should we store the resulting re-colored label images at all, as they can be recomputed from the stored tracking results? Resolving mergers is a time consuming process though, which would slow down project loading if this needs to be done at startup.

JaimeIvanCervantes:
You could store the colored label image of the first frame for display purposes only, since of the rest of the labels can be recomputed.

chaubold:
I've just noticed that if I compress my 600mb tracking `ilp` file, it reduces to 5mb! We must be doing something wrong when it comes to compression, probably of the label images. 

One pecularity about tracking is that we store the `coordinateMap` (a C++ object), serialized as binary string from C++ and then saved as python string. We should make sure that this is not huge.

PR #2401, opened 2021-02-22
made sure compression is passed through to the serialization of data

k-dominik:
looks like we didn't compress when we really should've :/

Anecdotal space saving for a project I have:
From 6.9 GB -> 40 MB...

and yeah the added test now suddenly using fixtures, I know. But I couldn't get myself to hacking it into one of the other tests (especially because a  larger block size is required, or compression will not have an effect on size of the final file...
fixes https://github.com/ilastik/ilastik/issues/1205

### DIFF
--- a/tests/test_ilastik/test_applets/base/testSerializer.py
+++ b/tests/test_ilastik/test_applets/base/testSerializer.py
@@ -26,6 +26,7 @@ import unittest
 import shutil
 import tempfile
 import pytest
+from pathlib import Path
 from copy import deepcopy
 from ilastik.applets.base.appletSerializer import SerialObjectFeatureNamesSlot
 from lazyflow.graph import Graph, Operator, InputSlot, Slot, OperatorWrapper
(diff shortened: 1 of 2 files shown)


Tomaz-Vieira (approved):
This is great =)
I think I've stumbled upon this a while ago while posting .ilp files to the job runner, but we just ended up increasing the size limit on the server; but the server will not take a 6GB file, nor would any user have the patience to ever upload that =)

tests/test_ilastik/test_applets/base/testSerializer.py
```diff
@@ -535,6 +536,51 @@ def iter_dataset_names(name):
...
+
+    assert h5_filepath_no_compression.exists()
+
+    with h5py.File(h5_filepath_compressed, "w") as f:
+        label_group = f.create_group("label_data")
+        slotSerializer_compressed.serialize(label_group)
+
+    assert h5_filepath_compressed.exists()
+
+    assert h5_filepath_compressed.stat().size * 10 < h5_filepath_no_compression.stat().size
```
emilmelnikov, line 581: I'm feeling a bit uneasy of such tests with approximate pass thresholds...

emilmelnikov (approved):
LGTM

tests/test_ilastik/test_applets/base/testSerializer.py
k-dominik: hmmm, Would you have another suggestion? It's not exactly straight forward to calculate the expected file size :/ But maybe a mere `<` then?

emilmelnikov: This is not exactly the same, but what about something like the following?

```python
compressed = True

def all_compressed(_name, obj):
    if isinstance(obj, h5py.Dataset) and obj.compression is None:
        compressed = False
        return False

with h5py.File(h5_filepath_no_compression, "w") as f:
    label_group = f.create_group("label_data")
    slotSerializer_no_compression.serialize(label_group)
    label_group.visititems(all_compressed)
    assert compressed
```

Ditto for the other case.

k-dominik: cool thanks, I am inspired already :)

k-dominik: opted for a slightly simpler version, but otherwise in line with your suggestion.

tests/test_ilastik/test_applets/base/testSerializer.py
```diff
@@ -535,6 +536,56 @@ def iter_dataset_names(name):
...
+    opLabelArrays.Input[0].setValue(raw_data)
+    opLabelArrays.shape.setValue(raw_data.shape)
+    opLabelArrays.eraser.setValue(255)
+    opLabelArrays.deleteLabel.setValue(-1)
+    opLabelArrays.blockShape.setValue((64, 64, 64, 1))
+    return opLabelArrays
+
+
+def testCompression(tmpdir, opLabelArray):
+    def has_uncompressed_dataset(obj_name, obj):
```
emilmelnikov, line 554: This is a nitpick, but:

```python
def uncompressed_dataset_name(name, obj):
    dataset_name_matches = name == "data" or name.startswith("block")
    if isinstance(obj, h5py.Dataset) and dataset_name_matches and obj.compression is None:
        return name
    return None
```

tests/test_ilastik/test_applets/base/testSerializer.py
```diff
@@ -535,6 +536,56 @@ def iter_dataset_names(name):
...
+
+    with h5py.File(h5_filepath_compressed, "w") as f:
+        label_group = f.create_group("label_data")
+        slotSerializer_compressed.serialize(label_group)
+        uncompressed = label_group.visititems(has_uncompressed_dataset)
+        assert uncompressed is None, f"Found unexpected uncompressed dataset {uncompressed}"
+
+    assert h5_filepath_compressed.exists()
+
+    assert h5_filepath_compressed.stat().size < h5_filepath_no_compression.stat().size
```
emilmelnikov, line 586: Do we still need this? Strictly speaking, compressed data could be larger (the maximum entropy plus compression bookkeeping case).

emilmelnikov (approved):

tests/test_ilastik/test_applets/base/testSerializer.py
k-dominik: thx, no we don't!

PR #2403, opened 2021-02-25
Stack selection improvements

Tomaz-Vieira:
Fixes error message of "variable used before assignment" when selecting a bad directory in the Add Stack dialogue. Also disables the Ok button when no files are selected.


k-dominik (approved):
lgtm

PR #2404, opened 2021-02-26
Fixes globbing issues

Tomaz-Vieira:
Currently ilastik will not allow using files whose names can be interpreted as globs. This  PR makes it so that every file added via the GUI is interpreted as-is, and adds the `--skip-deglobbing` command line flag to to the same in headless mode (since our documentation already instructs users to use globbing in headless mode).

TODO:
Add tests

### DIFF
--- a/ilastik/applets/dataSelection/dataSelectionApplet.py
+++ b/ilastik/applets/dataSelection/dataSelectionApplet.py
@@ -159,6 +159,14 @@ class DataSelectionApplet(Applet):
         )
 
         arg_parser.add_argument(
+            "--skip-deglobbing",
+            "--skip_deglobbing",
+            help="Interpret file names as is, without deglobbing characters like []*?",
+            action="store_true",
+            default=False,
+        )
+
+        arg_parser.add_argument(
             "--preconvert-stacks",
             "--preconvert_stacks",
             help="Convert image stacks to temporary hdf5 files before loading them.",
@@ -243,14 +251,18 @@ class DataSelectionApplet(Applet):
         return role_name.lower().replace(" ", "_").replace("-", "_")
 
     def create_dataset_info(
-        self, url: Union[Path, str], axistags: Optional[vigra.AxisTags] = None, sequence_axis: str = "z"
+        self,
+        url: Union[Path, str],
+        skip_deglobbing: bool = False,
+        axistags: Optional[vigra.AxisTags] = None,
+        sequence_axis: str = "z",
     ) -> DatasetInfo:
         url = str(url)
         if isUrl(url):
             return UrlDatasetInfo(url=url, axistags=axistags)
         else:
             return RelativeFilesystemDatasetInfo.create_or_fallback_to_absolute(
-                filePath=url, axistags=axistags, sequence_axis=sequence_axis
+                filePath=url, skip_deglobbing=skip_deglobbing, axistags=axistags, sequence_axis=sequence_axis
             )
 
     def convert_info_to_h5(self, info: DatasetInfo) -> DatasetInfo:
@@ -269,6 +281,7 @@ class DataSelectionApplet(Applet):
         preconvert_stacks: bool = False,
         ignore_training_axistags: bool = False,
         stack_along: str = "z",
+        skip_deglobbing: bool = False,
     ) -> List[Dict[str, DatasetInfo]]:
         if not input_axes or not any(input_axes):
             if ignore_training_axistags or self.num_lanes == 0:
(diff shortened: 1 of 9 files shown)


k-dominik:
I know it's a draft, so just a quick questions: the `--skip-deglobbing` parameter would only be applicable if there were a file with the exact name as the glob expression, right? So the situation is this, I have files that match the glob expression, but I also have this file with the same name, and I want to make sure that the single file is used, rather than the globbed ones. I wonder if the other way round wouldn't make more sense, since it is very unlikely that the situation above is the case.
So, in general, check if a file with the name from the cmd exists, if so, use it, if not, go globbing. Then  for this one, pathological case, where we'd actually have both present, I would add something like a `--force-globbing`... But I don't know. I just think that files that have glob like file names should be more common, than having both, file names with glob like file names, and more files that match this glob expression. With the `--force-globbing` flag, we'd cater for the more likely case.

Tomaz-Vieira:
So here are our options in regards to globbing in the command line, as discussed in the previous meeting:

## 1. Always deglob unless `--skip-deglobbing` is also passed in
### Pros:
- Backwards compatible;
- Unambiguous behavior with no magic, no special cases: it either is or isn't a glob.
### Cons:
- Potentially confusing;
- Yet another command line flag.

## 2. Always deglob, no special flags. Preventing deglobbing is done by escaping the special characters.
### Pros:
- Very consistent with the shell's globbing behavior;
- Unambiguous behavior with no magic, no special cases.
### Cons:
 - Escaping globs is hard and hard to read ( `glob.escape("file[1].tiff") == 'file[[]1].tiff' `)

## 3. Check if a file with a globlike name exists, otherwise deglob
### Pros:
 - It is what the user wants 95% of the time
### Cons:
 - Not strictly backwards compatible
 - Magic behavior; more branches to debug, harder to specify, more things to explain in the documentation and in help emails;
 - Pathological cases like having 4 files named `file[123].txt`, `file1.txt`, `file2.txt`and `file3.txt` and it being impossible to specify a stack of the files `file1.txt`, `file2.txt` and `file3.txt`

## 4. Deglob only if `--stack-along` is also present
### Pros:
 - Incentivises a sensible usage of `--stack-along` i.e., to always use it for stacks
### Cons:
 - Not backwards compatible (`--stack-along` is optional now)
 - One could argue that an image of shape `{x: 100, y: 100}` is a special/degenerate case of a 3D volume with `z: 1`, and that expanding a glob to a list of single file (like in the shell) is the correct behavior.

What do you guys thing?

m-novikov:
My vote is for 3rd option try filename and deglob.

k-dominik:
Option 3 ftw ;)

emilmelnikov:
I vote for 2, but disable glob expansion by setting a special environment variable instead of escaping and/or `skip_deglobbing`. This is not pretty, but it is very rarely required anyway.

k-dominik:
| option | votes |
|:---:|:---:|
|1|0|
|2|1|
|3|2|

so far it looks like option 3, but @Tomaz-Vieira has sneakily not voiced a strong opinion yet.

ilastik/utility/data_url.py
```diff
@@ -0,0 +1,522 @@
...
+import enum
+
+import numpy as np
+from typing_extensions import final
+import z5py
+import h5py
+
+from lazyflow.utility.pathHelpers import lsH5N5, globH5N5, globNpz
+
+# pyright: strict
```
k-dominik, line 18: looks like you introduce new tooling here. Is it comparable to mypy? Is it better?

Tomaz-Vieira: It's... complementary to mypy, I guess. It does some things better, some things worse. Having that comment there shouldn't hurt anyone, but I can mark that file to be strictly checked elsewhere (outside of the source code) if it's bothering you =)

ilastik/utility/data_url.py
```diff
@@ -0,0 +1,522 @@
...
+    def from_string(cls, value: str) -> "Scheme":
+        if value == "":
+            return Scheme.FILE
+        for item in cls:
+            if item.value == value:
+                return item
+        raise ValueError(f"Could not convert {value} to a valid Scheme")
+
+    @classmethod
+    def contains(cls, value: str) -> bool:
```
k-dominik, line 36: This method seems to be used only once, and there it could also be done by constructing in a `try... except``` with raise from there... 

I think it's a bit weird that his enables:

```Python
>>> x = Scheme("http")
>>> x.contains("https")
True
```

k-dominik:
Heyo, got a quick glimpse at the code which looks great and then decided, before going into details there, to look at it from a user perspective...

Some of the following issues might have been there before...

the updated stacking dialog:
1. it is easy to get into an inconsistent state there. If I select some files, then decide to do some globbing, unsuccessfully, it will still show the old files.
1. Why is there the string "New Dataset" in the list field? It could be in the label that says "Selection" right now.
1. now that sequence axis is a drop down, it should probably not have a default value. In my experience users might overlook it
1. The field for the separator is very long. Are _any_ separators supported? so example `myverylongcustomseparator`? Also, is this ever useful? What is the use-case? I think this complicates more than it helps (but of course am open to be proven wrong :)) Wasn't the goal to mimic shell behavior here?
1. the pattern field will correctly _not_ accept something like `/this/is/my/folder` (with `folder` being a folder) but will happily accept `/this/is/my.h5/group`, where `group` is a group.
1. adding a single file to the stack, will open an image, but not add the singleton axis. I'd expect to either
   * not open single images there
   * add the appropriate singleton axis
1. While we are at it, there should be a note in the stack import that this will save the data to the project file
1. import is "silent", no progress, no indication of busyness
1. the layout with the buttons could be improved: the "ok" button stretches weirdly
   stack

ilastik/utility/data_url.py
k-dominik: Do you check other parts of the code without the strict option? Or did you only check this one file at all? When would you have non-strict checks (strict sounds like the right thing to do? :D)...

In any case, maybe that is better suited for `pyproject.toml`?

PR #1910, opened 2018-12-10
Sync buttons in EdgeTrainingWithMulticut applet and fix lazy requesting for multiple lanes

paulhfu:
This fixes #1858, fixes #2085 and fixes #1842. 
- "Live Predict" and "Live Multicut" buttons are now synced with each other and across lanes.
- Having only red labeled edges does result in a edge training prediction with all edges predicted a 0 proba
- Requests are issued only for slots in currently active lane.

PR #2412, opened 2021-03-24
short circuit prediction if only one class is annotated

### LINKED ISSUE #2085: multicut edgetraining inconsistent label prediction
## Describe the bug
In multicut, doing edgetraining with red labeled edges only results in all predicted labels being green.

## Expected behavior
The predicted labels should all be red.

## To Reproduce
Within the multicutworkflow
1. Add some [2d/3d/4d/5d] data in the data selection applet
2. Go to "Training and Multicut" applet
3. Label some edges with red. Make sure that there is no green label in all lanes.
4. click "Live Predict"

 - OS: Windows 10
 
ilastik ilastik/ilastik@beeba50b5674ccacac799d1e9b3428785867cc11
lazyflow ilastik/lazyflow@8d58bc148a5a8f8588187fab4aa3bf644b678090
volumina ilastik/volumina@ae28b3162e7737f17513870766dda3b843041ed9

k-dominik:
Salvaged from PR #1910 (4ece7b62ccd3c73aa007915cdd282601102e6b61)
fixes #2085

if only one class is labeled, the matching probabilities are returned right away without going through prediction first.

CC: @paulhfu

I'd say we can close #1910 after this is merged

### DIFF
--- a/ilastik/applets/edgeTraining/opEdgeTraining.py
+++ b/ilastik/applets/edgeTraining/opEdgeTraining.py
@@ -411,8 +411,19 @@ class OpPredictEdgeProbabilities(Operator):
             classifier = self.EdgeClassifier.value
 
             # Classifier can be None if no labels have been selected
-            if classifier is None or len(classifier.known_classes) < 2:
+            if classifier is None or len(classifier.known_classes) == 0:
                 result[0] = np.zeros((len(edge_features_df),), dtype=np.float32)
+                return
+
+            known_classes = classifier.known_classes
+            assert set(known_classes).issubset([1, 2])
+            if len(known_classes) == 1:
+                logger.info("Returning edge probabilities for only annotated class...")
+                if known_classes[0] == 1:
+                    probabilities = np.zeros((len(edge_features_df),), dtype=np.float32)
+                else:
+                    probabilities = np.ones((len(edge_features_df),), dtype=np.float32)
+                result[0] = probabilities
                 return
 
             logger.info("Predicting edge probabilities...")
--- a/ilastik/applets/edgeTraining/opEdgeTraining.py
+++ b/ilastik/applets/edgeTraining/opEdgeTraining.py
@@ -398,48 +398,53 @@ class OpPredictEdgeProbabilities(Operator):
     TrainRandomForest = InputSlot(value=False)
     EdgeClassifier = InputSlot()
     EdgeFeaturesDataFrame = InputSlot()
+    # historically named Edge"Probabilities" because multicut used predictions from
+    # random forest as edge weights. With neural networks, edge probabilities are often already good
+    # enough to act as weights.
     EdgeProbabilities = OutputSlot()  # A 1D array of probabilities, in same order as EdgeFeaturesDataFrame
 
     def setupOutputs(self):
         self.EdgeProbabilities.meta.shape = (1,)
         self.EdgeProbabilities.meta.dtype = object
 
+    def _edge_weights_from_random_forest_predictions(self):
+        edge_features_df = self.EdgeFeaturesDataFrame.value
+        classifier = self.EdgeClassifier.value
+
+        # Classifier can be None if no labels have been selected
+        if len(getattr(classifier, "known_classes", [])) == 0:
+            return np.zeros(len(edge_features_df), dtype=np.float32)
+
+        known_classes = classifier.known_classes
+        assert set(known_classes).issubset([1, 2])
+        if len(known_classes) == 1:
+            logger.info("Returning edge probabilities for only annotated class...")
+            if known_classes[0] == 1:
+                return np.zeros(len(edge_features_df), dtype=np.float32)
+            else:
+                return np.ones(len(edge_features_df), dtype=np.float32)
+
+        logger.info("Predicting edge probabilities...")
+        feature_matrix = edge_features_df.iloc[:, 2:].values  # Discard [sp1, sp2]
+        assert feature_matrix.dtype == np.float32, "Unexpected feature dtype: {}".format(feature_matrix.dtype)
+        probabilities = classifier.predict_probabilities(feature_matrix)[:, 1]
+        return probabilities
+
+    def _edge_weights_from_probability_map(self):
+        BEST_FEATURE = "standard_edge_mean"
+        edge_features_df = self.EdgeFeaturesDataFrame.value
+        edge_features_df = edge_features_df.iloc[:, 2:]  # Discard columns [sp1, sp2]
+        edge_weights = edge_features_df[BEST_FEATURE]
+        return edge_weights
+
     def execute(self, slot, subindex, roi, result):
         pd.set_option("display.expand_frame_repr", False)
         if self.TrainRandomForest.value:
-            edge_features_df = self.EdgeFeaturesDataFrame.value
-            classifier = self.EdgeClassifier.value
-
-            # Classifier can be None if no labels have been selected
-            if classifier is None or len(classifier.known_classes) == 0:
-                result[0] = np.zeros((len(edge_features_df),), dtype=np.float32)
-                return
-
-            known_classes = classifier.known_classes
-            assert set(known_classes).issubset([1, 2])
-            if len(known_classes) == 1:
-                logger.info("Returning edge probabilities for only annotated class...")
-                if known_classes[0] == 1:
-                    probabilities = np.zeros((len(edge_features_df),), dtype=np.float32)
-                else:
-                    probabilities = np.ones((len(edge_features_df),), dtype=np.float32)
-                result[0] = probabilities
-                return
-
-            logger.info("Predicting edge probabilities...")
-            feature_matrix = edge_features_df.iloc[:, 2:].values  # Discard [sp1, sp2]
-            assert feature_matrix.dtype == np.float32, "Unexpected feature dtype: {}".format(feature_matrix.dtype)
-            probabilities = classifier.predict_probabilities(feature_matrix)[:, 1]
-
-            assert len(probabilities) == len(edge_features_df)
-
+            edge_weights = self._edge_weights_from_random_forest_predictions()
         else:
-            BEST_FEATURE = "standard_edge_mean"
-            edge_features_df = self.EdgeFeaturesDataFrame.value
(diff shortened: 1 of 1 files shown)


ilastik/applets/edgeTraining/opEdgeTraining.py
```diff
@@ -400,10 +400,21 @@ def execute(self, slot, subindex, roi, result):
...
 
             # Classifier can be None if no labels have been selected
-            if classifier is None or len(classifier.known_classes) < 2:
+            if classifier is None or len(classifier.known_classes) == 0:
                 result[0] = np.zeros((len(edge_features_df),), dtype=np.float32)
                 return
 
+            known_classes = classifier.known_classes
+            assert set(known_classes).issubset([1, 2])
+            if len(known_classes) == 1:
```
emilmelnikov, line 409: I think it would be easier to follow the control flow if we extract all logic into `_execute` helper method that checks various conditions and early-returns `probabilities`, instead of multiple nested conditions in the original `execute`:

```python
def _execute(self):
    if not self.TrainRandomForest.value:
        ...
        return classifier.predict_probabilities(feature_matrix)[:, 1]

    if classifier is None or ...:
        return np.zeros((len(edge_features_df),), dtype=np.float32)

    ...
    return some_default_value

def execute(self, ..., result):
    result[0] = self._execute()
```

ilastik/applets/edgeTraining/opEdgeTraining.py
```diff
@@ -400,10 +400,21 @@ def execute(self, slot, subindex, roi, result):
...
+                logger.info("Returning edge probabilities for only annotated class...")
+                if known_classes[0] == 1:
+                    probabilities = np.zeros((len(edge_features_df),), dtype=np.float32)
```
emilmelnikov, line 412: `np.zeros((n,))` == `np.zeros(n)`

ilastik/applets/edgeTraining/opEdgeTraining.py
```diff
@@ -400,10 +400,21 @@ def execute(self, slot, subindex, roi, result):
             classifier = self.EdgeClassifier.value
 
             # Classifier can be None if no labels have been selected
-            if classifier is None or len(classifier.known_classes) < 2:
+            if classifier is None or len(classifier.known_classes) == 0:
```
emilmelnikov, line 403:
```suggestion
            if getattr(classifier, "known_classes", 0) == 0:
```

ilastik/applets/edgeTraining/opEdgeTraining.py
```diff
@@ -387,37 +387,53 @@ class OpPredictEdgeProbabilities(Operator):
...
+                return np.zeros(len(edge_features_df), dtype=np.float32)
+            else:
+                return np.ones(len(edge_features_df), dtype=np.float32)
 
-            logger.info("Predicting edge probabilities...")
-            feature_matrix = edge_features_df.iloc[:, 2:].values  # Discard [sp1, sp2]
-            assert feature_matrix.dtype == np.float32, "Unexpected feature dtype: {}".format(feature_matrix.dtype)
-            probabilities = classifier.predict_probabilities(feature_matrix)[:, 1]
+        logger.info("Predicting edge probabilities...")
+        feature_matrix = edge_features_df.iloc[:, 2:].values  # Discard [sp1, sp2]
```
emilmelnikov, line 417: Is it possible to select the required columns for the feature matrix by their names?

`feature_matrix = edge_features_df[:, ["col1", "col2"]].values`

ilastik/applets/edgeTraining/opEdgeTraining.py
```diff
@@ -387,37 +387,53 @@ class OpPredictEdgeProbabilities(Operator):
...
+        assert feature_matrix.dtype == np.float32, "Unexpected feature dtype: {}".format(feature_matrix.dtype)
+        probabilities = classifier.predict_probabilities(feature_matrix)[:, 1]
+        return probabilities
```
emilmelnikov, line 420:
```suggestion
        return classifier.predict_probabilities(feature_matrix)[:, 1]
```

ilastik/applets/edgeTraining/opEdgeTraining.py
```diff
@@ -387,37 +387,53 @@ class OpPredictEdgeProbabilities(Operator):
...
+    def _edge_weights_from_random_forest_predictions(self):
+        edge_features_df = self.EdgeFeaturesDataFrame.value
+        classifier = self.EdgeClassifier.value
 
-            # Classifier can be None if no labels have been selected
-            if classifier is None or len(classifier.known_classes) < 2:
-                result[0] = np.zeros((len(edge_features_df),), dtype=np.float32)
-                return
+        # Classifier can be None if no labels have been selected
+        if len(getattr(classifier, "known_classes", [])) == 0:
```
emilmelnikov, line 404:
```suggestion
        if getattr(classifier, "known_classes", []):
```

I didn't suggest this before because IMO `if n == 0:` was better than `if not n:`.

ilastik/applets/edgeTraining/opEdgeTraining.py
```diff
@@ -387,37 +387,53 @@ class OpPredictEdgeProbabilities(Operator):
...
+            return np.zeros(len(edge_features_df), dtype=np.float32)
+
+        known_classes = classifier.known_classes
+        assert set(known_classes).issubset([1, 2])
```
emilmelnikov, line 408: What are the possible values of `known_classes`? Is it one of `{1}`, `{2}`, and `{1, 2}`?

If yes, I think it'd be better to:

```python
known_classes = set(classifier.known_classes)
if known_classes == {1}:
    return np.zeros(...)
if known_classes == {2}:
    return np.ones(...)
if known_classes == {1, 2}:
    return ...
raise AssertionError(...)
```

ilastik/applets/edgeTraining/opEdgeTraining.py
```diff
@@ -387,37 +387,53 @@ class OpPredictEdgeProbabilities(Operator):
...
-            assert len(probabilities) == len(edge_features_df)
+    def _edge_weights_from_probability_map(self):
+        BEST_FEATURE = "standard_edge_mean"
+        edge_features_df = self.EdgeFeaturesDataFrame.value
+        edge_features_df = edge_features_df.iloc[:, 2:]  # Discard columns [sp1, sp2]
+        edge_weights = edge_features_df[BEST_FEATURE]
+        return edge_weights
 
+    def execute(self, slot, subindex, roi, result):
+        pd.set_option("display.expand_frame_repr", False)
```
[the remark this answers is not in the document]
k-dominik: Ha, I was really afraid to change that, because some part of the code might implicitly rely on it. But at some point we have to find out :D

ilastik/applets/edgeTraining/opEdgeTraining.py
```diff
@@ -387,37 +387,53 @@ class OpPredictEdgeProbabilities(Operator):
...
+        probabilities = classifier.predict_probabilities(feature_matrix)[:, 1]
+        return probabilities
 
-            assert len(probabilities) == len(edge_features_df)
+    def _edge_weights_from_probability_map(self):
+        BEST_FEATURE = "standard_edge_mean"
+        edge_features_df = self.EdgeFeaturesDataFrame.value
+        edge_features_df = edge_features_df.iloc[:, 2:]  # Discard columns [sp1, sp2]
+        edge_weights = edge_features_df[BEST_FEATURE]
+        return edge_weights
```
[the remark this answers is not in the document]
k-dominik: I'll do a variation of that... I like the message of calling it `BEST_FEATURE` :)

ilastik/applets/edgeTraining/opEdgeTraining.py
k-dominik: we do the names we want to exclude, the ones we want to include are not known in this operator (=all)

ilastik/applets/edgeTraining/opEdgeTraining.py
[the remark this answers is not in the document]
emilmelnikov: Disabling this option makes sense in non-interactive "non-notebook" settings. Maybe we should set this globally during ilastik initialization?

ilastik/applets/edgeTraining/opEdgeTraining.py
emilmelnikov: Then we could discard them with `edge_features_df.drop(columns=("sp1", "sp2")).values`.

ilastik/applets/edgeTraining/opEdgeTraining.py
k-dominik: with `if n`.. I'll have to reverse the logic. I think I'll leave it with `len`.

ilastik/applets/edgeTraining/opEdgeTraining.py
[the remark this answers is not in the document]
k-dominik: hmm I wonder if it's needed here at all...

ilastik/applets/edgeTraining/opEdgeTraining.py
emilmelnikov: Now I see an obvious refactoring:

```python
known_classes = set(getattr(classifier, "known_classes", []))
if known_classes == {}:
    ...
elif known_classes == {1}:
    ...
...
```

k-dominik: niiiice! :)

k-dominik: side note: `{}` will be an empty dict and `!=set()`

emilmelnikov: Another color for the bike shed:

```python
known_classes = sorted(getattr(classifier, "known_classes", []))
if known_classes in ([], [1]):
    ...
elif known_classes  == [2]:
    ...
elif known_classes == [1, 2]:
    ...
else:
    ...
```

This fails if `known_classes` has duplicates, which is probably good.

k-dominik: `known_classes` should not have duplicates. :)

Issue #2414, opened 2021-03-25
Multicut: only enable "live multicut" when training a RF


k-dominik:
Having training the RF optional, "live multicut" does not always make sense anymore. It should be disabled (or hidden) if RF is not trained but bondary probs are used directly.

PR #2415, opened 2021-03-25
fix doing "live" multicut without trained classifier

k-dominik:
quick fix for a problem I found using multicut (didn't bother to open an issue).

clicking "live multicut" before training a classifier would result in an error.

We should probably be more user-friendly here, too, and not have that button enabled if RF is not trained (https://github.com/ilastik/ilastik/issues/2414).


emilmelnikov (approved):
LGTM, but maybe move this just after the `if node_labels is None` check.

PR #2417, opened 2021-04-07
restore live update value after updating multicut

k-dominik:
That's a brief one.

before, live update would be turned off silently leaving the button in an inconsistent state when "update multicut" was pressed.

### DIFF
--- a/ilastik/applets/edgeTrainingWithMulticut/edgeTrainingWithMulticutGui.py
+++ b/ilastik/applets/edgeTrainingWithMulticut/edgeTrainingWithMulticutGui.py
@@ -1,5 +1,6 @@
 from PyQt5.QtWidgets import QWidget, QVBoxLayout, QGroupBox, QSpacerItem, QSizePolicy, QCheckBox
 
+from lazyflow.slot import valueContext
 from ilastik.applets.edgeTraining.edgeTrainingGui import EdgeTrainingMixin
 from ilastik.applets.multicut.multicutGui import MulticutGuiMixin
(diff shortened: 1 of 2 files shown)


emilmelnikov:
I wonder if this is the time to add a method for this pattern in `Slot`?..

```python
@contextmanager
def valueContext(self, value):
    """Temporarily change value in the *with* statement context, yielding an old value."""
    old = self.value
    self.setValue(value)
    try:
        yield old
    finally:
        self.setValue(old)
```

emilmelnikov (commented):
I wonder if this is the time to add a method for this pattern to `Slot`?..

```python
@contextmanager
def valueContext(self, value):
    """Temporarily change value in the *with* statement context, yielding an old value."""
    old = self.value
    self.setValue(value)
    try:
        yield old
    finally:
        self.setValue(old)
```

k-dominik:
> I wonder if this is the time to add a method for this pattern to Slot?..

I really wish we had a separate `ValueSlot` (sub)class. Then it would be a no-brainer to add it. Right now I am a little concerned about side effects doing this on a "regular" slot. Doing `self.value` could trigger a heavy/costly computation. I don't know if it's too inviting to use it, once it's there.

emilmelnikov:
> Right now I am a little concerned about side effects doing this on a "regular" slot. Doing `self.value` could trigger a heavy/costly computation. I don't know if it's too inviting to use it, once it's there.

We could move this thing out of the class: a helper function seems less inviting to use.

PR #2418, opened 2021-04-13
Fix OpMultiArraySlicer2 dirty to early

### LINKED ISSUE #2239: dirty ROI - Could not load project file
## Describe the bug
Failure to load a previously saved project for pixel classification. 

Error message:  Could not load project file. The dirty ROI (Subregion:start '[0,0,0]' stop '[1,3372,8820']) appears incorrect for the InputSlot, whihc has shape (1,1,3372,8820,1).

This is particularly frustrating as can now not access the classifier at all. 

## To Reproduce
Its possible this occurred following setting the export settings to export just the prediction (and not the raw data) as a tif ( having to change some of the setting there). 

 - ilastik version:1.3.3(post2)
 - OS: Windows 7

k-dominik:
Removed calls to `setDirty` for outputslots in `setupOutputs` of `OpMultiArraySlicer2`. I can not think of a reason why this was there in the first place. `Operator._setupOutputs` has the necessary logic in place.

~~while I couldn't exactly reproduce the reported issue~~, I could create a similar one when enabling users to add singleton axis in export (see https://github.com/ilastik/volumina/pull/263).
Update: I managed to reproduce the original issue - with a bit of trickery. I also made sure this does not happen anymore with this branch/pr:
1. In pixel classification, use 5d dataset with singletons in `t` and `z`.
2. train a classifier
3. In export: remove axes `t` and `z` (or just one of them), when closing the dialog, you'll get an error. Reopen the dialog and change some other setting (e.g. normalization), accept again, save the project.

this produces a project that results in #2239 ...

Fixes #2239 

I'm not sure this causes any side effects. Touching code that was in place for so long... So far I have not encountered any side effects that I could reproduce with this branch.

### DIFF
--- a/tests/test_lazyflow/test_operators/testOpMultiArraySlicer2.py
+++ b/tests/test_lazyflow/test_operators/testOpMultiArraySlicer2.py
@@ -25,7 +25,7 @@ from builtins import object
 from functools import partial
 import numpy
 import vigra
-from lazyflow.graph import Graph, OperatorWrapper
+from lazyflow.graph import Graph, OperatorWrapper, InputSlot, OutputSlot, Operator
 from lazyflow.rtype import SubRegion
 from lazyflow.operators import OpArrayPiper, OpMultiArraySlicer2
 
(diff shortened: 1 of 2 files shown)


tests/test_lazyflow/test_operators/testOpMultiArraySlicer2.py
```diff
@@ -193,3 +193,38 @@ def testWrapped(self):
...
+                # this was encountered when setDirty was called in setupOutputs
+                if not (len(in_roi.start) == len(in_roi.stop) == len(self.Input.meta.shape)):
+                    self.invalid_prob_dirty_count += 1
+
+        test_op = DirtyCountingOp(graph=self.graph)
+        test_op.Input.connect(self.opSlicer.Slices[0])
+        assert test_op.invalid_prob_dirty_count == 0
+        assert test_op.prob_dirty_count == 0
+
+        data = self.opProvider.Input.value
```
emilmelnikov, line 226: Does it make sense to test with a fresh `DirtyCountingOp` here?

tests/test_lazyflow/test_operators/testOpMultiArraySlicer2.py
```diff
@@ -193,3 +193,38 @@ def testWrapped(self):
...
+        reshaped = data.withAxes("tzyxc")
+        self.opProvider.Input.setValue(reshaped)
+        assert test_op.invalid_prob_dirty_count == 0
+        assert test_op.prob_dirty_count == 1
```
emilmelnikov, line 230: `invalid_prob_dirty_count` doesn't change anywhere. Should we add another "arrange, act, assert" with Input having a different ndim?

emilmelnikov (approved):

tests/test_lazyflow/test_operators/testOpMultiArraySlicer2.py
k-dominik: do you mean why does it even do the asserts before?

tests/test_lazyflow/test_operators/testOpMultiArraySlicer2.py
k-dominik: it does not change _now_, before it would've. I think that's one of the main points of the test.

emilmelnikov: Ah, OK, now I understand what is being actually tested.

tests/test_lazyflow/test_operators/testOpMultiArraySlicer2.py
emilmelnikov: Nevermind, now I see that the test is supposed to reused `test_op`.

k-dominik:
thx for the review @emilmelnikov

PR #2419, opened 2021-04-14
PC + OC workflow: fix loading projects

### LINKED ISSUE #2244: Labels lost on loading
## Describe the bug
Save, close, and then reopen a Pixel Classification + Object Classification project at step 6/7/8. Now the first labelled (object classified) image becomes unlabelled. If you leave it like this, save and reopen again, then the first labelled image (which is the second image) will become unlabelled.

## Expected behaviour
When a user reopens a saved project, everything in the project stays the same as it was.

## To Reproduce
Steps to reproduce the behaviour:
Within the Pixel Classification + Object Classification workflow
1. Add some 2d data in the data selection applet
2. Train the pixel classifier to segment the objects out
3. Train the object classifier by labelling 5 classes
4. Save the project at "object classification" applet or "object information export" applet
5. Quit ilastik and reopen ilastik
6. Load the saved project
7. "object classification" applet automatically shows up and the first image with binary image masked also shows up. But now the labels are all gone

 - ilastik version: ilastik-1.3.3post2-Linux
 - Ubuntu 18.04

k-dominik:
PixelClassification + Object Classification, while not recommended, should work properly. Right now projects are not loaded correctly (see #2244)

this prevents dirty notifications bubbling up. In the combined workflow, the pixel classification classifier is set after the object classifier is added.
Hence, it is marked as dirty and discarded. Even worse, it sets the segmentation dirty, which in turn discards the object annotations.

Setting classifiers "silently" is exactly what you want there. No dirty notifications bubbling ~~up~~ down on added lanes.

fixes #2244

### DIFF
--- a/ilastik/workflows/objectClassification/objectClassificationWorkflow.py
+++ b/ilastik/workflows/objectClassification/objectClassificationWorkflow.py
@@ -253,7 +253,7 @@ class ObjectClassificationWorkflow(Workflow):
         """
         if self.stored_object_classifier:
             opObjectClassification = self.objectClassificationApplet.topLevelOperator
-            opObjectClassification.classifier_cache.forceValue(self.stored_object_classifier)
+            opObjectClassification.classifier_cache.forceValue(self.stored_object_classifier, set_dirty=False)
             # Release reference
             self.stored_object_classifier = None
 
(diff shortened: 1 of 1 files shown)


Tomaz-Vieira:
Great that you found a fix for that. Seems to me, though, that we shouldn't be firing dirty notifications at all on added lanes (either that or just outright ignoring dirtyness-by-way-of-new-lanes). In fact, "adding lanes and not changing anything on the graph" is exactly what headless mode does. Do you have any idea if that is feasible?

k-dominik:
> adding lanes and not changing anything on the graph

I can think already of cases where you need dirtyness when adding lanes (e.g. something relating to statistics over all lanes, ...), not that we have it right now - anyhow, those are special cases I think, and in the general case, there shouldn't be dirtiness involved.

In the current architecture, it is simply not foreseen to temporarily remove certain signals, hence the workaround here.

Issue #1033, opened 2014-11-07
PixelClassification: Segmentation layer should not show anything on dirty slices


martinsch:
To reproduce:
1. Label and predict on one slice/timestep.
2. Turn on live prediction
3. Set segmentation layers to visible
4. Turn off live prediction
5. Go to a different slice/timestep => the segmentation is all green/red. This looks wrong, I would expect an empty image on those layers.

stuarteberg:
The problem is that the segmentation layers are not produced with a cache: They are always computed on-demand using cached *input*.  When the cached input is zero across all channels, the output is undefined.

See [here](https://github.com/ilastik/ilastik/blob/f1fc32d75c523556816dd8a1de0b8d991a39695a/ilastik/applets/pixelClassification/opPixelClassification.py#L514)

Solution could be to have `OpMaxChannelIndicatorOperator` check for the special case of all-zero input, and output all-zeros in that case.

PR #2423, opened 2021-04-20
Simplify logic in ArgMaxChannelIndicator; get rid of transparent pixels

### LINKED ISSUE #2025: White/Transparent pixels in prediction (3rd class with 2 defined classes)
## Describe the bug

Using the release version `1.3.2post1-win64.exe` on Win 10:

 With only two classes defined (label colors yellow/green), the segmentation output 
shows a few white (or transparent) pixels. When I turn on the visibility for the "raw image"
the pixel values of the raw image show through at these locations.  Possibly this is related to https://github.com/ilastik/ilastik/issues/72 ?

image

## Expected behavior
* Each pixel belongs to one of the two defined classes
* raw image data not visible at any pixel location if segmentation view is enabled

## To Reproduce
Just labelled a volume with 2 classes

k-dominik:
* removed logic that would return `0` pixels for tied predictions.
* unified behavior with the Simple Segmentation export slot (no `0` pixels there).
* made sure not to reintroduce #1033

fixes #2025

### DIFF
--- a/lazyflow/operators/generic.py
+++ b/lazyflow/operators/generic.py
@@ -39,7 +39,7 @@ import vigra
 # lazyflow
 from lazyflow.graph import Operator, InputSlot, OutputSlot
 from lazyflow import roi
-from lazyflow.roi import roiToSlice, sliceToRoi, TinyVector, getIntersection
+from lazyflow.roi import roiToSlice, sliceToRoi, TinyVector, getIntersection, InvalidRoiException
 from lazyflow.request import RequestPool
 
 
@@ -533,6 +533,9 @@ class OpMaxChannelIndicatorOperator(Operator):
 
     def execute(self, slot, subindex, roi, result):
         key = roi.toSlice()
+        n_channels_requested = key[-1].stop - key[-1].start
+        if n_channels_requested != 1:
+            raise InvalidRoiException(f"This operator only accepts slices of size 1 for c! Got {n_channels_requested}.")
         data = self.inputs["Input"][key[:-1] + (slice(None),)].wait()
 
         # special case, when data is all zeros (e.g. directly from frozen cache w/o trained classifier)
--- a/lazyflow/operators/generic.py
+++ b/lazyflow/operators/generic.py
@@ -520,6 +520,8 @@ class OpMaxChannelIndicatorOperator(Operator):
     Produces a bool image where each value is either 0 or 1, depending on whether
     or not that channel of the input is the max value at that pixel compared
     to the other channels.
+
+    Note: it is expected that Output Rois are always single channel.
     """
 
     Input = InputSlot()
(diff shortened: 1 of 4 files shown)


lazyflow/operators/generic.py
```diff
@@ -535,19 +535,13 @@ def execute(self, slot, subindex, roi, result):
...
-        # We could use numpy.concatenate():
-        #  tied_pixels = numpy.concatenate((tied_pixels, tied_pixels), axis=-1)
-        # But that's not as fast as duplicating the array with a stride trick.
-        # (This only works because channel is the last axis.)
-        tied_pixels = numpy.lib.stride_tricks.as_strided(tied_pixels, shape=res.shape, strides=tied_pixels.strides)
-        res[tied_pixels] = 0
+        res = numpy.zeros(data.shape, numpy.uint8)
+        numpy.put_along_axis(res, numpy.expand_dims(numpy.argmax(data, axis=-1), axis=-1), 1, axis=-1)
 
         result[:] = res[..., key[-1]]
```
emilmelnikov, line 546: Assuming that we always have pointwise channel indexing here (that is, if `key[-1]` is int, and not slice):

```suggestion
        result[:] = (numpy.argmax(data, axis=-1) == key[-1]).astype(numpy.uint8)
```

lazyflow/operators/generic.py
```diff
@@ -535,19 +535,13 @@ def execute(self, slot, subindex, roi, result):
         key = roi.toSlice()
         data = self.inputs["Input"][key[:-1] + (slice(None),)].wait()
```
emilmelnikov, line 536:
```suggestion
        *key, c = roi.toSlice()
        data = self.Input[(*key, slice(None))].wait()
```

... and use `c` instead of `key[-1]`.

lazyflow/operators/generic.py
```diff
@@ -535,19 +535,13 @@ def execute(self, slot, subindex, roi, result):
...
 
-        dm = numpy.max(data, axis=data.ndim - 1, keepdims=True)
-        res = numpy.zeros(data.shape, numpy.uint8)
-        numpy.equal(data, dm, out=res)
-
-        # Special case: If all channels are tied, then none of them win.
-        tied_pixels = numpy.logical_and.reduce(res, axis=-1, keepdims=True, dtype=numpy.uint8)
+        # special case, when data is all zeros (e.g. directly from frozen cache w/o trained classifier)
```
emilmelnikov, line 538: Is it significantly faster compared to the argmax branch?

k-dominik: problem is the argmax branch will return `1`s, which will be rendered as solid color

emilmelnikov: That's right, sorry, got confused by the previous "not fast enough" comments.

But what if the input just happens to be all zeroes?

Also, what if, later, the input changes from all-zeroes to a single non-zero pixel? Suddenly the output jumps from all-zeroes to almost-all-ones.

k-dominik: when there are probabilities we assume they sum to 1. So if there is all zeros, we draw from cache. If there is one non-zero value, you can assume you have probabilities.

k-dominik:
thx for the review @emilmelnikov. This op should probably be moved and renamed. generic... if something does not accept rois of the full shape... Also it seems to be tightly bound to gui functionality.

lazyflow/operators/generic.py
```diff
@@ -532,24 +534,20 @@ def setupOutputs(self):
...
 
     def execute(self, slot, subindex, roi, result):
-        key = roi.toSlice()
-        data = self.inputs["Input"][key[:-1] + (slice(None),)].wait()
+        *key, c = roi.toSlice()
 
-        dm = numpy.max(data, axis=data.ndim - 1, keepdims=True)
-        res = numpy.zeros(data.shape, numpy.uint8)
-        numpy.equal(data, dm, out=res)
+        n_channels_requested = c.stop - c.start
```
emilmelnikov, line 539: Is it possible for our ROIs to have `c.start is None` or `c.stop is None`?

emilmelnikov (approved):

lazyflow/operators/generic.py
k-dominik: thx to your type annotations I can confidently say: no :)
```Python
def roiToSlice(start: Sequence[numbers.Integral], stop: Sequence[numbers.Integral])
```

PR #2430, opened 2021-05-19
make sure to pass project file path to dataset info

### LINKED ISSUE #2429: Relative paths stored relative to cwd :o // Relative paths for image files not possible in Windows
## Describe the bug
I think this is just an oversight, fixing it seems straight forward. But it opens up deeper questions, like why would we ever store project data relative to the cwd, or why is this fallback even there?

One side effect of this is, that for windows users, that have installed ilastik in the default location, it will never (well, the data is usually never further down in the hierarchy wrt the ilastik installation on `C:\Program Files....`

## Expected behavior
Links to data should be relative to the project file

## To Reproduce
just store your project in a location different from the cwd for ilasitk, and observe the behavior in the data selection applet.

 - ilastik version: 1.4.0x

k-dominik:
this makes sure links are relative to the project file, not cwd

fixes #2429

### DIFF
--- a/ilastik/applets/dataSelection/dataSelectionApplet.py
+++ b/ilastik/applets/dataSelection/dataSelectionApplet.py
@@ -250,7 +250,10 @@ class DataSelectionApplet(Applet):
             return UrlDatasetInfo(url=url, axistags=axistags)
         else:
             return RelativeFilesystemDatasetInfo.create_or_fallback_to_absolute(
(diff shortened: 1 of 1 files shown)


k-dominik:
okay, as cleared up in an "offline" discussion with @Tomaz-Vieira, this code path will be gone once #2404 is merged. Hence will not investigate further whether it makes sense to have paths related to pwd at all or not.

PR #2444, opened 2021-07-08
Neural Network Classification workflow cleanup

k-dominik:
Summary:

* This sets the classification applet up as a "predict" only for now.
* initially hides prediction layers in export (would start predicting there right away otherwise)
* model "upload" rephrased to model "initialization"

NN Drawer:

before:
before

after:

after

### DIFF
--- a/ilastik/applets/neuralNetwork/nnClassApplet.py
+++ b/ilastik/applets/neuralNetwork/nnClassApplet.py
@@ -44,9 +44,9 @@ class NNClassApplet(StandardApplet):
 
         self._topLevelOperator.classifier_cache.Output.notifyDirty(on_classifier_changed)
 
-        applet_name = ALLOW_TRAINING and "NN Training" or "NN Prediction"
+        applet_name = "NN Training" if ALLOW_TRAINING else "NN Prediction"
 
-        super(NNClassApplet, self).__init__(applet_name, workflow=workflow)
+        super().__init__(applet_name, workflow=workflow)
 
         self._serializableItems = [NNClassificationSerializer(self.topLevelOperator, projectFileGroupName)]
         self._gui = None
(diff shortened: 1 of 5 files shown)


ilastik/applets/neuralNetwork/nnClassApplet.py
```diff
@@ -44,7 +44,9 @@ def on_classifier_changed(slot, roi):
 
         self._topLevelOperator.classifier_cache.Output.notifyDirty(on_classifier_changed)
 
-        super(NNClassApplet, self).__init__("NN Training", workflow=workflow)
+        applet_name = ALLOW_TRAINING and "NN Training" or "NN Prediction"
```
emilmelnikov, line 47: Nitpick, but I think this reads somewhat better:
```suggestion
        applet_name = "NN Training" if ALLOW_TRAINING else "NN Prediction"
```

ilastik/applets/neuralNetwork/nnClassApplet.py
```diff
@@ -44,7 +44,9 @@ def on_classifier_changed(slot, roi):
...
+
+        super(NNClassApplet, self).__init__(applet_name, workflow=workflow)
```
emilmelnikov, line 49:
```suggestion
        super().__init__(applet_name, workflow=workflow)
```

ilastik/applets/neuralNetwork/nnClassGui.py
```diff
@@ -413,6 +413,21 @@ def validationMenu():
 
         return menus
 
+    @threadRouted
+    def enableLabellingUI(self, enabled: bool = False):
```
emilmelnikov, line 417: For consistency with the Qt-style `setThing` methods:
```suggestion
    def setLabelingUIEnabled(self, enabled: bool):
```

emilmelnikov (approved):

PR #2450, opened 2021-07-13
Handle outputshape returned by server

k-dominik:
Previously, the output shape was hard-coded. This breaks many networks ;).

Derive output shape from model information  returned by server.

### DIFF
--- a/lazyflow/operators/tiktorch/classifier.py
+++ b/lazyflow/operators/tiktorch/classifier.py
@@ -21,11 +21,8 @@
 ###############################################################################
 import logging
 import socket
-import threading
 import numpy
 import warnings
-import numpy as np
-from concurrent.futures import CancelledError
 from typing import Iterable, List, Callable
 
 import vigra
@@ -90,6 +87,9 @@ class ModelSession:
         scale_by_name = {d.name: d.size for d in self.scale}
 
         result = []
+        # FIXME: currently we don't make use of having potentially multiple valid
+        # output shapes. It can make sense, to choose it dynamically, e.g.
+        # a smaller one for "live" prediction, a larger one for batch
         for shape in self.__session.validShapes:
             dim_size_by_name = {d.name: d.size for d in shape.dims}
             valid_shape = {}
@@ -124,7 +124,7 @@ class ModelSession:
 
     @property
     def training_shape(self):
-        logger.warning("HARDCODED training shape, this might not do what you want")
+        warnings.warn("HARDCODED training shape, this might not do what you want.")
         return [0, 0, 0, 128, 128]
 
     @property
--- a/lazyflow/operators/tiktorch/classifier.py
+++ b/lazyflow/operators/tiktorch/classifier.py
@@ -74,6 +74,44 @@ class ModelSession:
         return self.__session.outputAxes
 
     @property
+    def offset(self):
+        return self.__session.offset
+
+    @property
+    def scale(self):
+        return self.__session.scale
+
+    @property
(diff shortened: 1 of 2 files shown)


lazyflow/operators/tiktorch/classifier.py
```diff
@@ -73,6 +73,44 @@ def input_axes(self):
...
+    @property
+    def offset(self):
+        return self.__session.offset
+
+    @property
+    def scale(self):
+        return self.__session.scale
+
+    @property
+    def output_shapes(self):
```
k-dominik, line 85: okay not sure what i wanted with that one...

lazyflow/operators/tiktorch/classifier.py
```diff
@@ -97,11 +135,13 @@ def get_valid_shapes(self, axes: str = "zyx"):
 
     @property
     def training_shape(self):
+        logger.warning("HARDCODED training shape, this might not do what you want")
```
emilmelnikov, line 138: Maybe use `warnings.warn` here?

lazyflow/operators/tiktorch/classifier.py
```diff
@@ -73,6 +73,44 @@ def input_axes(self):
...
+    @property
+    def output_shape(self):
+        """
+        shape = shape(input_tensor) * scale + 2 * offset
+        """
+        offset_by_name = {d.name: d.size for d in self.offset}
+        scale_by_name = {d.name: d.size for d in self.scale}
+
+        result = []
+        for shape in self.__session.validShapes:
```
emilmelnikov, line 97: We don't need to loop through everything if only `result[0]` is returned (unless there's something else going on here).

k-dominik: that's True. Honestly, I'm not sure how to best handle this. The model will give a list of valid shapes, and on the ilastik side, we might choose one depending on whether we do batch prediction, interactive prediction and so on. I implemented it like this in order to be able to switch quickly once we figure out what we want to do with it.

emilmelnikov: Fair point. I think adding a FIXME or some other comment about this would be nice.

lazyflow/operators/tiktorch/classifier.py
```diff
@@ -73,6 +73,33 @@ def input_axes(self):
...
+            dim_size_by_name = {d.name: d.size for d in shape.dims}
+            valid_shape = {}
+            for dim in dim_size_by_name:
+                size = int(dim_size_by_name.get(dim, 1) * scale_by_name.get(dim, 1.0) + 2 * offset_by_name.get(dim, 0))
+                valid_shape[dim] = size
+
+            result.append(valid_shape)
+        return result[0]
```
emilmelnikov: This has a nested comprehension, but I think it shows the resulting data structure a bit better.
```suggestion
        offsets = defaultdict({d.name: d.size for d in self.offset}, 0)
        scales = defaultdict({d.name: d.size for d in self.scale}, 1.0)

        output_shapes = [
            {
                dim: int(dim_size * scales[dim] + 2 * offsets[dim])
                for dim, dim_size in shape.dims.items()
            }
            for shape in self.__session.validShapes
        ]
        return output_shapes[0]
```

k-dominik: Hm, I have to admit that for me the current version is easier to parse. But I'm gonna add the `defaultdicts` for `scales` and `offsets`!

PR #2451, opened 2021-07-13
adds some margin around the startup screen logo

k-dominik:
that will be controversial ;) I didn't like how crammed our startup screen looks, also the fadout of the logo looked questionable.

Before:
startup_before

After: 
\
startup-after


emilmelnikov:
That' great! However, is it possible to add margins via Qt?

k-dominik:
well, if we want to use the "proper" fading out style logo then we have to use this image. Further margins, could be added with qt, yes, but I don't think that is necessary.

PR #2458, opened 2021-07-17
Prevent Ctrl+A shortcut toggling 2d/3d state in feature table

m-novikov:
out

### DIFF
--- a/tests/test_ilastik/widgets/test_featureTableWidget.py
+++ b/tests/test_ilastik/widgets/test_featureTableWidget.py
@@ -1,0 +1,43 @@
+from unittest import mock
+
+import pytest
+from PyQt5.QtCore import Qt
+
+from ilastik.widgets.featureTableWidget import FeatureTableWidget, FeatureEntry
+
+
+@pytest.fixture
+def widget(qtbot):
+    w = FeatureTableWidget(
+        None,
+        (
+            ("Color", [FeatureEntry("Banana", minimum_scale=0.3)]),
+            ("Edge", [FeatureEntry("Mango"), FeatureEntry("Cherry")]),
+        ),
+        [0.3, 0.7, 1, 1.6, 3.5, 5.0, 10.0],
+        computeIn2d=[True, False, True, True, True, False, False],
+    )
+    qtbot.addWidget(w)
+    w.show()
+
+    qtbot.waitForWindowShown(w)
+
+    return w
+
+
+def test_ctrl_a_shortcut_should_select_everything(qtbot, widget: FeatureTableWidget):
+    assert widget.featureMatrix == [
+        [False, False, False, False, False, False, False],
+        [False, False, False, False, False, False, False],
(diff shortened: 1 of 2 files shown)


tests/test_ilastik/widgets/test_featureTableWidget.py
```diff
@@ -0,0 +1,43 @@
...
+@pytest.fixture
+def widget(qtbot):
+    w = FeatureTableWidget(
+        None,
+        (
+            ("Color", [FeatureEntry("Banana", minimum_scale=0.3)]),
+            ("Edge", [FeatureEntry("Mango"), FeatureEntry("Cherry")]),
+        ),
+        [0.3, 0.7, 1, 1.6, 3.5, 5.0, 10.0],
+        computeIn2d=[True, True, True, True, True, True, True],
```
k-dominik, line 18: For the `computeIn2d` it would be interesting to have at least _some_ `False` values in place. Previous incarnation of "Select all" would set them all to `True`

k-dominik (commented):
This is awesome! I hope I'll unlearn the inhibition of `Ctrl+A` on 3D data soon :)

tests/test_ilastik/widgets/test_featureTableWidget.py
m-novikov: Updated to use a mix of the feature dimensionality.

m-novikov:
> Side question: what did you use for the screencap?

vokosreenNg + screenkey 

> This is awesome! I hope I'll unlearn the inhibition of Ctrl+A on 3D data soon :)

I also thought about selecting the row/column by clicking on the header. What do you think? (Not in the scope of this PR)

k-dominik (approved):
LGTM, thx a lot!

k-dominik:
> > Side question: what did you use for the screencap?
> 
> vokosreenNg + screenkey
>
thx!

> > This is awesome! I hope I'll unlearn the inhibition of Ctrl+A on 3D data soon :)
> 
> I also thought about selecting the row/column by clicking on the header. What do you think? (Not in the scope of this PR)

At first thought this definitely sounds like something nice to have. On the other hand we usually recommend using all... I don't have a good intuition whether this is behavior users would find on their own (then I'd say really good to have).

emilmelnikov:
What do you think about 
1. adding explicit `Select All` and `Deselect All` buttons to the dialog?
2. selecting all features by default?

I did this in napari-ilastik, and it seems convenient.

k-dominik:
> What do you think about
> 
>     1. adding explicit `Select All` and `Deselect All` buttons to the dialog?
> 
>     2. selecting all features by default?
> 
> 
> I did this in napari-ilastik, and it seems convenient.

It sure seems convenient, but I think there was a "didactic" reason to have users select something in the first place, right @akreshuk? Also with "nothing" selected, it is consistent throughout ilastik. Take e.g. Autocontext with 3D data, there selecting all in both stages will most likely break.

PR #2459, opened 2021-07-21
Fix notebook ex

k-dominik:
This (hopefully) fixes the notebook example by adding missing package and a note about strict channel priority.

Also added setting up conda with strict channel priority as a first setup step. It's not default yet, but I think we've been using it for so long, not realizing that this is something worth mentioning.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,10 +29,16 @@ The following text equips you with knowledge that makes contributing to ilastik 
    gh repo clone ilastik
    ```
 
+1. Configure conda to use [strict channel priority](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-channels.html#strict-channel-priority):
+
+   ```
+   conda config --set channel_priority strict
+   ```
+
 1. Install [conda-build][conda-build] in order to access the [conda develop][conda-develop] command:
 
    ```
-   conda install --name base conda-build
+   conda install --name base -c conda-forge conda-build
    ```
 
 1. Create a new environment and install dependencies:
--- a/notebooks/Readme.md
+++ b/notebooks/Readme.md
@@ -12,7 +12,7 @@ notebooks/
 ├── ...
 ```
 
-We use _conda_ to develop Python and recommend it for scientific Python development.
+We use _conda_ for Python-based projects and recommend it for scientific Python development.
 It is assumed that you have already installed _conda_ and are familiar with using a Terminal.
 
 In order to run the notebook type the following in a Terminal/Command Line:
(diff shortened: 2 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -29,10 +29,16 @@ The following text equips you with knowledge that makes contributing to ilastik
    gh repo clone ilastik
    ```
 
+1. Configure conda to use [strict channel priority](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-channels.html#strict-channel-priority):
+
+   ```
+   conda config --set channel_priority strict
+   ```
+
 1. Install [conda-build][conda-build] in order to access the [conda develop][conda-develop] command:
```
emilmelnikov, line 38: What do you think about moving to [`miniforge`](https://github.com/conda-forge/miniforge/#miniforge3)? I have no idea  whether it is stable enough though...

notebooks/Readme.md
```diff
@@ -13,11 +13,17 @@ notebooks/
 ```
 
 We use _conda_ to develop Python and recommend it for scientific Python development.
```
emilmelnikov, line 15:
```suggestion
We use _conda_ for Python-based projects and recommend it for scientific Python development.
```

notebooks/pixel_classification_api/environment.yml
```diff
@@ -4,5 +4,6 @@ channels:
   - conda-forge
 dependencies:
   - ilastik-dependencies-no-solvers
+  - ilastik-meta >=1.4.0b14
   - notebook
```
emilmelnikov, line 8: Does it make sense to recommend JupyterLab instead of just notebooks?

emilmelnikov (approved):

CONTRIBUTING.md
k-dominik: Looks interesting, are you using it? I'm not. I'd wait until most of us are using it before recommending it.

notebooks/pixel_classification_api/environment.yml
k-dominik: I don't see added value in jupyterlab for the purpose of this notebook and would in general try to keep this minimal.

CONTRIBUTING.md
emilmelnikov: Me too, but use the default installed just out of the habit, and I'd rather use only conda-forge channel if possible.

k-dominik: okay, I'll try to switch before Friday :_). Then let's maybe give it a go a bit by ourselves before recommending it. I definitely see the appeal.

emilmelnikov: I tried to install a fresh ilastik env with miniforge, and it seems to work.

PR #2460, opened 2021-07-22
Update devenv.py installation script

emilmelnikov:
Since lazyflow repository has been merged into the ilastik repository,
and @k-dominik improved the ilastik conda package, the `devenv.py`
script is no longer needed.

Use instructions from `CONTRIBUTING.md` instead.


k-dominik:
I'm not in favor - creating a devenv might become more tricky in the future, depending on one wants do dev in tiktorch synchronously and stuff. Did not have time to think about it.

emilmelnikov:
> I'm not in favor - creating a devenv might become more tricky in the future, depending on one wants do dev in tiktorch synchronously and stuff. Did not have time to think about it.

OK, but at least we need to update this.

scripts/devenv.py
```diff
@@ -1,272 +1,94 @@
...
 
-    env = CondaEnv(args.name)
-    execution_time, _res = timed(args.func, args, env)
-    logging.info(f"finished {args.command} ({execution_time:.3f} seconds)")
+def action_create(args):
+    run(["conda", "install", "--yes", "--name", "base", "--channel", "conda-forge", "conda-build"])
 
+    chans = ["--channel", "ilastik-forge", "--channel", "conda-forge"]
+    pkgs = ["ilastik-dependencies-no-solvers", "pre-commit"]
+    run(["conda", "create", "--yes", "--name", args.name] + chans + pkgs)
```
k-dominik, line 59: should have `--override-channels` and `--strict-channel-priority` (just in case someone has a `.condarc` stating something else.

scripts/devenv.py
```diff
@@ -1,272 +1,94 @@
...
 
-def command_create(args: argparse.Namespace, env: CondaEnv) -> None:
-    if env.exists():
-        logging.error(f"environment {env.name} already exists")
-        return
+    for repo in REPOSITORIES:
+        run(["conda", "develop", "--name", args.name, repo])
+        run(["conda", "run", "--name", args.name, "pre-commit", "install"], cwd=repo)
```
k-dominik, line 63: didn't know about `conda run` (which at least my version claims to be experimental - so maybe not good to rely on it?)

scripts/devenv.py
```diff
@@ -1,272 +1,94 @@
...
 
-    logging.info(f"Installing local ilastik packages in development mode.")
-    for pkg in packages:
-        env.install_pacakge_dev(location / pkg)
 
+def is_git_repo(cwd):
+    if not os.path.isdir(cwd):
+        return False
+    proc = subprocess.run(["git", "rev-parse"], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, cwd=cwd)
+    return proc.returncode == 0
```
k-dominik, line 82: I see you're adopting the unified `run` from subprocess.

subprocess docs [state](https://docs.python.org/3/library/subprocess.html#subprocess.check_call)
> Code needing to capture stdout or stderr should use run() instead:

so not sure here.

k-dominik (commented):
Do I see correctly, that it is no longer possible to override channels, add custom packages? I for one have used it a lot.

scripts/devenv.py
emilmelnikov: Even though it's experimental, I don't think too complicated for it to break: https://github.com/conda/conda/blob/master/conda/cli/main_run.py

scripts/devenv.py
emilmelnikov: > Do I see correctly, that it is no longer possible to override channels, add custom packages? I for one have used it a lot.

I wanted to simplify this script to the point that you could only use it to automate steps from `CONTRIBUTING.md`. My intention was that if someone wants to have other stuff installed, they could do themselves later in the environment.

But if you were using this a lot, I'll add this back.

emilmelnikov:
> Do I see correctly, that it is no longer possible to override channels, add custom packages? I for one have used it a lot.

I wanted to simplify this script to the point that you could only use it to automate steps from `CONTRIBUTING.md`. My intention was that if someone wants to have other stuff installed, they could do themselves later in the environment.

But if you were using this a lot, I'll add this back.

scripts/devenv.py
emilmelnikov: In this specific instance `check_call` is good because we don't need to capture.

k-dominik:
> But if you were using this a lot, I'll add this back.

on average I think I use it once a week. I'm really a fan of this script :), simplifies stuff for me a lot (say trying out different version of packages, adding packages)

scripts/devenv.py
```diff
@@ -1,272 +1,109 @@
...
-            logging.error(f"{location} is not a valid directory")
-            return
 
-    logging.info(
-        f"create environment {env.name}"
-        f" with packages [{', '.join(args.package)}]"
-        f" from channels [{', '.join(args.channel)}]"
-    )
-    env.create(packages=args.package, channels=args.channel)
+def cmd_create(args, rest):
```
k-dominik, line 75: just so I understand correctly, the syntax will slightly change (I'm fine with that if it is the case):

Before:

```
python devev.py create -n idev -c all my channels here -p all my packackages here
```

new:

```
python devenv.py create -n idev -c all -c my -c channels  here all my  packages  here
```
is this how it's supposed to be? I actually like it, because it's the same as in conda.

scripts/devenv.py
```diff
@@ -1,272 +1,109 @@
...
+    if not rest:
+        rest = list(DEFAULT_CREATE_ARGS)
+
+    chan_args = ["--override-channels", "--strict-channel-priority"]
+    run(["conda", "install", "--yes", "--name", "base"] + chan_args + ["--channel", "conda-forge", "conda-build"])
```
k-dominik, line 80: this first run modifies the base environment of users - should they be made aware? Or alternatively install conda-build in ilastik environment?

scripts/devenv.py
emilmelnikov: Yes, it just passes everything to `conda create`.

scripts/devenv.py
emilmelnikov: This is a good idea, although I wonder if this is a good opportunity to make `pip install -e .` work since [using `conda develop` is not recommended](https://github.com/conda/conda-build/issues/2633#issuecomment-359572215).

k-dominik: Since we have no setup.py [all conda develop does is add the directory to `conda.pth`](https://github.com/conda/conda-build/blob/cd761345ee4542eb19b989a1f44ff9cea457bdc2/conda_build/develop.py#L175) in site-packages. I don't think there's any pressure to change now. There isn't any deprecation notice for conda-develop.

k-dominik (commented):
Awesome, could you also please adapt CI (it's using the devenv script)

https://github.com/ilastik/ilastik/blob/ecddddc90b80727f896a1453190e77e6eafd499d/.circleci/config.yml#L68-L70

https://github.com/ilastik/ilastik/blob/ecddddc90b80727f896a1453190e77e6eafd499d/appveyor.yml#L31

scripts/devenv.py
```diff
@@ -1,272 +1,108 @@
...
-    env.create(packages=args.package, channels=args.channel)
+def cmd_create(args, rest):
+    chan_args = ["--override-channels", "--strict-channel-priority"]
+    create_args = list(rest if rest else DEFAULT_CREATE_ARGS)
+
+    run(["conda", "create", "--yes", "--name", args.name] + chan_args + create_args)
+    run(["conda", "install", "--yes", "--name", args.name] + chan_args + ["--channel", "conda-forge", "conda-build"])
+
+    for repo in REPOSITORIES:
+        run(["conda", "develop", "--name", args.name, repo])
```
k-dominik, line 83: my guess is that conda develop did not run in the right env

appveyor.yml
```diff
@@ -28,7 +28,7 @@ install:
   # replace with whatever version of ilastik triggered the appveyor
   - ps: cp -recurse C:\projects\ilastik $env:IlASTIK_ROOT\ilastik-meta\ilastik
   # Point to the current ilastik-meta
-  - python ilastik/scripts/devenv.py create -n %ENV_NAME% -c ilastik-forge conda-forge defaults -p ilastik-dependencies-no-solvers
+  - python ilastik/scripts/devenv.py --name %ENV_NAME% create
```
k-dominik, line 31: aww I see. So this PR introduces a slightly different syntax from conda, but I guess we can live with that.

emilmelnikov: I've moved `--name` into separate subcommands,

k-dominik: awesome

k-dominik (approved):
lgtm. awesome, how so much code was removed, but no functionality lost (well at least non that I used) and even convenience gained (pre-commit hooks).

Issue #2464, opened 2021-07-28
Using ilastik api fails for 2d image


constantinpape:
I tried using the ilastik api to run prediction for a 2d image, following the example notebook https://github.com/ilastik/ilastik/tree/main/notebooks/pixel_classification_api. Prediction fails with an assertion error (see full traceback below)

Minimal reproducible example
```python
import numpy as np
from ilastik.experimental.api import from_project_file
from xarray import DataArray

# project trained on 2D image data
project = "/home/pape/Work/my_projects/shilas_data/nuclei_foreground.ilp"
ilp = from_project_file(project)
# random example data of correct shape
image = np.random.rand(2048, 2048)
pred = ilp.predict(DataArray(image, dims=("y", "x")))
```

```
Traceback (most recent call last):
  File "test.py", line 10, in <module>
    pred = ilp.predict(DataArray(image, dims=("y", "x")))
  File "/home/pape/Work/software/conda/miniconda3/envs/shila/ilastik-meta/ilastik/ilastik/experimental/api/_base.py", line 61, in predict
    self._reorder_op.Input.setValue(data)
  File "/home/pape/Work/software/conda/miniconda3/envs/shila/ilastik-meta/ilastik/lazyflow/slot.py", line 73, in call_in_setup_context
    return func(self, *args, **kwargs)
  File "/home/pape/Work/software/conda/miniconda3/envs/shila/ilastik-meta/ilastik/lazyflow/slot.py", line 1143, in setValue
    self._changed()
  File "/home/pape/Work/software/conda/miniconda3/envs/shila/ilastik-meta/ilastik/lazyflow/slot.py", line 1328, in _changed
    self.maybe_call_within_transaction(self._changed_impl)
  File "/home/pape/Work/software/conda/miniconda3/envs/shila/ilastik-meta/ilastik/lazyflow/slot.py", line 1323, in maybe_call_within_transaction
    self.graph.maybe_call_within_transaction(fn)
  File "/home/pape/Work/software/conda/miniconda3/envs/shila/ilastik-meta/ilastik/lazyflow/graph.py", line 143, in maybe_call_within_transaction
    fn()
  File "/home/pape/Work/software/conda/miniconda3/envs/shila/ilastik-meta/ilastik/lazyflow/slot.py", line 1361, in _changed_impl
    self._configureOperator(self)
[...]

PR #2465, opened 2021-07-28
Ensure correct inference of the channel axis in api

m-novikov:
When project trained on 2d data loaded we should add singleton channel axis
to dimensions of the input, otherwise feature computation won't work

Closes: #2464


tests/test_ilastik/test_experimental/types.py
```diff
@@ -29,7 +31,10 @@ class TestProjects(enum.Enum):
 @dataclasses.dataclass
 class Dataset:
     path: str
+    #: Axes to use for api prediction
     axes: str
+    #: Axes to use for headless run
+    headless_axes: str
```
m-novikov, line 37: When running headless project (trained on xy image) on the image (x, y, c) where c is 1 it will fail. So for such case axis order is explicitly specified

k-dominik: headless refers to the "headless" mode in ilastik, and is not a good fit here. Also I don't find it very intuitive in this context. Maybe something like `training_axes` or `project_axes` would be a better fit.

k-dominik (commented):
Awesome, fixed before I even looked at the issue :D. Slight naming issue (not high priority, as iit concerns tests-only)

tests/test_ilastik/test_experimental/types.py
m-novikov: Those axis not about project, but about dataset. This what would be passed to `--input_axes` during the headless prediction run.

k-dominik: as far as I understand the `--input_axes` parameter it should represent the axes in the data. Are you adding singleton `c` here?

m-novikov: Not exactly as it's a png it's already there. Axes of dataset when added to ilastik are (x, y, c = 1) but if I try to run project trained on (x, y) axes without explicitly specifying input axes (x, y, c) using this dataset it will fail with an error.

k-dominik: ah, alright, then it's fine. I'd still not call it `headless_axes` but `data_axes` then

k-dominik (approved):
lgtm

PR #2466, opened 2021-07-29
Improve feature dialog rendering

m-novikov:
feature_seleciton

Left is patched, right is current

Old feature dialog icon rendering used incorrect sizeHints
which resulten in upscaling and pixelization of checkmarks and
3d / 2d toggles


k-dominik:
looks great on osx, too! Haven't looked at the code yet (and will not get to it today), but it would be cool if the feature window could be resized without fixed aspect ratio (which is currently the case)

k-dominik (approved):
Awesome. Always fun to fiddle with sizes, absolute values and the likes. For reference screenshots from osx and win:

osx

win

looks really presentable now. Thank you very much!

PR #2479, opened 2021-09-26
Allow for thicker lines when hovering over segmentation edge

m-novikov:
Currently, an edge is drawn as a transparent line.

Using the same pen type as in "Superpixel Edges" allows making non-labeled
lines interactive.

Related to: https://github.com/ilastik/volumina/pull/286


k-dominik:
Hm, @stuarteberg's comment (https://github.com/ilastik/volumina/pull/222#issuecomment-472864966)  suggests, that transparent edges are needed to make them clickable?

m-novikov:
> Hm, @stuarteberg's comment ([ilastik/volumina#222 (comment)](https://github.com/ilastik/volumina/pull/222#issuecomment-472864966)) suggests, that transparent edges are needed to make them clickable?

I think what is meant here is regardless of the color it should be drawn to be clickable. Why transparent has been chosen is a mystery, maybe it was a performance optimization.

Issue #2472, opened 2021-08-30
adding multiple files as secondary inputs via drag and drop broken


k-dominik:
we got a [user report](https://forum.image.sc/t/ilastik-input-data-bulk-segmentation-loading/56869?u=k-dominik) about `1.3.3` still supporting adding multiple files as secondary inputs via drag and drop (e.g. probability maps in object classification), and it not working in `1.4.0b15`. I don't recall any intentional change that would result in this behavior.

PR #2486, opened 2021-09-30
fix dnd behavior on secondary roles

k-dominik:
in the past it was possible to drag n drop files into all roles.

Reason for loss of this functionality was that the startingLaneNum would not be
would not be incremented.

Now we increment this with each file path.

fixes: https://github.com/ilastik/ilastik/issues/2472

### DIFF
--- a/ilastik/applets/dataSelection/dataSelectionGui.py
+++ b/ilastik/applets/dataSelection/dataSelectionGui.py
@@ -66,7 +66,7 @@ from lazyflow.slot import Slot
 
 
 class LocationOptions(object):
-    """ Enum for location menu options """
+    """Enum for location menu options"""
 
     Project = 0
     AbsolutePath = 1
@@ -415,11 +415,17 @@ class DataSelectionGui(QWidget):
 
     def addFileNames(self, paths: List[Path], startingLaneNum: int, roleIndex: int):
         # If the user didn't cancel
-        for path in paths or []:
+        # we iterate through every path and add it. now bulk addition.
+        for index, path in enumerate(paths or []):
             try:
+                # startlingLaneNum = -1 if dragged onto the empty table
+                if startingLaneNum is not None and startingLaneNum >= 0:
+                    lane_index = startingLaneNum + index
+                else:
+                    lane_index = startingLaneNum
                 full_path = self._get_dataset_full_path(path, roleIndex=roleIndex)
                 info = self.instantiate_dataset_info(url=str(full_path), role=roleIndex)
-                self.addLanes([info], roleIndex=roleIndex, startingLaneNum=startingLaneNum)
+                self.addLanes([info], roleIndex=roleIndex, startingLaneNum=lane_index)
             except DataSelectionGui.UserCancelledError:
                 pass
             except Exception as ex:
--- a/ilastik/applets/dataSelection/dataSelectionGui.py
+++ b/ilastik/applets/dataSelection/dataSelectionGui.py
@@ -28,6 +28,7 @@ from vigra import AxisTags
 import threading
 import h5py
 from functools import partial
+import itertools
 import logging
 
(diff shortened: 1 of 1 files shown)


ilastik/applets/dataSelection/dataSelectionGui.py
```diff
@@ -415,11 +415,17 @@ def addFiles(self, roleIndex, startingLaneNum=None):
 
     def addFileNames(self, paths: List[Path], startingLaneNum: int, roleIndex: int):
```
emilmelnikov, line 416: Assuming that null checks are really needed:

```suggestion
    def addFileNames(self, paths: Optional[List[Path]], startingLaneNum: Optional[int], roleIndex: int):
```

ilastik/applets/dataSelection/dataSelectionGui.py
```diff
@@ -415,11 +415,17 @@ def addFiles(self, roleIndex, startingLaneNum=None):
...
         # If the user didn't cancel
-        for path in paths or []:
+        # we iterate through every path and add it. now bulk addition.
```
emilmelnikov, line 418: I think this comment is unnecessary because it just reiterates what is already happening in code.

ilastik/applets/dataSelection/dataSelectionGui.py
```diff
@@ -415,11 +415,17 @@ def addFiles(self, roleIndex, startingLaneNum=None):
...
+        for index, path in enumerate(paths or []):
```
emilmelnikov, line 419: I'd replace default-checking and indexing code with something like this:

```python
if paths is None:
    paths = []

# startlingLaneNum == -1 if dragged onto the empty table.
if startingLaneNum is None:
   startingLaneNum = -1

lane_indices = itertools.count(0) if startingLaneNum >= 0 else itertools.repeat(startingLaneNum)

for lane_index, path in zip(lane_indices, paths):
    ...
```

ilastik/applets/dataSelection/dataSelectionGui.py
```diff
@@ -413,13 +414,23 @@ def addFiles(self, roleIndex, startingLaneNum=None):
...
         # If the user didn't cancel
-        for path in paths or []:
+        if paths is None:
+            paths = []
+
+        # startlingLaneNum == -1 if dragged onto the empty table.
+        if startingLaneNum is not None and startingLaneNum >= 0:
+            lane_indices = itertools.count(startingLaneNum)
+        else:
+            # startlingLaneNum = -1 if dragged onto the empty table
```
emilmelnikov, line 426:
```suggestion
```

ilastik/applets/dataSelection/dataSelectionGui.py
```diff
@@ -413,13 +414,23 @@ def addFiles(self, roleIndex, startingLaneNum=None):
         paths = ImageFileDialog(self).getSelectedPaths()
         self.addFileNames(paths, startingLaneNum, roleIndex)
 
-    def addFileNames(self, paths: List[Path], startingLaneNum: int, roleIndex: int):
+    def addFileNames(self, paths: Optional[List[Path]], startingLaneNum: Optional[int], roleIndex: int):
         # If the user didn't cancel
```
emilmelnikov, line 418:
```suggestion
```

emilmelnikov (changes requested):

ilastik/applets/dataSelection/dataSelectionGui.py
k-dominik: that comment I'd leave in because it's not obvious from code

emilmelnikov: It's a duplicate; or you can delete the other one.

emilmelnikov (approved):

PR #2496, opened 2021-11-11
Add conda recipe

k-dominik:
This PR got bigger than I expected, but I chose to fix some issues related to dependency upgrades right away, some had to be fixed upstream..., specifically

* vigra incompatibility with newer h5py versions (at least in some builds) -> replaced usages of `vigra.impex.writeHDF5` with `h5py`, directly
* some tests invoked the headless mode via `ilastik-meta/ilastik/ilastik.py`. This file is only really necessary for the release binary, and I don't want to include it in the packge. Hence, switched to using `ilastik.__main__` instead
* we had a pin on a (old) version of sklearn - definitely wanted to get rid of that. had to remove a test that used a serialized sklearn model - these models are not compatible between versions, unfortunately (https://scikit-learn.org/stable/modules/model_persistence.html).
* `testCarvingGui` has always been flaky, but it got worse now, so I disabled to tackle it separately, #2512
* also #2513 - also needs to be tackled separately

## Checklist


conda-recipe/meta.yaml
```diff
@@ -0,0 +1,79 @@
...
+  host:
+    - python >=3.7,<3.10
+    - pip
+    - setuptools >=40.0
+    - setuptools_scm
+  run:
+    - python >=3.7,<3.10
+    # pin for numpy necessary because of vigra issue
+    # https://github.com/conda-forge/vigra-feedstock/pull/80#issuecomment-965105882
+    - numpy <1.20
```
constantinpape, line 20: This pin is really unfortunate, as it will make this recipe incompatible with many existing environments. 
I see the test error in the conda build you linked to, but I have been using vigra with numpy>=1.20 for quite some time withouth any issues. For example this
```python
import numpy as np
import vigra
x = np.random.rand(64, 64).astype("float32")
y = vigra.filters.gaussianSmoothing(x, 1.0)
```
works in an env with vigra 1.11.1 and numpy 1.21.4 (both from conda-forge) and I have used many other "common" vigra functionality (mostly filters and segmentation functionality like watershed, connected components and relabeling) in that env without issues.

Have you ever run the relevant ilastik tests with a numpy>=1.20? I would argue that we don't need this pin if these tests pass, despite the failing conda tests. (Ofc it would be better to properly fix this, but I am not sure how easy that would be, esp. since vigra is pretty much unmaintained)

k-dominik: yes, sure did. add a `x + y` to your code and you're there...

k-dominik: version that will run into issues:

```python
import numpy as np
import vigra
x = vigra.taggedView(np.random.rand(64, 64).astype("float32"), axistags="xy")
y = vigra.filters.gaussianSmoothing(x, 1.0)
x + y

Comment (constantinpape) [reply conda-recipe/meta.yaml:20]: This also fails:
```python
import numpy as np
import vigra
x = vigra.taggedView(np.random.rand(64, 64).astype("float32"), axistags="xy")
y = vigra.taggedView(np.random.rand(64, 64).astype("float32"), axistags="xy")
x + y
```
but this passes
```python
import numpy as np
import vigra
x = vigra.taggedView(np.random.rand(64, 64).astype("float32"), axistags="xy")
y = np.random.rand(64, 64).astype("float32")
x + y
print("Passed")
```
So the issue seems to be adding two vigra arrays with axistags (of course it was the axistags...).

> I think this is in practice only an issue with ilastikrag, so multicut.

Is this code that we could change to fix these issues? I think that this package will be much more useful if we don't need this numpy pin ...

tests/test_ilastik/test_workflows/test_HeadlessPixelClassificationWorkflow.py
```diff
@@ -78,10 +78,10 @@ def run_headless_pixel_classification(
...
     assert raw_data.parent.exists()
 
-    ilastik_dot_py = Path(__file__).parent.parent.parent.parent / "ilastik.py"
     subprocess_args = [
         "python",
-        str(ilastik_dot_py),
+        "-m",
+        "ilastik",
         "--headless",
         "--project=" + str(project),
```
emilmelnikov, line 86: While we're here:

```suggestion
        f"--project={project}",
```

Same for other occurrences.

tests/test_ilastik/test_workflows/testCarvingGui.py
```diff
@@ -29,6 +29,10 @@
 import pytest
 
 
+if os.environ.get("ON_CIRCLE_CI", False):
```
emilmelnikov, line 32: `None` is falsey too:

```suggestion
if os.environ.get("ON_CIRCLE_CI"):
```

tests/test_ilastik/test_workflows/axesOrderPreservation/testAxesOrderPreservation.py
```diff
@@ -111,15 +106,17 @@ def create_simple_input(cls, name, data):
...
+                ds = f.create_dataset(name="simple", data=data)
+                ds.attrs["axistags"] = data.axistags.toJSON()
             cls.created_data.append(write_at)
 
         return write_at
 
     @classmethod
     def create_input(cls, filepath, input_axes, outmin=None, outmax=None, dtype=None):
-        """ Creates a file from the data at 'filepath' that has 'input_axes' """
+        """Creates a file from the data at 'filepath' that has 'input_axes'"""
```
emilmelnikov, line 119:
```suggestion
        """Creates a file from the data at ``filepath`` that has ``input_axes``."""
```

setup.cfg
```diff
@@ -1,3 +1,15 @@
...
+url = https://github.com/ilastik/ilastik
+
+[options]
+packages = find:
+include_package_data = True
+
+[options.packages.find] #optional
+include=ilastik*, lazyflow*
+
 [tool:pytest]
```
emilmelnikov, line 13: Move  pytest and isort configs to `pyproject.toml`?

conda-recipe/meta.yaml
```diff
@@ -0,0 +1,78 @@
+{% set setup_py_data = load_setup_py_data(setup_file='../setup.py', from_recipe_dir=True) %}
+
+package:
+  name: ilastik-core
+  version: {{ setup_py_data.version }}
```
emilmelnikov, line 5: I wonder whether it's generally a good idea to reference a version from setuptools here...

Conda packages and wheels/sdists are really different things with different dependencies, even though names of those dependencies are often identical. So, maybe we should explicitly specify conda package version here too, like we do with conda requirements? Do we really benefit from using `setuptools_scm`? (I know I advocated for using it, but I'm not so sure anymore, esp. given that we release only on conda).

.github/workflows/core-conda-bld.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: install xvfb/deps
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          sudo apt-get update
+          sudo apt-get install -yy libgl1-mesa-dev xvfb curl
+      - name: linux conda build test
+        if: matrix.os == 'ubuntu-latest'
+        shell: bash -l {0}
+        run: |
+          conda install -n base -c conda-forge mamba boa setuptools_scm -y
```
emilmelnikov, line 35: Extract `conda install` into a separate, shared step?

.github/workflows/core-conda-bld.yml
```diff
@@ -0,0 +1,48 @@
...
+    branches: [ main ]
+
+jobs:
+  test-w-conda-recipe:
+    strategy:
+      matrix:
+        os: [macos-latest, windows-latest, ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: checkout
```
emilmelnikov, line 16: A very minor point, but I'd skip naming steps that are already described by their `uses` names.

emilmelnikov (approved):

setup.cfg
k-dominik: note, pytest config still needs to stay in setup.cfg (we're still stuck with <4 - pytproject is supported in 6+)

conda-recipe/meta.yaml
k-dominik: I have to say that I appreciate the convenience that comes with it - we can add automatic builds from master with proper versions and everything in sync without basically any effort. So far `setuptools_scm` has been working nicely for me and I'm happy you recommended it :)

k-dominik:
allright - merging this now. Will follow up with PR for automatic builds

PR #2497, opened 2021-11-12
looks like big sur went beyond 10.16 already

k-dominik:
Big Sur is 10.16.1
Monterey is 12.0.1 (thx @emilmelnikov)

## Checklist

### DIFF
--- a/ilastik.py
+++ b/ilastik.py
@@ -109,9 +109,11 @@ def main():
     path_setup()
 
     # https://bugreports.qt.io/browse/QTBUG-87014
-    if platform.mac_ver()[0] == "10.16":
-        os.environ["QT_MAC_WANTS_LAYER"] = "1"
-        os.environ["VOLUMINA_SHOW_3D_WIDGET"] = "0"
+    if platform.system().lower() == "darwin":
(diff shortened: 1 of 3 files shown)


emilmelnikov (changes requested):
On Monterey, `platform.mac_ver()[0]` returns `12.0.1`. `tuple(map(int, platform.mac_ver()[0].split("."))) >= (10, 16)` should work for all affected versions.

However, since we're moving to Python 3.9 anyway, maybe it makes sense to remove these workarounds.

k-dominik:
> On Monterey, `platform.mac_ver()[0]` returns `12.0.1`. `tuple(map(int, platform.mac_ver()[0].split("."))) >= (10, 16)` should work for all affected versions.
> 
> However, since we're moving to Python 3.9 anyway, maybe it makes sense to remove these workarounds.

awesome that you had a look, I have honestly missed Monterey release :o

emilmelnikov:
> with the `map(int...` are we sure they'll never append any letters to version numbers at apple?

No, but I think it's reasonable assumption to make, at least for the released versions.

emilmelnikov (approved):

PR #2499, opened 2021-11-23
removed deprecated Iterable from collections

k-dominik:
collections.Iterable -> collections.abc.Iterable (since Python 3.3, deprecated in 3.9)

## Checklist

### DIFF
--- a/ilastik/applets/objectExtraction/opObjectExtraction.py
+++ b/ilastik/applets/objectExtraction/opObjectExtraction.py
@@ -23,6 +23,7 @@ from __future__ import division
 from builtins import range
 from copy import copy, deepcopy
 import collections
+from collections.abc import Iterable
 from functools import partial
 
 # SciPy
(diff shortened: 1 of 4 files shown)


emilmelnikov (approved):
Iterable protocol also [tries to call `__getitem__`][iter] on an object if it doesn't have `__iter__`. Not sure whether we want to handle these cases too, but if we do, we could use something like this:

```python
def isiter(obj):
    try:
        iter(obj)
        return True
    except TypeError:
        return False
```

[iter]: https://docs.python.org/3/library/functions.html#iter

k-dominik:
> Iterable protocol also [tries to call `__getitem__`](https://docs.python.org/3/library/functions.html#iter) on an object if it doesn't have `__iter__`. Not sure whether we want to handle these cases too, but if we do, we could use something like this:
> 
> ```python
> def isiter(obj):
>     try:
>         iter(obj)
>         return True
>     except TypeError:
>         return False
> ```

the code probably means `itertools.abc.Sequence`? I can't see how it would, e.g., work with a dict?!

PR #2503, opened 2021-12-02
removed check for sklearn version in OpInterpMissingData

k-dominik:
we're checking there against sklearn 0.11, which is not even available on
conda-forge anymore (oldest version at time of writing 0.17).
Furthermore this check doesn't work for versions 1.* :)

## Checklist

### DIFF
--- a/lazyflow/operators/opDetectMissingData.py
+++ b/lazyflow/operators/opDetectMissingData.py
@@ -1,4 +1,3 @@
-from __future__ import print_function
 from __future__ import division
 from future import standard_library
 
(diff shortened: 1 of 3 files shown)


lazyflow/operators/opDetectMissingData.py
```diff
@@ -51,42 +51,15 @@
...
-        return int(m[1])
-
-
 try:
-    from sklearn import __version__ as sklearnVersion
-
-    svcTakesScaleC = extractVersion(sklearnVersion) < 11
-except ImportError as VersionError:
+    from sklearn.svm import SVC
+except ImportError:
```
emilmelnikov, line 56: Can we have sklearn as a required dep? It's tiny compared to ilastik itself.

emilmelnikov (approved):

PR #2506, opened 2021-12-03
Fix ram estimation, prepare numpy upgrade

k-dominik:
numpy deprecated checking `numpy.issubdtype(whatever, numpy.dtype)` with `numpy.dtype` as second argument (which was coerced to `numpy.generic` in the past). The warning for this was there for quite some time. I took the opportunity and removed some code duplication in the process of fixing this.

Also fixed a typo that has been there from 2014 that prevented the metadata for ram_usage_per_requested_pixel to be set. To be fair, this wasn't an issue up until the numpy deprecation as most operators had some sort of safeguard around this.

## Checklist

- [x] Add tests.

### DIFF
--- a/lazyflow/utility/helpers.py
+++ b/lazyflow/utility/helpers.py
@@ -22,8 +22,10 @@
 
 from functools import reduce
 from operator import mul
-from typing import Tuple, Iterable
+from typing import Iterable, Tuple, Type, Union
+import sys
 import numbers
+import numpy
 
 
 def itersubclasses(cls, _seen=None):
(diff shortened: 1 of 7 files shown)


lazyflow/utility/helpers.py
```diff
@@ -127,3 +129,30 @@ def bigintprod(nums: Iterable[numbers.Integral]) -> int:
...
+    8
+    """
+    dtype = numpy.dtype(dtype)
+
+    if numpy.issubdtype(dtype, numpy.number):
+        return dtype.itemsize
+    elif numpy.issubdtype(dtype, numpy.bool_):
+        return dtype.itemsize
+    else:
+        return sys.getsizeof(None)
```
emilmelnikov, line 158: Why are we using the size of `None` here? Are we trying to guesstimate the required size of an object that is not a numpy number/boolean without creating an instance?

k-dominik: I have really no clue (see commit comment: "in principle we don't need this - we could simply
`numpy.dtype(whatever).itemsize`. For some reason the old code treats `object`
and `numpy.object_` differently (numpy will give itemsize = 8 on my machine,
vs `sys.getsizeof(None) = 16). Keeping this behavior for now.")

emilmelnikov: Hmm, it would be helpful to put this in a comment then. Commit messages are great, but I think this is quite obscure and should be more visible.

emilmelnikov (approved):

lazyflow/utility/helpers.py
```diff
@@ -127,3 +129,37 @@ def bigintprod(nums: Iterable[numbers.Integral]) -> int:
     array shapes seen in practice on some operating systems (windows: long -> 32 bit).
     """
     return reduce(mul, map(int, nums))
+
+
+def get_ram_per_element(dtype: Union[object, numpy.dtype]) -> int:
```
emilmelnikov, line 134: By the way, the type signature is somewhat incorrect: `numpy.dtype()` accepts a lot of stuff, and also `numpy.dtype` value is an `object` too.

k-dominik: yeah  right. It should be... `Type[object]` I suppose

emilmelnikov: Strictly speaking. it should be [`numpy.typing.DtypeLike`](https://numpy.org/doc/stable/reference/typing.html#numpy.typing.DTypeLike).

k-dominik: Awesome, thanks for the heads-up there! Looks like it was only introduced in numpy 1.20 - which we currently don't support because of vigra. Let's keep this in mind until we can support numpy 1.20. Will leave it at that for the time being.

PR #2511, opened 2021-12-20
configure logging only once

k-dominik:
logging is configure whenever the main ilastik main function is executed
(`ilastik.app.main`). In tests this happens multiple times.

This change prevents multiple initialization of logging. For me this
reduced the log size per test run from 2.7MB to 270kB

## Checklist

### DIFF
--- a/ilastik/ilastik_logging/default_config.py
+++ b/ilastik/ilastik_logging/default_config.py
@@ -1,5 +1,3 @@
-from __future__ import absolute_import
-
 ###############################################################################
 #   ilastik: interactive learning and segmentation toolkit
 #
@@ -218,7 +216,15 @@ def get_default_config(
     return default_log_config
 
 
+_LOGGING_CONFIGURED = False
+
+
 def init(format_prefix="", output_mode=OutputMode.LOGFILE_WITH_CONSOLE_ERRORS, logfile_path=DEFAULT_LOGFILE_PATH):
+    global _LOGGING_CONFIGURED
+    if _LOGGING_CONFIGURED:
+        warnings.warn("logging has been already initialized; skipping further initialization calls", RuntimeWarning)
+        return
+
     if logfile_path == "/dev/null":
         assert output_mode != OutputMode.LOGFILE, "Must enable a logging mode."
         output_mode = OutputMode.CONSOLE
@@ -262,3 +268,5 @@ def init(format_prefix="", output_mode=OutputMode.LOGFILE_WITH_CONSOLE_ERRORS, l
         return filename + "(" + str(lineno) + "): " + category.__name__ + ": " + message.args[0]
 
     warnings.formatwarning = simple_warning_format
+
+    _LOGGING_CONFIGURED = True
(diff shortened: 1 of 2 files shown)


emilmelnikov:
Is it possible to avoid multiple log configuration calls in tests? It seems that calling `init` multiple times should either reconfigure the logging system with the new parameters, throw an error, or at least issue some warning.

ilastik/ilastik_logging/default_config.py
```diff
@@ -218,7 +218,14 @@ def get_default_config(
     return default_log_config
 
 
+LOGGING_CONFIGURED = False
+
+
 def init(format_prefix="", output_mode=OutputMode.LOGFILE_WITH_CONSOLE_ERRORS, logfile_path=DEFAULT_LOGFILE_PATH):
+    global LOGGING_CONFIGURED
```
emilmelnikov, line 225: Maybe this is an overkill, but there is a way to avoid the global variable:

```python
def call_once(func):
    def wrapper(*args, **kwargs):
        if not wrapper.called_once:
            func(*args, **kwargs)
            wrapper.called_once = True

    wrapper.called_once = False
    functools.update_wrapper(wrapper, func)
    return wrapper


@call_once
def init(...):
    ...
```

k-dominik: yeah I found something very similar on stackoverflow, but indeed thought it was overkill, but maybe I'll make the global variable private...

k-dominik:
> Is it possible to avoid multiple log configuration calls in tests? It seems that calling `init` multiple times should either reconfigure the logging system with the new parameters, throw an error, or at least issue some warning.

hey, thanks for having a look!

It is never called intentionally (as far as I have seen) and currently I cannot think of a reason to allow for that. But sure, can issue a warning to at least give a hint in case someone actually wants to do that.

ilastik/ilastik_logging/default_config.py
```diff
@@ -218,7 +216,15 @@ def get_default_config(
     return default_log_config
 
 
+_LOGGING_CONFIGURED = False
+
+
 def init(format_prefix="", output_mode=OutputMode.LOGFILE_WITH_CONSOLE_ERRORS, logfile_path=DEFAULT_LOGFILE_PATH):
+    global _LOGGING_CONFIGURED
+    if _LOGGING_CONFIGURED:
+        warnings.warn("logging already configure - repeat configuration ignored", RuntimeWarning)
```
emilmelnikov, line 225:
```suggestion
        warnings.warn("logging has been already initialized; skipping further initialization calls", RuntimeWarning)
```

emilmelnikov (approved):

ilastik/ilastik_logging/default_config.py
k-dominik: thank you!!

k-dominik:
this builds just fine on my linux machine - circleci kills the process during solving of the environment. I'll probably switch it to the "in progress" new dependencies package....

PR #2515, opened 2021-12-23
Api add object classification

k-dominik:
Adding object classification API here. To enable this I had to introduce a few changes...

* deprecate `pipeline.predict(data=...)` in favor of following naming conventions in headless mode: `pipeline.get<export_source>(raw_data=..., <secondary_input>=...)`
* Object Classification pipelines can be constructed from trained projects (both variants, from segmentation, from probabilities).

* currently all pipelines are constructed via `from_project_file` - not sure yet whether constructing a pipeline should be "magic" for all project types. Would be probably make sense to have explicit constructors for specific project types to clearly communicate intention - it would also fail earlier at runtime (during construction) and not when accessing a method that is not part of that pipeline's interface.

I uploaded `ilastik-core` package built at the tip of this branch to ilastik-forge (version `1.4.0b21.dev38+g980b8d0a8-py_0`)
It can be installed via

```bash
conda install -c ilastik-forge -c conda-forge ilastik-core
```

## Checklist

- [x] Add tests.


ilastik/experimental/api/_pipelines.py
```diff
@@ -0,0 +1,256 @@
...
+        self._thresholding_op.CurOperator.setValue(thresh_settings.method)
+        self._thresholding_op.InputImage.connect(self._normalize_probmaps.Output)
+
+        self._obj_extraction_op.RawImage.connect(self._reorder_raw_op.Output)
+        self._obj_extraction_op.BinaryImage.connect(self._thresholding_op.Output)
+
+    def get_object_probabilities(self, raw_data, prediction_maps):
+        return self._process(raw_data, prediction_maps, output_slot=self._stack_prob_channels.Output)
+
+    def _setup_normalization_function(self, prediction_maps: numpy.ndarray):
```
Tomaz-Vieira, line 214: This makes the pipeline not thread-safe (but then again, maybe more stuff inside classic ilastik already makes it not thread-safe when used as a library like this anyway). Are we already settled in the concurrency limitations we want to impose on the consumer of the API?

ilastik/experimental/api/_pipelines.py
```diff
@@ -0,0 +1,256 @@
...
+        num_spatial_in_data = sum(a.isSpatial() for a in data.axistags)
+        if num_spatial_in_data != self._num_spatial_dims:
+            raise ValueError(
+                "Number of spatial dims doesn't match. "
+                f"Classifier trained for {self._num_spatial_dims} but input has {num_spatial_in_data}"
+            )
+
+
+class _ObjectClassificationFromSegmentationPipelineImpl(
+    _ObjectClassificationPipelineBase, ObjectClassificationFromSegmentationPipeline
```
Tomaz-Vieira, line 161: Not sure if I follow why we need to have two versions of `ObjectClassificationFromSegmentationPipeline`, one public and one private. I'm guessing it's to hide the private methods and potentially also the `__init__`? If so, the private methods are private anyway (or just as private as the private versions of these classes), and the `__init__` seems to be good enough to be public too.

ilastik/experimental/parser/_base.py
```diff
@@ -33,44 +31,50 @@ def __exit__(self, exc_type, exc, tb):
...
+
+class _PixelClassificationKeys:
+    ROOT = "PixelClassification"
+    TYPE = "pickled_type"
+    FACTORY = "ClassifierFactory"
+    FORESTS = "ClassifierForests"
     LABEL_NAMES = "LabelNames"
 
 
-class _PixelClassProjectImpl(types.PixelClassificationProject):
```
emilmelnikov, line 61: Any reason for holding onto HDF5 file handle and not parsing all data into a dataclass? This way, there is no need for ABCs to exist.

```python
@dataclass
class Project:
    _registry: ClassVar[MutableMapping[str, "Project"]] = {}

    def __init_subclass__(cls, workflow_name: str = "", **kwargs):
        if workflow_name:
            Project._registry[workflow_name] = cls

    common_prop_1: ...
    common_prop_2: ...

    @classmethod
    def _from_ilp(cls, f: h5py.Group):
        return cls(
            common_prop_1=f[...],
            common_prop_2=f[...],
            ...
        )


@dataclass
class ObjectClassificationProject(Project):
    common_for_oc_prop: ...

    @classmethod
    def _from_ilp(cls, f: h5py.Group):
       return cls(
           common_for_oc_prop=f[...],
           **dataclasses.asdict(super()._from_ilp(f)),
       )


@dataclass
class PredictionsObjectClassificationProject(ObjectClassificationProject, workflow_name="Object Classification (from prediction image)"):
    ...


def from_ilp(path):
    with h5py.File(path, "r") as f:
        workflow_name = f[...]
        try:
            return Project._registry[workflow_name]._from_ilp(f)
       except KeyError:
            raise ValueError(f"unsupported workflow {workflow_name}") from None
```

ilastik/experimental/api/_pipelines.py
```diff
@@ -0,0 +1,256 @@
...
+    raise ValueError("numpy arrays don't provide information about axistags expecting xarray")
+
+
+@convert_to_vigra.register
+def _(data: xarray.DataArray):
+    axistags = "".join(data.dims)
+    return vigra.taggedView(data.values, axistags)
+
+
+class _PixelClassificationPipelineImpl(PixelClassificationPipeline):
```
emilmelnikov, line 61: I'm not sure that we really need an ABC and it's implementation, just for the sake of a single method...

Do we really need to keep all operator state between calls? If no, then all prediction logic could be placed into project classes themselves since they already have workflow-specific types. Those methods would prepare all the required operators, run the computation and return the result, discarding all operator state.

tests/test_ilastik/test_experimental/test_parser/test_base.py
```diff
@@ -86,11 +86,111 @@ def test_parse_project_classifier(self, test_data_lookup, proj, expected_factory
...
+    """Helper function to compare dicts that might have numpy arrays as values"""
+    if isinstance(a, dict):
+        assert isinstance(b, dict), f"Expected dictionary for second operand, got {type(b)}"
+        assert a.keys() == b.keys()
+        for k in a:
+            assert_dict_equal(a[k], b[k])
+    elif isinstance(a, np.ndarray):
+        np.testing.assert_array_equal(a, b)
+    else:
+        assert a == b, f"{a} not equal to {b}"
```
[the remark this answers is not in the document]
k-dominik:
```python
>>> a = {"a": 1, "b": numpy.arange(10)}
>>> b = {"a": 1, "b": numpy.arange(10)}
>>> assert a == b
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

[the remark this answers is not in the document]
emilmelnikov: Yes, of course, I just wanted to note that the name itself could be a little bit confusing.

PR #2524, opened 2022-01-25
Fix drag n drop in batchprocessing on windows

### LINKED ISSUE #2523: batchprocessing drag and drop wrong file paths on windows
## Describe the bug
adding files by drag and drop on windows results in broken file paths with leading `/` on windows, so something like `/C:/blah/blah.h5`. Those then aren't found and batch processing will fail.

## Expected behavior
Clearly the paths should be valid

## To Reproduce
Within the any workflow on windows
1.Do whatever is necessary to get to batch processing
2. From windows explorer drag and drop some files -> :boom: 

we check whether dnd events are local files but then don't use `toLocalFile()` for no apparent reason (other than that it works on linux and osx and probably never tested this on win?!) https://github.com/ilastik/ilastik/blob/6acd2c554bc517e9c8ddad3623a7aaa2e6970c28/ilastik/applets/batchProcessing/batchProcessingGui.py#L56

## Screenshots

dnd-issue-win

 - ilastik version: 1.4.0b21 (but also at least in 1.3.3post3)
 - OS: Windows

k-dominik:
treat drag and dropped paths as local files.

- [x] Add tests.

fixes #2523

### DIFF
--- a/ilastik/applets/batchProcessing/batchProcessingGui.py
+++ b/ilastik/applets/batchProcessing/batchProcessingGui.py
@@ -53,7 +53,7 @@ class FileListWidget(QListWidget):
     def dropEvent(self, dropEvent):
         urls = dropEvent.mimeData().urls()
         self.clear()
-        self.addItems(qurl.path() for qurl in urls)
+        self.addItems(qurl.toLocalFile() for qurl in urls)
 
     def dragEnterEvent(self, event):
         # Only accept drag-and-drop events that consist of urls to local files.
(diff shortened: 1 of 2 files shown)


Tomaz-Vieira:
This would work even for funny filenames, like `my file:name.txt`, right? Hopefully Qt isn't reinterpreting strings to determine what the correct course of action would be

k-dominik:
> This would work even for funny filenames, like `my file:name.txt`, right? Hopefully Qt isn't reinterpreting strings to determine what the correct course of action would be

Hey, thanks for looking into this. This goes down to OS/Source of drag/drop event. But I'm pretty sure only known sources are converted to URLs (depending on the protocol, no conversion is needed - xdnd should already have the correct MIME type)...

I think on windows such a filename is not even possible. On linux it works as expected not overinterpreting the filename.

PR #2534, opened 2022-02-07
adding deploy job for core-package on merges to main

k-dominik:
currently this also builds on every PR commit (in order to demonstrate that it works), which is of course not intended. Before merging, this has to be changed of course.

files are uploaded here: https://anaconda.org/ilastik-forge/ilastik-core/files

side note: will tackle the black thing separately, which is cause by empty changeset (python-wise)

Also, maybe it would be better still to do releases via tags, and not build _every_ push to main as a conda-recipe.

Todos:
- [x] remove build on PRs


k-dominik:
nah :)

PR #2535, opened 2022-02-07
test release recipe

k-dominik:
This adds "release" recipes that enable conda as a distribution channel for ilastik.
As a side effect, gui tests are now run on all platforms.

In the conda recipe, I added "variants" for the cpu (`""`) and gpu (`"-gpu"`) builds via environment variables instead of having conda manage variants via `cbc.yaml`, mainly to enable parallel builds. I think in our case this is also simpler to read.

Once this is merged, we can probably get rid of appveyor and circleci builds

## Checklist

- [ ] Add tests.

### DIFF
--- a/.github/workflows/test-releases.yml
+++ b/.github/workflows/test-releases.yml
@@ -1,0 +1,51 @@
+name: test-releases
+
+on:
+  push:
+    branches: [ main ]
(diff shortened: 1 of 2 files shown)


k-dominik:
nah

PR #2537, opened 2022-02-08
Add release recipe

k-dominik:
Summary:

* added multi-output (meta) recipe, which ensures core and "release" packages are in sync
* added build/test to gh-actions
  * legacy-gui-tests currently not run there (would hang, need to investigate, will open issue after merge)
  * windows-gpu variant not tested - didn't manage to find a working combination of packages yet (issue after merge)
  * in order to speedup tests, avoid unnecessary builds, conda build is done in two stages, ones to only build noarch recipe on linux, and then runs the tests in a second stage on all osses.

### DIFF
--- a/.github/workflows/core-conda-bld.yml
+++ b/.github/workflows/core-conda-bld.yml
@@ -1,4 +1,4 @@
-name: test
+name: conda-package
 
 on:
   push:
@@ -7,11 +7,10 @@ on:
     branches: [ main ]
 
 jobs:
-  test-w-conda-recipe:
-    strategy:
-      matrix:
-        os: [macos-latest, windows-latest, ubuntu-latest]
-    runs-on: ${{ matrix.os }}
+  conda-noarch-build:
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.version.outputs.version }}
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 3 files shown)


tests/test_ilastik/widgets/test_ImageFileDialog.py
```diff
@@ -23,7 +23,7 @@ def tmp_preferences(tmp_path) -> pathlib.Path:
 def image(tmp_path) -> Path:
     image_path = Path(tmp_path) / "some_picture.png"
     image_path.touch()
-    return image_path
+    return image_path.resolve()
```
emilmelnikov, line 26: Leftover code change?.. This doesn't seem to be build-related.

emilmelnikov (approved):

.github/workflows/core-conda-bld.yml
```diff
@@ -21,23 +80,42 @@ jobs:
...
+          key: ${{ github.sha }}-packages
+      - name: linux test
         if: matrix.os == 'ubuntu-latest'
         shell: bash -l {0}
-        run: xvfb-run mamba build -c ilastik-forge -c conda-forge conda-recipe
-      - name: osx conda build test
+        run: |
+          xvfb-run --server-args="-screen 0 1024x768x24" conda mambabuild --test --override-channels \
+            -c ./pkgs -c pytorch -c ilastik-forge -c conda-forge \
+            ./pkgs/noarch/${ILASTIK_VARIANT}-${ILASTIK_VERSION}-${ILASTIK_BLD_PREF}0.tar.bz2
```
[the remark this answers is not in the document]
k-dominik: that's the build number, it's always appended in [conda package names](https://docs.conda.io/projects/conda-build/en/latest/concepts/package-naming-conv.html) unless build string is overridden, which I don't want to do.

tests/test_ilastik/widgets/test_ImageFileDialog.py
k-dominik: it's test related :o. On windows runners on gitlab ci, the username is so long that tmp_path returns a "short name" (windows stuff... like `BLAH~1` instead of `blah-my-long-name`). ImageFileDialog seems to resolve those internally..

emilmelnikov: Are 8.3 filename restrictions still relevant these days?..

k-dominik: yeah :/

PR #2545, opened 2022-02-16
use darker for black on gh actions

k-dominik:
the black check started failing recently with empty change sets.
Moved this to a gh action with darker.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -85,39 +85,6 @@ jobs:
         paths:
         - /opt/conda/pkgs
 
-  black:
-    working_directory: /root/ilastik/ilastik
-    parallelism: 1
-    shell: /bin/bash --login
-    docker:
-    - image: continuumio/miniconda3
-    steps:
-    - restore_cache:
-        name: restore ilastik-repo cache
-        key: v1-repo-{{ .Environment.CIRCLE_SHA1 }}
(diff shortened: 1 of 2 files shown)


.github/workflows/statitc-coda-analysis.yml
```diff
@@ -0,0 +1,21 @@
+name: static-code-analysis
```
emilmelnikov, line 1: Typo in the file name. Also, maybe rename it to "code formatting"? We don't really do static analysis here (no mypy, no linters). Or do we intend to add these later?

.github/workflows/statitc-coda-analysis.yml
```diff
@@ -0,0 +1,21 @@
...
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+
+jobs:
+  black:
```
emilmelnikov, line 11: A minor point, but I'd rename this to `lint` or `format`, without referencing a particular tool.

emilmelnikov (approved):

.github/workflows/statitc-coda-analysis.yml
k-dominik: awww thanks :)

and yeah, I'd say that we should slowly introduce more static checking!

PR #2547, opened 2022-02-17
moved startup script to own module

k-dominik:
I forgot to include the startup script in the conda package for a reason ;)

I did try to use the setup.py scripts directive, but this will not work reliably as an entrypoint in conda.

note: this entrypoint has to run before any import of an ilastik module.


emilmelnikov:
Should we create an issue in the [`conda-build`](https://github.com/conda/conda-build) repo? This behaviour looks like a bug...

k-dominik:
okay this one is really annoying. I tried different variants, also following the conda-build tests that test that it actually should be working, but I never got an exe file on windows for the script when supplied as `scripts` in `setup.cfg` (hence I closed #2549, and reopened this one.). I think this is related to https://github.com/conda/conda-build/issues/3965.

emilmelnikov:
Do we even need an entrypoint for non-packaged ilastik variants? I think devs and people who want to use conda-based ilastik should be fine with `python ilastik.py ...` instead of `./ilastik ...`.

k-dominik:
I'm thinking not of devs but people who install ilastik using conda. They should be able to `ilastik` to start.

emilmelnikov (approved):
Maybe rename the module to `ilastik.main`, or at least `ilastik.startup`?

Issue #2557, opened 2022-03-08
Carving workflow on OSX (observed in Monterey)


k-dominik:
## Describe the bug
When starting the carving workflow it will complain about loading openGL

## Expected behavior
We know that we can't show the GL view, but normal carving should work.

Monkeypatching `ctypes` as described [here](https://github.com/PixarAnimationStudios/USD/issues/1372#issuecomment-912254279) makes it work again/

## To Reproduce

Create a new carving project

## Error message/traceback

If applicable, please add tracebacks with formatting, such as

```pytb
Starting ilastik from "/.../mambaforge/envs/ilastik-app/lib/python3.7".
ERROR 2022-03-08 08:35:49,451 log_exception 69078 8641803776 Traceback (most recent call last):
  File "/.../mambaforge/envs/ilastik-app/lib/python3.7/site-packages/OpenGL/platform/darwin.py", line 38, in GL
    mode=ctypes.RTLD_GLOBAL
  File "/.../mambaforge/envs/ilastik-app/lib/python3.7/site-packages/OpenGL/platform/ctypesloader.py", line 36, in loadLibrary
    return _loadLibraryWindows(dllType, name, mode)
  File "/.../mambaforge/envs/ilastik-app/lib/python3.7/site-packages/OpenGL/platform/ctypesloader.py", line 89, in _loadLibraryWindows
    return dllType( name, mode )
  File "/.../mambaforge/envs/ilastik-app/lib/python3.7/ctypes/__init__.py", line 364, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: ("dlopen(OpenGL, 0x000A): tried: '/.../mambaforge/envs/ilastik-app/lib/OpenGL' (no such file), '/.../mambaforge/envs/ilastik-app/lib/OpenGL' (no such file), '/.../mambaforge/envs/ilastik-app/lib/python3.7/lib-dynload/../../OpenGL' (no such file), '/.../mambaforge/envs/ilastik-app/lib/OpenGL' (no such file), '/.../mambaforge/envs/ilastik-app/bin/../lib/OpenGL' (no such file), 'OpenGL' (no such file), '/usr/local/lib/OpenGL' (no such file), '/usr/lib/OpenGL' (no such file), '/.../OpenGL' (no such file), '/usr/local/lib/OpenGL' (no such file), '/usr/lib/OpenGL' (no such file)", 'OpenGL', None)

During handling of the above exception, another exception occurred:

```

[...]

PR #2558, opened 2022-03-08
Fix OpenGL shared library lookup on macOS Monterey

emilmelnikov:
Closes https://github.com/ilastik/ilastik/issues/2557


k-dominik (approved):
:+1:  cool! Thank you for the quick fix!!

PR #2559, opened 2022-03-09
fix path setup for windows binary

k-dominik:
package structure has changed - `ilastik-meta` folder is not in `ilastik_root` anymore. Use `python.exe`, which is in `ilastik_root`, as an anchor.

This command line flag is not documented and will not show up with `ilastik --help`. It is only called in the context of `ilastik.exe`.

## Checklist

### DIFF
--- a/ilastik_scripts/ilastik_startup.py
+++ b/ilastik_scripts/ilastik_startup.py
@@ -4,6 +4,7 @@ import pathlib
 import platform
 import shlex
 import sys
+import warnings
 from typing import Iterable, Mapping, Sequence, Tuple, Union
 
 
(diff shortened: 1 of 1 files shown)


ilastik_scripts/ilastik_startup.py
```diff
@@ -93,8 +94,14 @@ def _parse_internal_config(path: Union[str, os.PathLike]) -> Tuple[Sequence[str]
...
-        ilastik_root = script_dir.parent.parent
+        # this is only used in the windows binary - there python.exe is
+        # directly in ilastik_root
+        if platform.system != "Windows":
+            warnings.warn(
+                f"--clean_paths argument only supported on windows and should only be used in the binary distribution"
+            )
+        ilastik_root = pathlib.Path(sys.executable).parent
+        assert (ilastik_root / "ilastik.exe").exists()
         _clean_paths(ilastik_root)
```
emilmelnikov, line 105: I think it's better to do nothing if preconditions for `--clean_paths` are not met:
```suggestion
        ilastik_root = pathlib.Path(sys.executable).parent
        if platform.system() == "Windows" and (ilastik_root / "ilastik.exe").is_file():
            _clean_paths(ilastik_root)
        else:
            warnings.warn("'--clean_paths' is supported only on Windows binary distributions")
```

PR #2562, opened 2022-03-23
replaced license dialog

k-dominik:
* removed old `buildem` references - `buildem` has not been used in eons
* `QDialog` as base in order to add custom btns
* open long licenses via `webbrowser`: the old way of showing as "details" was sorta ugly.
* corrected path for license when in source distribution/conda/development mode
* remove duplicate ilastik logo
* make logo look more smooth in dialog
* added tests

once this is merged, I'll add the environment variables during release process

## Checklist

- [x] Add tests.

before:
Screenshot 2022-03-23 at 07 56 17

this PR:

Screenshot 2022-03-23 at 07 55 39


ilastik/shell/gui/licenseDialog.py
```diff
@@ -20,98 +20,119 @@
...
+            partial(show_license, license_path=license_path, error_message=self.LICENSE_ERROR)
+        )
+        self._show_details_btn = show_details_btn
+        show_3rd_party_btn = QPushButton("Show third party licenses")
+        show_3rd_party_btn.clicked.connect(
+            partial(show_license, license_path=license_3rd_party_path, error_message=self.LICENSE_3RD_PARTY_ERROR)
+        )
+        self._show_3rd_party_btn = show_3rd_party_btn
+        btnbox = QHBoxLayout()
+        btnbox.addItem(QSpacerItem(0, 0, QSizePolicy.Expanding, QSizePolicy.Minimum))
```
emilmelnikov, line 122:
```suggestion
        btnbox.addStretch()
```

k-dominik: looks like I've been using the designer too much...

ilastik/shell/gui/licenseDialog.py
```diff
@@ -20,98 +20,119 @@
...
+        super().__init__(parent=parent)
 
-        self.setText(self.LICENSE_SHORT)
+        ilastik_path = Path(ilastik.__file__).parent.parent
 
-        global ilastik_path
-        logo_path = os.path.join(ilastik_path, "shell", "gui", "ilastik-logo-alternate-colors.png")
+        # environment variable set in release binary
+        license_path = Path(os.environ.get("ILASTIK_LICENSE_PATH", ilastik_path / "LICENSE"))
+        license_3rd_party_path = Path(os.environ.get("ILASTIK_LICENSE_3RD_PARTY_PATH", "NOTFOUND"))
```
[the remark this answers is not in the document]
k-dominik: the license info is generated (in fact into a single 30k+ lines textfile) during packaging the binary - different OSs and configurations result in different third party packages being included. I searched around the internet a bit in order to see how other people do it and it seems pretty common to keep the main license and third-party stuff separate. But maybe my search was biased towards this kind of arrangement. While IANAL, to me it seems to be the way.

Generating some additional package that has all the licenses is more complicated than having _two_ env variables when in release mode, imo.

Given that the 3rd party bit is 30k+ lines long, I'd at least want people to have the ability to search - hence I thought it would make sense to open information in a browser (or... default program - that behavior is a bit undefined for local files).

PR #2566, opened 2022-04-19
fix version str

k-dominik:
previously build str was not included, so version would currently be `1.4.0` as opposed to `1.4.0b27`.

Also added setuptools_scm to dev dependencies and made sure there is a proper version set in dev mode, too.

## Checklist

### DIFF
--- a/ilastik/__init__.py
+++ b/ilastik/__init__.py
@@ -25,29 +25,29 @@ import re
 ################################
 # # Add Submodules to sys.path ##
 ################################
-import os
 import h5py
-import time
-from typing import Optional, Iterable, List
+from typing import Optional, List
 from pkg_resources import parse_version
 
 try:
-    from ._version import version_tuple
-except ImportError:
-    version_tuple = (1, 4, "0b20")
+    from ._version import version
+except ModuleNotFoundError:
+    raise RuntimeError(
+        "Couldn't determine ilastik version - if you are developing ilastik please "
+        "make sure to use an editable install (`pip install -e .`). "
+        "Otherwise, please report this issue to the ilastik development team: team@ilastik.org"
+    )
+
 ##################
 # # Version info ##
 ##################
+
+__version__ = version
 
 
 def _format_version(t):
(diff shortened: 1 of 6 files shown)


emilmelnikov:
I managed to configure automatic version strings in conda packages using the following approach:

1. Add `setuptools-scm` to `host` dependencies, and `importlib-metadata` to `runtime` dependencies in `meta.yaml`.

2. Create `setup.py` which uses `setuptools_scm`:

    ```python
    from setuptools import setup
    from setuptools_scm import get_version

    setup(version=get_version())
    ```

3. Get version string from the package metadata:
    ```python
    from importlib_metadata import version

    __version__ = version("package-name")

    del version
    ```

No need for runtime dependency on `setuptools_scm` or generated `_version.py`.

Not exactly sure how to use this in the ilastik case, but the approach of getting version string from package metadata seems more correct to me.

k-dominik:
Hey thanks @emilmelnikov!

Currently, metadata, as well as the `_version` file are created in the same line (when not in development "mode"). So I think packaging is fine, extra file or not, while I also agree that it might be more pure to only rely on metadata.

I wanted to avoid having to do editable installs - but I definitely also want to avoid having to bump any version manually when developing, so I guess there is no way around that. I updated the `CONTRIBUTING` guidelines accordingly.

emilmelnikov:
> I wanted to avoid having to do editable installs

What could be the reasons for that?

ilastik/__init__.py
```diff
@@ -25,31 +25,31 @@
 ################################
 # # Add Submodules to sys.path ##
 ################################
-import os
 import h5py
-import time
-from typing import Optional, Iterable, List
+from typing import Optional, List
 from pkg_resources import parse_version
```
emilmelnikov, line 30: We probably shouldn't depend on modules from `setuptools` in runtime, such as `pkg_resources`. It could be replaced by the [`packaging`](https://packaging.pypa.io/en/latest/version.html) package.

Issue #2561, opened 2022-03-10
Headless Mode Multicut broken


constantinpape:
Running multicut in headless mode fails with a (very long exception). I have tried this with `1.4b021` and `1.4b021` on two linux systems.

## To reproduce

The data (and script) is located here:  `/g/kreshuk/pape/Work/bioimageio/use-cases/case1`
(The call is wrapped into python to be able to run this for multiple blocks.)

```python
import subprocess

def segment_mc(block_id):
    ilastik_exe = "/g/kreshuk/pape/Work/software/src/ilastik/ilastik-1.4.0b25-Linux/run_ilastik.sh"
    ilp = "./segmentation.ilp"

    raw_path = f"./data/blocks/raw_block{block_id}.h5"
    pmap_path = f"./data/blocks/pred_block{block_id}.h5"
    out_path = f"./data/blocks/segmentation_block{block_id}.h5"
    cmd = [ilastik_exe, "--headless",
           f'--project={ilp}', f'--raw_data={raw_path}',
           f'--probabilities={pmap_path}', '--export_source=Multicut Segmentation',
           f'--output_filename_format={out_path}']
    subprocess.run(cmd)

segment_mc(block_id=0)
```

## Exception

The exception is so long that it exceeds my shell buffer...
Here is the very last bit (that's likely not very relevant):
```
  File "/g/kreshuk/pape/Work/software/src/ilastik/ilastik-1.4.0b25-Linux/lib/python3.7/site-packages/lazyflow/request/request.py", line 569, in _wait                                                      
    self._wait_within_request(current_request)                                                                                                                                                             
  File "/g/kreshuk/pape/Work/software/src/ilastik/ilastik-1.4.0b25-Linux/lib/python3.7/site-packages/lazyflow/request/request.py", line 654, in _wait_within_request                                       
    raise_with_traceback(exc_type(exc_value), exc_tb)                                                                                                                                                      
TypeError: __init__() takes 1 positional argument but 2 were given     
```

A warning that pops up at some point before the exception is thrown and might be more relevant:
```
WARNING 2022-03-10 19:51:25,894 opEdgeTraining 21077 140449621325632 Superpixels changed.  Deleting all labels in lane 0.                                                                                  
[...]

PR #2569, opened 2022-04-20
Fix project deserialization in multicut

k-dominik:
makes sure the classifier is properly deserialized last, so that it is not set dirty by accident (and then retrained)...

update: I looked into why the labels are removed at the end of headless batch processing (and also when closing the project) and it happens during cleanup, - in principle we don't care about the removal there, so it's now skipped during cleanup.

partly addresses #2561

this PR doesn't improve command line handling or error messages, that are also part of #2561

### DIFF
--- a/ilastik/workflows/edgeTrainingWithMulticut/edgeTrainingWithMulticutWorkflow.py
+++ b/ilastik/workflows/edgeTrainingWithMulticut/edgeTrainingWithMulticutWorkflow.py
@@ -20,6 +20,7 @@
 ###############################################################################
 import argparse
 import numpy as np
+import sys
 
 from ilastik.workflow import Workflow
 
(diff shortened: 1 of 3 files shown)


ilastik/workflows/edgeTrainingWithMulticut/edgeTrainingWithMulticutWorkflow.py
```diff
@@ -20,6 +20,7 @@
 ###############################################################################
 import argparse
 import numpy as np
+import sys
```
emilmelnikov, line 23:
```suggestion
```

emilmelnikov (approved):

ilastik/workflows/edgeTrainingWithMulticut/edgeTrainingWithMulticutWorkflow.py
k-dominik: sys is actually used later in this file - I guess that import was missing

PR #2577, opened 2022-05-11
bump vigra, default to cf pytorch-cpu, remove pytorch channel

k-dominik:
* vigra: use the latest build
* use pytorch-cpu from cf in favor of pytorch + cpuonly from pytorch channel
* remove pytorch channel from osx, linux ci
* only windows gpu build requires pytorch channel


k-dominik:
:/ cf pytorch version for windows don't go back far enough :/

PR #2583, opened 2022-06-02
Update CONTRIBUTING.md

jonashund:
I encountered some issues attempting to install `mamba` following the how-to. Added the alternative commands that worked in my case.
Also removed references to `conda` after the installation of `mamba`.

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,7 +33,7 @@ The following text equips you with knowledge that makes contributing to ilastik 
 
    ```
    conda config --set channel_priority strict
-   conda install --name base -c conda-forge mamba
+   conda install --name base --channel conda-forge mamba
    ```
 
 1. Create a new environment and install dependencies:
@@ -50,8 +50,8 @@ The following text equips you with knowledge that makes contributing to ilastik 
    ```
    conda activate ilastik
    # from the folder containing ilastik and volumina
-   pip install -e ilastik
-   pip install -e volumina
+   pip install --editable ilastik
+   pip install --editable volumina
    ```
 
 1. Install pre-commit hooks:


CONTRIBUTING.md
```diff
@@ -33,7 +34,34 @@ The following text equips you with knowledge that makes contributing to ilastik
...
+   conda install mamba -n base -c conda-forge
+   ```
+
+   If [strict channel priority][strict-channel-priority] does not allow installation of [Mamba][mamba]:
+
+   ```
+   conda config --set channel_priority false
+   conda install mamba -n base -c conda-forge
+   ```
+
```
k-dominik, line 46: In general we would like to keep our Contributing guidelines a help for setting up ilastik development environments. While it might be helpful in some situations, I think troubleshooting conda/mamba should be done in the appropriate places. For this particular suggestions I would also want to add that not using strict channel priority is almost always a bad idea, as you might pull together incompatible builds.

CONTRIBUTING.md
```diff
@@ -33,7 +34,34 @@ The following text equips you with knowledge that makes contributing to ilastik
 
    ```
    conda config --set channel_priority strict
-   conda install --name base -c conda-forge mamba
+   conda install mamba -n base -c conda-forge
```
k-dominik, line 37: In general this guide is thought to be a set of copy-paste operations.

not sure why short options would be better than long options. "Experts" already know them both and can probably replace while typing, while novice users profit from the added verbosity of long options.

CONTRIBUTING.md
```diff
@@ -33,7 +34,34 @@ The following text equips you with knowledge that makes contributing to ilastik
...
+   ```shell
+   if [ -f "/opt/conda/etc/profile.d/mamba.sh" ]; then
+      . "/opt/conda/etc/profile.d/mamba.sh"
+   fi
+   ```
+
+   Restart shell or source:
+
+   ```shell
+   source ~/.zshrc
```
k-dominik, line 64: This is a very peculiar case of someone having done an installation of conda-/mini-/mambaforge without adding it to their profile. While this fix is useful, I think it is too much of a special case to be included in the contributing guidelines.

k-dominik (changes requested):
Hello @jonashund,

thank you very much for opening this PR! Making ilastik easier to contribute to is super important to us so problems with the contributing guidelines have a high priority!

I left some comments. A lot of the modifications seem to be related to a (maybe old/outdated/cluttered) conda base environment.  I tried with fresh conda-forge and mamba-forge installations and the guide can be followed without isses, which makes me think that the solutions you add are very specific to the state of your conda installation/configuration. Rather than adding those, maybe a link to general conda troubleshooting would be useful to a wider range of users.

Thank you again, cheers
Dominik

CONTRIBUTING.md
jonashund: Didn't think about that.
Yet, the usage of both short `-c` and long option `--name` also is inconsistent in that regard.

So maybe `conda install --name base --channel conda-forge` mamba would be best suited here?

k-dominik: yes, very much so!

k-dominik (approved):
thank you very much @jonashund!

k-dominik:
hey @jonashund, I removed the two first commits (as their content was removed in the third) and also changed the committer and author of the commit to be associated with your github account (I think you might have recently changed the configuration there, or are were committing from machines that had different author settings). If that is okay with you, I will merge :)

Issue #2590, opened 2022-07-13
Adding a reset button to export path


sebherbert:
## Description
Adding a reset button for the export path

## Scenario
Simply click on the button and the export path is restored to 
{dataset_dir}/{nickname}_{result_type}.h5

## Rationale
If I modify the path and want to come back to the magic descriptors, I have to fetch them from the website.
Also when selecting a folder for the output the path automatically changes to my/selected/path/{dataset_dir}.h5 and removes the last {nickname} and {result_type}.

Thanks!

PR #2599, opened 2022-08-04
Added pin for click in black pre-commit hook

sebherbert:
https://github.com/psf/black/issues/2964

Hi,
I'm working with @k-dominik on the #2590 issue and we realized pre-commit was crashing on my machine unless I fixed the click version.
This PR modifies the .pre-commit-config.yaml file to fix _click_ at version 8.0.4 that solves the pre-commit problem.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,3 +3,4 @@ repos:
     rev: 21.7b0
     hooks:
       - id: black
+        additional_dependencies: ['click==8.0.4']


k-dominik (approved):
Dear @sebherbert, thank you so much for your first contribution! Looks all good.

One note though: there will be a conflict between your fork's main branch (`sebherbert:main`) and ours after the merge (`ilastik:main`) because you committed to main directly for this PR. This means you will have to do some reset hackery and force-pushery on your fork to sync them up again ..

or you remove your fork afterwards and create a new one :slightly_smiling_face:

Thank you, again!

imagejan:
Instead of pinning to an old version of `click`, how about pinning to a newer version of `black` instead? According to https://github.com/psf/black/issues/2964#issuecomment-1081035937, `black>=22.3.0` should contain a fix for this issue, and this (also pinning to a version _range_ instead of a fixed version) would also create less problems going forward in the future, when existing environment might bring in newer dependencies...

k-dominik:
Hi @imagejan,

thanks for the comment. We discussed this offline and updating black always involves slight style changes, so doing that would be a more conscious effort, probably involving running black on the whole repo again - and we opted against it for now. But it's definitely planned for the future!

Cheers
Dominik

PR #2632, opened 2022-11-14
Neural Network Worfklow UI improvements

k-dominik:
Summary:

* removed the "Labels" layer, as it cannot be interacted with (currently) but is still shown in the layerstack
* added right-click interaction on Prediction layers to change layer color (it uses the same method as going via the label list. Note: Labellist is currently not shown -> no other way to change colors).
* fixed a regression introduced in https://github.com/ilastik/ilastik/commit/1d4eb62deaf3a890435304e27a34994a75e0a210 that caused model information from a model saved in the project file not being shown when project was loaded.

## Checklist

### DIFF
--- a/ilastik/applets/neuralNetwork/nnClassGui.py
+++ b/ilastik/applets/neuralNetwork/nnClassGui.py
@@ -39,6 +39,8 @@ from PyQt5.QtCore import (
 )
 from PyQt5.QtGui import QColor, QIcon
 from PyQt5.QtWidgets import (
+    QAction,
+    QColorDialog,
     QComboBox,
     QDesktopWidget,
     QDialog,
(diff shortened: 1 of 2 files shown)


ilastik/applets/neuralNetwork/nnClassGui.py
```diff
@@ -568,6 +570,46 @@ def _changeInteractionMode(self, toolId):
...
+        def setPredLayerName(n, predictLayer_=predictionLayer, initializing=False):
+            """
+            function for setting the names for every Channel
+            """
+            if not initializing and predictLayer_ not in self.layerstack:
+                # This layer has been removed from the layerstack already.
+                # Don't touch it.
+                return
+            newName = "Prediction for %s" % n
+            predictLayer_.name = newName
```
emilmelnikov, line 596:
```suggestion
        def setLayerName(name, layer=predictionLayer, initializing=False):
            if initializing or layer in self.layerstack:
                layer.name = f"Prediction for {name}"
```

emilmelnikov (commented):
These are obvious simplifications, however I'm not sure we should apply them because other `*Gui` classes have very similar code.

k-dominik:
thx @emilmelnikov, no need to copy bad examples, your improvements are appreciated :)

PR #2633, opened 2022-11-15
release pin on pytest

k-dominik:
Final blocking thing preventing us from updating Python was pytest. This updates pytest from `3.x` pin to 7.2.0 (latest, not pinned)

There were some small real issues

* `tmp_path` magic that would resolve symlinks vs `tempfile.mk*` not doing it.
* in one of the tests, we use [`pytester`](https://docs.pytest.org/en/6.2.x/reference.html#pytest.Pytester), which is supposed to run tests in a more isolated manner (guess that was the motivation for using it, from the docs I take it is used to test pytest itself). Behavior of `pytester` `testdir` fixture seems to have changed - since pytest 3.x. I didn't dig into the pytest source code, but somehow invoking `run` seems to alter the environment for subsequent calls to `run` in the same test function. Splitting those functions means better reporting anyway. (In addition we could think about dropping the use of `pytester` here - I'm not sure it's supposed to be used outside the context of testing pytest itself).

while debugging I simplified `appveyor.yml` a bit - but [these changes](https://github.com/ilastik/ilastik/pull/2633/commits/293caf0344eae8ec25791ea0fbf52562ca16d362) are unrelated and I can also remove them.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,7 +11,7 @@ environment:
   APPVEYOR_CACHE_ENTRY_ZIP_ARGS: -xr!*/ -ir-!*.tar.bz2 -ir-!*.conda  # Exclude directories only cache downloaded tars
 
 cache:
-  - C:\Miniconda37-x64\pkgs -> appveyor.yml
+  - C:\Miniconda38-x64\pkgs -> appveyor.yml
 
 install:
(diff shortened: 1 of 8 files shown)


btbest (approved):
Looks like everything runs, and some neat improvements in the tests 👍

k-dominik:
thank you for the review @btbest !

PR #2636, opened 2022-11-29
Fix suggest features for narrow stacks

### LINKED ISSUE #1280: Automated feature selection should gracefully handle data sets that are too small for the largest filters in at least one dimension
> When I first ran Ilastik, I loaded my images and selected all features.
> Then, Ilastik complained that the largest (sigma 10) features are too 
> big for my data. No problem with that, just deselected it.
> 
> But when I want to run the "suggest features", I get the very same 
> error message.

k-dominik:
Enable _SuggestFeatures_ Dialog to deal with "narrow" stacks. Previously it would error out and might even produce segfaults :o.

On the way there was
* refactoring - naming _is_ important. It is frustrating every time I am looking for this particular dialog. Now naming reflects what is shown in the UI.
* made the dialog properly modal and made it run in the same eventloop
* made its `_run_selection` method somewhat more robust to errors - at least the graph is reverted to the same state as before
* by adding some attributes to the error that is raised by the `OpFeatureSelection` when incompatible scales are used, the dialog now automatically deselects the ones that do not work (for features that only don't fit in _z_, they are now calculated in 2D).

See also #2637 

Bottom line: Ideally this whole functionality would be reworked from the ground up. This is just a patch.

fixes #1280 

## Checklist

- [x] Add tests.

### DIFF
--- a/ilastik/applets/featureSelection/__init__.py
+++ b/ilastik/applets/featureSelection/__init__.py
@@ -1,5 +1,3 @@
-from __future__ import absolute_import
-
 ###############################################################################
 #   ilastik: interactive learning and segmentation toolkit
 #
@@ -20,4 +18,44 @@ from __future__ import absolute_import
 # on the ilastik web site at:
 # 		   http://ilastik.org/license.html
 ###############################################################################
+from typing import Callable, Iterable, Union
+
+from ilastik.applets.base.applet import DatasetConstraintError
+
+_SCALES = Iterable[Union[int, float]]
+
+
+class FeatureSelectionConstraintError(DatasetConstraintError):
+    def __init__(
+        self,
+        appletName: str,
+        invalid_scales: _SCALES,
+        invalid_z_scales: _SCALES,
+        fixing_dialogs=Iterable[Callable],
+    ):
+        """
+        Args:
+          appletName: applet where the exception is happening
+          invalid_scales: list of scales that are not compatible in the x-y plane
+          invalid_z_scales: list of scales that are not compatible in the z-plane
+          fixing_dialogs: list of functions to show dialogs which can alleviate the dataset constraint.
+        """
+        message = "\nSome of your selected feature scales are too large for your dataset.\n"
+        if invalid_scales:
+            message += f"Reduce or remove these scales:\n{invalid_scales}\n\n"
+
+        if invalid_z_scales:
+            message += f"Reduce, remove or switch to 2D computation for these scales:\n{invalid_z_scales}\n\n"
+
+        message += "Alternatively use another dataset."
+
+        super().__init__(appletName, message)
+        self.fixing_dialogs = fixing_dialogs
+        self.invalid_scales = invalid_scales
+        self.invalid_z_scales = invalid_z_scales
+
+    def __str__(self):
+        return f"Constraint of '{self.appletName}' applet was violated: {self.message}"
+
+
 from .featureSelectionApplet import FeatureSelectionApplet
--- a/ilastik/applets/featureSelection/__init__.py
+++ b/ilastik/applets/featureSelection/__init__.py
@@ -18,11 +18,11 @@
 # on the ilastik web site at:
 # 		   http://ilastik.org/license.html
 ###############################################################################
-from typing import Callable, Iterable, Union
+from typing import Callable, Iterable, Sequence, Union
 
 from ilastik.applets.base.applet import DatasetConstraintError
 
-_SCALES = Iterable[Union[int, float]]
+_SCALES = Sequence[Union[int, float]]
 
 
 class FeatureSelectionConstraintError(DatasetConstraintError):
@@ -30,8 +30,8 @@ class FeatureSelectionConstraintError(DatasetConstraintError):
         self,
         appletName: str,
         invalid_scales: _SCALES,
-        invalid_z_scales: _SCALES,
-        fixing_dialogs=Iterable[Callable],
+        invalid_z_scales: _SCALES = (),
+        fixing_dialogs: Iterable[Callable] = (),
     ):
         """
         Args:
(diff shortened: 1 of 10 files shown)


tests/test_ilastik/test_applets/pixelClassification/testPixelClassificationGui.py
```diff
@@ -266,8 +267,8 @@ def impl():
             selections = numpy.array(
                 [
                     [True, False, False, False, False, False, False],
-                    [True, False, False, False, False, False, False],
-                    [True, False, False, False, False, False, False],
+                    [False, True, False, False, False, False, False],
+                    [False, True, False, False, False, False, False],
```
k-dominik, line 271: previous feature set was "illegal" for ilastik. It is not possible to achieve using the gui.

But main reason for changing it is that the SuggestFeatures Dialog that later is used in this test, makes this assumption

ilastik/applets/pixelClassification/suggestFeaturesDialog.py
```diff
@@ -767,157 +777,186 @@ def _auto_select_num_features(self, feature_order):
...
+                    compute_in_2d[numpy.isin(scales, err.invalid_z_scales)] = True
+                    self.opFeatureSelection.ComputeIn2d.setValue(compute_in_2d.tolist())
+                    self.compute_in_2d_compat = compute_in_2d.tolist()
+                    msg += f"Calculating features at sigmas {err.invalid_z_scales} in 2D only."
+
+                logger.info(msg)
+
+                # suppress the popup in pytest:
+                if "PYTEST_CURRENT_TEST" not in os.environ:
+                    QMessageBox.information(self, "Not all features are compatible with your data!", msg)
```
k-dominik, line 819: I am unhappy to have pytest-related shenanigans in production code, but alternative would be to keep a handle to the message box around in the dialog itself (I didn't manage to get a handle on it with `.activeModalWidget()`...

tests/test_ilastik/test_applets/pixelClassification/testPixelClassificationGui.py
```diff
@@ -678,3 +679,76 @@ def impl():
...
+            gui = pixClassApplet.getMultiLaneGui()
+            opFeatures = workflow.featureSelectionApplet.topLevelOperator
+
+            # this time we want to trigger the message box that informs us,
+            # that some sigmas can only be computed in 2D
+            # so we add a sigma that doesn't fit into size of 50 along z:
+            scales = [0.3, 0.7, 15.0]
+            selections = numpy.array(
+                [
+                    [True, False, False],
```
emilmelnikov, line 718: A readability idea: use `0` and `1` instead of `False` and `True`, and then `.astype(bool)` the array. Single-digit numbers are shorter and line up better than boolean constants. Could use this trick in other places too.

tests/test_ilastik/test_applets/pixelClassification/testPixelClassificationGui.py
```diff
@@ -678,3 +679,76 @@ def impl():
...
+                    [False, True, False],
+                    [False, False, False],
+                    [False, False, False],
+                    [False, False, False],
+                ]
+            )
+
+            opFeatures.SelectionMatrix.setValue(None)
+            opFeatures.Scales.setValue(scales)
+            opFeatures.ComputeIn2d.setValue([False for _ in scales])
```
emilmelnikov, line 729:
```suggestion
            opFeatures.ComputeIn2d.setValue([False] * len(scales))
```

ilastik/applets/featureSelection/__init__.py
```diff
@@ -20,4 +18,44 @@
...
+_SCALES = Iterable[Union[int, float]]
+
+
+class FeatureSelectionConstraintError(DatasetConstraintError):
+    def __init__(
+        self,
+        appletName: str,
+        invalid_scales: _SCALES,
+        invalid_z_scales: _SCALES,
+        fixing_dialogs=Iterable[Callable],
```
emilmelnikov, line 34: Default arguments for a seemingly common case:
```suggestion
        invalid_z_scales: _SCALES = (),
        fixing_dialogs=Iterable[Callable] = (),
```

ilastik/applets/pixelClassification/suggestFeaturesDialog.py
```diff
@@ -1,33 +1,33 @@
...
-import re
+
+import logging
+
+logger = logging.getLogger(__name__)
 
 
 # just a container class, nothing fancy here
-class FeatureSelectionResult(object):
+class SuggestFeaturesResult(object):
```
emilmelnikov, line 30:
```suggestion
class SuggestFeaturesResult:
```

ilastik/applets/pixelClassification/suggestFeaturesDialog.py
```diff
@@ -767,157 +777,186 @@ def _auto_select_num_features(self, feature_order):
...
+                )
+                self._add_feature_set_to_results(current_features_result)
+                self._initialized_current_features_segmentation_layer = True
 
-        # revert changes to matrix
-        self.opFeatureSelection.SelectionMatrix.setValue(user_defined_matrix)
-        self.opFeatureSelection.SelectionMatrix.setDirty()  # this does not do anything!?!?
-        self.opFeatureSelection.setupOutputs()
-        QtWidgets.QApplication.instance().restoreOverrideCursor()
+        finally:
```
emilmelnikov, line 941: I wonder if a context manager that reverts slot values on `__exit__` would be more readable and robust.

emilmelnikov (commented):
I've reviewed some parts of the new code, but left alone other parts that were unchanged, because we agree that it required a rewrite anyway.

ilastik/applets/featureSelection/__init__.py
```diff
@@ -20,4 +18,44 @@
 # on the ilastik web site at:
 # 		   http://ilastik.org/license.html
 ###############################################################################
+from typing import Callable, Iterable, Union
+
+from ilastik.applets.base.applet import DatasetConstraintError
+
+_SCALES = Iterable[Union[int, float]]
```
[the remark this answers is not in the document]
k-dominik: thx!

ilastik/applets/pixelClassification/suggestFeaturesDialog.py
k-dominik: agreed, I'll give it a go and see how it looks. Thx! :)

k-dominik:
thx for the review @emilmelnikov

PR #2644, opened 2022-12-20
remove unnecessary setupOutputs

k-dominik:
The operator doesn't need `setupOutputs`, inner ops are fully connected. Metadata is already set in `OpSimpleBlockedArrayCache`, nothing should be necessary in this Op.

Probably leftover code from refactoring to using `OpSimpleBlockedArrayCache`

`setupOutputs` here also had a strange behavior of conditionally setting the value of an input slot: `Blockshape`. When changing the input to this op, the Blockshape would not update 💣 

Also updated the documentation of this op and removed commented code for clarity.

### DIFF
--- a/lazyflow/operators/opBlockedArrayCache.py
+++ b/lazyflow/operators/opBlockedArrayCache.py
@@ -58,7 +58,7 @@ class OpBlockedArrayCache(Operator, ManagedBlockedCache):
 
         # SCHEMATIC WHEN BypassModeEnabled == False:
         #
-        # Input ---------> opCacheFixer -> opSimpleBlockedArrayCache -> (indirectly via execute) -> Output
+        # Input ---------> opCacheFixer -> opSimpleBlockedArrayCache -> Output
         #                 /               /
         # fixAtCurrent --                /
         #                               /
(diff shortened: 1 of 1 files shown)


emilmelnikov (approved):

lazyflow/operators/opBlockedArrayCache.py
```diff
@@ -80,23 +80,12 @@ def __init__(self, *args, **kwargs):
...
         # This member is used by tests that check RAM usage.
         self.setup_ram_context = RamMeasurementContext()
         self.registerWithMemoryManager()
 
     def setupOutputs(self):
-        if not self.BlockShape.connected() and not self.BlockShape.ready():
-            self.BlockShape.setValue(self.Input.meta.shape)
-        # Copy metadata from the internal pipeline to the output
-        self.Output.meta.assignFrom(self._opSimpleBlockedArrayCache.Output.meta)
+        pass
```
btbest, line 88: Is there a reason to keep this setupOutputs stub?

btbest (approved):

k-dominik:
thx for having a look @emilmelnikov and @btbest

Issue #2081, opened 2019-08-15
Update Watershed functionality


constantinpape:
We had a meeting with @wolny @akreshuk and @boykovdn today to discuss updating the multicut workflow to reflect the developments of recent years and make it more user-friendly.
We settled onupdating the watershed functionality first, tackling 3 issues:

- [ ] Use new distance transform watershed implementation.
- [ ] Use common watershed functionality in carving and multicut.
- [ ] Enable lazy preview of watershed results.

I will add issues about the individual points later.

boykovdn:
- [ ] Make training a Random Forest Classifier optional in Boundary-based Segmentation with Multicut workflow.

PR #2645, opened 2022-12-20
parallel dt watershed in multicut using elf

k-dominik:
**Summary**:

This PR enables parallel waterershed in multicut following the pattern outlined in [carvingtools.py](https://github.com/ilastik/ilastik/blob/4f673a68ff5b49b52c6b911d67ceec6234260810/ilastik/workflows/carving/carvingTools.py#L108)

Note, that in future, we want to [use the same watershed methods in Carving and Multicut](https://github.com/ilastik/ilastik/issues/2081), but this will not be addressed in this PR.

Quirks to find here:
* in order to make this backwards compatible I added an additional slot - `BlockwiseWatershed` to the op. In order to not do it blockwise for legacy projects, I didn't see another way then reviving the deprecated default value for serialSlots. For now I put it into a separate class. The deprecation [needs a separate investigation](). If someone has a better idea how to solve this without losing annotations due to dirtyness (say, post-deserialization) I'd be happy to adapt the code.
* needed to adapt the gui multicut test, as the parallel watershed used here will relabel the volume (after removing halo) and hence introduce different superpixel indices.
* needed to adapt `test_Consistency.py`, which is a bit of a weird test on its own - not sure it achieves something of value.

Todo:

- [x] add compatibility mode for legacy projects - should not use parallel watershed (classifier is not trained on such data and probably hasn't seen stuff like block boundaries
- [x] add test for compatibility mode

### DIFF
--- a/ilastik/applets/wsdt/opWsdt.py
+++ b/ilastik/applets/wsdt/opWsdt.py
@@ -1,13 +1,142 @@
 from collections import OrderedDict
 import numpy as np
+from typing import Optional, Sequence
 
 from elf.segmentation.watershed import distance_transform_watershed
+from elf.parallel.common import get_blocking
+
+import vigra
 
 from lazyflow.utility import OrderedSignal
+from lazyflow.request import Request
 from lazyflow.graph import Operator, InputSlot, OutputSlot
 from lazyflow.roi import roiToSlice, sliceToRoi
-from lazyflow.operators import OpBlockedArrayCache, OpValueCache, OpMetadataInjector
-from lazyflow.operators.generic import OpPixelOperator, OpSingleChannelSelector
+from lazyflow.operators import OpBlockedArrayCache, OpMetadataInjector
+from lazyflow.operators.generic import OpPixelOperator
+from lazyflow.utility.timer import Timer
+
+from concurrent.futures import ThreadPoolExecutor
+import concurrent.futures
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def parallel_watershed(
+    data: np.ndarray,
+    threshold: float,
+    sigma_seeds: float,
+    sigma_weights: float,
+    minsize: int,
+    alpha: float,
+    pixel_pitch: Sequence[float],
+    non_max_suppression: bool,
+    block_shape: Optional[Sequence[int]] = None,
+    halo: Optional[Sequence[int]] = None,
+    max_workers: Optional[int] = None,
+):
+    """Parallel dt watershed with hard block boundaries.
+
+    parallel wrapper around elf.segmentation.watershed.distance_transform_watershed
+
+    Args:
+      data: data to run watershed on, ndarray with either 2 or 3 dims
+      threshold: data will be thresholded at this value, distance transform will be calculated from
+        the remaining mask
+      sigma_seeds: smoothing factor that is applied to the watershed seed map
+      sigma_weights: smoothing factor that is applied to the watershed weight map
+      minsize: minimal boundary size in pixels
+      alpha: how to blend data and distance transform
+      pixel_pitch: anisotropy factor: pixel distance along the three axes
+      non_max_suppression: flag to enable apply non-maxmimum suppression to filter out seeds
+      block_shape: size of blocks to process, defaults to 128 for 3D, 512 for 2D in each spacial dimension
+      halo: portion of each block to discard after processing for smoother boundary regions
+        if not specified: 10 voxels around the block in each direction
+      max_workers: if not specified or None, will use number of workers in the global Requests threadpool
+
+    """
+
+    logger.info(f"blockwise watershed with {max_workers} threads.")
+    shape = data.shape
+    ndim = len(shape)
+
+    assert ndim in [2, 3], "Watershed segmentor will only work on 2D and 3D data"
+
+    base_block = 512 if ndim == 2 else 128
+    # check for None arguments and set to default values
+    block_shape = (base_block,) * ndim if block_shape is None else block_shape
+    # nifty requires the halo shape to be of type list
+    halo = [10] * ndim if halo is None else halo
+    blocking = get_blocking(data, block_shape, roi=None)
+
+    if max_workers is None:
+        max_workers = max(1, Request.global_thread_pool.num_workers)
+
+    n_blocks = blocking.numberOfBlocks
+
+    labels = np.zeros_like(data, dtype=np.uint32)
+
+    # watershed for a single block
+    def ws_block(block_index):
+        nonlocal labels
+
+        # get the block with halo and the slicings corresponding to
+        # the block with halo, the block without halo and the
+        # block without halo in loocal coordinates
+        block = blocking.getBlockWithHalo(blockIndex=block_index, halo=halo)
+        inner_slicing = roiToSlice(block.innerBlock.begin, block.innerBlock.end)
+        outer_slicing = roiToSlice(block.outerBlock.begin, block.outerBlock.end)
+        inner_local_slicing = roiToSlice(block.innerBlockLocal.begin, block.innerBlockLocal.end)
+
+        with Timer() as btimer:
+            # write watershed result to the label array
+            ws_outer, _ = distance_transform_watershed(
+                data[outer_slicing],
+                threshold,
+                sigma_seeds,
+                sigma_weights,
+                minsize,
+                alpha,
+                pixel_pitch,
+                non_max_suppression,
+            )
+
+        logger.debug(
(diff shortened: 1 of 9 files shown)


ilastik/applets/wsdt/opWsdt.py
```diff
@@ -1,13 +1,141 @@
...
+
+        labels[inner_slicing] = ws_inner
+        # return the max-id for this block, that will be used as offset
+        return ws_inner.max()
+
+    with Timer() as wstimer:
+        # run the watershed blocks in parallel
+        with ThreadPoolExecutor(max_workers=max_workers) as executor:
+            tasks = [executor.submit(ws_block, block_index) for block_index in range(n_blocks)]
+            offsets = np.array([t.result() for t in tasks], dtype="int64")  # TODO uint32
```
emilmelnikov, line 116:
```suggestion
        with ThreadPoolExecutor(max_workers=max_workers) as executor:
            offsets = np.fromiter(executor.map(ws_block, range(n_blocks)), dtype=np.int64)
```

ilastik/applets/wsdt/opWsdt.py
```diff
@@ -1,13 +1,141 @@
...
+
+    logger.info(f"parallel ws took {wstimer.seconds()} s")
+    # compute the block offsets and the max id
+    last_max_id = offsets[-1]
```
emilmelnikov, line 120: If I'm not mistaken, we could simplify offset computation like this:

```python
def ws_block(...):
    # Return the next possible ID for this block.
    return ws_inner.max() + 1

...

numpy.cumsum(offsets, out=offsets)

def add_offset_block(block_index):
    ...
    labels[block] += offsets[block_index - 1]

with ThreadPoolExecutor(max_workers=max_workers) as executor:
    # Labels in the first block don't need to change.
    [executor.submit(add_offset_block, block_index) for block_index in range(1, n_blocks)]

return labels, offsets[-1]
```

ilastik/applets/wsdt/opWsdt.py
```diff
@@ -1,13 +1,141 @@
...
+                sigma_seeds,
+                sigma_weights,
+                minsize,
+                alpha,
+                pixel_pitch,
+                non_max_suppression,
+            )
+
+        logger.debug(
+            f"\tprocessing block {block_index} {block.outerBlock.begin}-{block.outerBlock.end} took {btimer.seconds()}"
```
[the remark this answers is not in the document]
k-dominik: ah thx, that was a debugging leftover when I used print statements :o

ilastik/applets/wsdt/opWsdt.py
```diff
@@ -1,13 +1,141 @@
...
+    assert ndim in [2, 3], "Watershed segmentor will only work on 2D and 3D data"
+
+    base_block = 512 if ndim == 2 else 128
+    # check for None arguments and set to default values
+    block_shape = (base_block,) * ndim if block_shape is None else block_shape
+    # NOTE nifty expects a list for the halo parameter although a tuple would be more logical
+    halo = [10] * ndim if halo is None else halo
+    blocking = get_blocking(data, block_shape, roi=None)
+
+    max_workers = max_workers = max(1, Request.global_thread_pool.num_workers) if max_workers is None else max_workers
```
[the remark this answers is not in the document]
k-dominik: thumbsup for the twoliner, refactor as proposed postponed to next release

ilastik/applets/base/appletSerializer.py
```diff
@@ -311,6 +311,31 @@ def _deserialize(self, subgroup, slot):
...
+
+        actually need to override this method to make use of the default value
+
+        :param group: The parent group in which to create this slot's
+            group.
+        :type group: h5py.Group
+
+        """
+        if self.name not in group:
+            if self.inslot.level == 0:
```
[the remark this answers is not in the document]
k-dominik: As discussed in the meeting with @emilmelnikov, and @btbest we'll accept the current solution as a temporary one (with added exception) and try to refactor this properly in the next cycle.

ilastik/applets/wsdt/opWsdt.py
```diff
@@ -1,13 +1,141 @@
...
+    # add the offset to blocks to make ids unique
+    def add_offset_block(block_index):
+        block = blocking.getBlock(block_index)
+        block = roiToSlice(block.begin, block.end)
+        labels[block] += offsets[block_index]
+
+    # add offsets in parallel
+    with ThreadPoolExecutor(max_workers=max_workers) as executor:
+        tasks = [executor.submit(add_offset_block, block_index) for block_index in range(n_blocks)]
+        [t.result() for t in tasks]
```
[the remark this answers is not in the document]
k-dominik: side note: in Python 3.7 `concurrent.futures.wait` expects a sequence, not an iterable ...

ilastik/applets/wsdt/opWsdt.py
k-dominik: hmm, not exactly. In the current code, 1 is added to the first block, this would not be the case with this code (as you acknowledged indirectly with the comment that first block doesn't need any changing). While this is not necessary for multicut, it is for carving. And at some point I'd like to reuse this function there... I'll update it with a comment.

ilastik/applets/wsdt/opWsdt.py
```diff
@@ -1,13 +1,142 @@
...
+    """Parallel dt watershed with hard block boundaries.
+
+    parallel wrapper around elf.segmentation.watershed.distance_transform_watershed
+
+    Args:
+      data: data to run watershed on, ndarray with either 2 or 3 dims
+      threshold: data will be thresholded at this value, distance transform will be calculated from
+        the remaining mask
+      sigma_seeds: smoothing factor that is applied to the watershed seed map
+      sigma_seeds: smoothing factor that is applied to the watershed weight map
```
btbest, line 47:
```suggestion
      sigma_weights: smoothing factor that is applied to the watershed weight map
```

btbest (approved):
I don't feel able to give very deep feedback here yet, but thankfully you also have Emil's feedback :) as far as I can tell without spending days looking into all the background here, I don't see any issues, or code that could gain much from a rework.

ilastik/applets/wsdt/opWsdt.py
emilmelnikov: Hmm, I see. In this case, do we even need offsets as numpy arrays? We already collect them sequentially from `Executor.map`, so we might as well prefix-sum them right there.
```python
offsets = itertools.accumulate(executor.map(ws_block, range(n_blocks)), initial=1)
```
Obviously, in this case we return `int(ws_inner.max())`, and index `offsets` with just `block_index` without `-1`.

At least it would be nice to avoid explicit `max_id/last_max_id` variables because it's normal for prefix sum idioms to have some starting element (1 in our case) and have something like `prefix_sum[-1] == total_max`.

k-dominik: initial was sadly only added in  3.8...

k-dominik: well, in the end:

https://github.com/ukoethe/vigra/blob/9b514fa00a136f5fd81bb57ee9f6293c333ffc1f/include/vigra/multi_watersheds.hxx#L456-L457

vigra already returns the labels starting at 1... I added a trivial test to confirm this.

tests/test_ilastik/test_applets/multicutWsdt/test_parallel_watershed.py
```diff
@@ -0,0 +1,51 @@
...
+import pytest
+
+from elf.parallel.common import get_blocking
+from lazyflow.roi import roiToSlice
+
+from ilastik.applets.wsdt.opWsdt import parallel_watershed
+
+
+@pytest.fixture
+def data():
```
emilmelnikov, line 11: This could probably be created in the test itself without fixtures.

tests/test_ilastik/test_applets/multicutWsdt/test_parallel_watershed.py
```diff
@@ -0,0 +1,51 @@
...
+
+    blocking = get_blocking(data, block_shape, roi=None)
+    running_max = 1
+    for block_index in range(blocking.numberOfBlocks):
+        block = blocking.getBlockWithHalo(blockIndex=block_index, halo=halo)
+        inner_slicing = roiToSlice(block.innerBlock.begin, block.innerBlock.end)
+
+        block_data = ws[inner_slicing]
+        assert block_data.min() == running_max
+        running_max = block_data.max() + 1
```
emilmelnikov, line 51: 👍

k-dominik:
thank you both for the reviews @emilmelnikov @btbest . Useful stuf as usual :)

PR #2646, opened 2022-12-20
add "i" shortcut some preprocessing applets

### LINKED ISSUE #2628: Shortcut `i` should be available in any applet
The `i`-keyboard shortcut for toggling between "raw data"-only and whatever else was on before is very useful with ilastik, as it's adding so many different layers. It's easy to forget about the raw data otherwise ;).

Currently it is available in many applets, but not in all.

Currently missing:

- [ ] _Object Classification_ thresholding
- [ ] _Carving_ preprocessing
- [ ] _Multicut_ DT Watershed

k-dominik:
homogenizes user experience by making the "i" shortcut available in
* thresholding (object classification)
* preprocessing (carving)
* watershed (multicut)
* Tracking
* Manual Tracking
* Blockwise Object Classification
* Object Extraction

needs `volumina>=1.3.8`

fixes #2628

Missing is only Carving, where the existence of the Overlay layer makes it impossible to use with the `toggleTopToBottom` method.

### DIFF
--- a/ilastik/applets/thresholdTwoLevels/thresholdTwoLevelsGui.py
+++ b/ilastik/applets/thresholdTwoLevels/thresholdTwoLevelsGui.py
@@ -432,21 +432,13 @@ class ThresholdTwoLevelsGui(LayerViewerGui):
             rawLayer.opacity = 1.0
             layers.append(rawLayer)
 
-            def toggleTopToBottom():
-                index = self.layerstack.layerIndex(rawLayer)
-                self.layerstack.selectRow(index)
-                if index == 0:
-                    self.layerstack.moveSelectedToBottom()
-                else:
-                    self.layerstack.moveSelectedToTop()
-
             rawLayer.shortcutRegistration = (
                 "i",
                 ShortcutManager.ActionInfo(
                     "Thresholding Layers",
                     "Bring Input To Top/Bottom",
                     "Bring Input To Top/Bottom",
-                    toggleTopToBottom,
+                    partial(self.layerstack.toggleTopToBottom, rawLayer),
                     self.viewerControlWidget(),
                     rawLayer,
                 ),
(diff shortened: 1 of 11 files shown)


btbest:
At this point we go from 2 to 5 duplications of this code, so definitely a point where refactoring instead of duplicating should be considered.

Does this functionality belong in the base class LayerViewerGui? There are a dozen or so other classes that use it without implementing the "i" shortcut, so I'm not sure.

But maybe it would at this point make sense to subclass LayerViewerGui for the shared functionality (of which there may be more than this shortcut?), and then let these three plus the other two (voxelSegmentationGui and labelingGui) use the new one as their base class.

k-dominik:
> At this point we go from 2 to 5 duplications of this code, so definitely a point where refactoring instead of duplicating should be considered.

absolutely - I deemed a refactor this close to the release too dangerous though :) It's gui functionality and hard to make sure it will work everywhere once refactor is introduced. Good time to do this would be _after_ the next release!

ilastik/applets/thresholdTwoLevels/thresholdTwoLevelsGui.py
```diff
@@ -431,4 +432,24 @@ def setupLayers(self):
             rawLayer.opacity = 1.0
             layers.append(rawLayer)
 
+            def toggleTopToBottom():
```
emilmelnikov, line 435: This could be a method of the `LayerStackModel` class:

```python
def toggleTopToBottom(target: Union[int, Layer]) -> None:
    if isinstance(target, Layer):
        target = self.layerIndex(target)
    ...
```

emilmelnikov (commented):
It looks like the `shortcutRegistration` attribute could be reworked into a `shortcut` property (or `setShortcut` method), and `ActionInfo` could be turned into a dataclass instead of namedtuple since we require at least Python 3.7, but this probably should be addressed later.

ilastik/applets/thresholdTwoLevels/thresholdTwoLevelsGui.py
k-dominik: thx @emilmelnikov, I don't get sweaty hands with this refactor. We can address a more general solution (maybe also a la `LabelingGui` which adds raw layer automatically and assigns the shortcut) in a dedicated subclass of `LayerViewerGui`  as @btbest suggested, in the next release cycle.

emilmelnikov (approved):

PR #2647, opened 2022-12-21
avoid showing multiple overlapping edge layers

k-dominik:
This is a bit of a tiny optimization, but it feels like it is doubling the fps you get in multicut.

Painting overlapping paths is probably quite costly (I guess to check what parts need to be painted, basically the whole path has to be compared).

There is no visible difference doing that, just faster rendering.


btbest (approved):
Was there anything in particular you wanted me to check here?

In my previous project, I would have said this comment reads as "I want to be a unit test", but I'm not sure about the cost/benefit of that for us here :)

k-dominik:
thx for the review @btbest :)

PR #2654, opened 2023-01-18
propagate dirtyness to layers

k-dominik:
Enables immediate dirty propagation once super pixel change via a `FreezeCache` cycle  for the edge-training/multicut caches.
Fixes display of outdated multicut solution after superpixels have changed.

### DIFF
--- a/ilastik/workflows/edgeTrainingWithMulticut/edgeTrainingWithMulticutWorkflow.py
+++ b/ilastik/workflows/edgeTrainingWithMulticut/edgeTrainingWithMulticutWorkflow.py
@@ -262,6 +262,10 @@ class EdgeTrainingWithMulticutWorkflow(Workflow):
         def _invalidate_cache_on_sp_change(*args, **kwargs):
             op = opEdgeTrainingWithMulticut
             opEdgeTrainingWithMulticut.clear_caches(op.current_view_index())
+            # allow for dirty propagation to the volumina layers
+            if op.FreezeCache.value:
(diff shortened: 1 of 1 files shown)


ilastik/workflows/edgeTrainingWithMulticut/edgeTrainingWithMulticutWorkflow.py
```diff
@@ -262,6 +262,10 @@ def handle_new_superpixels(*args):
         def _invalidate_cache_on_sp_change(*args, **kwargs):
             op = opEdgeTrainingWithMulticut
             opEdgeTrainingWithMulticut.clear_caches(op.current_view_index())
+            # allow for dirty propagation to the volumina layers
```
emilmelnikov, line 265: Do we need to do propagate dirtyness every time operator's caches are cleared? If yes, then this change should go to [`OpEdgeTrainingWithMulticut.clear_caches()`](
https://github.com/ilastik/ilastik/blob/068d8d709d74626e59a471bb50d9fcd607cecf27/ilastik/applets/edgeTrainingWithMulticut/opEdgeTrainingWithMulticut.py#L98).

k-dominik: moving this change there would mean the operator itself would modify its slot (`FreezeCache`) which is not something that should happen.

btbest (approved):

emilmelnikov (approved):

PR #2655, opened 2023-01-18
delay bioimageio imports

k-dominik:
bioimageio.spec will try to fetch some data from github and bioimageio. This is necessary for its operation, but shouldn't be done on _every_ startup of ilastik.
This delays those imports until NN-related workflows are actually used.

[1c50af6eb7e664bf024a0c3e50f828d5d3325f09](https://github.com/ilastik/ilastik/pull/2655/commits/1c50af6eb7e664bf024a0c3e50f828d5d3325f09) is unrelated - it's a tiny change to applet readiness that I didn't want to bother anyone reviewing independently. Now Export and Batch processing are consistently enabled once a valid model is in the project.

### DIFF
--- a/ilastik/workflows/neuralNetwork/_nnWorkflowBase.py
+++ b/ilastik/workflows/neuralNetwork/_nnWorkflowBase.py
@@ -166,14 +166,11 @@ class _NNWorkflowBase(Workflow):
         input_ready = len(opDataSelection.ImageGroup) > 0 and not self.dataSelectionApplet.busy
 
         opNNClassification = self.nnClassificationApplet.topLevelOperator
+        nn_ready = input_ready and opNNClassification.ModelSession.ready()
 
         opDataExport = self.dataExportApplet.topLevelOperator
 
-        predictions_ready = input_ready and len(opDataExport.Inputs) > 0
-
-        # Problems can occur if the features or input data are changed during live update mode.
-        # Don't let the user do that.
-        live_update_active = not opNNClassification.FreezePredictions.value
+        predictions_ready = nn_ready and len(opDataExport.Inputs) > 0
 
         # The user isn't allowed to touch anything while batch processing is running.
         batch_processing_busy = self.batchProcessingApplet.busy
@@ -185,7 +182,7 @@ class _NNWorkflowBase(Workflow):
         )
         self._shell.setAppletEnabled(
             self.dataExportApplet,
-            predictions_ready and not batch_processing_busy and not live_update_active and upstream_ready,
+            predictions_ready and not batch_processing_busy and upstream_ready,
         )
 
         if self.batchProcessingApplet is not None:
--- a/ilastik/applets/neuralNetwork/bioimageiodl.py
+++ b/ilastik/applets/neuralNetwork/bioimageiodl.py
@@ -2,10 +2,6 @@ import pathlib
 from io import BytesIO
 from zipfile import ZIP_DEFLATED
 
-from bioimageio.core import load_raw_resource_description
-from bioimageio.core.resource_io.io_ import make_zip
-from bioimageio.spec import get_resource_package_content
-from bioimageio.spec.shared import resolve_source, raw_nodes, DownloadCancelled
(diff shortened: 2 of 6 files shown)


emilmelnikov:
It is somehow possible to group all bioimageio imports somewhere around the top level of the module?.. I am a little bit worried that these imports will be hard to find and manage later. Maybe extract some code that uses bioimageio into top-level functions and do imports there?

ilastik/workflows/neuralNetwork/_nnWorkflowBase.py
```diff
@@ -166,14 +166,11 @@ def handleAppletStateUpdateRequested(self, upstream_ready=True):
         input_ready = len(opDataSelection.ImageGroup) > 0 and not self.dataSelectionApplet.busy
```
emilmelnikov, line 166: It would be nice to group op shortcuts, flags, and applet changes separately, and also replace `busy` with `ready` where possible.

```python
opDataSelection = self.dataSelectionApplet.topLevelOperator
opNNClassification = self.nnClassificationApplet.topLevelOperator
opDataExport = self.dataExportApplet.topLevelOperator

data_selection_ready = not self.dataSelectionApplet.busy
input_ready = data_selection_ready and len(opDataSelection.ImageGroup) > 0
nn_ready = input_ready and opNNClassification.ModelSession.ready()
predictions_ready = nn_ready and len(opDataExport.Inputs) > 0
batch_processing_ready = not self.batchProcessingApplet.busy

self._shell.setAppletEnabled(..., upstream_ready and batch_processing_ready)
...
```

Maybe it would also make sense to merge `batch_processing_ready` into `upstream_ready`, but I'm not sure about this:
```python
upstream_ready &= not self.batchProcessingApplet.busy
```

btbest:
> It is somehow possible to group all bioimageio imports somewhere around the top level of the module?

I agree that it would be nicer to have some sort of top-level place to collect such delayed imports, but I don't see a realistic way this can be done.

Though in absence of such a solution, I would prefer the explanatory comment to be with the actual import statement, and not in the top-of-the-file import block.

emilmelnikov:
> > It is somehow possible to group all bioimageio imports somewhere around the top level of the module?
> 
> I agree that it would be nicer to have some sort of top-level place to collect such delayed imports, but I don't see a realistic way this can be done.

I was thinking about the ways to implement with the Pyhton's import machinery and found out that stdlib already has [`LazyLoader`][lazy]. Python's bugtracker even has [an example][bpo].

There is also [a pretty interesting blog post by Brett Cannon][post] that is even packaged in a library, but I think for our purposes `LazyLoader` is better.

[lazy]: https://docs.python.org/3/library/importlib.html#importlib.util.LazyLoader
[bpo]: https://bugs.python.org/issue17621
[post]: https://snarky.ca/lazy-importing-in-python-3-7/

k-dominik:
allright, concerning lazy imports...

Unfortunately the lazy-loading technique described in [the python docs](https://docs.python.org/3/library/importlib.html#implementing-lazy-imports) does not work with namespace packages (that's my observation at least). Currently I don't want to dive deeply into the Python import system.

Another option would be to use the [`lazy_import` module](https://github.com/mnmelo/lazy_import), which does work as expected as far as I can see. This would mean adding it as a dependency (it's available on cf :yay:).  I'm dreaming of faster ilastik startups already and think that such a package makes sense in general (and I don't think there is a different one).
Downside of using `lazy_import`:
 * there might be windows issues: https://github.com/mnmelo/lazy_import/pull/23
 * it seems unmaintained (no comment on issues since 2019, no commits in 5 years)
 * it interferes with logging: https://github.com/mnmelo/lazy_import/issues/8

So it seems we'd have to resort to fix these things and maintain our own fork. Whether we really want to do this depends on how much of a speedup we can get from applying lazy loading to other modules. I'd say we investigate this for the next iteration, not for the current release.

Bottom line: I'll move the comments to the imports of the current state of the PR as @btbest suggested.

btbest (approved):

PR #2659, opened 2023-01-23
Gh2657 carving does not save all preprocessing settings to project

btbest:
Added the missing Preprocessing settings to the serializer and changed the naming as the "reset" mechanism no longer exists (removed in 2013 - see 89e5559ae8534b4b231f4385347965984be99b4b).

This is branched from https://github.com/ilastik/ilastik/pull/2658 to save myself some merge conflicts, so we should merge that first if it hasn't happened yet. The commits to be reviewed in this PR are only the last three.

## Checklist

### DIFF
--- a/ilastik/workflows/carving/preprocessingSerializer.py
+++ b/ilastik/workflows/carving/preprocessingSerializer.py
@@ -47,10 +47,16 @@ class PreprocessingSerializer(AppletSerializer):
 
                 deleteIfPresent(preproc, "sigma")
                 deleteIfPresent(preproc, "filter")
+                deleteIfPresent(preproc, "do_agglomeration")
+                deleteIfPresent(preproc, "size_regularizer")
+                deleteIfPresent(preproc, "reduce_to")
                 deleteIfPresent(preproc, "graph")
 
                 preproc.create_dataset("sigma", data=opPre.initialSigma)
                 preproc.create_dataset("filter", data=opPre.initialFilter)
+                preproc.create_dataset("do_agglomeration", data=opPre.initialDoAgglo)
+                preproc.create_dataset("size_regularizer", data=opPre.initialSizeRegularizer)
+                preproc.create_dataset("reduce_to", data=opPre.initialReduceTo)
 
                 preprocgraph = getOrCreateGroup(preproc, "graph")
                 mst.saveH5G(preprocgraph)
@@ -64,6 +70,11 @@ class PreprocessingSerializer(AppletSerializer):
 
         sigma = topGroup["sigma"][()]
         sfilter = topGroup["filter"][()]
+
+        # Pre-1.4.0rc9 project files do not contain doAgglo, sizeRegularizer and reduceTo - use defaults
+        doAgglo = topGroup["do_agglomeration"][()] if "do_agglomeration" in list(topGroup.keys()) else 1
+        sizeRegularizer = topGroup["size_regularizer"][()] if "size_regularizer" in list(topGroup.keys()) else 0.5
+        reduceTo = topGroup["reduce_to"][()] if "reduce_to" in list(topGroup.keys()) else 0.2
 
         if "graph" in list(topGroup.keys()):
             graphgroup = topGroup["graph"]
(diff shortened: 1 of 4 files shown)


ilastik/workflows/carving/opCarving.py
```diff
@@ -231,12 +229,10 @@ def dataIsStorable(self):
...
-        fg_seedNum = len(numpy.where(nodeSeeds == 2)[0])
-        bg_seedNum = len(numpy.where(nodeSeeds == 1)[0])
-        if not (fg_seedNum > 0 and bg_seedNum > 0):
-            return False
-        else:
-            return True
+        fg_seedNum = len(numpy.where(nodeSeeds == Labels.FOREGROUND)[0])
+        bg_seedNum = len(numpy.where(nodeSeeds == Labels.BACKGROUND)[0])
+        has_bg_and_fg_seeds = fg_seedNum > 0 and bg_seedNum > 0
+        return has_bg_and_fg_seeds
```
emilmelnikov, line 235:
```suggestion
        return numpy.any(numpy.isin(nodeSeeds, [l.value for l in Labels]))
```

ilastik/workflows/carving/opCarving.py
```diff
@@ -540,53 +506,32 @@ def saveCurrentObjectAs(self, name):
...
-            w1 = [w1[i] + sl[i].start for i in range(1, 4)]
-            w2 = [w2[i] + sl[i].start for i in range(1, 4)]
+            return None, None
+
+        bg = [[], [], []]
+        fg = [[], [], []]
+        for slicing in self.opLabelArray.NonzeroBlocks[:].wait()[0]:
+            label = self.opLabelArray.Output[slicing].wait()
+            labels_bg = numpy.where(label == Labels.BACKGROUND)
+            labels_fg = numpy.where(label == Labels.FOREGROUND)
```
emilmelnikov, line 521: [Numpy recommends to use `nonzero` instead of `where`](https://numpy.org/doc/stable/reference/generated/numpy.where.html#numpy.where) if 2nd and 3rd arguments are omitted.

ilastik/workflows/carving/opCarving.py
```diff
@@ -540,53 +506,32 @@ def saveCurrentObjectAs(self, name):
...
-            if len(coors1[i]) > 0:
-                coors1[i] = numpy.concatenate(coors1[i], 0)
-            else:
-                coors1[i] = numpy.ndarray((0,), numpy.int32)
-            if len(coors2[i]) > 0:
-                coors2[i] = numpy.concatenate(coors2[i], 0)
-            else:
-                coors2[i] = numpy.ndarray((0,), numpy.int32)
-        return (coors2, coors1)
+            bg[i] = numpy.concatenate(bg[i], 0) if len(bg[i]) > 0 else numpy.ndarray((0,), numpy.int32)
```
emilmelnikov, line 529: It's a nitpick, but I think keyword arguments and standard `numpy.array` are a little bit nicer:
```suggestion
            bg[i] = numpy.concatenate(bg[i], axis=0) if len(bg[i]) > 0 else numpy.array((), dtype=numpy.int32)
```

ilastik/workflows/carving/preprocessingSerializer.py
```diff
@@ -95,7 +112,7 @@ def _deserializeFromHdf5(self, topGroup, groupVersion, hdf5File, projectFilePath
 
     def isDirty(self):
         for opPre in self._o.innerOperators:
-            if opPre._unsavedData:
+            if opPre.hasUnsavedData:
                 return True
         return False
```
emilmelnikov, line 117:
```python
return any(op.hasUnsavedData for op in self._o.innerOperators)
```

ilastik/workflows/carving/preprocessingSerializer.py
```diff
@@ -37,25 +37,30 @@ def _serializeToHdf5(self, topGroup, hdf5File, projectFilePath):
         preproc = topGroup
 
         for opPre in self._o.innerOperators:
-            mst = opPre._prepData[0]
+            mst = opPre.cachedResult[0]
```
emilmelnikov, line 40: It would be nice to extract all cached values in a dataclass and pass it around instead of individual values. It's also unclear why `cachedResult` is a 1-element list and not just an MST value itself... OTOH, I'm not sure we should do these changes right now.

PR #2660, opened 2023-01-24
move applet index saving to save method

k-dominik:
previously, closing the project would trigger update of `currentApplet` setting in project file. This had the unintentional side effect, that opening a file, doing something without saving, closing (without saving) would still alter the project file. This should not happen.

Now the `currentApplet` setting is only updated on the `saveProject` action (not on `saveProjectAs`, or `saveProjectSnapshot`).

- [ ] Add tests.

### DIFF
--- a/ilastik/shell/gui/ilastikShell.py
+++ b/ilastik/shell/gui/ilastikShell.py
@@ -1696,16 +1696,6 @@ class IlastikShell(QMainWindow):
         assert threading.current_thread().name == "MainThread"
         if self.projectManager is not None:
 
-            projectFile = self.projectManager.currentProjectFile
-            if (
-                not self.projectManager.closed
-                and projectFile is not None
-                and not self.projectManager.currentProjectIsReadOnly
-            ):
-                if "currentApplet" in list(projectFile.keys()):
-                    del projectFile["currentApplet"]
-                self.projectManager.currentProjectFile.create_dataset("currentApplet", data=self.currentAppletIndex)
-
             self.removeAllAppletWidgets()
             for f in self.cleanupFunctions:
                 f()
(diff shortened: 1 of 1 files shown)


ilastik/shell/gui/ilastikShell.py
```diff
@@ -1755,6 +1745,17 @@ def ensureNoCurrentProject(self, assertClean=False):
...
+    def _save_current_applet(self):
+        projectFile = self.projectManager.currentProjectFile
+        if (
+            not self.projectManager.closed
+            and projectFile is not None
+            and not self.projectManager.currentProjectIsReadOnly
+        ):
+            if "currentApplet" in list(projectFile.keys()):
+                del projectFile["currentApplet"]
+            self.projectManager.currentProjectFile.create_dataset("currentApplet", data=self.currentAppletIndex)
```
emilmelnikov, line 1757:
```suggestion
        pm = self.projectManager
        if pm.closed or pm.currentProjectIsReadOnly or pm.currentProjectFile is None:
            return
        ds = pm.currentProjectFile.require_dataset("currentApplet", shape=(), dtype=int)
        ds[()] = self.currentAppletIndex
```

emilmelnikov (approved):

ilastik/shell/gui/ilastikShell.py
btbest: What does an empty tuple as an index do? I find this notation rather obscure and would prefer to avoid it.

btbest (approved):

ilastik/shell/gui/ilastikShell.py
emilmelnikov: [This is now you use scalar datasets in HDF5](https://docs.h5py.org/en/stable/high/dataset.html#reading-writing-data). It looks weird, but actually makes sense: scalar dataset is 0-dimensional, so you use 0-dim tuple to reference a value.

k-dominik: Heya, I agree wtith both of you :). Probably the right thing to use would be `require_dataset`, despite it's weird syntax... In ilastik `require_dataset` has not been used before - the other pattern is all over the place. So for now I'll keep the ilastik way of "requireing a dataset". We can discuss a refactor of the codebase to the different pattern of course.

The change to the conditions make a lot of sense, I'll  add those.

emilmelnikov: One could also `group.pop("name", None)` instead of `if "name" in group: del group["name"]`, but we use the latter throughout ilastik too.

PR #2663, opened 2023-01-26
Avoid dirtyness during batch processing

k-dominik:
setting bypassmode would invalidate object features and classifier would be retrained. This can alter the tracking result a lot, especially if classifiers are not trained well.

## Checklist

- [ ] Add tests.

### DIFF
--- a/ilastik/workflows/tracking/conservation/conservationTrackingWorkflow.py
+++ b/ilastik/workflows/tracking/conservation/conservationTrackingWorkflow.py
@@ -219,14 +219,14 @@ class ConservationTrackingWorkflowBase(Workflow):
         # If we have stored division and cell classifiers, restore them into the workflow now.
         if self.stored_division_classifier:
             opDivisionClassification = self.divisionDetectionApplet.topLevelOperator
-            opDivisionClassification.classifier_cache.forceValue(self.stored_division_classifier)
+            opDivisionClassification.classifier_cache.forceValue(self.stored_division_classifier, set_dirty=False)
             # Release reference
             self.stored_division_classifier = None
 
         # If we have stored division and cell classifiers, restore them into the workflow now.
         if self.stored_cell_classifier:
             opCellClassification = self.cellClassificationApplet.topLevelOperator
-            opCellClassification.classifier_cache.forceValue(self.stored_cell_classifier)
+            opCellClassification.classifier_cache.forceValue(self.stored_cell_classifier, set_dirty=False)
             # Release reference
             self.stored_cell_classifier = None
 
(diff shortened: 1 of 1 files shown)


ilastik/applets/tracking/conservation/opConservationTracking.py
```diff
@@ -328,7 +328,11 @@ def _resolveMergers(self, hypothesesGraph, model):
             withFullGraph = True
             logger.info("Computing full graph on merger resolver (Only enabled on animal tracking workflow)")
 
-        mergerResolver = IlastikMergerResolver(originalGraph, pluginPaths=self.pluginPaths, withFullGraph=withFullGraph)
+        # supply random_state for reproducible merger resolving
+        # random_state will be used in gmm-based merger resolving
+        mergerResolver = IlastikMergerResolver(
+            originalGraph, pluginPaths=self.pluginPaths, withFullGraph=withFullGraph, random_state=42
```
emilmelnikov, line 334: Move `random_state` to a keyword parameter in `track()`? It probably doesn't matter, but the `parameters` dict already contains a lot of configurable values with defaults.

emilmelnikov (approved):

ilastik/applets/tracking/conservation/opConservationTracking.py
k-dominik: you mean like in future we (users) might want to modify the seed? E.g. on a per project basis?

emilmelnikov: I'm not sure to what extent that would be useful. Just thought about lifting hardcoded constants up to the config. But this is probably unnecessary in this case.

k-dominik: hmmm.... on the _other_ hand...

Tomaz-Vieira:
We've been using the `set_dirty=False` trick a lot lately, every time to go around the `OpValueCache` not doing what we want. What if, instead of using an `OpValueCache`, we move the training logic out `OpObjectClassification.execute` and into something like a standalone cached function:

```python3
@lru_cache
def do_training(featMatrix, labels):
    ...
```

and then just call that from  `OpObjectClassification.execute` ? Then whatever we do upstream, if the classifier gets the same inputs, we'd still get the same classifier again, and therefore the same results.

k-dominik:
Hey @Tomaz-Vieira,

that is absolutely sensible and I'd definitely want to try this, but not as part of this PR. Could you maybe move your comment to a new issue?

emilmelnikov:
> ```python
> @lru_cache
> ```

We need some kind of cache management strategy though: Python's default LRU cache is unbounded.

Issue #2011, opened 2019-05-07
Carving: update of superpixels invalidates objects, but objects are still present


k-dominik:
## Describe the bug
Right now, if one already has some saved objects and decides to change the superpixel segmentation, already carved objects are still present (associated with superpixel ids that don't make sense anymore).
Furthermore, if the number of superpixels is reduced, it is very likely that there will be `IndexErrors` accessing the superpixels from saved objects (as superpixels with large ids are not present anymore).

## Expected behavior
Updating the superpixel segmentation should not result in inconsistent, non-sensical objects. Safest for now would be to remove all objects and add a fair warning about this!

## To Reproduce
Within the Carving workflow
1. Add some 2d/3d, do preprocessing, carve/save some objects
2. Go back to the preprocessing applet, make changes to the superpixel segmentation
3. In the labeling applet you can now, load an object and see that it doesn't make sense at all anymore

## Screenshots
To illustrate, you go from this:
![export](https://user-images.githubusercontent.com/24434157/57307097-3e349a80-70e4-11e9-947c-c98dc1b84551.png)

to this 
![export2](https://user-images.githubusercontent.com/24434157/57307116-455ba880-70e4-11e9-89bf-5115f2911cfa.png)

## Desktop (please complete the following information):
 - ilastik version: 1.3.2post1, 1.3.3b1
 - OS: all

k-dominik:
xref: #918

btbest:
As of #2664, we now discard all saved objects when the user changes preprocessing. The user is warned about this before they re-execute preprocessing.

Cross-posting the consideration regarding an alternative solution from #918: We could consider keeping the labels of saved objects and discarding only their segmentations, but doing this properly will require reworking the handling of saved objects quite a bit. One would also need to extend Carving with a sub-workflow to guide the user to re-segment all saved objects after changing preprocessing. The user would have to do this manually for each object. And they _should_ do it manually, as it's not sound to just assume that all saved labels are semantically valid with the new preprocessing.

PR #2664, opened 2023-01-27
Carving drop saved objects when changing preprocessing

btbest:
Sorry this one became so big. The commits are probably easier to review individually than the PR as a whole.

The original aim was to delete existing saved objects when the user re-runs the preprocessing, and to warn the user before doing so (#2011).

It turned out there were some more issues around saving and loading objects, particularly with tracking whether the current state is ready to be saved as an object. This should be improved now, which should make it easier in the future if we go for keeping the labels of saved objects and only throw away their segmentation when changing preprocessing.

## Checklist

### DIFF
--- a/ilastik/workflows/carving/carvingGui.py
+++ b/ilastik/workflows/carving/carvingGui.py
@@ -45,7 +45,6 @@ from volumina.view3d.volumeRendering import RenderingManager
 # ilastik
 from ilastik.utility import bind
 from ilastik.applets.labeling.labelingGui import LabelingGui
-from ilastik.workflows.carving.opCarving import DEFAULT_OBJECT_NAME
 
 
 import logging
@@ -53,6 +52,7 @@ import logging
 logger = logging.getLogger(__name__)
 
 CURRENT_SEGMENTATION_NAME = "__current_segmentation__"
+DEFAULT_OBJECT_NAME = "<not saved yet>"
 # ===----------------------------------------------------------------------------------------------------------------===
 
 
@@ -280,7 +280,7 @@ class CarvingGui(LabelingGui):
             logger.error("object not saved due to faulty data.")
             return
 
-        old_name = self.topLevelOperatorView.getCurrentObjectName()
+        old_name = self.topLevelOperatorView.CurrentObjectName.value
         saved_object_names = self.getObjectNames()
         was_object_previously_saved = old_name in saved_object_names
 
@@ -655,8 +655,9 @@ class CarvingGui(LabelingGui):
         def onButtonsEnabled(slot, roi):
             currObj = self.topLevelOperatorView.CurrentObjectName.value
             canSave = self.topLevelOperatorView.CanObjectBeSaved.value
+            label = currObj if currObj else DEFAULT_OBJECT_NAME
 
-            self.labelingDrawerUi.currentObjectLabel.setText(currObj)
+            self.labelingDrawerUi.currentObjectLabel.setText(label)
             self.labelingDrawerUi.save.setEnabled(canSave)
 
         self.topLevelOperatorView.CurrentObjectName.notifyDirty(onButtonsEnabled)
--- a/ilastik/workflows/carving/carvingGui.py
+++ b/ilastik/workflows/carving/carvingGui.py
@@ -132,7 +132,7 @@ class CarvingGui(LabelingGui):
         self.labelingDrawerUi.segment.setEnabled(True)
 
         self.topLevelOperatorView.Segmentation.notifyDirty(bind(self._segmentation_dirty))
-        self.topLevelOperatorView.HasSegmentation.notifyValueChanged(bind(self._updateGui))
+        self.topLevelOperatorView.CanObjectBeSaved.notifyValueChanged(bind(self._updateGui))
 
         self.labelingDrawerUi.objPrefix.setText(self.objectPrefix)
         self.labelingDrawerUi.objPrefix.textChanged.connect(self.setObjectPrefix)
@@ -637,13 +637,13 @@ class CarvingGui(LabelingGui):
 
         def onButtonsEnabled(slot, roi):
             currObj = self.topLevelOperatorView.CurrentObjectName.value
-            hasSeg = self.topLevelOperatorView.HasSegmentation.value
+            canSave = self.topLevelOperatorView.CanObjectBeSaved.value
 
             self.labelingDrawerUi.currentObjectLabel.setText(currObj)
-            self.labelingDrawerUi.save.setEnabled(hasSeg)
+            self.labelingDrawerUi.save.setEnabled(canSave)
 
         self.topLevelOperatorView.CurrentObjectName.notifyDirty(onButtonsEnabled)
-        self.topLevelOperatorView.HasSegmentation.notifyDirty(onButtonsEnabled)
+        self.topLevelOperatorView.CanObjectBeSaved.notifyDirty(onButtonsEnabled)
         self.topLevelOperatorView.opLabelArray.NonzeroBlocks.notifyDirty(onButtonsEnabled)
 
         # Labels
--- a/ilastik/workflows/carving/carvingGui.py
+++ b/ilastik/workflows/carving/carvingGui.py
@@ -316,8 +316,8 @@ class CarvingGui(LabelingGui):
             if len(selected) != 1:
                 return
             name = selected[0].text()
+            dialog.close()
             if self.confirmLoadObject():
-                dialog.close()
                 self.topLevelOperatorView.loadObject(name)
 
         def deleteSelection():
--- a/ilastik/workflows/carving/carvingGui.py
+++ b/ilastik/workflows/carving/carvingGui.py
@@ -240,17 +240,14 @@ class CarvingGui(LabelingGui):
 
         return last + 1
 
-    def saveAsDialog(self, name=""):
-        """special functionality: reject names given to other objects"""
-        namesInUse = self.getObjectNames()
-
+    def saveAsDialog(self, name_input_default: str, existing_names: List[str]):
         def generateObjectName():
             return f"{self.objectPrefix}{self.findNextPrefixNumber()}"
 
-        name = name or generateObjectName()
+        name_input_default = name_input_default or generateObjectName()
 
         dialog = uic.loadUi(self.dialogdirSAD)
-        dialog.lineEdit.setText(name)
+        dialog.lineEdit.setText(name_input_default)
         dialog.lineEdit.selectAll()
         dialog.warning.setVisible(False)
         dialog.Ok.clicked.connect(dialog.accept)
@@ -259,7 +256,7 @@ class CarvingGui(LabelingGui):
 
         def validate():
             name = dialog.lineEdit.text()
-            if name in namesInUse:
+            if name in existing_names:
                 dialog.Ok.setEnabled(False)
                 dialog.warning.setVisible(True)
                 dialog.isDisabled = True
@@ -268,6 +265,7 @@ class CarvingGui(LabelingGui):
                 dialog.warning.setVisible(False)
                 dialog.isDisabled = False
 
+        validate()
         dialog.lineEdit.textChanged.connect(validate)
         result = dialog.exec_()
         if result:
@@ -284,7 +282,7 @@ class CarvingGui(LabelingGui):
         saved_object_names = self.getObjectNames()
         was_object_previously_saved = old_name in saved_object_names
 
-        new_name = self.saveAsDialog(name=(old_name if was_object_previously_saved else ""))
+        new_name = self.saveAsDialog(old_name, saved_object_names)
         if new_name is None:
             return
 
--- a/ilastik/workflows/carving/carvingGui.py
+++ b/ilastik/workflows/carving/carvingGui.py
@@ -314,10 +314,13 @@ class CarvingGui(LabelingGui):
         dialog.objectNames.addItems(sorted(names, key=_humansort_key))
 
         def loadSelection():
-            selected = [str(name.text()) for name in dialog.objectNames.selectedItems()]
-            dialog.close()
-            for objectname in selected:
-                self.topLevelOperatorView.loadObject(objectname)
+            selected = dialog.objectNames.selectedItems()
+            if len(selected) != 1:
+                return
+            name = selected[0].text()
+            if self.confirmLoadObject():
+                dialog.close()
+                self.topLevelOperatorView.loadObject(name)
 
         def deleteSelection():
             items = dialog.objectNames.selectedItems()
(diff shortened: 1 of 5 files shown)


ilastik/workflows/carving/opCarving.py
```diff
@@ -263,29 +262,12 @@ def setupOutputs(self):
...
-    def hasCurrentObject(self):
-        """
-        Returns current object name. None if it is not set.
-        """
-        # FIXME: This is misleading. Having a current object and that object having
-        # a name is not the same thing.
-        return self._currObjectName
-
     def getCurrentObjectName(self):
+        f"""
```
k-dominik, line 270: fstrings cannot be used as docstrings. See: https://docs.python.org/3/reference/lexical_analysis.html#formatted-string-literals (right at the end of that paragraph)

ilastik/workflows/carving/watershed_segmentor.py
```diff
@@ -136,3 +138,4 @@ def saveH5G(self, h5g):
 
     def setResulFgObj(self, fgNodes):
         self.gridSegmentor.setResulFgObj(fgNodes)
+        self.hasSeg = True
```
k-dominik, line 141: It seems like `self.hasSeg` was not properly utilized before (was not in sync sometimes). For some reason I would feel safer if self.hasSeg could be a dynamic property, rather than relying on every  method setting it correctly. But I take it that you ironed it out appropriately.

ilastik/workflows/carving/opCarving.py
```diff
@@ -348,20 +335,21 @@ def restore_and_get_labels_for_object(self, name):
...
-            logger.info("  --> no such object '%s'" % name)
-            return False
+        logger.info(f"want to load object with name = {name}")
+        if name not in self._mst.object_lut:
+            logger.info("  --> no object with this name")
+            return
 
-        if self.hasCurrentObject():
-            self.saveCurrentObject()
+        if self.CanObjectBeSaved.value:
```
k-dominik, line 350: Should we do this silently without giving the user to intervene? This will save an object with the default name...

ilastik/workflows/carving/opCarving.py
```diff
@@ -224,20 +225,22 @@ def _buildDone(self):
             for name, objectSupervoxels in self._mst.object_lut.items():
                 if name == self._currObjectName:
                     continue
-                assert name in self._mst.object_names, "%s not in self._mst.object_names, keys are %r" % (
-                    name,
-                    list(self._mst.object_names.keys()),
-                )
+                assert (
+                    name in self._mst.object_names
+                ), f"{name} not in self._mst.object_names, keys are {list(self._mst.object_names.keys())!r}"
```
emilmelnikov, line 230:
```suggestion
                ), f"{name} not in self._mst.object_names, keys are {list(self._mst.object_names)!r}"
```

ilastik/workflows/carving/opCarving.py
```diff
@@ -611,24 +587,23 @@ def execute(self, slot, subindex, roi, result):
...
-
-            self.has_seeds = True
-        else:
-            raise RuntimeError("unknown slots")
+        assert slot == self.WriteSeeds, f"Invalid input slot: {slot.name}"
+
+        with Timer() as timer:
+            logger.info("Writing seeds to label array")
+            self.opLabelArray.LabelSinkInput[roi.toSlice()] = value
+            logger.info("Writing seeds to label array took {} seconds".format(timer.seconds()))
```
emilmelnikov, line 595:
```suggestion
            logger.info(f"Writing seeds to label array took {timer.seconds()} seconds")
```

ilastik/workflows/carving/preprocessingGui.py
```diff
@@ -126,6 +124,24 @@ def onFailed(self, exception, exc_info):
...
+                "Confirm Deleting Saved Objects",
+                f"This project already contains {n} saved segmented objects. The existing segmentations "
+                "become invalid if you change preprocessing settings, and will be deleted.\n\n"
+                "Please consider creating a copy of the project file for the different preprocessing instead.\n\n"
+                "Run preprocessing and delete all saved objects?",
+                buttons,
+                defaultButton=QMessageBox.Cancel,
+            )
+            if response == QMessageBox.Cancel:
+                return
```
emilmelnikov, line 144:
```suggestion
        result = self.topLevelOperatorView.cachedResult[0]
        nsaved = len(result.object_names) if result is not None else 0
        if nsaved:
            response = QMessageBox.warning(
                self,
                "Confirm Deleting Saved Objects",
                (
                    f"<p>This project already contains {nsaved} saved segmented objects. "
                    "The existing segmentations become invalid if you change preprocessing settings, and will be deleted.</p>"
                    "<p>Please consider creating a copy of the project file for the different preprocessing instead.</p>"
                    "<p>Run preprocessing and delete all saved objects?</p>"
                    
                ),
                buttons=QMessageBox.Yes | QMessageBox.Cancel,
                defaultButton=QMessageBox.Cancel,
            )
            if response == QMessageBox.Cancel:
                return
```

ilastik/workflows/carving/carvingGui.py
```diff
@@ -277,38 +275,41 @@ def validate():
...
+        old_name = self.topLevelOperatorView.getCurrentObjectName()
+        saved_object_names = self.getObjectNames()
+        was_object_previously_saved = old_name in saved_object_names
+
+        new_name = self.saveAsDialog(name=(old_name if was_object_previously_saved else ""))
+        if new_name is None:
+            return
+
+        is_name_changed = new_name != old_name
+        if is_name_changed and new_name in saved_object_names:
```
emilmelnikov, line 296: It seems impossible to have `new_name in saved_object_names` because `saveAsDialog` doesn't let you to proceed if the name one entered is in `getObjectNames()`. However, it may be possible that `validate()` in `saveAsDialog()` is not triggered if `QLineEdit.textChanged` signal is not emitted, which is possible if one never changes text in line input (I'm not sure whether it's true or not...)

ilastik/workflows/carving/opCarving.py
emilmelnikov: I wonder if it's possible to make `<not saved yet>` purely a view thing, and deal with an "empty" name (`""` or `None`) in all other logic code?..

ilastik/workflows/carving/opCarving.py
btbest: 😅

ilastik/workflows/carving/watershed_segmentor.py
btbest: Agree. I'm still not 100% confident this works as intended and refrained from reworking it for now to not bloat this PR even more, but the most obvious issue should be fixed for now...

ilastik/workflows/carving/carvingGui.py
btbest: Nice catch. The text input is pre-filled with the old name (if there was one), so to obtain a name from another already saved object, the user has to change the text. And I think we should be able to trust QT to ensure the signal gets emitted.

Would you prefer the lines thrown out in that case?

ilastik/workflows/carving/opCarving.py
btbest: I agree; I've pushed an update to get a user confirmation to discard current (potentially unsaved) work instead.

Also agree that I'd prefer the default string to be strictly GUI-side

ilastik/workflows/carving/carvingGui.py
emilmelnikov: Hmm, I think the dialog function as least needs to call `validate()` itself in order to set up GUI state properly. Maybe also receive `namesInUse` as a parameter, and don't generate a unique object name by itself.

k-dominik:
Heya, I gave this one a go and it's already so much better than before (from a user perspective). One minor observation:

when loading a project while segmenting and deciding to cancel the loading in order to finish the segmentation first, the "browse objects" dialog is still opened and needs to be closed manually. I think it would make sense to close it too, so that the user can go on segmenting.

and another one that should probably be a separate issue (that also has nothing to do with your changes): In carving, after loading an object, it is unclear how to start a new obejct. In practice I guess on is expected to "clear" and go on - but then the old name is still active...

ilastik/workflows/carving/carvingGui.py
```diff
@@ -259,7 +256,7 @@ def generateObjectName():
 
         def validate():
             name = dialog.lineEdit.text()
```
emilmelnikov, line 258: I think it's fine to unconditionally change widget properties here, and do this even if `isDisabled` is false.
```python
valid = name not in existing_names
# Could also remove double negation here and rename the field to isEnabled.
dialog.isDisabled = not valid
dialog.Ok.setEnabled(valid)
dialog.warning.setVisible(not valid)
```

ilastik/workflows/carving/carvingGui.py
```diff
@@ -240,17 +240,14 @@ def findNextPrefixNumber(self):
 
         return last + 1
 
-    def saveAsDialog(self, name=""):
-        """special functionality: reject names given to other objects"""
-        namesInUse = self.getObjectNames()
-
+    def saveAsDialog(self, name_input_default: str, existing_names: List[str]):
         def generateObjectName():
```
emilmelnikov, line 244: I think moving `generateObjectName` into `findNextPrefixNumber` would make sense here: we probably only care about generating fresh, unique names.

Also, if you move `name_input_default = name_input_default or generateObjectName()` out of the dialog, it would make the dialog independent of the enclosing class.

ilastik/workflows/carving/carvingGui.py
emilmelnikov: We probably should `name = name.strip()` and treat an empty name as invalid.

emilmelnikov (approved):

k-dominik:
I got carried away. Carving is so much fun to use that I segmented a whole embryo 🐙 . Of course I encountered an issue (that was not there before). With the changes here, one cannot overwrite objects. That's a bit sad, as I wanted to correct one of the early objects lateron, so I loaded it, added one stroke and wanted to resave. This cannot be done anymore. So the workaround would be saving the modified one as an extra object an deleting the old one... Maybe instead of not allowing to overwrite, we could warn?

btbest:
> it would make the dialog independent of the enclosing class.
Sure, though I don't quite follow what benefit that brings.

> Maybe instead of not allowing to overwrite, we could warn?
It feels much more intuitive for me to just save the object when I confirm saving it under the same name as before. What I don't like right now is that we (a) don't have a way to "get out" of a currently loaded object without saving it, combined with (b) when saving a loaded object under a new name, that we delete the old object (which feels very unintuitive and I'm just assuming there is some good reason for this)

btbest:
> there is the "clear labels"... leaves the previously saved object untouched, allows for starting a new one (but probably the internal logic still preserves the name...?!)

It was always only a "clear all labels" button, but I think I will then change it to "clear the current object". If the user just wants to clear labels on a currently loaded object for some reason, they can just delete the old one and make a new one (maybe in a new PR though)

> ah, wow - this is terrible :D . Just to clarify - once we save an object under a different name, the old one gets deleted?!

Yes, there is explicit logic for this. Which is why I assumed that it was a wanted deviation from the "intuitive" thing to do, to push the user into a certain way of working.

```
if was_object_previously_saved:
    if is_name_changed:
        self.topLevelOperatorView.deleteObject(old_name)
    else:
        self._renderMgr.removeObject(old_name)
        self._renderMgr.invalidateObject(old_name)
        self._shownObjects3D.pop(old_name, None)
```

btbest:
The last thing I'm now fixing before merging this is that adding PreprocessedData.setDirty() in OpPreprocessing.propagateDirty now causes the preprocessing to be executed when loading a project file.

That happens because Filter.setValue() somehow leads to OpPreprocessing.propagateDirty being executed, and I don't yet know how (the other InputSlots like e.g. Sigma don't). Either Filter.setValue needs to stop calling propagateDirty, or I need to add another "trigger" flag  similar to how OpCarving does it, to allow the deserializer to block execution until it is done loading.

k-dominik:
why is `setDirty` here even needed?

btbest:
> why is setDirty here even needed?

The downstream opCarving needs to be notified after generating new superpixels to carry over its potentially stored user-drawn labels into the new MST.

Ironically, it seems this is the wrong place though. Due to the caching mechanism, I think here the right place to do this is actually in execute(), in the case where a new MST is generated rather than returning the cached one.

Admittedly, I got a bit lost with various things I tried and I could swear at one point the exact change I just committed didn't work. Could you try on your side whether now during project load, preprocessing isn't executed, but when you change some settings and re-run it, already-drawn labels are carried over?

k-dominik:
hey, it works now (for me). Loading a project is now basically instant.

ilastik/workflows/carving/opPreprocessing.py
```diff
@@ -413,18 +413,11 @@ def execute(self, slot, subindex, roi, result):
...
-            mst.object_seeds_bg_voxels = self.cachedResult[0].object_seeds_bg_voxels
-            mst.object_seeds_fg_voxels = self.cachedResult[0].object_seeds_fg_voxels
-            mst.bg_priority = self.cachedResult[0].bg_priority
-            mst.no_bias_below = self.cachedResult[0].no_bias_below
-
         self.cachedResult = result
 
         result[0] = mst
+
+        self.PreprocessedData.setDirty(slice(None))
```
k-dominik, line 420: not sure this is necessary, or useful - since we are already in `execute` and given the assert in the beginning of this function, we are requesting `self.PreprocessedData`. So in principle we would _not_ want to send any dirtyness upstream here (we just gave the data, setting the slot dirty invalidates it).

btbest: PreprocessedData is the output of opPreprocessing, and feeds into opCarving's input `MST`.

I don't see any other way for opCarving to know that opPreprocessing.execute generated a new MST object. The question then is who tells opCarving that there is a new MST, and when:

1. When one of opPreprocessing's inputs change (in `opPreprocessing.propagateDirty`). This happens either when the Run button is pressed, or when we load a project file. Problem is that changed input implies a new MST needs to be generated, so this is why previously loading a project also caused it to re-run.
2. When the Run button is pressed. The preprocessingGui could give the request to opPreprocessing a callback to notify opCarving once it's done, or maybe the Workflow should be attaching this as a listener to the output slot?
3. Whenever a new MST is generated - which is what I did here

k-dominik: > I don't see any other way for opCarving to know that opPreprocessing.execute generated a new MST object

when you are in execute, upstream already requested that data, so it "knew" it was dirty. So requesting the data from the slot is the reason we get into `execute` at all. Doing any dirty setting _after_ basically invalidates the result immediately - not sure why it even currently works.

btbest: > we just gave the data

The caching implementation here basically runs against the intended use I think. execute hands over the currently cached data when `_dirty==False`, which is supposed to be the case whenever the downstream opCarving requests its MST input's value - i.e. the default use of execute().

It's only the special case when the Run button in preprocessingGui is clicked that opPreprocessing inputs change, hence `_dirty=True` and when the Gui itself makes a request for opPreprocessing's output.

k-dominik: thx for the clarification (over zoom) @btbest :)

k-dominik (approved):

Issue #2626, opened 2022-10-24
Carving, object names of only numbers error out


k-dominik:
## Describe the bug
When saving an object and giving it only a numerical name, some error will be raised (probably we expect it to be in the form of \w+ \d? or something like that).
Naming an object with a single number should also work (so - empty prefix).
Prefix could also change during the course of the processing.

At least two objects need to be saved.

Error appears when clicking on `Browse objects`

## Expected behavior
Single digit object names should work (or otherwise not be possible to save)

## To Reproduce
Within the Carving workflow
1. Add some  in the data selection applet
2. Do appropriate preprocessing
3. In the labeling applet, carve something out, and save it e.g. as `1`
4. Carve something different and save it with _any_ name
5. Click on `Browse objects` --> 💥 

## Error message/traceback

If applicable, please add tracebacks with formatting, such as

```pytb
ERROR 2022-10-24 09:45:37,027 excepthooks 19396 8593368576 Unhandled exception in thread: 'MainThread'
ERROR 2022-10-24 09:45:37,028 excepthooks 19396 8593368576 Traceback (most recent call last):
  File "/.../ilastik-1.4.0rc6-OSX.app/Contents/Resources/lib/python3.7/site-packages/ilastik/workflows/carving/carvingGui.py", line 359, in onShowObjectNames
    dialog.objectNames.addItems(sorted(names, key=_humansort_key))
TypeError: '<' not supported between instances of 'int' and 'str'
```

## Screenshots

<img width="431" alt="image" src="https://user-images.githubusercontent.com/24434157/197474236-45b3a815-1933-4890-bb51-fae438510e68.png">

## Desktop (please complete the following information):
 - ilastik version: `1.4.0rc6`
 - OS: all

btbest:
Using only digits as object names actually works. It breaks once you mix purely-digit names with names containing string bits.

This is due to a custom sorting key we have here, which splits strings and converts digit-only chunks into integer values. So to be specific, the sorting of object names will work as long as all object names follow the same pattern of digits-vs-letters ordering. E.g. objects named "1", "1a1", "2b2a", "3c3c" will be handled just fine, as long as you don't add a name that starts with a non-digit.

PR #2665, opened 2023-01-30
Carving: Enable saved-objects-list sorting to handle digits and letters

btbest:
Fix for #2626.

This is for sorting the list of saved carved objects when the user wants to browse and maybe load. The previous implementation converts any chunks of digits within object names into int, which can break sorting when it leads to comparisons between str and int. This happens when some objects start with a digit and others start with a non-digit.

To avoid building something more sophisticated that will sort numbers numerically and non-digits alphabetically, we instead now do numerical sorting only for a trailing number if it exists, or if the entire name is a number.

## Checklist
- [ ] Add tests.

### DIFF
--- a/ilastik/workflows/carving/carvingGui.py
+++ b/ilastik/workflows/carving/carvingGui.py
@@ -21,9 +21,10 @@
 # Python
 import os
 import re
+import string
 from functools import partial
 from collections import defaultdict
-from typing import List, Union
+from typing import List
 import numpy
 
 # PyQt
@@ -314,7 +315,7 @@ class CarvingGui(LabelingGui):
         """show object names and allow user to load/delete them"""
         dialog = uic.loadUi(self.dialogdirCOM)
         names = self.getObjectNames()
-        dialog.objectNames.addItems(sorted(names, key=_humansort_key))
+        dialog.objectNames.addItems(human_sorted(names))
 
         def loadSelection():
             selected = [str(name.text()) for name in dialog.objectNames.selectedItems()]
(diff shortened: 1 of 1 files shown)


ilastik/workflows/carving/carvingGui.py
```diff
@@ -793,23 +793,23 @@ def onButtonsEnabled(slot, roi):
...
+    ['3','a 1', 'a 10', 'a 9', 'b 2', 'c']
     >>> sorted(lst, key=_humansort_key)
-    ['a 1', 'a 9', 'a 10', 'b 2']
+    ['3','a 1', 'a 9', 'a 10', 'b 2', 'c']
     """
     if not (elem and isinstance(elem, str)):
         return tuple()
-
-    return tuple(_str_to_int(token) for token in re.split("(\d+)", elem) if token)
+    split = [token for token in re.split(r"(\d+)$", elem) if token]
```
emilmelnikov, line 807: Awesome!

However, this will give (somewhat) strange result on `["x1", "x0", "x"]` -> `["x0", "x", "x1"]` because `x0` and `x` map to the same key, and Python sort is stable. It seems more reasonable to sort this input as `["x", "x0", "x1"]`.

Instead of explicitly converting to int, we could try to sort numeric suffixes as strings too, but compare their lengths first.

Also, I think it's better to precompute all keys since this key func is not just data projection (probably won't make a huge difference though).

```python
def humansorted(items):
    digits = "0123456789"
    keys = {}
    for item in items:
        i = next((i for i in range(len(item), 0, -1) if item[i - 1] not in digits), 0)
        keys[item] = item[:i], -i, item[i:]
    return sorted(items, key=keys.__getitem__)
```

emilmelnikov: Totally forgot about the `strip` methods! Obviously, they are much better.

k-dominik (approved):
LGTM

emilmelnikov (approved):

PR #2667, opened 2023-02-01
License data copied to ilastik root folder during release process

k-dominik:
setting env variables reliably on app startup tended to be error prone. now we copy license files statically into the ilastik root

## Checklist

- [x] Add tests.

### DIFF
--- a/ilastik/shell/gui/licenseDialog.py
+++ b/ilastik/shell/gui/licenseDialog.py
@@ -80,13 +80,13 @@ class LicenseDialog(QDialog):
     def __init__(self, parent=None):
         super().__init__(parent=parent)
 
-        ilastik_path = Path(ilastik.__file__).parent.parent
+        ilastik_path = Path(ilastik.__file__).parent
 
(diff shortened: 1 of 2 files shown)


emilmelnikov:
Do we really need to list all licenses?.. We are open source, so everyone could go to our repository and find out the exact code we are pulling into our project. I've only seen those screens listing licenses of open-source software in closed-source applications (it makes sense in those cases because there is no public way to know what is being used under the hood unless they tell you).

k-dominik:
hm basically look at _any_ license. The minimum requirement is distribution of the license along the code (which we do of course). I don't think it's about being closed source, per se, it's more for binary distributions where you have to jump additional hoops. We technically link against everything dynamically and distribute the accompanying licenses.

Given all that I'm still not sure what speaks against proper attribution here. Users should know that a lot of other work goes also into ilastik, indirectly. Also why make it hard. If anything, it's not done enough - also in open source circles :)

btbest (approved):

PR #2668, opened 2023-02-06
Autosave before batch

k-dominik:
This PR adds autosave before doing batch processing on clicking of the _Process all files_ button. Furthermore, __dirty tracking is disabled__ for the duration of the batch processing. This results in a "clean" graph after batch. Otherwise it will get inevitably dirty because lanes are added/removed...

- [ ] Add tests.

### DIFF
--- a/ilastik/applets/base/applet.py
+++ b/ilastik/applets/base/applet.py
@@ -18,9 +18,10 @@
 # on the ilastik web site at:
 # 		   http://ilastik.org/license.html
 ###############################################################################
-from lazyflow.utility.orderedSignal import OrderedSignal
 from abc import ABCMeta, abstractproperty, abstractmethod
+from enum import Enum, auto
 from future.utils import with_metaclass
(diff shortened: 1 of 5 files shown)


btbest:
Looks fine.

Odd that the file doesn't get written on hard disk when "saving project", but only when closing ilastik afterwards, but that's unrelated here 😅

k-dominik:
> lgtm
> 
> Odd that the file doesn't get written on hard disk when "saving project", but only when closing ilastik afterwards, but that's unrelated here 😅

hm, there is the call to `.flush` in `saveProject`... 

https://github.com/ilastik/ilastik/blob/02b0ed5052dfe52efdc93bf19bf965b061d0dc0d/ilastik/shell/projectManager.py#L313

btbest:
In that case I'm guessing that Windows doesn't update the last-modified timestamp until the handle is released by ilastik, even if the file was written in the meantime...

PR #2669, opened 2023-02-06
Better default value for maxNearestNeighbors

k-dominik:
because default values matter!

in most cases, the maxNearestNeighbors value should be larger than 1, to allow for greater flexibility in the tracking solution.

New default is 3 (discussed with @akreshuk)


btbest (approved):
Defaults do matter, and I trust Anna's judgment on what makes a good default here much more than anything I could add :)

I also don't see any code occurrences that would obviously break with the new default value, so lgtm.

PR #2670, opened 2023-02-07
Change startup screen to the 3-column layout

emilmelnikov:
Because v1.4.0 is imminent, don't change too much, just move _Open Project_ and _Open Recent Project_ buttons, make the logo smaller, and add some external links.

image

### DIFF
--- a/ilastik/shell/gui/ui/ilastikShell.ui
+++ b/ilastik/shell/gui/ui/ilastikShell.ui
@@ -103,6 +103,48 @@
              <item>
               <layout class="QVBoxLayout" name="verticalLayout_6">
                <item>
+                <spacer name="verticalSpacer_5">
+                 <property name="orientation">
+                  <enum>Qt::Vertical</enum>
+                 </property>
+                 <property name="sizeType">
+                  <enum>QSizePolicy::Fixed</enum>
+                 </property>
+                 <property name="sizeHint" stdset="0">
+                  <size>
+                   <width>20</width>
+                   <height>32</height>
+                  </size>
+                 </property>
+                </spacer>
+               </item>
+               <item>
+                <widget class="QLabel" name="appNameLabel">
+                 <property name="text">
+                  <string>&lt;h1&gt;ilastik&lt;/h1&gt;</string>
+                 </property>
+                 <property name="alignment">
+                  <set>Qt::AlignCenter</set>
+                 </property>
+                </widget>
+               </item>
+               <item>
+                <spacer name="verticalSpacer">
+                 <property name="orientation">
+                  <enum>Qt::Vertical</enum>
+                 </property>
+                 <property name="sizeType">
+                  <enum>QSizePolicy::Fixed</enum>
+                 </property>
+                 <property name="sizeHint" stdset="0">
+                  <size>
+                   <width>20</width>
+                   <height>32</height>
+                  </size>
+                 </property>
+                </spacer>
+               </item>
+               <item>
                 <widget class="QLabel" name="label">
                  <property name="maximumSize">
                   <size>
@@ -122,6 +164,9 @@
                  <property name="scaledContents">
                   <bool>true</bool>
                  </property>
+                 <property name="alignment">
+                  <set>Qt::AlignCenter</set>
+                 </property>
                 </widget>
                </item>
                <item>
@@ -129,13 +174,10 @@
                  <property name="orientation">
                   <enum>Qt::Vertical</enum>
                  </property>
-                 <property name="sizeType">
-                  <enum>QSizePolicy::Minimum</enum>
-                 </property>
                  <property name="sizeHint" stdset="0">
                   <size>
-                   <width>20</width>
-                   <height>16</height>
+                   <width>0</width>
+                   <height>0</height>
                   </size>
                  </property>
                 </spacer>
@@ -154,12 +196,12 @@
                 </widget>
                </item>
                <item>
-                <spacer name="verticalSpacer_5">
+                <spacer name="verticalSpacer_9">
                  <property name="orientation">
                   <enum>Qt::Vertical</enum>
                  </property>
                  <property name="sizeType">
-                  <enum>QSizePolicy::Minimum</enum>
+                  <enum>QSizePolicy::Fixed</enum>
                  </property>
                  <property name="sizeHint" stdset="0">
                   <size>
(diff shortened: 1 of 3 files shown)


k-dominik:
Okay, I think it's fair to do the minimal thing now, and the more intricate change directly in the beginning of the next cycle.

I had a look on all platforms that are available to me:

<details>

Windows 10:
Screenshot 2023-02-07 at 14 10 32

Macbook pro 2022, osx 12.5.1
Screenshot 2023-02-07 at 13 53 26

Macbook air 2015, osx 10.13:
osx-10 13

Ubuntu 20.04 i3
i3

Ubuntu 20.04 X

ubuntu-x

Ubuntu 20.04 wayland
wayland

</details>

I think all look fine. Small observations:
* Once filename _and_ workflow name get abbreviated the buttons for old projects no longer resize - and the whole area will get a horizontal scrollbar. While this was always the case, this now can happen much earlier.
* my sense for aesthetics wants to move the logo down so its upper edge aligns with maybe the lower edge of the first headings of the content columns ("Create New Project", "Open Project").

emilmelnikov:
> Once filename _and_ workflow name get abbreviated the buttons for old projects no longer resize - and the whole area will get a horizontal scrollbar. While this was always the case, this now can happen much earlier.  **--> probably not easy to change**

I tried to just break file path and workflow name into 2 lines, and it seems to look better.

> my sense for aesthetics wants to move the logo down so its upper edge aligns with maybe the lower edge of the first headings of the content columns ("Create New Project", "Open Project"). --> **probably also not easy to do in a portable way. It wasn't so unnerving when the logo was in the center, imo.**.

I tried a different layout with a new title label at the top and links moved to the bottom. The title and some additional spacings visually move the logo a bit down. Let me know if you like it or not.

Screenshot 2023-02-07 at 15 07 13

btbest:
Win11 for completeness. I'm ok with it.

Nice-to-haves for me would be:

- vertically centered title/icon column
- not sure what's up with the spacing for the line-break character separating file path and workflow name

<details>

image

</details>

akreshuk:
The placement of the logo doesn't seem that ugly to me, so I wouldn't spend much time optimizing it. But what about the empty space in the lower part of the window on all platforms except Mac? Can we get rid of it? Overall, Windows10 is just beyond ugly, really hurts the eyes, but people who use it must be desensitized to aesthetics already.

k-dominik:
> not sure what's up with the spacing for the line-break character separating file path and workflow name

that _would_ be a nice fix.

Otherwise LGTM.

btbest:
> But what about the empty space in the lower part of the window on all platforms except Mac

Which space are you referring to? The space between "Other workflows" and the bottom seems consistent (and anyway depends on window size)

emilmelnikov:
I wanted to use the line-break character (technically _Arrow Pointing Downwards Then Curving Rightwards_) as a way to break up file path and workflow name a bit. Otherwise they are left-aligned and kind of merge together. Unfortunately, Qt buttons don't understand HTML, so using explicit line breaks and Unicode symbols was the quick and dirty way to do something similar.

emilmelnikov:
The arrow was replaced with a fixed-width 1em space.
image

btbest:
Looks good! (Well, as good as it will on Win11...)

image

k-dominik:
it's kinda cool how compact the startup screen now can be

<details>

win 10:
image


</details>

k-dominik (approved):

emilmelnikov:
> it's kinda cool how compact the startup screen now can be

Now it should be even more compact if we keep the second column from expanding.

k-dominik:
> Now it should be even more compact if we keep the second column from expanding.

hm, not sure I get the advantage - now it'll look weird on wider setups:

<details>

image

before:

image


</details>


I guess the optics are "only" a problem on osx - as the outlines are not visible on windows ...

emilmelnikov:
> * I really think we should try a stylesheet here for osx

I tried a more lightweight design with a flat look in a normal state, and hover/pressed indications.

image

image

The new buttons don't stretch to the maximum, except for `FilePathButton` ones: I haven't been able to convince the Qt layout system to properly size them in our custom code.

btbest:
Nice, looks great!

The non-maxwidth sizing of the buttons (other than filepaths) is unfortunate but I'd be willing to take it for the cross-platform consistency.

k-dominik:
I love the look! Do you think it is possible to reduce vertical spacing between buttons there a bit? 

Side note, we should look into styling all ilastik more like this :)

emilmelnikov:
> Nice, looks great!
> 
> The non-maxwidth sizing of the buttons (other than filepaths) is unfortunate but I'd be willing to take it for the cross-platform consistency.

I actually wanted buttons to not to fill their containers' widths first (IMO it looks cleaner), but because "recent project" buttons fill the whole width, I guess we're better to do the same for all buttons.

emilmelnikov:
> Do you think it is possible to reduce vertical spacing between buttons there a bit?

image

emilmelnikov:
Final version:

https://user-images.githubusercontent.com/1649961/217839056-b41635ed-20a7-4824-9ce1-b0ef33aaf242.mov

ilastik/widgets/filePathButton.py
```diff
@@ -1,91 +1,82 @@
...
+        self._path = PurePath(path)
+        self.setToolTip(str(self._path))
+
+        self._suffix = f"\n{subtext}" if subtext else ""
+
+        # Icon should be set before the text in order to be accounted for in the size hint.
+        if icon is not None:
+            super().setIcon(icon)
+
+        self._abbrevs = [(PurePath(), 0)]
```
k-dominik, line 40: any reason to initialize `self._abbrevs` here, also with a value type (`tuple`) that is otherwise not used in this variable.

```suggestion
```

emilmelnikov: This is just to always have an initialized field in `__init__` with some default value.

k-dominik: Okay, then it should probably go into the else of the above if clause, as this already calls `_update_abbrevs` if there is an icon.

ilastik/shell/gui/ui/ilastikShell.ui
```diff
@@ -88,7 +88,16 @@
              <property name="spacing">
               <number>0</number>
              </property>
-             <property name="margin">
```
k-dominik, line 91: always beautiful to see the noise QtDesigner introduces ;)

ilastik/widgets/filePathButton.py
emilmelnikov: It calls `setIcon()` from `super()`, so it just sets an icon.

i agree that this class doesn't behave in the most intuitive way, but I haven't been able to come up with a better solution (`resizeEvent` needs to be fast, so we can't recompute `_abbrevs` every time there).

ilastik/shell/gui/ui/ilastikShell.ui
emilmelnikov: I've given up on manually correcting elements from `.ui` files...

ilastik/widgets/filePathButton.py
k-dominik: ah right, I'll not loose sleep over it.

k-dominik:
k, let's get it in!

PR #2671, opened 2023-02-09
Carving: Also clear object when clearing labels

btbest:
This allows the user to use the "clear" button to not only remove all currently drawn labels, but also to get "out" of a currently loaded object (see [previous comment](https://github.com/ilastik/ilastik/pull/2664#issuecomment-1410503826)).

I've also inlined a couple of extracted functions that were previously used by an alternative saving mechanism. There used to be "Save" and "Save As", but we got rid of the former a few years ago.

### DIFF
--- a/ilastik/workflows/carving/opCarving.py
+++ b/ilastik/workflows/carving/opCarving.py
@@ -282,14 +282,6 @@ class OpCarving(Operator):
         return names
 
     @Operator.forbidParallelExecute
-    def attachVoxelLabelsToObject(self, name, fgVoxels, bgVoxels):
-        """
-        Attaches Voxellabes to an object called name.
-        """
-        self._mst.object_seeds_fg_voxels[name] = fgVoxels
-        self._mst.object_seeds_bg_voxels[name] = bgVoxels
-
-    @Operator.forbidParallelExecute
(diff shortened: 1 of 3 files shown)


ilastik/workflows/carving/opCarving.py
```diff
@@ -498,27 +454,45 @@ def get_label_voxels(self):
...
-        self.attachVoxelLabelsToObject(name, fgVoxels=fgVoxels, bgVoxels=bgVoxels)
+        logger.info(f"   --> Saving object {name!r}")
+        if name in self._mst.object_names:
+            objNr = self._mst.object_names[name]
+        else:
+            # find free objNr
+            if len(list(self._mst.object_names.values())) > 0:
+                objNr = numpy.max(numpy.array(list(self._mst.object_names.values()))) + 1
+            else:
+                objNr = 1
```
emilmelnikov, line 476:
```suggestion
        if name in self._mst.object_names:
            objNr = self._mst.object_names[name]
        else:
            # Find next free object number.
            objNr = max(self._mst.object_names.values(), default=0) + 1
```

k-dominik (approved):
Makes more sense now from a user perspective. We should make sure the docs match the current behavior.

PR #2672, opened 2023-02-09
Carving: Prevent running segmentation when fore- or background labels are missing

### LINKED ISSUE #2656: Carving is buggy when running segmentation with only foreground or only background labels
## Describe the bug
In the Carving workflow, when running segmentation with only foreground or only background labels, bad things happen.

When only background labels were drawn, the segmentation "works", i.e. the software is fine technically, but the result is nonsense.

When only foreground labels were drawn, ilastik errors out and the applet freezes. The program finally crashes if I go back to the previous step (applet) to redo the rest of the workflow from there.

## Expected behavior
Segmentation should either work, or the program should prevent me from doing this.

Logically it seems it should be sufficient to have only one labels. If there are only foreground labels, only the labelled superpixels are part of the object; if there are only background labels, all superpixels except the labelled ones are part of the object.

## To Reproduce
Within the Carving workflow
1. Open some image, choose some preprocessing settings
2. Go to Labeling (Carving) applet
3. Draw either some foreground labels or some background labels
4. Click "Segment"

## Error message/traceback
With only foreground labels (no error with only background labels):
```pytb
  File "%USER%\AppData\Roaming\JetBrains\IntelliJIdea2022.3\plugins\python\helpers\pydev\_pydev_bundle\pydev_monkey_qt.py", line 234, in _new_run
    return self._original_run()
  File "%USER%\Code\ilastik-group\volumina\volumina\view3d\meshgenerator.py", line 126, in run
    for label, mesh in labeling_to_mesh(self._labeling, self._labels):
  File "%USER%\Code\ilastik-group\volumina\volumina\view3d\meshgenerator.py", line 62, in labeling_to_mesh
    vertices, normals, faces = march(where(labeling == label, 2, 0).astype(int).T, 3)
ValueError: vertex or face count are zero: terminating marching cubes
```

## Screenshots
With only foreground labels: (the applet is frozen with an infinite progress bar under the bottom-right 3D volumina panel showing the segmentation)
bug-foreground-only

With only background labels, I managed to get one time some segmentation that was nonsense (seemed a few random superpixels were selected as part of the object). I didn't manage to reproduce this though for a screenshot. On further attempts, clicking "Segment" just froze ilastik for a second and then it recovered without producing any segmentation (and without console output). Log just receives the two lines "compute new carving results --> ... carving took x sec" as normal.

 - ilastik version: ilastik-1.4.0rc8
 - OS: Win11

btbest:
This closes #2656 

The watershed can't handle either fore- or background labels being missing currently. I don't see why it couldn't, but fixing that in ilastiktools is a different scale of work.

So instead we quick-fix by preventing the user from running segmentation if one of the label types is missing.

### DIFF
--- a/ilastik/workflows/carving/carvingGui.py
+++ b/ilastik/workflows/carving/carvingGui.py
@@ -219,6 +219,13 @@ class CarvingGui(LabelingGui):
 
     def onSegmentButton(self):
         logger.debug("segment button clicked")
+        if not self.topLevelOperatorView.CanRunSegmentation.value:
+            QMessageBox.critical(
+                self,
+                "Unable to Run Segmentation",
+                "Both foreground and background labels are required to run segmentation.",
+            )
+            return
         bkPriorityValue = self.labelingDrawerUi.backgroundPrioritySpin.value()
         self.topLevelOperatorView.BackgroundPriority.setValue(bkPriorityValue)
         biasValue = self.labelingDrawerUi.noBiasBelowSpin.value()
--- a/ilastik/workflows/carving/opCarving.py
+++ b/ilastik/workflows/carving/opCarving.py
@@ -107,6 +107,7 @@ class OpCarving(Operator):
     AllObjectNames = OutputSlot(rtype=List, stype=Opaque)
 
     CanObjectBeSaved = OutputSlot(stype="bool")
+    CanRunSegmentation = OutputSlot(stype="bool")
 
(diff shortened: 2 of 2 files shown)


k-dominik (commented):
Right now you can get into an inconsistent state by adding annotations, segmenting, and removing annotations. Then you'll get the error, but the old segmentation is still showing. I would say this is acceptable (well and much better than erroring out) :)

PR #2677, opened 2023-02-20
Set stylesheet as early as possible

k-dominik:
this was necessary for correct size calculation on linux

before:

image

with this PR:

image

### DIFF
--- a/ilastik/widgets/filePathButton.py
+++ b/ilastik/widgets/filePathButton.py
@@ -27,8 +27,8 @@ class FilePathButton(QPushButton):
         The subtext, if not empty, is displayed below the path.
         """
         super().__init__(parent)
-        # Setting stylesheet early is necessary for correct size calculation
-        # on all platforms
+        # Setting stylesheet early is necessary for correct size calculation on all
+        # platforms (linux would otherwise result in sizes which were too small).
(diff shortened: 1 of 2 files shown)


ilastik/widgets/filePathButton.py
```diff
@@ -17,6 +17,7 @@ def __init__(
         path: Union[str, os.PathLike],
         subtext: str = "",
         icon: Optional[QIcon] = None,
+        styleSheet: Optional[str] = None,
```
emilmelnikov, line 20:
```suggestion
        styleSheet: str = "",
```

ilastik/widgets/filePathButton.py
```diff
@@ -26,6 +27,10 @@ def __init__(
         The subtext, if not empty, is displayed below the path.
         """
         super().__init__(parent)
+        # Setting stylesheet early is necessary for correct size calculation
+        # on all platforms
```
emilmelnikov, line 31: Maybe explicitly mention Linux, or link to this PR in a comment? But I'm not sure whether it's really necessary.

emilmelnikov (approved):
For once I'm glad that an issue has been caused by the underlying platform :)

PR #2690, opened 2023-04-05
fix elf ws now returns uint64

k-dominik:
since 0.46, elf watershed returns `uint64`. As we do some further processing with `vigra` we need to cast to `uint32` (`uint64` not supported by `vigra`).

xref: https://github.com/constantinpape/elf/issues/68

(also removed an unused import in the tests)

### DIFF
--- a/ilastik/applets/wsdt/opWsdt.py
+++ b/ilastik/applets/wsdt/opWsdt.py
@@ -105,6 +105,8 @@ def parallel_watershed(
             f"processing block {block_index} {block.outerBlock.begin}-{block.outerBlock.end} took {btimer.seconds()}"
         )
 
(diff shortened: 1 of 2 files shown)


btbest (approved):
Looks good to me; and as far as I'm also happy that this would remain compatible with older elf versions where the return is already uint32.

The comment in line 108 isn't needed imo (given that line 228 also remains uncommented), but eh, just nitpicking. Or am I misunderstanding the comment and what it's trying to say is that at some future point, this cast (but not the one on line 228) would need to be removed?

k-dominik:
awww I wanted to safe some lines and thought one comment would be enough, but you're right, I'll add it in both places :) thx for the review!

PR #2692, opened 2023-04-13
mnt: update pytorch versions

### LINKED ISSUE #2691: investigate pytorch dependency issues (with cuda)
seems that recent changes to pytorch (on pytorch channel) seem to lead to some insolvable envs (at least on windows).

k-dominik:
fixes #2691

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -108,7 +108,6 @@ outputs:
         - pytorch >=1.6
         - tensorflow 1.14.*
         - tiktorch 23.2.0*
-        - cpuonly
         - inferno
         - torchvision
     test:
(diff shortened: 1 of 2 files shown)


conda-recipe/meta.yaml
```diff
@@ -120,7 +120,7 @@ outputs:
         - pytest >=3,<4
         - pytest-qt
         # need to help mamba here a bit
```
btbest, line 122: I guess this comment can go now?

btbest (approved):

conda-recipe/meta.yaml
k-dominik: ah crap I moved stuff around - this should actually stay, but need to move down the `ilastik-pytorch-version-helper-cpu` thing...

k-dominik: or the comment up :)

k-dominik: ah, nah, this should actually be in test reqs - good catch thx :)

PR #2694, opened 2023-04-14
functionality to ignore multiple dirty propagations from same source

k-dominik:
Dirty propagation in ilastik goes depth-first. It is possible for dirtiness to propagate to the same target via multiple upstream paths. This commit adds the concept of a "modification time" (`mod_time` in the code) to encode sources of dirtiness.
Operators will keep track of the last `mod_time` they've seen (during dirty propagation) and can ignore it, if it is already known (see `Operator.setAllDirty`).

In more complicated graphs like in autocontext, this solves heavy delays when annotating.

image

To illustrate the problem, consider the above DAG, a "simplified" auto-context-like setup. In normal dirty prop, when setting the input dirty, the following sequence would occur:

```
                                    _____________________ BLOCK A ____
                                   |                                  |
Input --> Filter1 --> Stacker1 --> Filter4 --> Stacker2               |
                               --> Filter5 --> Stacker2               |
                               --> Filter6 --> Stacker2 ______________|
      --> Filter 2 --> Stacker1 --> BLOCK A
      --> Filter 3 --> Stacker1 --> BLOCK A
```

Here, via BlockA Stacker2 will in total emit 9 dirty signals.

With the changes from this PR, any change in Input1 will be labeled with `mod_time` - downstream operators can choose to ignore subsequent dirty propagation:

```
Input --> Filter1 --> Stacker1 --> BLOCK A
      --> Filter 2 --> Stacker1: no Dirty prop IGNORED
      --> Filter 3 --> Stacker1: no Dirty prop IGNORED
```
Here, via BlockA, Stacker2 will only propagate dirtiness once.

For reference, with a real world Autocontext project with 6 lanes, the latency for annoations in the first stage went down from 16 seconds to 0.25 seconds.

Despite being very defensive here with this new feature, I want to keep this PR in a single commit, to make it easily revertible (meaning I'll squash any changes, if necessary).

One of the side effects here is that `MultiArrayStacker` will _not_ emit any fine grained dirty signals anymore, and always set complete output roi dirty.

For more simple workflows this optimization doesn't make a big difference. It's really only apparent if there are multiple "merging" operators, like `OpMultiArrayStacker` and the classifier operators.

For operators that do not use `setAllDirty`, nothing changes in behavior (still, the `mod_time` is propagated through the whole graph).

### DIFF
--- a/lazyflow/operator.py
+++ b/lazyflow/operator.py
@@ -631,21 +631,20 @@ class Operator(metaclass=OperatorMetaClass):
         # return self._debug_text
         return "setups: {}".format(self._setup_count)
 
-    def setAllDirty(self):
-        """set all outputs dirty and ignore subsequent dirty prop from same source
+    def propagateDirtyIfNewModTime(self):
+        """set all outputs dirty and ignore if _pending_dirty_mod_time is new
 
-        should never be called outside `propagateDirty`.
+        This ignores subsequent dirty prop from same source (=same _pending_dirty_mod_time)
+
+        Should only be called within `Operator.propagateDirty`.
+        See https://github.com/ilastik/ilastik/pull/2694
         """
+        assert self._pending_dirty_mod_time != -1
         if self._pending_dirty_mod_time <= self._previous_dirty_mod_time_buffer:
             return
 
         for slot in self.outputs.values():
             slot.setDirty((), _mod_time=self._pending_dirty_mod_time)
-
-
-#    @debug_text.setter
-#    def debug_text(self, text):
-#        self._debug_text = text
 
 
 def format_operator_stack(tb):
--- a/lazyflow/operator.py
+++ b/lazyflow/operator.py
@@ -258,7 +258,7 @@ class Operator(metaclass=OperatorMetaClass):
         # keeps track of any setDirty calls on inputs slots
         self._previous_dirty_mod_time_buffer = -1
         # temporary variable during dirty notification to buffer the previous
-        # value. Used in `setAllDirty` in order to ignore multiple dirty
+        # value. Used in `propagateDirtyIfNewModTime` in order to ignore multiple dirty
         # notifications cause by the same upstream source
         self._pending_dirty_mod_time = -1
 
--- a/lazyflow/operator.py
+++ b/lazyflow/operator.py
@@ -256,11 +256,11 @@ class Operator(metaclass=OperatorMetaClass):
         self._setup_count = 0
 
         # keeps track of any setDirty calls on inputs slots
-        self._last_dirty = -1
+        self._previous_dirty_mod_time_buffer = -1
         # temporary variable during dirty notification to buffer the previous
         # value. Used in `setAllDirty` in order to ignore multiple dirty
         # notifications cause by the same upstream source
-        self._previous_dirty = -1
+        self._pending_dirty_mod_time = -1
 
     @property
     def children(self):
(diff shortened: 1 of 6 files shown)


emilmelnikov:
Is it possible to add a wrapper operator that somehow delays dirtiness propagation until all required input slots report? Something similar to [barriers or memory fences](https://en.wikipedia.org/wiki/Barrier_(computer_science)). If it's possible and not very complicated, I think this approach will be better for debugging later because dirty propagation delays are explicit.

k-dominik:
> Is it possible to add a wrapper operator that somehow delays dirtiness propagation until all required input slots report? Something similar to [barriers or memory fences](https://en.wikipedia.org/wiki/Barrier_(computer_science)). If it's possible and not very complicated, I think this approach will be better for debugging later because dirty propagation delays are explicit.

Idea sounds nice. Not quite sure how to define "required" slots. For me it almost sounds like you'd have to know in downstream ops about upstream ops (some might set only a roi dirty, some the whole thing, some will gather dirty prop from multiple lanes). It would also completely change dirty propagation flow (in a good way though). I'll think about it a bit.

I still think the above approach has its appeal - very little code changes, changed behavior only in a few selected cases.

k-dominik:
Okay, I've added some more changes:

* clearer naming of variables:
  * `op._pending_dirty_mod_time` is only "active" during dirty prop
  * `op._previous_dirty_mod_time_buffer` holds the last completed dirtiness
    `_mod_time`.
* removed the explicit `reset_modtime` parameter of `setDirty`: not needed anymore - all calls to
  `setDirty` outside of dirty propagation now will generate a new `_mod_time`.

btbest:
I've been thinking about how a barrier could be built here, without needing orchestration from the Workflow (i.e. without actually knowing what operators exist and how many inputs they have). What I came up with:

```
class DirtinessBufferingOperator(Operator):
    buffered_dirty_calls: List[Tuple[InputSlot, Any, Any]]

    def propagateDirty(self, slot, subindex, roi):
        self.buffered_dirty_inputs.append((slot, subindex, roi))
        self.queuePropInitiation()

    def queuePropInitiation(self):
        # TODO: queue up a request that will call self.flushBuffer

    def flushBuffer(self):
        for slot, subindex, roi in self.buffered_dirty_calls:
            slot.setDirty(roi) # ?
        self.buffered_dirty_calls = []
```

(I'm not able to "off the top of my head" implement that request-enqueuing, i.e. the TODO).

I am assuming that the flushBuffer requests are handled AFTER all dirty propagation from upstream the barrier is done. The first flushBuffer request would then actually flush, and all subsequent requests queued up by the same dirty propagation would run into the empty buffer.

Depending on how the implementation of the request-enqueuing looks, this could be more self-contained than the current approach. Inheriting operators wouldn't be allowed to do much more than call super.propagateDirty in their own propagateDirty. But it looks like that would be fine for the two operators in question.

k-dominik:
Cool that you're exploring that, in your example, how, when would an operator decide that upstream is done propagating dirtiness? (maybe thinking of the example illustrated above in the first comment?)

btbest:
The idea here is that the operator doesn't decide that at all - it queues up asynchronous function calls to be executed after the dirtiness propagation is finished.

My current understanding of lazyflow is that it effectively implements async; and the dirty-propagation is part of handling one lazyflow request. So we basically cut that request short in order to gather information about the dirty-propagation (in the form of queued-up requests), and then resolve the rest once we know what's coming up.

Though as I noted, I'm not sure we would actually be able to use it, as I assume these calls would be in the same queue as the GUI requests that the dirtiness propagation is trying to handle. That could be problematic as then whatever triggered the dirtiness propagation would leave the workflow in an inconsistent state until the rest of the request queue is worked through and gets to resolving the calls queued up during the dirty-propagation.

btbest (approved):
Much better naming now, and I'm very happy you were able to remove the extra parameter in setDirty!

PR #2705, opened 2023-05-09
Rearrange labeling UI

btbest:
I padded/stretched the tool icons to be as close to the same size as possible with the current designs. It's not perfect, but better. Replacing them with prettier ones tbd in the future when we have time to deal with licensing.

Also rearranged the buttons to get rid of awkwardly expanding buttons and dropdown-selects to the drawer width. Plus I think this just makes a lot more sense intuitively - having everything to do with labeling together, and the "continue in the workflow" buttons visually separate at the bottom.

image

I've increased the minimum width of the applet drawer to 310px to make sure that the buttons never get squished. Except... In long workflows like Object Classification, the tall drawer stack can be too tall for the window. That introduces a scroll bar, which takes up space and eats up the buttons' paddings:

image

Would be cool to see how this looks on Mac and Linux now.


k-dominik:
OSX:

image

this is at minimum width... not sure about the buttons on the bottom - should they maybe also be left/right aligned?
Otherwise looks great :)

btbest:
Not a fan of aligning the buttons in Windows, but I can live with it. On Mac this would probably look better because the drop-down isn't perfectly right-aligned 🤦 What's your opinion?

image

The tool buttons aren't square for you because in great Qt fashion, the paddings only apply to top and bottom. Left- and right-paddings that I tried to set just get ignored and replaced with apparently whatever Qt likes as your OS's default style. I don't know if this would be different with QPushButtons, but at least with QToolButtons, I have the impression that we're forced to accept the tool buttons being rectangular either on Mac (too tall) or on Windows (too wide)...

k-dominik:
for reference, on OSX, the current state:
image


before (main branch):
image

weird that before, the tool buttons were nice and square...

btbest:
They were square before because they didn't have an explicit padding for top and bottom, so it used the default, which on Mac is clearly identical on all sides.. But for Windows having default on all sides means rectangular with width > height. So now with the explicit padding, Windows is square but Mac is rectangular in the other axis :shrug:

Still looks like an overall improvement also for Mac to me...

btbest:
I'm talking nonsense and this is really weird.

New (left) vs. old:
image

I needed the paddings to make the buttons with text look ok, but it looks like the label-related buttons are fine without styling. Does that make any difference for OSX?

k-dominik:
It does make a difference on the mac:
image

I think this looks good :)

k-dominik (approved):

PR #2710, opened 2023-05-22
Restyle label table

btbest:
Finishing up the prettifying work, here are the improvements to the label list.

I also included the tool icon swap, though we will need to include proper attribution if we want to keep those (the eraser and the gradient icon to be specific, the mouse pointer icon is made by yours truly).

Old (left, 1.4.1b2) vs new (right):

image


btbest:
I dropped a commit that removed a bunch of unused imports in random files unrelated to this PR. It confused codecov and probably would confuse others looking at the history.

ilastik/widgets/listView.py
```diff
@@ -61,38 +60,38 @@ def tableViewCellClicked(self, modelIndex):
...
-    #             #print "brush color = {}".format(self._colorDialog.brushColor().name())
-    #             #print "pmap color  = {}".format(self._colorDialog.pmapColor().name())
-    #             self._table.model().setData(modelIndex, (self._colorDialog.brushColor(),
-    #                                               self._colorDialog.pmapColor ()))
-
     def rowMovedTest(self, logicalIndex, oldVisualIndex, newVisualIndex):
         logger.debug("{} {} {}".format(logicalIndex, oldVisualIndex, newVisualIndex))
 
     def _setListViewLook(self):
+        if self.model.columnCount() < 3:
```
emilmelnikov, line 73: Looks like this class is just an extension of `ListView` that shows a different screen when there are no items in it's associated model, so putting label-specific logic here might not be appropriate.

btbest: This is a custom function that was probably put in this ListView because two of the three inheriting sub-ListViews use it. So the original author probably wanted to de-duplicate but also still have the third sub-ListView inheriting from the same base class.

I added the check as a minimal sanity guard, but in retrospect one might as well take it out. Without it, the function would just throw further down in the scenario this check would catch...

ilastik/shell/gui/aboutDialog.py
```diff
@@ -62,6 +62,11 @@
         volumina contributors</a></li>
     </ul>
     </p>
+
+    <p>ilastik uses icons made by
+        <a href="https://www.gimp.org/">the GIMP team</a> under
+        <a href="http://creativecommons.org/licenses/by-sa/4.0/">CC-BY-SA-4.0</a> license.
+    </p>
```
k-dominik, line 69: Maybe also mention the other icons license (even if public domain) there as well. Even if not required.

k-dominik (commented):
I think it looks great. I like that long names are now abbreviated and don't expand the widget towards infinity... The gimp icons are also definitely an improvement!

Are the `.png` versions still required given there are some svgs that at least to my eye look very similar. Along that line is there a reason `cursor.svg` has a corresponding `arrow.png`?

I am still a little reluctant to add a new section to our about dialog for using 4 icons, but it seems that the license might really require it (for me it looks like having a textfile next to the icons could also do the job).

For reference, it looks also good on osx :):

Screenshot 2023-05-22 at 20 20 03 Screenshot 2023-05-22 at 20 19 24

ilastik/widgets/listView.py
btbest: To revise that a bit: This might not be the best place, but understanding and refactoring the inheritance structure around boxListView, cropListView and labelListView wasn't the point of this miniproject. If you know something about it and can suggest a better place, please do.

I checked how the guis end up looking in the workflows Pixel Classification, Pixel + Object Classification, Cell Density Counting and NN (local). I could do that check more systematically as there are I think a couple more workflows that inherit this; though my intention with adding the `raise` here was to break tests so I don't have to do this manually.

btbest:
> For reference, it looks also good on osx :):

yay :D

> Are the `.png` versions still required given there are some svgs that at least to my eye look very similar. Along that line is there a reason `cursor.svg` has a corresponding `arrow.png`?

I committed the .svgs that I exported the .pngs from for my convenience (and/or future reference), but I could of course remove one copy of each image. I would prefer to keep the .svgs, but to get rid of the .pngs, it seems I would have to try out how we can even use .svgs directly. At least it doesn't look like we use .svgs anywhere else in the GUI. And if I remove the .svgs, we end up in that situation again where nobody has the original .svg when you want it rescaled.

> I am still a little reluctant to add a new section to our about dialog for using 4 icons, but it seems that the license might really require it (for me it looks like having a textfile next to the icons could also do the job).

Same, I expected we make a decision here how to handle the attribution and then amend that commit. Or leave the icons out of this PR and do it >>some other time<<<sup>TM</sup>

btbest:
Put this into draft as there are oustanding TODOs relating to icons:

* icon files: Remove either .pngs or .svgs. Ideally keep the .svgs, though this requires figuring out how to use them in the GUI.
* icon license: Decide how/where to handle attribution.

btbest:
I've dropped the commit with the icon changes, so this should be good to merge now.

PR #2716, opened 2023-06-13
Fix conservation tracking tests

btbest:
The six tests in testConservationTrackingHeadless.py have been xfailing for quite a while because file paths were broken. Having fixed that, it turned out also the main function import didn't work as implemented.


btbest:
(see these six tests xfailing e.g. in this run: https://github.com/ilastik/ilastik/actions/runs/4698963212/jobs/8331947378 or most recently in https://app.circleci.com/pipelines/github/ilastik/ilastik/1340/workflows/130b9cc3-3761-43b7-bd3a-ea7609e32c4e/jobs/5647)

btbest:
Draft because (a) there's more tests with the same problem and (b) these shouldn't xfail when they can't find their input files, they should plain old fail.

emilmelnikov:
A small tip: Github now supports [draft pull requests](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request#converting-a-pull-request-to-a-draft), so it's not necessary to add `[DRAFT]` or `Draft:` to PR's name

btbest:
I've found and fixed the remaining tests that had the same file path issues, but those tests suffer from their own additional problem: They contained a switch that would pass all tests if they are unable to import either `multiHypoTracking_with_cplex` or `multiHypoTracking_with_gurobi`; and that's why they were passing locally for me.

I changed the pass-if-missing to xfail-if-missing, and now all of the `structuredLearning` tests are xfailing on CI 😄

emilmelnikov:
> I've found and fixed the remaining tests that had the same file path issues, but those tests suffer from their own additional problem: They contained a switch that would pass all tests if they are unable to import either `multiHypoTracking_with_cplex` or `multiHypoTracking_with_gurobi`; and that's why they were passing locally for me.

Instead of xfailing, we probably should [conditionally skip those tests if a module cannot be imported](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#skipping-on-a-missing-import-dependency).

btbest:
I managed to create a new input .ilp that passes (with Gurobi 9.5.2 on my system)

Turns out the GUI gives you no way to actually see all existing track labels that have been put in Training, and I only found the ones that weren't visible when I inspected the h5 manually.

k-dominik:
Hi @btbest,

sorry this took longer as I had to build the tracking with learning dependencies on apple silicon first (was pushing this off, so thanks for the nudge).

Two tests fail for me:

```
tests/test_ilastik/test_applets/structuredLearningTracking/testStructuredLearningTrackingHytraWithMergers.py::TestStructuredLearningTrackingHeadless::testStructuredLearningTrackingHeadless FAILED [ 50%]
tests/test_ilastik/test_applets/structuredLearningTracking/testStructuredLearningTrackingHytraWithMergers.py::TestStructuredLearningTrackingHeadless::testCSVExport FAILED                        [100%]
```
it appears that the outputs cannot be found, can you double check that in a clean repo, you can run these tests on windows?

e.g. I get an error wrt to not being able to find the following file:
```
ilastik/tests/test_ilastik/test_applets/structuredLearningTracking/../../data/inputdata/mitocheck_2d+t/mitocheck_small_2D+t_mergers-crop_data_Tracking-Result.h5
```

PR #2723, opened 2023-07-07
unfreeze cache before exporting data from tda workflow

k-dominik:
Exports could be incomplete previously :(. Now cache is unfrozen before doing any export.

cc: @Buglakova

### DIFF
--- a/ilastik/workflows/trainableDomainAdaptation/_localTrainableDomainAdaptationWorkflow.py
+++ b/ilastik/workflows/trainableDomainAdaptation/_localTrainableDomainAdaptationWorkflow.py
@@ -103,6 +103,9 @@ class LocalTrainableDomainAdaptationWorkflow(_NNWorkflowBase):
         opDataExport.PmapColors.connect(opClassify.PmapColors)
         opDataExport.LabelNames.connect(opClassify.LabelNames)
 
+        self.dataExportApplet.prepare_for_entire_export = self.prepare_for_entire_export
+        self.dataExportApplet.post_process_entire_export = self.post_process_entire_export
+
         self._applets.append(self.dataExportApplet)
 
     def connectLane(self, laneIndex):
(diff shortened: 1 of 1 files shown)


ilastik/workflows/trainableDomainAdaptation/_localTrainableDomainAdaptationWorkflow.py
```diff
@@ -103,6 +103,9 @@ def _createDataExportApplet(self):
         opDataExport.PmapColors.connect(opClassify.PmapColors)
         opDataExport.LabelNames.connect(opClassify.LabelNames)
 
+        self.dataExportApplet.prepare_for_entire_export = self.prepare_for_entire_export
```
emilmelnikov, line 106: Shouldn't `prepare_*` methods be defined in the `TdaExportApplet` itself?

k-dominik: well, that way, export applet does not have to know implementation details of the workflow. It's a pattern used in a couple of workflows. We could change this such that it would call `parent.prepare_for_entire_export` or something, is a refactor I'm not sure is worth doing right now. Effectively it's doing that right now.

emilmelnikov: Yes, maybe changing this right now is not the best approach.

However, I still don't quite like this pattern (adding/removing/replacing instance methods) being used outside of some specific curcumstances. In Python it's fine, but still could be confusing because `self` in the reassigned method refers to a different object instance, and also it's not obvious that this happens just from `TdaExportApplet` source code.

In order to avoid leaking these implementation details into `ExportApplet`, something like this could be used instead:

```python
class ValueSaver:
    def __init__(self, pairs: Iterable[Tuple[Slot, Any]]):
        self.slots, self.values = zip(*pairs)
        self.saved = []
    
    def save(self):
        self.saved.clear()
        for slot, value in zip(self.slots, self.values):
            self.saved.append(slot.value)
            slot.setValue(value)
    
    def restore(self):
        for slot, value in zip(self.slots, self.saved):
            slot.setValue(value)

    # Could also add __enter__ and __exit__ if desired etc.

...

opNN = self.nnClassificationApplet.topLevelOperator
self.dataExportApplet = TdaDataExportApplet(self, "Data Export")
self.dataExportApplet.exportSaver = ValueSaver([(opNN.FreezePredictions, False)])

...

def prepare_for_entire_export(self):
    self.exportSaver.save()

def post_process_entire_export(self):
    self.exportSaver.restore()
```

k-dominik: I agree about the above points. I also like the approach above, but I think rather than it would probably belong into the applet api to do any preparations for exports, if necessary. Such that one can compose workflows without always modifying the export op. I don't see different workflows using different approaches to that with the same ops/applets.

k-dominik:
current circleci build failing because of: https://github.com/conda-forge/python-libarchive-c-feedstock/issues/35

k-dominik:
okay I merging this despite _unrelated_ failure on circleci

PR #2724, opened 2023-07-13
hacky fix for elf limiting threads

k-dominik:
elf limits threads on import, which has side effects even for the tiktorch process (via env variables). This resulted in single threaded neural network prediction on CPU.
In order to get proper performance out of neural networks on CPU, we unset these variables before starting the process.

ref: https://github.com/constantinpape/elf/issues/75


btbest:
Looks good, my only request would be to add the ref to the elf issue in the commit message along the lines of "We can revert this hack once this issue is closed"

PR #2726, opened 2023-07-13
Fix devenv

k-dominik:
What began as a "simple" update of dependencies (https://github.com/ilastik/ilastik/pull/2726/commits/b06dfe433f8ef4a5ca04b3f64a24edbc3ac75fe8) sent me into a ghost hunt for the cause of segfaults at the end of the tests.

I tried various things, like finally fixing parent/child relationships in volumina and ilastik, but nothing really helped. In the end, the first guess was probably right, that multiple instances of `QApplications` were created. https://github.com/ilastik/ilastik/pull/2726/commits/33f59942c69c51ec38a92490afea3133fbc9ece9 solves this...

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -190,22 +190,35 @@ def pytest_runtestloop(session):
     return True
 
 
-def run_gui_tests(tstcls, gui_test_bag):
-    assert tstcls.app is None, "Should only encounter every class once. Check Sorting"
-    tst_queue = queue.Queue()
-    app = tstcls.app = QApplication.instance() or QApplication([])
+_qapplication = None
+
+
+def qapplication():
+    global _qapplication
+    if not _qapplication:
+        _qapplication = QApplication.instance() or QApplication([])
+
+    app = _qapplication
     app.setQuitOnLastWindowClosed(False)
     app.thread_router = ThreadRouter(app)
-    tstcls.shell = launchShell(None, [], [])
 
-    QApplication.setAttribute(Qt.AA_DontUseNativeDialogs, True)
+    app.setAttribute(Qt.AA_DontUseNativeDialogs, True)
 
     platform_str = platform.platform().lower()
     if "ubuntu" in platform_str or "fedora" in platform_str:
-        QApplication.setAttribute(Qt.AA_X11InitThreads, True)
+        app.setAttribute(Qt.AA_X11InitThreads, True)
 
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -193,7 +193,7 @@ def pytest_runtestloop(session):
 def run_gui_tests(tstcls, gui_test_bag):
     assert tstcls.app is None, "Should only encounter every class once. Check Sorting"
     tst_queue = queue.Queue()
-    app = tstcls.app = QApplication([])
+    app = tstcls.app = QApplication.instance() or QApplication([])
```
emilmelnikov, line 196: That looks quite tricky, awesome that you've figured this out!

`QApplication` instantiation and initial configuration could be moved outside this function into `pytest_runtestloop`:

```python
if QApplication.instance() is not None:
    raise RuntimeError("QApplication is already instantiated")
app = QApplication([])

# This is probably unnecessary, but...
assert QApplication.instance() is app

# Could set all common attributes here or somewhere else just once.
# This should be fine unless some code changes these too.
QApplication.setAttribute(..., ...)


def run_gui_tests(...):
    ...

    app = QApplication.instance()
    # Use app
```

emilmelnikov (approved):
LGTM

btbest:
I don't get segfaults, but I do have test_ImageFileDialog breaking the test run because it opens a file picker dialog that waits for user input (the two tests in that file failed as I cancelled the file picker)... There's already a skipif on the test for `GH_ACTIONS and WIN` - I guess for some reason now it also hangs on WIN without GH_ACTIONS 😅 

When I add `--run-legacy-gui` I also get this, which looks odd but I suppose hopefully harmless. Looks like that also appears on CI.
image

More harmful is that the second test in `testPixelClassificationGui.py`, which is skipped on CI, hangs infinitely on my run with --run-legacy-gui.

This was after `mamba create` with the new env.yml already threw some clobber errors and a SafetyError for hytra, so my guess is that the env is botched for reasons I would have to debug tomorrow.

tests/conftest.py
k-dominik: hm yeah true. Ideally it would use the pytest-qt `qapp` fixture in any case, but this seems like a more major change. Good point with setting the attributes multiple times, that is not intuitive - I'll try pulling it outside and see how it looks :)

PR #2732, opened 2023-07-26
make NN-related operators aware of tensor ids

k-dominik:
In preparation for NN-like workflows to handle models with multiple in- and outputs and support tensor names/ids [which are mandatory in the spec](https://github.com/bioimage-io/spec-bioimage-io/blob/gh-pages/model_spec_latest.md#bioimageio-model-resource-description-file-specification-049). These changes still work only for single in- and output models. Will make this operator properly work with multiple inputs/outputs in follow up work with an example workflow.

Summary
* `OpNNclass` tensor id aware, but expects only a single in- and output
  tensor.
* `ModelSession`:
  * added convenience properties for in and output names.
  * getters for input and output shapes now return dicts with tensor ids
    as keys.
* added a few type annotations for peace of mind.

note: this requires [tiktorch 23.7.0](https://github.com/ilastik/tiktorch/pull/205)

## Checklist

- [x] Add tests.
- [x] remove temp tiktorch testing channel from ab1d49f5b2a86281ca7038de86e0e16ab98ab78d
- [x] move tiktorch `23.7.0` to main label on ilastik-forge

### DIFF
--- a/ilastik/config.py
+++ b/ilastik/config.py
@@ -24,7 +24,7 @@ import os
 import warnings
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Optional, Union
+from typing import List, Optional, Union
 
 import appdirs
 
@@ -83,7 +83,7 @@ filename: in
 
 @dataclass
 class RuntimeCfg:
-    tiktorch_executable: Optional[str] = None
+    tiktorch_executable: Optional[List[str]] = None
     preferred_cuda_device_id: Optional[str] = None
 
 
--- a/lazyflow/operators/tiktorch/classifier.py
+++ b/lazyflow/operators/tiktorch/classifier.py
@@ -24,7 +24,7 @@ import socket
 import numpy
 import warnings
 from collections import defaultdict
-from typing import Iterable, List, Callable, Tuple, Dict
+from typing import Callable, Dict, Iterable, Sequence, Tuple, Union
 
 import xarray
 import grpc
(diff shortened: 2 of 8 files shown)


ilastik/config.py
```diff
@@ -83,7 +83,7 @@
 
 @dataclass
 class RuntimeCfg:
-    tiktorch_executable: Optional[str] = None
+    tiktorch_executable: Optional[List[str]] = None
```
emilmelnikov, line 86: Is it possible to have multiple tiktorch executables? If this is really possible, then it probably makes sense to drop `Optional` and default to `field(default_factory=list)`.

lazyflow/operators/tiktorch/classifier.py
```diff
@@ -91,104 +91,148 @@ def name(self) -> str:
...
-    def output_axes(self) -> List[str]:
+    def output_axes(self) -> Sequence[str]:
+        """Get axes for all model inputs
+
+        linked via index to `output_names`
+        """
         return self.__session.outputAxes
 
-    def get_output_shapes(self) -> List[List[Dict[str, int]]]:
+    def get_output_shapes(self) -> Dict[str, Sequence[Dict[str, int]]]:
```
emilmelnikov, line 119: Is it possible and/or makes sense to simplify this method and `get_input_shapes` by extracting the inner loop code into something like `input_shapes_of(name: str, axes: str) -> Generator[Dict[str, int]]`? It seems like we don't do much across different input/output names.

btbest (approved):
Only thing that jumps out at me is the now ubiquitous asserting on length==1, which feels like we're validating user input rather than values provided by our own code. Which feels like responsibilities could be better defined.

I'm inclined to say it's WIP anyway, and at least the checks provide an anchor for implementing the final product...

k-dominik:
> Only thing that jumps out at me is the now ubiquitous asserting on length==1, which feels like we're validating user input rather than values provided by our own code. Which feels like responsibilities could be better defined.
> 
> I'm inclined to say it's WIP anyway, and at least the checks provide an anchor for implementing the final product...

Hey Thx for looking into this. Having asserts in python code is of course an opinionated topic. To clarify: Part of the operator can already deal with multiple inputs/outputs, other parts cannot. Here they really are slightly better `TODO` comments that are there to help me with the follow up PR(s).

ilastik/config.py
k-dominik: The singular here is correct - it is a single exectutable. But it could be something like `['python', '-m', 'tiktorch']`, or anything alike. But sure, can definitely use the default factory - as far as I can see the code does not rely on it being `None` if not provided, something "falsy" is good enough.

lazyflow/operators/tiktorch/classifier.py
k-dominik: Good point. I'll leave it like this for now, but will keep this in mind once dealing with a multi-input/multi-output model. It would for sure be more efficient if one accesses only a single input/output.

emilmelnikov: I was actually thinking more about readability, not performance: this shouldn't be a performance bottleneck, but tracking multiple shapes/axes and their potential relationships was a little bit difficult for me to do.

PR #2734, opened 2023-07-28
Undo elf fix

### LINKED ISSUE #2725: Revert hacky change mitigating elf overeaching thread limiting
once https://github.com/constantinpape/elf/issues/75 is merged and we've updated our elf dependency, we can revert https://github.com/ilastik/ilastik/commit/921efda35ee718f9644e76477e20955dfe93e30f

k-dominik:
our hacky fix is not necessary when ensuring elf > 0.4.8

Summary:
* updated pin for `elf`
* updated dependencies for pandas2 (elf --> skan2 --> pandas2)
* removed the hacky fix from #2724
* plus only very vaguely related changes
  * watershed in muticut is sensitive to small numerical inaccuracies in filters. The superpixels are not the same on OSX arm-64. In order to still test the RF training there, added a test that uses ground truth to autolabel. Related to that, we require a slightly lower accuracy (0.98 -> 0.97) to compare superpixels to ground truth.
  * Moved some of the testing utility functions into a single module (`tests.test_ilastik.helpers.wait`) in order to reuse them in the multicut test. There is some overlap in the functions there, but will not solve this now.
  * re-enabled Carving tests on osx - they work now (with python>=3.8) 

xref: #2724
fixes #2725

### DIFF
--- a/tests/test_ilastik/helpers/wait.py
+++ b/tests/test_ilastik/helpers/wait.py
@@ -49,10 +49,11 @@ def wait_signal(signal, timeout=1000):
     evtLoop = QEventLoop()
     # QueuedConnection so we don't receive signal before entering the loop
     signal.connect(evtLoop.quit, Qt.QueuedConnection)
-
-    yield
-
-    _exec_with_timeout(evtLoop, timeout, signal)
+    try:
+        yield
+        _exec_with_timeout(evtLoop, timeout, signal)
+    finally:
+        evtLoop.quit()
 
 
 @contextlib.contextmanager
@@ -74,9 +75,11 @@ def wait_slot_ready(slot, timeout=1000):
     # Use QTimer to avoid event firing before waiting has started
     QTimer.singleShot(0, _register)
 
-    yield
-
-    _exec_with_timeout(evtLoop, timeout, slot)
+    try:
+        yield
+        _exec_with_timeout(evtLoop, timeout, slot)
+    finally:
+        evtLoop.quit()
 
 
 def _exec_with_timeout(loop, timeout, timeout_obj):
@@ -88,7 +91,7 @@ def _exec_with_timeout(loop, timeout, timeout_obj):
         loop.quit()
 
     QTimer.singleShot(timeout, _timeout)
-    loop.exec_()
+    loop.exec()
 
     if timeout_exceeded:
         raise RuntimeError(f"Timeout exceeded for {timeout_obj}")
--- a/tests/test_ilastik/helpers/wait.py
+++ b/tests/test_ilastik/helpers/wait.py
@@ -1,0 +1,112 @@
+###############################################################################
+#   ilastik: interactive learning and segmentation toolkit
+#
+#       Copyright (C) 2011-2023, the ilastik developers
+#                                <team@ilastik.org>
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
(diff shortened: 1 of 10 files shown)


tests/test_ilastik/test_workflows/testCarvingGui.py
```diff
@@ -34,14 +34,15 @@
 
 
 # https://bugreports.qt.io/browse/QTBUG-87014
-is_darwin_bigsur = False
+must_skip_darwing_bigsur = False
 if platform.system().lower() == "darwin":
     mac_ver = tuple(map(int, platform.mac_ver()[0].split(".")))
-    if mac_ver >= (10, 6):
-        is_darwin_bigsur = True
+    python_ver = tuple(map(int, platform.python_version().split(".")))
```
emilmelnikov, line 40: I wonder if it's time to move this into some `version` utility module.

tests/test_ilastik/helpers/wait.py
```diff
@@ -0,0 +1,112 @@
...
+
+def wait_until(f, timeout=10):
+    result = f()
+    while not result and timeout:
+        timeout -= 1
+        time.sleep(1)
+        result = f()
+    if not result:
+        raise Exception(f"TIMEOUT! waiting for {f}")
+    return result
```
emilmelnikov, line 38:
```suggestion
def wait_until(f, seconds=10):
    for t in range(seconds + 1):
        time.sleep(t > 0)
        if result := f():
            return result
    raise Exception(f"TIMEOUT! waiting for {f}")
```

tests/test_ilastik/helpers/wait.py
```diff
@@ -0,0 +1,112 @@
...
+    :param timeout: timeout in ms
+    :param signal:
+    """
+    evtLoop = QEventLoop()
+    # QueuedConnection so we don't receive signal before entering the loop
+    signal.connect(evtLoop.quit, Qt.QueuedConnection)
+
+    yield
+
+    _exec_with_timeout(evtLoop, timeout, signal)
```
emilmelnikov, line 55: Wrap `yield` in `try...finally`?

emilmelnikov (approved):
🧝‍♂️🧝‍♀️

tests/test_ilastik/helpers/wait.py
k-dominik: :walrus:

tests/test_ilastik/helpers/wait.py
k-dominik: yeah will be nicer to handle quitting of loop in this function

btbest (approved):
Agree with Emil's comments and nothing further 🆒

tests/test_ilastik/test_workflows/testCarvingGui.py
```diff
@@ -25,24 +25,15 @@
 logger.addHandler(logging.StreamHandler(sys.stdout))
 logger.setLevel(logging.DEBUG)
 
-import platform
 import pytest
 
 
 if os.environ.get("ON_CIRCLE_CI"):
     pytest.skip("These tests are too flaky on circleCI", allow_module_level=True)
+elif not is_mac_py_compatible():
```
emilmelnikov, line 33: Hmm, now looking at this I wonder whether it'd be better to check for `os.environ.get("VOLUMINA_SHOW_3D_VIDGET", "1") == "0"`...

tests/test_ilastik/test_workflows/testCarvingGui.py
k-dominik: this change introduced some side effects, so with @emilmelnikov, we decided to leave it as is...

PR #2736, opened 2023-07-28
fix test mouse move

k-dominik:
**Summary**

* fixed test "mouse moves" by sending event to the viewport
* fixed getting pixel values from viewports by hiding the cursor.

**Details**

tests relying on getting pixel values from one of the viewports seemed to be involving mouse moves (not strokes) in order to move the cursor out of the fov. I think these might have worked by chance before. At least on osx, where `TestPixelClassificationGui.test_7_EraseCompleteLabel` stopped working.

Mouse move works if sent to the viewport imageview itself, which is fixed in https://github.com/ilastik/ilastik/pull/2736/commits/0857884d4c53e3aaf4cd30e4a0aba1e97f0f4470

I have confirmed that this issue also affects windows. There the test passes by coincidence, due to maybe some slightly different rendering of the crosshair.

In order to get pixel values more reliably than moving the mouse cursor (hopefully) away from the sampled pixel, I changed the `getPixelColor` method to hide the cursor prior to sampling the data. I made sure to confirm that all tests using this method do in fact _not_ want to sample the cursor. I left the `moveMouseFromCenter` method in, as it might be useful in the future, although it's currently not really needed anymore. (https://github.com/ilastik/ilastik/pull/2736/commits/0fa48c761f76cc38856f1722e2f0a07fdefb42a7)

### DIFF
--- a/tests/test_ilastik/helpers/shellGuiTestCaseBase.py
+++ b/tests/test_ilastik/helpers/shellGuiTestCaseBase.py
@@ -135,9 +135,10 @@ class ShellGuiTestCaseBase(object):
     def hiddenCursor(self, imgView: ImageView2D):
         """Context manager that hides the given image view's cursor
 
-        Useful, e.g. when accessing the viewports rendered image - where
+        Useful, e.g. when accessing the viewport's rendered image - where
         sampling the cursor would be unwanted.
         """
+        wasVisible = imgView._crossHairCursor.isVisible()
         try:
             imgView._crossHairCursor.setVisible(False)
             # Wait for the gui to catch up
@@ -146,7 +147,7 @@ class ShellGuiTestCaseBase(object):
 
             yield
         finally:
-            imgView._crossHairCursor.setVisible(True)
+            imgView._crossHairCursor.setVisible(wasVisible)
             # Wait for the gui to catch up
             QApplication.processEvents()
             self.waitForViews([imgView])
--- a/tests/test_ilastik/helpers/shellGuiTestCaseBase.py
+++ b/tests/test_ilastik/helpers/shellGuiTestCaseBase.py
@@ -26,7 +26,6 @@ from typing import Iterable, Union
 
 import pytest
 from ilastik.ilastik_logging import default_config
-from past.utils import old_div
 from PyQt5.QtCore import QEvent, QPoint, QPointF, Qt
 from PyQt5.QtGui import QMouseEvent
 from PyQt5.QtWidgets import QAbstractScrollArea, QApplication, qApp
@@ -151,11 +150,12 @@ class ShellGuiTestCaseBase(object):
         return img.pixel(point)
 
     def moveMouseFromCenter(self, imgView, coords, modifier=Qt.NoModifier):
-        centerPoint = old_div(imgView.rect().bottomRight(), 2)
-        point = QPoint(*coords) + centerPoint
+        centerPoint = imgView.rect().bottomRight() / 2
+        point = _asQPointF(coords) + centerPoint
         move = QMouseEvent(QEvent.MouseMove, point, Qt.NoButton, Qt.NoButton, modifier)
-        QApplication.sendEvent(imgView, move)
+        QApplication.sendEvent(imgView.viewport(), move)
         QApplication.processEvents()
+        self.waitForViews([imgView])
 
     def strokeMouse(
         self,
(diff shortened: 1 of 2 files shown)


tests/test_ilastik/helpers/shellGuiTestCaseBase.py
```diff
@@ -150,12 +174,26 @@ def getPixelColor(self, imgView, coordinates, debugFileName=None, relativeToCent
 
         return img.pixel(point)
 
-    def moveMouseFromCenter(self, imgView, coords, modifier=Qt.NoModifier):
-        centerPoint = old_div(imgView.rect().bottomRight(), 2)
-        point = QPoint(*coords) + centerPoint
+    def moveMouseFromCenter(
```
emilmelnikov, line 177: The comment around the sole usage of this method says: ["Move the cursor out of the way so we can sample the image"](https://github.com/k-dominik/ilastik/blob/e01f1f5962ed0bf7369bbf529259fc567ca0f135/tests/test_ilastik/test_applets/pixelClassification/testPixelClassificationGui.py#L613). Now we can just hide the cursor, do we need this method at all?

tests/test_ilastik/helpers/shellGuiTestCaseBase.py
```diff
@@ -150,12 +174,26 @@ def getPixelColor(self, imgView, coordinates, debugFileName=None, relativeToCent
...
+        """Move the mouse to a specific point.
+
+        Args:
+            imgView: View that will receive mouse events.
+            point: Target coordinate in relation to imageView center.
+            modifier: This modifier will be active when pressing, moving and releasing.
+
+        """
+        centerPoint = imgView.rect().bottomRight() / 2
+        point = _asQPointF(point) + centerPoint
```
emilmelnikov, line 192:
```suggestion
        absolutePoint = _asQPointF(point) + imgView.rect().center()
```

tests/test_ilastik/helpers/shellGuiTestCaseBase.py
```diff
@@ -130,15 +131,38 @@ def waitForViews(self, views):
...
+            # Wait for the gui to catch up
+            QApplication.processEvents()
+            self.waitForViews([imgView])
+
+            yield
+        finally:
+            imgView._crossHairCursor.setVisible(True)
+            # Wait for the gui to catch up
+            QApplication.processEvents()
+            self.waitForViews([imgView])
```
emilmelnikov, line 152: If the cursor could be already hidden for some reason:
```suggestion
        wasVisible = imgView._crossHairCursor.isVisible()
        try:
            imgView._crossHairCursor.setVisible(False)
            # Wait for the gui to catch up
            QApplication.processEvents()
            self.waitForViews([imgView])
            yield
        finally:
            imgView._crossHairCursor.setVisible(wasVisible)
            # Wait for the gui to catch up
            QApplication.processEvents()
            self.waitForViews([imgView])
```

emilmelnikov (approved):
This bug is quite subtle... Awesome that is's fixed now!

tests/test_ilastik/helpers/shellGuiTestCaseBase.py
k-dominik: hm, I oriented myself at the `strokeMouseFromCernter` method, that also uses `...Point` for the start and end points, despite them being kind of deltas, too...

tests/test_ilastik/helpers/shellGuiTestCaseBase.py
k-dominik: yeah true, that could make sense (e.g. when doing some nesting with the contextmanager...)

tests/test_ilastik/helpers/shellGuiTestCaseBase.py
k-dominik: You mean for the `movemouseFromCenter`, you'd always want to hide the cursor?

emilmelnikov: > "Move the cursor out of the way so we can sample the image"

[At the call site](https://github.com/k-dominik/ilastik/blob/e01f1f5962ed0bf7369bbf529259fc567ca0f135/tests/test_ilastik/test_applets/pixelClassification/testPixelClassificationGui.py#L613), it seems the mouse is moved out of the way because the crosshair interferes with image sampling, exactly the same problem you fixed in this PR.

k-dominik: yeah, that's why I removed all calls to the `moveMouseFromCenter` method. Because it's not adequate for the use case here... But I can imagine other use-cases that would not necessarily involve hiding the cursor, but actually moving it.

PR #2737, opened 2023-08-03
bigger shape in 'z' for tiffs means faster browsing

k-dominik:
I thought about overengineering this, taking into account available ram and such... but bottom line is, if you look at any 3D multipage tiff in the quadview, chances are that you're accessing maaany pages (and effectively, caching them):

quadview(1)

In this PR the cache size along z is increased from 1 to 32 (arbitrary value, in trials this allowed for smooth scrolling). In most cases this should not even increase the size of used ram (given that the pages are already cached in z=1 blocks).

I don't think there is any benefit of doing something smarter there. If data really is so large that reading it poses a problem, having z=1 would also not be much of a benefit (when accessing in Z) and performance still would be way worse accessing any of the other views (and would equally blow ram). I think in those cases we should advice to use hdf5.

### DIFF
--- a/lazyflow/operators/ioOperators/opTiffReader.py
+++ b/lazyflow/operators/ioOperators/opTiffReader.py
@@ -102,8 +102,8 @@ class OpTiffReader(Operator):
 
         blockshape = defaultdict(lambda: 1, zip(self._page_axes, self._page_shape))
         # optimization: reading bigger blockshapes in z means much smoother user experience
-        if "z" not in blockshape and "z" in axes:
-            blockshape["z"] = 32
+        # but don't change z if it's part of the page shape
+        blockshape.setdefault("z", 32)
         self.Output.meta.ideal_blockshape = tuple(blockshape[k] for k in axes)
 
     def execute(self, slot, subindex, roi, result):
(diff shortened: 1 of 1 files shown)


lazyflow/operators/ioOperators/opTiffReader.py
```diff
@@ -78,7 +99,12 @@ def setupOutputs(self):
         self.Output.meta.shape = shape
         self.Output.meta.axistags = vigra.defaultAxistags(axes)
         self.Output.meta.dtype = numpy.dtype(dtype_code).type
-        self.Output.meta.ideal_blockshape = (1,) * len(self._non_page_shape) + self._page_shape
+
+        blockshape = defaultdict(lambda: 1, zip(self._page_axes, self._page_shape))
+        # optimization: reading bigger blockshapes in z means much smoother user experience
+        if "z" not in blockshape and "z" in axes:
+            blockshape["z"] = 32
```
emilmelnikov, line 106:
```suggestion
        blockshape.setdefault("z", 32)
```

emilmelnikov (approved):

lazyflow/operators/ioOperators/opTiffReader.py
k-dominik: thanks! I didn't know about `setdefault`!

PR #2740, opened 2023-08-23
reduce number of deprecation warnings

k-dominik:
#### Summary:
* made sure to type cast explicitly in many instance
* fixed deprecation of non-raw special sequences in regular expressions (e.g. `"\d"` --> `r"\d"`) 
* fixed deprecation of importing abcs directly from `collections`, not from `collections.abc`
* fixed complaints about comparing with `is not` to a literal
* updated deprecated gh actions (checkout, cache) to v3

on my machine this reduces the number of warnings when running the tests from `9849` --> `1055` (together with https://github.com/ilastik/volumina/pull/303)

#### Details for the change of the colormap generation in counting:

the manual colormap resulted in a lot of warnings for implicit float to int conversions. Instead of simply making them ints, I implemented a good guess of how the original colormap was generated. The colormaps don't match perfectly, but for all purposes well enough I'd say.

<details>
<summary>"old" colormap</summary>

counting-00-old-colortable

</details>

<details>
<summary>"new" colormap</summary>

counting-01-jet-colortable-tanh

</details>

### DIFF
--- a/ilastik/applets/counting/countingGui.py
+++ b/ilastik/applets/counting/countingGui.py
@@ -18,7 +18,6 @@
 # on the ilastik web site at:
 # 		   http://ilastik.org/license.html
 ###############################################################################
-from functools import lru_cache
 import importlib
 import logging
 import os
@@ -45,6 +44,25 @@ from volumina.colortables import matplotlib_to_qt4_colortable
 
 logger = logging.getLogger(__name__)
 traceLogger = logging.getLogger("TRACE." + __name__)
+
+
+def _countingColorTable():
+    """
+    Create a colortable for the counting applet with gradual alpha increase
+    for the first third, then two thirds red, max alpha.
+    """
+    n_total = 256
+    n_gradual = 84
+    colortable = matplotlib_to_qt4_colortable("jet", N=n_gradual, asLong=False)
+    weights = numpy.tanh(numpy.linspace(0, 1.0, n_gradual))
+    for w, c in zip(weights, colortable):
+        c.setAlpha(w * 255)
+    colortable = [c.rgba() for c in colortable]
+
+    return [*colortable, *[colortable[-1]] * (n_total - n_gradual)]
+
+
+countingColorTable = _countingColorTable()
 
 
 def _listReplace(old, new):
@@ -651,7 +669,7 @@ class CountingGui(LabelingGui):
         for name, (slot, opacity) in list(slots.items()):
             if slot.ready():
                 layer = ColortableLayer(
-                    createDataSource(slot), colorTable=countingColorTable(), normalize=(0, self.upperBound)
+                    createDataSource(slot), colorTable=countingColorTable, normalize=(0, self.upperBound)
                 )
                 layer.name = name
                 layer.opacity = opacity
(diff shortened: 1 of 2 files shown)


btbest (approved):
I'm again running into problems with the file picker dialog on my machine. The dialog opens and then the tests hangs, my guess is because the dialog opened on my most recently used directory rather than whatever the test expects. This isn't reproducible though, in different runs it happens during different tests (I had it once during testEdgeTrainingWithMulticutGui.py and once during testPixelClassificationGui.py).

I will dig a bit and see if I can figure out what's going on.

ilastik/workflows/carving/carvingGui.py
```diff
@@ -654,7 +654,7 @@ def setupLayers(self):
 
         def onButtonsEnabled(slot, roi):
             currObj = self.topLevelOperatorView.CurrentObjectName.value
-            canSave = self.topLevelOperatorView.CanObjectBeSaved.value
+            canSave = bool(self.topLevelOperatorView.CanObjectBeSaved.value)
```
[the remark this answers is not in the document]
k-dominik: unfortunately the value is `numpy.bool_` and our qt version doesn't like it...

ilastik/applets/counting/countingGui.py
```diff
@@ -783,7 +784,7 @@ def handleLabelSelectionChange(self):
         if self.topLevelOperatorView.MaxLabelValue.ready():
             enabled = True
             enabled &= self.topLevelOperatorView.MaxLabelValue.value >= 2
-            enabled &= numpy.all(numpy.asarray(self.topLevelOperatorView.CachedFeatureImages.meta.shape) > 0)
+            enabled &= bool(numpy.all(numpy.asarray(self.topLevelOperatorView.CachedFeatureImages.meta.shape) > 0))
```
emilmelnikov, line 787: Maybe this is not the right PR to change this, but...
```python
value = self.topLevelOperatorView.MaxLabelValue.value
shape = self.topLevelOperatorView.CachedFeatureImages.meta.shape
enabled = value >= 2 and all(d > 0 for d in shape)
```

ilastik/applets/counting/countingGui.py
```diff
@@ -1102,521 +1103,19 @@ def updateSum(self, *args, **kw):
...
-    QColor(127.0, 0.0, 0.0, 255.0).rgba(),
-    QColor(127.0, 0.0, 0.0, 255.0).rgba(),
-    QColor(127.0, 0.0, 0.0, 255.0).rgba(),
-    QColor(127.0, 0.0, 0.0, 255.0).rgba(),
-    QColor(127.0, 0.0, 0.0, 255.0).rgba(),
-    QColor(127.0, 0.0, 0.0, 255.0).rgba(),
-    QColor(127.0, 0.0, 0.0, 255.0).rgba(),
-]
+@lru_cache(maxsize=1)
+def countingColorTable():
```
emilmelnikov, line 1107: Why keep this as a function instead of just computing it once and storing it as a module attribute? It shouldn't be that expensive to calculate...

ilastik/applets/counting/countingGui.py
```diff
@@ -1102,521 +1103,19 @@ def updateSum(self, *args, **kw):
...
+    """
+    Create a colortable for the counting applet with gradual alpha increase
+    for the first third, then two thirds red, max alpha.
+    """
+    n_max_alpha = 172
```
emilmelnikov, line 1112:
```python
n_total = 256
...
return [*colortable, colortable[-1] * (n_total - n_gradual)]

ilastik/applets/counting/countingGui.py
k-dominik: Ok I started with this parametrized and thought how cool it would be to have this more generic, and adapt the colors based on label colors and so on, but soon found that the changes proliferate quite a bit - so in the end I removed params and set the cache size to 1 :D. Not _really_ needed in that state, I agree.

ilastik/applets/counting/countingGui.py
k-dominik: ah, I'll add this in as it avoids the explicit cast to bool!

PR #2743, opened 2023-08-25
mnt: update gh actions

k-dominik:
checkout and cache are outdated. Bumped both to v3


k-dominik:
will add this to #2740

PR #2745, opened 2023-08-29
let pytest use pytest.ini to discover tests

k-dominik:
in the conda recipe we had practically overridden the paths for tests by specifying `.` as path. So in the gh actions it worked before as no source folders were present (e.g. collection errors out for me locally when the `bin` folder is checked for tests...)

Also add source folders for doctests.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -73,6 +73,8 @@ outputs:
         - lazyflow
 
       source_files:
+        - ilastik
+        - lazyflow
         - tests
         - pytest.ini
 
@@ -83,7 +85,7 @@ outputs:
         - tiktorch
 
       commands:
-        - pytest -v .
+        - pytest -v
 
     about:
       home: https://github.com/ilastik/ilastik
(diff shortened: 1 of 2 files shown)


btbest (approved):
Looks fine on my machine, though it seems we have a known problem test where the `skipif WIN and GH_ACTIONS` just needs to be `skipif WIN`...

image

k-dominik:
thanks for looking into this @btbest. Bummer about the test not working on windows at all anymore...

k-dominik:
A bit of a bummer, but I didn't find a (quick) way to fix the tests for windows. Did these not run before, or why does this surface with this PR?

btbest:
I see now that it actually passes when I run it within a full run, but fails when I run it individually. So we don't have to skip it in CI after all.

When I debug into the function during individual execution, the path and selection are fine, and if I click "Ok" manually, the test passes. In contrast, when I debug into the function from a full test run, `dialog.getSelectedPaths()` doesn't even open a dialog anymore at this point. My first assumption would have been that it must be working directory related, since that's the main difference between full run and individual execution that I'm aware of...

k-dominik:
oh wow...

Maybe that ImageFileDialog class should be refactored so that the functionality can be tested without testing QFileDialog functionality, which I think we can assume to work.

Issue #2714, opened 2023-05-23
object information export (with plugins) runs into hdf5 header size limit


k-dominik:
## Describe the bug

There is a [limit to the hdf5 header length for datasets (64kb)](https://davis.lbl.gov/Manuals/HDF5-1.8.7/UG/13_Attributes.html) (header itself is an attribute), which can be reached with the structured array types we're saving the object features in. I've read different accounts on how many columns one can save, but it depends on column names and dtypes.

## Error message/traceback

If applicable, please add tracebacks with formatting, such as

```pytb
ERROR 2023-05-23 12:05:18,292 log_exception 5138 12942405632 Export failed
ERROR 2023-05-23 12:05:18,293 log_exception 5138 12990263296 Traceback (most recent call last):
  File "/.../ilastik/ilastik/applets/dataExport/dataExportGui.py", line 429, in exportSync
    self.parentApplet.post_process_lane_export(lane_index)
  File "/.../ilastik/ilastik/applets/objectClassification/objectClassificationDataExportApplet.py", line 63, in post_process_lane_export
    req = self._tableExporter.export_object_data(lane_index, show_gui=False).wait()
  File "/.../ilastik/lazyflow/request/request.py", line 541, in wait
    return self._wait(timeout)
  File "/.../ilastik/lazyflow/request/request.py", line 564, in _wait
    self._wait_within_foreign_thread(timeout)
  File "/.../ilastik/lazyflow/request/request.py", line 624, in _wait_within_foreign_thread
    raise_with_traceback(exc_value, exc_tb)
  File "/.../mambaforge/envs/ilastik/lib/python3.7/site-packages/future/utils/__init__.py", line 446, in raise_with_traceback
    raise exc.with_traceback(traceback)
  File "/.../ilastik/lazyflow/request/request.py", line 360, in _execute
    self._result = self.fn()
  File "/.../ilastik/ilastik/applets/objectClassification/opObjectClassification.py", line 185, in do_export
    export_file.write_all(settings["file type"], settings["compression"])
  File "/.../ilastik/ilastik/utility/exportFile.py", line 457, in write_all
    compression if compression is not None else {},
  File "/.../ilastik/ilastik/utility/exportFile.py", line 493, in _make_h5_dataset
    dset = fout.create_dataset(table_name, sanitized_table.shape, data=sanitized_table, **compression)
[...]

PR #2746, opened 2023-08-31
First try most compatible H5 saving, otherwise use v1.8 

oanegros:
fix for #2714, only uses the v1.8 H5 if necessary

### DIFF
--- a/ilastik/utility/exportFile.py
+++ b/ilastik/utility/exportFile.py
@@ -447,17 +447,25 @@ class ExportFile(object):
         count = 0
         self.ExportProgress(0)
         if mode in ("h5", "hd5", "hdf5"):
-            with h5py.File(self.file_name, "w") as fout:
-                for table_name, table in self.table_dict.items():
-                    self._make_h5_dataset(
-                        fout,
-                        table_name,
-                        table,
-                        self.meta_dict.get(table_name, {}),
-                        compression if compression is not None else {},
-                    )
-                    count += 1
-                    self.ExportProgress(count * 100 / len(self.table_dict))
+            for libver in ["earliest", "v108"]:
+                try:
+                    with h5py.File(self.file_name, "w", libver=libver) as fout:
+                        for table_name, table in self.table_dict.items():
+                            self._make_h5_dataset(
+                                fout,
+                                table_name,
+                                table,
+                                self.meta_dict.get(table_name, {}),
+                                compression if compression is not None else {},
+                            )
+                            count += 1
+                            self.ExportProgress(count * 100 / len(self.table_dict))
+                except ValueError:
+                    logger.warning("Writing most compatible H5 failed, attempting HDF5 v1.8 format")
+                    count = 0
+                    self.ExportProgress(0)
+                    continue
+                break
         elif mode == "csv":
             f_name = self.file_name.rsplit(".", 1)
             if len(f_name) == 1:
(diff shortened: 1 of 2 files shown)


ilastik/utility/exportFile.py
```diff
@@ -447,17 +447,25 @@ def write_all(self, mode, compression=None):
...
-                    self._make_h5_dataset(
-                        fout,
-                        table_name,
-                        table,
-                        self.meta_dict.get(table_name, {}),
-                        compression if compression is not None else {},
-                    )
-                    count += 1
-                    self.ExportProgress(count * 100 / len(self.table_dict))
+            for libver in ["earliest", "v108"]:
```
k-dominik, line 450: in principle one could try all values in [`h5py._hl.files.libver_dict`](https://github.com/h5py/h5py/blob/8bed19bed35108190e598993d26bacb70879059c/h5py/_hl/files.py#L35) - do you know of any changes that occurred after `v108` that address the issue of object headers?

Otherwise I would prefer to check first if any of the tables can be written beforehand, but it seems not obvious to determine header size from a structured array.

ilastik/utility/exportFile.py
```diff
@@ -447,17 +447,25 @@ def write_all(self, mode, compression=None):
...
+                                fout,
+                                table_name,
+                                table,
+                                self.meta_dict.get(table_name, {}),
+                                compression if compression is not None else {},
+                            )
+                            count += 1
+                            self.ExportProgress(count * 100 / len(self.table_dict))
+                except ValueError:
+                    logger.warning("Writing most compatible H5 failed, attempting HDF5 v1.8 format")
```
k-dominik, line 464:
```suggestion
                except ValueError as e:
                    logger.warning(f"Writing most compatible H5 failed, attempting HDF5 v1.8 format. {e:r}")
```

might be useful to see the actual error here, too - just in case.

ilastik/utility/exportFile.py
```diff
@@ -447,17 +447,25 @@ def write_all(self, mode, compression=None):
...
+                    count = 0
+                    self.ExportProgress(0)
+                    continue
```
k-dominik, line 467: It would be nice to check here whether options of writing were exhausted - otherwise this routine will fail "quietly" with a warning and _not_ produce an output.

k-dominik (changes requested):
Hey @oanegros,

thx a lot for fixing the issue with feature tables with a large number of columns (~2000+). I left a few remarks that would be worth looking into.

Todo for myself: add a test.

oanegros:
Added 'latest' hdf5 version to still check if v108 fails to allow future improvements to h5py/HDF5 to allow more than the current v1.8 maximum header size, and added 'loud' failing. 

I don't check for table size before writing as I believe the size of the table header is quite complicated to calculate given datatype conversions.

ilastik/utility/exportFile.py
```diff
@@ -447,17 +447,28 @@ def write_all(self, mode, compression=None):
...
+                                table,
+                                self.meta_dict.get(table_name, {}),
+                                compression if compression is not None else {},
+                            )
+                            count += 1
+                            self.ExportProgress(count * 100 / len(self.table_dict))
+                except ValueError as e:
+                    if libver == "latest":
+                        logger.error("latest version of HDF5 export failed in export")
+                        raise ValueError(e)
```
k-dominik, line 466: this will generate a "During handling of the above exception, another exception occurred" message. Here it's better to do a bare `raise` (will automatically re-raise `e`).

PR #2747, opened 2023-08-31
Data input: Phrase axistags error message more user-friendly

btbest:
Inspired by https://forum.image.sc/t/error-loading-file-ilastik/81584, I was looking for error messages relating to axistags and image shape, as I recall seeing some during my own use that were quite ugly and not understandable.

I couldn't find any other ones though, so here's a suggestion for the one mentioned in the issue on the forum.

Old: 
Screenshot from image.sc

New (with a forced error, since I have no broken example file 🙃 ): 
Screenshot 2023-08-31 114309

### DIFF
--- a/ilastik/applets/dataSelection/opDataSelection.py
+++ b/ilastik/applets/dataSelection/opDataSelection.py
@@ -64,8 +64,15 @@ class CantSaveAsRelativePathsException(Exception):
 
 
 class UnsuitedAxistagsException(Exception):
-    def __init__(self, axistags, shape):
-        super().__init__(f"Axistags {axistags} don't fit data shape {shape}")
+    def __init__(self, axistags: AxisTags, shape):
+        if len(axistags) > len(shape):
+            problem = "Inconsistent metadata: The dataset reports dimensions for which it contains no data."
+        else:
+            problem = "Inconsistent metadata: The dataset has more dimensions than it reports."
+        super().__init__(
+            f"{problem}\nReported axes: {', '.join([tag.key for tag in axistags])}\n"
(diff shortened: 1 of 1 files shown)


ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -64,8 +64,15 @@ def __init__(self, file_path: str, base_dir: str):
...
-    def __init__(self, axistags, shape):
-        super().__init__(f"Axistags {axistags} don't fit data shape {shape}")
+    def __init__(self, axistags: AxisTags, shape):
+        if len(axistags) > len(shape):
+            problem = "Inconsistent metadata: The dataset reports dimensions for which it contains no data.\n"
+        else:
+            problem = "Inconsistent metadata: The dataset has more dimensions than it reports.\n"
+        super().__init__(
+            problem + f"""Reported axes: {", ".join([tag.key for tag in axistags])}\n"""
+            f"""Actual data dimensions: {", ".join([str(s) for s in shape])} (pixels/channels/timepoints)"""
```
k-dominik, line 74: not sure about the triple quote strings...

```suggestion
            f"{problem}Reported axes: {", ".join([tag.key for tag in axistags])}\n"
            f"Actual data dimensions: {", ".join([str(s) for s in shape])} (pixels/channels/timepoints)"
```

k-dominik (commented):
would you have a dataset that produces such an error?

Otherwise, better error messages, are definitely better error messages :)

ilastik/applets/dataSelection/opDataSelection.py
btbest: I had to convert the {}-enclosed quotes to single quotes for this one but yeah, nicer this way.

k-dominik (approved):
👍

PR #2750, opened 2023-09-11
Request object features per plugin in parallel

### LINKED ISSUE #2744: Rework ObjectFeaturesPlugin::compute_local
To support easy writing of stable and fast plugins for object feature computation, the bones exist in the ObjectFeaturesPlugin::compute_local method, but writing is hampered by some implementation details.

- By default this method gives the neighborhood along with the object, this should be another method or settable behaviour
- The compute_local loop is unparallelized and not integrated into the request architecture of ilastik multithreading
- compute_local does not get basic object information such as object_id and/or location, which make debugging difficult
- good example features that are not connected to vigra (which adds historical complexity) are not in the codebase

If these would be fixed, it opens up a very easy route to make any object quantifier directly attached to ilastik as a new one would generally only require duplicating the example and changing the name of the feature and the compute_local function, meaning that ilastik would be a viable place for any quantitative object extraction that you may want to pipe into a classifier.

k-dominik:
Use request system to allow for parallel computation of local object features.

Will speed up computation if gil is released in features `compute_local` somewhere.

fixes part of #2744 

Todos:

- [x] cleanup
- [x] benchmark our features (also with time-series)
   for our [drosophila example](https://data.ilastik.org/drosophila.zip) I found a 2-3 times speedup in feature computation (all excl location) on my M1 without any numbers changing.

### DIFF
--- a/ilastik/applets/objectExtraction/opObjectExtraction.py
+++ b/ilastik/applets/objectExtraction/opObjectExtraction.py
@@ -698,54 +698,40 @@ class OpRegionFeatures(Operator):
         maxcoords = extrafeats["Coord<Maximum>"].astype(int)
         nobj = mincoords.shape[0]
 
-        # local features: loop over all objects
-        def dictextend(a, b):
-            for key in b:
-                a[key].append(b[key])
-            return a
-
         local_features = collections.defaultdict(lambda: collections.defaultdict(list))
         margin = max_margin(feature_names)
-        has_local_features = {}
-        for plugin_name, feature_dict in feature_names.items():
-            has_local_features[plugin_name] = False
-            for features in feature_dict.values():
-                if "margin" in features:
-                    has_local_features[plugin_name] = True
-                    break
 
-        if numpy.any(margin) > 0:
+        if numpy.any(margin):
             bboxes = {}
             for plugin_name, feature_dict in feature_names.items():
-                if not has_local_features[plugin_name]:
+                if not any("margin" in features for features in feature_dict.values()):
                     continue
 
                 plugin = pluginManager.getPluginByName(plugin_name, "ObjectFeatures")
-                pool = RequestPool()
                 tmp_dicts = [None] * nobj
 
                 def _calc_single(i, raw_bbox, binary_bbox):
                     feats = plugin.plugin_object.compute_local(raw_bbox, binary_bbox, feature_dict, axes)
                     tmp_dicts[i] = feats
 
-                # starting from 0, we stripped 0th background object in global computation
-                for i in range(0, nobj):
-                    logger.debug("processing object {}".format(i))
-                    if i not in bboxes:
-                        extent = self.compute_extent(i, image, mincoords, maxcoords, axes, margin)
-                        raw_bbox = self.compute_rawbbox(image, extent, axes)
-                        # it's i+1 here, because the background has label 0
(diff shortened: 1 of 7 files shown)


ilastik/applets/objectExtraction/opObjectExtraction.py
```diff
@@ -708,19 +715,37 @@ def dictextend(a, b):
...
+                        bboxes[i] = (raw_bbox, binary_bbox)
+
+                    raw_bbox, binary_bbox = bboxes[i]
+                    pool.add(Request(partial(_calc_single, i, raw_bbox, binary_bbox)))
+
+                pool.wait()
+                pool.clean()
+                # merge the results
+                for i in range(0, nobj):
+                    local_features[plugin_name] = dictextend(local_features[plugin_name], tmp_dicts[i])
```
emilmelnikov, line 748:
```suggestion
                for feature_dict in tmp_dicts:
                    for feature_name, features in feature_dict.items():
                        local_features[plugin_name][feature_name].append(features)
```

...and there is no need for `dictextend`.

emilmelnikov (commented):
It's out of the scope of this PR, but this method could use some refactoring.

ilastik/applets/objectExtraction/opObjectExtraction.py
k-dominik: Hm, yeah, you could inline it. I'm not sure it's better for readability here. Do you think so?

emilmelnikov: It is used only in one place, the name doesn't explain what it actually does (at least to me), and the double loop has `dict1[key1][key2].append(value)` statement that shows how the main data structure is updated.

ilastik/applets/objectExtraction/opObjectExtraction.py
```diff
@@ -708,19 +715,37 @@ def dictextend(a, b):
...
+                        extent = self.compute_extent(i, image, mincoords, maxcoords, axes, margin)
+                        raw_bbox = self.compute_rawbbox(image, extent, axes)
+                        # it's i+1 here, because the background has label 0
+                        binary_bbox = numpy.where(labels[tuple(extent)] == i + 1, 1, 0).astype(bool)
+                        bboxes[i] = (raw_bbox, binary_bbox)
+
+                    raw_bbox, binary_bbox = bboxes[i]
+                    pool.add(Request(partial(_calc_single, i, raw_bbox, binary_bbox)))
+
+                pool.wait()
```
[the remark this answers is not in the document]
emilmelnikov: The following approach could also work if there is no need to cancel (the name could be better though):

```python
def run_in_pool(funcs: Sequence[Callable]) -> None:
    pool = RequestPool()
    for func in funcs:
        pool.add(Request(func))
    pool.wait()


funcs = []
for something in sequence:
    ...
    funcs.append(partial(func, *args))
    ...

run_in_pool(funcs)
```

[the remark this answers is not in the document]
k-dominik: also a nice pattern, but I kinda like using a contextmanager. It is a bit closer to concurrent.futures...

k-dominik:
merging this, there were no major changes after the review, mostly CI gymnastics...

PR #2752, opened 2023-09-20
experimental API refactor

k-dominik:
_partly_ addresses #2515. In this PR I added Object Classification and the discussion uncovered some weak points of the design. In order to proceed with the API, e.g. adding more workflows, I addressed these in this PR.

Summary:

* reworked pipelines: (along with deprecating the "generic" `from_project_file` function), exposed the `PixelClassificationPipeline` class with a static `from_ilp_file` method for object creation.
* reworked parsing, which is now done eagerly (project file will not be kept open). Also moved parsing logic to the respective types (`from_ilp_group`), out of the Project class (will enable reusability of types, e.g. for Autocontext).
* generic `predict` method will not hold for future workflows, even pixel classification can have multiple outputs.
   I propose to go on with `get_OUTPUT` methods, where `OUTPUT` would be the name of the export source, analogous to the headless mode.
* array conversion (to vigraarray, used internally) without singledispatch in a single function
* updated the notebook to reflect these changes
* existing projects using the api will still work, no code changes required for existing downstream users. They'll get an annoying deprecation warning though.

## Checklist

- [x] Add tests.


ilastik/experimental/parser/types.py
```diff
@@ -23,30 +34,87 @@ class FeatureMatrix:
     selections: numpy.ndarray
     compute_in_2d: numpy.ndarray
 
+    @staticmethod
```
emilmelnikov, line 37: Since it returns an instance of the enclosing class, this could be replaced with `@classmethod`.

ilastik/experimental/parser/_base.py
```diff
@@ -1,137 +1,41 @@
...
-    FEATURES_SCALES = "Scales"
-    FEATURES_COMPUTE_IN_2D = "ComputeIn2d"
-    FEATURES_SELECTION_MATRIX = "SelectionMatrix"
-    PIXEL_CLASSIFICATION = "PixelClassification"
-    PIXEL_CLASSIFICATION_TYPE = "pickled_type"
-    PIXEL_CLASSIFICATION_FACTORY = "ClassifierFactory"
-    PIXEL_CLASSIFICATION_FORESTS = "ClassifierForests"
-    LABEL_NAMES = "LabelNames"
+    def __init__(self, hdf5_file: h5py.File) -> None:
+        class ILPkeys(types.StrEnum):
```
emilmelnikov, line 12: What's the reason for using enum classes instead of local variables or class attributes?

ilastik/experimental/parser/_base.py
```diff
@@ -1,137 +1,41 @@
...
-    FEATURES_IDS = "FeatureIds"
-    FEATURES_SCALES = "Scales"
-    FEATURES_COMPUTE_IN_2D = "ComputeIn2d"
-    FEATURES_SELECTION_MATRIX = "SelectionMatrix"
-    PIXEL_CLASSIFICATION = "PixelClassification"
-    PIXEL_CLASSIFICATION_TYPE = "pickled_type"
-    PIXEL_CLASSIFICATION_FACTORY = "ClassifierFactory"
-    PIXEL_CLASSIFICATION_FORESTS = "ClassifierForests"
-    LABEL_NAMES = "LabelNames"
+    def __init__(self, hdf5_file: h5py.File) -> None:
```
emilmelnikov, line 11: If this is supposed to represent data from a project file, I'd vote for converting this into a dataclass with constructor taking `data_info`, `feature_matrix`, `classifier` etc. and adding classmethod or external function for parsing it from HDF5 file.

ilastik/experimental/parser/types.py
```diff
@@ -23,30 +34,87 @@ class FeatureMatrix:
...
+        for size, dim in zip(shape, tags_dict["axes"]):
+            if dim["key"] in "xyz":
+                spatial_axes += dim["key"]
+            elif dim["key"] == "c":
+                num_channels = size
 
-    @abc.abstractproperty
-    def classifier(self) -> Optional[Classifier]:
-        pass
+            axis_order += dim["key"]
```
emilmelnikov, line 118:
```python
axis_order = "".join(dim["key"] for dim in tags_dict["axes"])
spatial_axes = "".join(axis for axis in axis_order if axis in "xyz")
try:
    num_channels = shape[axis_order.index("c")]
except ValueError:
    num_channels = 1
```

ilastik/experimental/api/_pipelines.py
```diff
@@ -0,0 +1,97 @@
...
+
+    return PixelClassificationPipeline.from_project_file(path)
+
+
+def as_vigra_array(data: Union[vigra.VigraArray, xarray.DataArray]) -> vigra.VigraArray:
+    if isinstance(data, vigra.VigraArray):
+        return data
+    elif isinstance(data, xarray.DataArray):
+        axistags = "".join(data.dims)
+        return vigra.taggedView(data.values, axistags)
```
emilmelnikov, line 93: It seems vigra also accepts sequences for axistags:
```suggestion
        return vigra.taggedView(data.values, data.dims)
```

ilastik/experimental/api/_pipelines.py
```diff
@@ -0,0 +1,97 @@
...
+        DeprecationWarning,
+    )
+
+    return PixelClassificationPipeline.from_project_file(path)
+
+
+def as_vigra_array(data: Union[vigra.VigraArray, xarray.DataArray]) -> vigra.VigraArray:
+    if isinstance(data, vigra.VigraArray):
+        return data
+    elif isinstance(data, xarray.DataArray):
```
emilmelnikov, line 91: It's a little bit of personal preference, but...

```python
if some_condition:
    return case_1
# No need for elif: previous condition already returned.
if other_condition:
    return case_2
# No nesting in the last "default" case.
raise SomethingWrongHappened
```

Feel free to ignore this!

ilastik/experimental/parser/_base.py
k-dominik: convenience... with this one can 

```Python
if any(key not in hdf5_file for key in ILPkeys):
...
```

ilastik/experimental/parser/types.py
k-dominik: looks good, for spacial axes i'll opt for

```Python
spatial_axes = "".join(ax for ax in axis_order if ax in "xyz")
```
for subjectively better readability.

ilastik/experimental/parser/_base.py
emilmelnikov: In this case I think it's better to avoid `Enum` because values like `ILPKeys.WORKFLOW` have types that are different from their values (e.g. `ILPKeys.WORKFLOW` is not `str`), which could potentially break something downstream.

One could cook up a simple metaclass that does approximately the same but preserves actual types (really this just adds convenient iteration, nothing more):

```python
class BagMeta(type):
    def __iter__(cls):
        # Ignore built-in dunders
        return iter(v for k, v in cls.__dict__.items() if not k.startswith("_"))


# Just to make inheritance more convenient.
class bag(metaclass=BagMeta):
    pass


class values(bag):
    spam = "Lorem Ipsum"
    eggs = 42
    # Bonus feature: hide something from iteration
    _ham = "hidden!"


print(list(values))
# ['Lorem Ipsum', 42]
```

emilmelnikov: Note that it's possible to get the actual value from enum member with `ILPKeys.WORKFLOW.value`, but it's an additional attribute access.

k-dominik: hmm, I don't know, with `StrEnum` you get `isinstance(`ILPKeys.WORKFLOW`, "str") is True`...

k-dominik: from what I read in the docs, it's pretty much intended use: https://docs.python.org/3.9/library/enum.html#others

emilmelnikov: I think this is the intended use if you use enum members as unique symbols and maybe sometimes access their values via `MyEnum.MY_MEMBER.value`, but enum members and their values are still different:

```python
import enum

class StrEnum(str, enum.Enum):
   pass

class my(StrEnum):
   a = 'value of a'
   b = 'value of b'

print(list(my))
# [<my.a: 'value of a'>, <my.b: 'value of b'>]
```

But I'm not sure to what extent these differences matter in practice...

btbest:
I love every bit of it! I would only suggest to rename `_base.py` into `_projects.py` (what is it a base for?). I'm also unable to guess what the prefix underscores are supposed to express, as they don't seem to denote "module-internalness"; so unless they trigger some importer magic I would remove them.

k-dominik:
> I love every bit of it! I would only suggest to rename `_base.py` into `_projects.py` (what is it a base for?). I'm also unable to guess what the prefix underscores are supposed to express, as they don't seem to denote "module-internalness"; so unless they trigger some importer magic I would remove them.

well 😅 the underscores represent a bit the style of the previous author of the API (@m-novikov) for signalling that something is not part of the API. I kind of like it here and would want to keep this. I agree on the rename though. The naming with `_base` made more sense before...

btbest:
> signalling that something is not part of the API

Shouldn't `_pipelines.py` be without underscore in that case, since it contains the main export of the module?

btbest (approved):

ilastik/experimental/parser/_base.py
k-dominik: gonna stick with enums, they're only used in `from_project_file`, their behavior is quite predictable there, without the need for any custom meta classes...

k-dominik:
> > signalling that something is not part of the API
> 
> Shouldn't `_pipelines.py` be without underscore in that case, since it contains the main export of the module?

I renamed `_base.py` in  in parser to `_projects.py` - to have a consistent layout. Exports meant for user consumption are only exposed via `__init__.py`, that way it's a) clear how to use the api (importing `experimental.api` vs `experimental.api.pipelines`, and we have more flexibility inside these modules. Not a big deal really.

PR #2756, opened 2023-10-12
OpDataSelection: Ensure all children get cleaned up

### LINKED ISSUE #2755: File reader op does not get garbage collected when data-selection op's input changes
## Describe the bug
File loading involves an OpDataSelection instantiating (indirectly) an OpInputDataReader. This happens in `OpDataSelection.setupOutputs()`, so every time an input of OpDataSelection changes, a new OpInputDataReader is instantiated.

This occurs when the user right-clicks a file row and chooses "Replace with file". In this case, the OpDataSelection's `Dataset` input changes and re-triggers setupOutputs.
image

Afterwards, the original file's OpInputDataReader still persists in the graph, unconnected (see screenshot below)

Only an issue for wasted RAM, depending on how much data the reader had cached already. But not nice.

Also note that using the minus button to completely remove the file also fully cleans up the ops. This is specific to input changes like when clicking "Replace with file".

## Expected behavior
The loose op should be deleted.

## To Reproduce
2. Start any workflow
3. Add a file
4. Right-click and "Replace with file"
5. Choose any other file
6. Repeat 4-5 to your heart's content
7. Debug -> Export operator diagram -> Unlimited

## Screenshots
image

 - Based on commit history, all versions since 2014

btbest:
This implements option 2 to fix #2755 

Still todo (regardless of which option we go for):
- [] Add test to ensure all children are deleted


k-dominik (approved):
looks good.

Not sure how much you've looked into this file now, but on the off chance that you looked into this in detail, do you think you could also document the methods/parameters a bit. I tried to read this class now and I have to do some guessing here without looking into the code. E.g. what are `axistags` vs. `default_tags` in `DatasetInfo.__init__`...

Tomaz-Vieira:
> E.g. what are `axistags` vs. `default_tags` in `DatasetInfo.__init__`...

If I recall correctly:

`default_tags` are the axis tags that would be applied to this dataset if the user did not specify any other reinterpretation of the axis. Those can be derived form the file type, or from the file's metadata, or from  ilastik's best guess as to what the meaning of the axis would be.

The `default_tags` can be overriden by `axistags`, which are provided by the user by manipulating the fields in `datasetInfoEditorWidget.py` so that the axis are interpreted differently to what the `default_tags` would determine

btbest:
I will rename the axistags variables to make that a bit clearer.

Also, my favoured solution to the problem is this one: https://github.com/ilastik/ilastik/pull/2757. Though now I'm thinking of just doing both, because the changes in this PR remove some unnecessary complexity.

Issue #2184, opened 2020-02-27
Investigate timing test failure on circleci


k-dominik:
Did not menage to reproduce this anywhere else, but currently, vigra seems to be faster than fastfilters :o

See: https://circleci.com/gh/ilastik/ilastik/2102

once this resolved, https://github.com/ilastik/ilastik/pull/2191/commits/91e7ff823dfd4af343a80f1c18afae587208652c should be reverted

Tomaz-Vieira:
The fastfilters repo also has some performance tests comparing itself to vigra. Maybe we can use those for some benchmarking too

k-dominik:
closed via https://github.com/ilastik/ilastik/pull/2757

PR #2757, opened 2023-10-13
OpDataSelection: Directly obtain data-reader

### LINKED ISSUE #2755: File reader op does not get garbage collected when data-selection op's input changes

btbest:
This implements option 1 to fix #2755, and it's my preferred option as it cleans up the graph just a tiny bit.

Main concern is that nobody seems to know what was the reason for using OpMetadataInjector in the first place. But if tests pass we should be fine right? :smile:

Edit: I've added option 2 as described in the issue since they work quite well together.

Closes #2755 

## Todo (regardless of which option we go for)
- [x] Add test to ensure children are deleted


k-dominik:
now thinking of it I remember tales that there were plans once to let users add metadata in dataselection. Maybe it was added in preparation. Subtle difference of using the op vs `assignFrom` is that the metadataop can be update (and it would trigger dirtyness immediately), whereas `assignFrom` doesn't.

btbest:
Yeah, renaming those parameters isn't going to fly I think, it's breaking project loading. I guess somewhere during load, the hdf5 fields get fed into constructors in bulk while assuming `hdf-field-name == constructor-parameter-name`. I don't think I will bother to try to find out where exactly, since I don't think I'll get around renaming the params being a compatibility-breaking change.

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -715,105 +717,93 @@ def __init__(
...
+                )
+                raise DatasetConstraintError("DataSelection", msg)
+
+            output_order = sorted(candidate_orders, key=len)[0]  # the shortest one
+            output_order = "".join(output_order)
+        else:
+            # No forced axisorder is supplied. Use original axisorder as
+            # output order: it is assumed by the export-applet, that the
+            # an OpReorderAxes operator is added in the beginning
+            output_order = data_provider.meta.getAxisKeys()
```
k-dominik, line 798:
```suggestion
            output_order = "".join(data_provider.meta.getAxisKeys())
```

`getAxisKeys` returns `List[str]`

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -63,14 +62,28 @@ def __init__(self, file_path: str, base_dir: str):
...
+
+
 class UnsuitedAxistagsException(Exception):
     def __init__(self, axistags: AxisTags, shape):
         if len(axistags) > len(shape):
-            problem = "Inconsistent metadata: The dataset reports dimensions for which it contains no data."
+            problem = "No data for some axes."
         else:
-            problem = "Inconsistent metadata: The dataset has more dimensions than it reports."
+            problem = "Some data dimensions have no axis interpretation."
```
k-dominik, line 83: I know you didn't really change the logic here, but I would expect a single channel time series with all axistags to work in a project that was trained on a single channel xy image...

In dataselection those two datasets would coexist happily.


image

k-dominik (commented):
Always good to see less code being added than removed :) More specific Exceptions, earlier failing and better comments are a plus :)

I had some minor comments...

k-dominik:
note: also fixes https://github.com/ilastik/ilastik/issues/2184

ilastik/applets/dataSelection/opDataSelection.py
btbest: I guess this was never a problem because the downstream OpReorderAxes does the same join again in setupOutputs... but yeah this should be consistent between the different paths.

btbest: Entertainingly:
```
a = ["x", "y"]
a += "c" # This is fine
a = a + "c" # TypeError
```

ilastik/applets/dataSelection/opDataSelection.py
btbest: You mean they should coexist happily but currently don't..? I'd say make an issue for it...

btbest:
Used the fixtures as suggested, and also fixed a tiny change in logic that I accidentally introduced: Previously, user-provided `axistags` were able to override potentially erroneous `default_tags` because the dimensions were only checked after deciding whether to use `axistags` or `default_tags`.

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -715,105 +717,93 @@ def __init__(
...
+            candidate_orders = [order for order in candidate_orders if minimal_axes.issubset(set(order))]
+
+            if len(candidate_orders) == 0:
+                msg = (
+                    f"The axes of your dataset ({data_provider.meta.getAxisKeys()}) are not compatible with "
+                    f"any of the allowed axis configurations used by this workflow ({self.forceAxisOrder})."
+                )
+                raise DatasetConstraintError("DataSelection", msg)
+
+            output_order = sorted(candidate_orders, key=len)[0]  # the shortest one
```
emilmelnikov, line 792:
```python
# Any forced axis order must include all non-singleton axis from the original dataset.
required_axes = {k for k, v in data_provider.meta.getTaggedShape().items() if v > 1}
# Pick the shortest forced axis order that includes all non-singleton axes.
output_order = min(filter(required_axes.issuperset, self.forceAxisOrder), default=(), key=len)
if not output_order:
    raise DatasetConstraintError(...)
```

ilastik/applets/dataSelection/opDataSelection.py
k-dominik: k, let's open an issue, it's unintuitive that it can be added as input data, but not in batch.

btbest:
Force-push to (a) fix comparing a set to a string and (b) revert to the original list comprehension style because with the set-conversion it's just easier to read.

PR #2762, opened 2023-10-26
Remove obsolete OpTransposeSlots

btbest:
This removes OpTransposeSlots and all references to it.

The operator isn't used anywhere, so we could get rid of it. But besides a tiny contribution to bloat, its presence probably also doesn't harm anyone...


tests/test_lazyflow/test_graph/testMultislotResize.py
```diff
@@ -46,30 +42,6 @@ def propagateDirty(self, slot, subindex, roi):
...
-        opB.Inputs.connect(opA.Inputs)
-        opC.Inputs.connect(opB.Inputs)
-
-        opT = OpTransposeSlots(graph=graph)
-        opT.Inputs.connect(opA.Outputs)
-        opT.OutputLength.setValue(2)
-
-        assert len(opT.Outputs) == 2
-
-        opA.Inputs.resize(3)
```
Tomaz-Vieira, line 66: It kinda looks like this part of the test doesn't really need `OpTransposedSlots` to do its thing (looks like it could just use `OpA`). The name of the test, "`testMultiSlotResize`" seems to only be about checking that resizing multislots works, so maybe we don't need to nuke the entire test.

btbest: I didn't check in detail, but the test below also covers resizing. I deleted this one because it only asserts on the transposer, so I'd have to understand what value this test adds to decide what it needs to assert after removing that...

k-dominik: To me it looks fine to delete.

Issue #1830, opened 2018-09-12
enhancement: command line: ditch positional arguments


k-dominik:
since we have discovered https://github.com/ilastik/ilastik/issues/1422 , and waited for it to be fixed in cpython (which didn't happen until now) I am wondering whether we should adhere to

> Special cases aren't special enough to break the rules.

and get rid of the positional arguments that are _only_ applicable in workflows, that don't take a second type of image (so that leaves us with PixelClassification).
And of course one could argue that options are for optional things. However, we are already have _options_ like `project` that is not optional, so in this case i'd argue to continue with

> Although practicality beats purity.

and disable positional arguments

PR #2763, opened 2023-11-03
better "debug" for error in command line usage

k-dominik:
tiny bandaid...

it's sadly common to get not very helpful error messages when using ilastik from the command line.
For me this often comes back to this error message complaining about trailing file names.
It is not obvious what portion of the cmd is treated that way so now we at least give a hint on what is interpreted incorrectly.

xref: https://github.com/ilastik/ilastik/issues/1830


btbest (approved):
Better error messages 👍

PR #2767, opened 2023-11-06
try larger resource class for tests on circleci

### LINKED ISSUE #2766: investigate circleci test hangs
tests seem to hang on circleci without any output at an increasing rate. Maybe it's good enough to try a larger resource class.

k-dominik:
I'm honestly not sure if this fixes #2766, but it didn't hang in this PR ;). Also makes probably sense to go to larger resource as we are pretty much maxing out the ram usage on the medium ones.

For reference the memory graph for the large one with 8Gb of Ram (medium has 4Gb):

image

for the checkout part, I kept the medium resource - cause why not :)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,7 @@
 version: 2.1
 jobs:
   checkout_ilastik:
+    resource_class: medium
     working_directory: /root/ilastik/ilastik
     parallelism: 1
     shell: /bin/bash --login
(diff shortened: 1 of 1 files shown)


btbest (approved):
I don't see a reason why not, or are we worried about running into pricing issues?

k-dominik:
we have a lot of free credits in the open source plan - I don't think that's a danger right now :)

PR #2774, opened 2023-11-14
OpDataSelection: Default file exports to project file path

btbest:
This changes the default export location to the directory containing the project file, unless otherwise specified.

As far as I can tell, this affects (i.e. changes the default to project dir instead of user home dir):
- datasets sourced over URL
- potentially some legacy projects (those that end up deserialising into the fallback path that uses `PreloadedArrayDatasetInfo`, though I didn't investigate how a project file ends up there)

For all other dataset types (e.g. directory-is-dataset like for tif series, datasets stored project-internally, and of course datasets from a single file), the behaviour should not change.

I added a couple of assert messages in metaDict.py because they tripped me up in the GUI and without any message or stack trace, it took ages to figure out where I was running into issues.

### DIFF
--- a/ilastik/applets/dataSelection/opDataSelection.py
+++ b/ilastik/applets/dataSelection/opDataSelection.py
@@ -213,7 +213,7 @@ class DatasetInfo(ABC):
     @property
     def default_output_dir(self) -> Path:
         if self.project_file:
-            return Path(self.project_file.filename).parent
+            return Path(self.project_file.filename).absolute().parent
         return Path.home()
 
     @classmethod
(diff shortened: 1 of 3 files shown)


ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -204,6 +212,8 @@ def display_string(self) -> str:
 
     @property
     def default_output_dir(self) -> Path:
+        if self.project_file:
+            return Path(self.project_file.filename).parent
```
k-dominik, line 216: maybe ensure an absolute path here (or even better when saving the attribute in the class)

k-dominik (approved):
Looks good :), tests run fine and I also couldn't see how this could break something else!

ilastik/applets/dataSelection/opDataSelection.py
btbest: self.project_file keeps the actual h5py.File handle. I don't think we want to mess with that, so I left it at the readout point

btbest: If you're happy with the `.absolute()` solution, I would go ahead and merge 🚀

k-dominik: sure!

btbest:
Small note here: When right-clicking an individual layer in the bottom-left list, and then selecting "Export", this seems to be a different export path than the proper export applet. I think this second path might not be using `DatasetInfo.default_export_dir`, because iirc I saw it still defaulting to my user home dir. Work for another PR...

PR #2775, opened 2023-11-17
bump numpy 1.22

k-dominik:
I wanted to bump vigra, too, with this PR, but currently it is not as straight forward as I was hoping: #2776


btbest (approved):
Resolves on Windows and seems to run fine :) Sweet, sweet updates.

PR #2784, opened 2023-12-14
Object size filter: Increase permitted value range

### LINKED ISSUE #2778: unbounded/less bounded size filters for Object Classification from Pixel Prediction Map
Currently the size filters in the Object Classification from Pixel Prediction Map pipeline have bounds, where they could be less bounded, for example the minimum size maxes out at 100.000; shouldnt the only bounds these have be min < max && min >= 0 ?
Screenshot 2023-11-29 at 11 03 56

btbest:
See #2778 

The maximum on QSPinBox also determines its width, so increasing the max also makes the layout wider.

I arranged the Min and Max spin boxes vertically because this applet drawer is already too wide:

image 

In terms of computation, there are no issues with allowing arbitrary values - if the user chooses nonsensical min and max (e.g. with min > max), there are simply no objects that match the filter. The user can also do this in the current version already.

Fixes #2778

### DIFF
--- a/ilastik/applets/thresholdTwoLevels/drawer.ui
+++ b/ilastik/applets/thresholdTwoLevels/drawer.ui
@@ -368,63 +368,71 @@
       </spacer>
      </item>
      <item>
-      <widget class="QLabel" name="label_9">
-       <property name="text">
-        <string>Min:</string>
+      <layout class="QVBoxLayout" name="verticalLayout_2">
+       <property name="spacing">
+        <number>5</number>
        </property>
-      </widget>
-     </item>
-     <item>
-      <widget class="QSpinBox" name="minSizeSpinBox">
-       <property name="sizePolicy">
-        <sizepolicy hsizetype="Preferred" vsizetype="Fixed">
-         <horstretch>0</horstretch>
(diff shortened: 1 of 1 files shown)


k-dominik (approved):
Hey that's cool. Now that I see it in vertical arrangement - would it maybe make more sense/be more intuitive to have maximum above the minimum?

Minor point: On macOS, at least, these spin boxes are less helpful than I'd wish. Tooltip doesn't indicate what the maximum value could be. I'm not sure if that is the case on other platforms as well. I know it's a bit of extra work, but it would be nice, if this information would be added to the tooltips.

btbest:
I find Min above Max more intuitive, just as previously it was Min left of Max...

There is also no tooltip about the maximum in Windows, this is only telegraphed by the number resetting to the max if you try to enter a value above the max. Would you prefer if we add "Cannot be higher than the max value of a 32-bit integer (2147483647)." to the tooltip?

k-dominik:
I think it would be more user friendly to mention the maximum value in the tooltip.

btbest:
I force-pushed to remove an excess space that found its way in the tooltip 😶 Ready to merge once the pipeline is done.

Issue #2728, opened 2023-07-21
Changing label drawing color preserved during active sessing between projects


k-dominik:
## Describe the bug
When changing the label _drawing_ color in project `A`, closing the project, and creating a new project `B`, the label corresponding to the changed index in project `A` will have the same color in project `B`.

## Expected behavior
Project color changes should not be preserved between projects

## To Reproduce
1) Create a pixel classification project, change one of the label _drawing_ colors.
2) Close the project
3) Create a new project and go to training -> label color (for _drawing and probability_ display) is changed there, too

## Desktop (please complete the following information):
 - ilastik version: `1.4.0`, `1.4.1b6`
 - OS: all

btbest:
Even better: I reproduced this by selecting a new "Color for drawing", but left the "Color for probability display" untouched. In the new project, it took over my custom "Color for drawing" also as the "Color for probability display".

PR #2786, opened 2023-12-15
Labeling GUI: Copy colortable

btbest:
Not copying the list means that when a value is changed (in this case, by the user selecting another color in the ColorDialog), the original list in the colortables module import is modified. This persists until ilastik is restarted.

Fixes https://github.com/ilastik/ilastik/issues/2728

### DIFF
--- a/ilastik/applets/labeling/labelingGui.py
+++ b/ilastik/applets/labeling/labelingGui.py
@@ -160,9 +160,6 @@ class LabelingGui(LayerViewerGui):
         :param rawInputSlot: Data from the rawInputSlot parameter will be displayed directly underneath the elements
                              (if provided).
         """
-
-        self._colorTable16 = list(colortables.default16_new)
-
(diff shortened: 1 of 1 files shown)


k-dominik (commented):
I guess that is the shortest route for a fix one could take!

I would honestly prefer to make all the mutable colormap constants in [`volumina.colortables`](https://github.com/ilastik/volumina/blob/94ff01656736f39cd2417bf3fcce01651759375c/volumina/colortables.py#L98) private and do the copying in some `create_COLORTABLENAME` functions, just to make it harder to run into such an issue again. What do you think @btbest

btbest:
I agree with the volumina change. In the meantime, I found the fix here is even stupider than I realised. There's already `self._colorTable16 = list(colortables.default16_new)` literally 20 lines above, which was overwritten by the line that I'm now "fixing" in this PR 🤦

PR #2793, opened 2024-01-02
Prettify dataset table

btbest:
* Format internal paths for display
* Show tooltips
* Combine AxisOrder and Shape columns + default column width

Before:
image

After:
image


k-dominik (approved):
Changes look nice! Some things you have to experience in order to miss them - for me at least that is the case with the tooltip for the location...
One minor comment concerning naming: giving the new combined column a more descriptive name like `TaggedShape` would improve readability.

btbest:
> giving the new combined column a more descriptive name like `TaggedShape` would improve readability.

Like this? I still left the table header called "Shape" since "tagged shape" isn't really common terminology among users.

PR #2795, opened 2024-01-03
Import legacy urldatasetinfo

btbest:
* Change the "Import Project" flow to allow serializers to modify the imported project file in line with the current ilastik version.
* Implement the functionality for the `DataSelectionSerializer`, upgrading legacy hbp-projects to be compatible with ilastik > 1.4.0

## Checklist

- [x] Add tests.
- [-] Introduce new h5 utils for exists-and-equals, get+decode, replace? (skipped)

### DIFF
--- a/ilastik/applets/base/appletSerializer.py
+++ b/ilastik/applets/base/appletSerializer.py
@@ -1197,6 +1197,13 @@ class AppletSerializer(with_metaclass(ABCMeta, object)):
         if they store relative paths."""
         pass
 
+    def updateLegacyProjectFile(self, file: h5py.File):
+        """Optional override for subclasses. Called during "Import Project".
+        Should modify the file in-place to make it compatible with the current version of the applet.
+        :param file: h5py.File to be modified - a copy of the imported project file.
+        """
+        pass
+
 
 class JSONSerialSlot(SerialSlot):
     """
--- a/ilastik/applets/base/appletSerializer.py
+++ b/ilastik/applets/base/appletSerializer.py
@@ -1197,7 +1197,7 @@ class AppletSerializer(with_metaclass(ABCMeta, object)):
         if they store relative paths."""
         pass
 
-    def updateLegacyProjectFile(self, file: h5py.File):
+    def updateLegacyEntries(self, file: h5py.File):
         """Optional override for subclasses. Called during "Import Project".
         Should modify the file in-place to make it compatible with the current version of the applet.
         :param file: h5py.File to be modified - a copy of the imported project file.
--- a/ilastik/applets/dataSelection/dataSelectionSerializer.py
+++ b/ilastik/applets/dataSelection/dataSelectionSerializer.py
@@ -319,7 +319,7 @@ class DataSelectionSerializer(AppletSerializer):
         self.topLevelOperator.WorkingDirectory.setValue(newdir)
         self._projectFilePath = newdir
 
-    def updateLegacyProjectFile(self, file: File):
+    def updateLegacyEntries(self, file: File):
         """
         Presence of UrlDatasetInfo indicates this is a pre-multiscale project file created in HBP mode.
         These projects only supported working on the first scale in the list of resolutions
--- a/ilastik/applets/dataSelection/dataSelectionSerializer.py
+++ b/ilastik/applets/dataSelection/dataSelectionSerializer.py
@@ -332,19 +332,18 @@ class DataSelectionSerializer(AppletSerializer):
         except KeyError:
             return
 
-        for laneIndex, (_, laneGroup) in enumerate(sorted(info_dir.items())):
-            for roleName, infoGroup in sorted(laneGroup.items()):
-                if "__class__" not in infoGroup:
+        for lane_group in info_dir.values():
+            for info_group in lane_group.values():
+                if "__class__" not in info_group or info_group["__class__"][()] != b"UrlDatasetInfo":
                     continue
-                loaded_class_name = infoGroup["__class__"][()].decode("utf-8")
-                if loaded_class_name == "UrlDatasetInfo":
-                    old_nickname = infoGroup["nickname"][()].decode("utf-8")
-                    del infoGroup["__class__"]
-                    del infoGroup["nickname"]
-                    infoGroup["__class__"] = "MultiscaleUrlDatasetInfo".encode("utf-8")
-                    infoGroup["nickname"] = MultiscaleUrlDatasetInfo.nickname_from_url(old_nickname).encode("utf-8")
-                    infoGroup["working_scale"] = "-1".encode("utf-8")
-                    infoGroup["scale_locked"] = "True".encode("utf-8")
+                old_nickname = info_group["nickname"][()].decode()
+                new_nickname = MultiscaleUrlDatasetInfo.nickname_from_url(old_nickname).encode()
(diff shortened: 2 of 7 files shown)


ilastik/applets/dataSelection/dataSelectionSerializer.py
```diff
@@ -331,6 +319,33 @@ def updateWorkingDirectory(self, newpath, oldpath):
...
+                    continue
+                loaded_class_name = infoGroup["__class__"][()].decode("utf-8")
+                if loaded_class_name == "UrlDatasetInfo":
+                    old_nickname = infoGroup["nickname"][()].decode("utf-8")
+                    del infoGroup["__class__"]
+                    del infoGroup["nickname"]
+                    infoGroup["__class__"] = "MultiscaleUrlDatasetInfo".encode("utf-8")
+                    infoGroup["nickname"] = MultiscaleUrlDatasetInfo.nickname_from_url(old_nickname).encode("utf-8")
+                    infoGroup["working_scale"] = "-1".encode("utf-8")
+                    infoGroup["scale_locked"] = "True".encode("utf-8")
```
emilmelnikov, line 347: 1. `laneIndex` doesn't seem to be used
2. Sorting doesn't seem to be necessary (maybe for debugging?..)
3. `bytes.decode` and `str.encode` do `utf-8` by default
4. Byte literals could be used directly instead of encoding them from strings

```suggestion
        for laneGroup in info_dir.values():
            for infoGroup in laneGroup.values():
                try:
                    if infoGroup["__class__"][()] != b"UrlDatasetInfo":
                        continue
                except KeyError:
                    continue
    
                old_nickname = infoGroup["nickname"][()].decode()
                new_nickname = MultiscaleUrlDatasetInfo.nickname_from_url(old_nickname).encode()

                del infoGroup["__class__"]
                infoGroup["__class__"] = b"MultiscaleUrlDatasetInfo"
                
                del infoGroup["nickname"]
                infoGroup["nickname"] = new_nickname
                
                infoGroup["working_scale"] = b"-1"
                infoGroup["scale_locked"] = b"True"
```

emilmelnikov (commented):
As a side note, it would be nice to have some sort of HDF5 helper module, so we could say for example:

```python
for inner_group in outer_group.values():
    for my_group in inner_group.values():
        if not h5utils.equals(my_group, "__class__", b"MyInfo"):
            continue
         # Does "()" indexing automatically.
         old_nickname = h5utils.get(my_group, "nickname")
         new_nickname = ...
         # Does "()" indexing and calls __delitem__ first
         h5utils.replace(my_group, "nickname", new_nickname)
```

Not sure how helpful this would be... Each helper by itself is probably not worth it, but, presented together as a whole module, maybe they could be useful.

ilastik/applets/base/appletSerializer.py
```diff
@@ -1197,6 +1197,13 @@ def updateWorkingDirectory(self, newdir, olddir):
         if they store relative paths."""
         pass
 
+    def updateLegacyProjectFile(self, file: h5py.File):
```
k-dominik, line 1200: This method is now always called when importing but not necessarily does change things. I'd propose to either check outside this method whether it should be called, or rename it to `maybeUpdateProjectFile` or something like this. And also it could be private as it's only used from _importProjectFile, which itself is already a private method.

k-dominik (commented):
lgtm.

Having an dedicated entry point for serializer updates is great idea! Have you seen any other serializers that do it differently and could be migrated to that, it would be great to open an issue.

@Tomaz-Vieira's comment from out last meeting still resonates in my brain, about the -1 index being a python specialty and I'm trying to think how this helps us instead of at least specifying the actual index in the data. I guess main reason  is that this can be done without reading any of the file info, whereas getting the actual scale would be cumbersome. I assume loading a remote dataset in our current code base (new project) would save the actual index of the working scale?

ilastik/applets/base/appletSerializer.py
btbest: [The internet](https://softwareengineering.stackexchange.com/questions/365310/how-to-name-a-method-which-may-or-may-not-perform-an-action-depending-on-a-condi) seems to lean towards "just name the function what it does if it meets its conditions". Or something like `ensureProjectFileIsUpToDate`. `updateLegacyEntries`. I'm ok with any.

I'm not a fan of splitting this into
```
if serializer.isLegacyProject(self.currentProjectfile):
  seralizer.updateProjectFile(self.currentProjectFile)
```
because then serializer subclasses have to implement two functions and their relation to each other isn't necessarily obvious besides containing some duplicated code.

I tried to see if the name should be underscored or not, but it didn't seem to matter around these parts (ProjectManager._importProject and _loadProject are both called in Shell._loadProject 🙃 ). Will underscore.

btbest:
> Having an dedicated entry point for serializer updates is great idea! Have you seen any other serializers that do it differently and could be migrated to that, it would be great to open an issue.

We probably have some legacy compatibility mechanisms that could be simpler as a one-time fix, but requiring Import Project is less nice for our users than just loading their old project files. I haven't seen anything I immediately itched to implement with this new mechanism.

> @Tomaz-Vieira's comment from out last meeting still resonates in my brain, about the -1 index being a python specialty and I'm trying to think how this helps us instead of at least specifying the actual index in the data. I guess main reason is that this can be done without reading any of the file info, whereas getting the actual scale would be cumbersome.

Exactly that; but since I'm now changing to key-string indexing, this import-upgrade mechanism will have to instantiate the loader and check the info file for the key names anyway.

> I assume loading a remote dataset in our current code base (new project) would save the actual index of the working scale?

Yup. For a project imported with this mechanism, the `-1` would only be replaced when the user next saves the project though.

ilastik/applets/base/appletSerializer.py
k-dominik: I'm fine keeping it as is although I'd prefer some kind of conditional. The logic for this has to be implemented anyhow, but yes, you're right, it would mean implementing a second function which would be more complicated to spot.

btbest: This is what an implementation with an exit step based on storage version could look like:

```
    # (self.version = "0.2" now bumped to self.version = "0.3")

    def updateLegacyEntries(self, file: File):
        try:
            info_dir = file[self.topGroupName]["infos"]
            storage_version = file[self.topGroupName]["StorageVersion"][()].decode()
        except KeyError:
            return
        
        storage_version = parse_version(storage_version)  # parse() from packaging.version
        if not self._isOutdatedStorage(storage_version):
            return

        if storage_version == parse_version("0.2"):
            # Version 0.2 may contain UrlDatasetInfo, the HBP-mode predecessor of MultiscaleUrlDatasetInfo

            # ... (Check all entries for UrlDatasetInfo)
                    
        del file[self.topGroupName]["StorageVersion"]
        file[self.topGroupName]["StorageVersion"] = self.version.encode()
                
    def _isOutdatedStorage(self, version: Version):
        return version < parse_version(self.version)
```

It does provide an obvious point for where to put analogous changes for future version bumps, but it's also a bunch of extra basically unneeded code...

btbest:
Closed in favour of https://github.com/ilastik/ilastik/pull/2805

In the end while an "Upgrade legacy project to current format" mechanism sounds attractive, this seemed like a user-friendlier and technically more sensible solution than modifying project files in-place.

PR #2799, opened 2024-01-18
remove OpBigTiffReader

### LINKED ISSUE #2798: remove `OpBigTiffReader`
The `tifffile` library gained a lot of functionality over time, so much so that we stopped depending on `pytiff` (in the course of #2709). We currently still maintain the distinction for _tiff_ and _bigtiff_, although both readers rely on `tifffile` only. I tested reading a _bigtiff multipage_ file with our normal tiff reader and it worked out of the box (something that even the bigtiff reader claimed not to support).

I propose to remove the `OpBigTiffReader`. Since `tifffile` supports lazy access, we also need not to worry about memmapping (I hope).

k-dominik:
Our OpTiffReader can, thanks to the mighty power of tifffile, deal with bigtiff files at least as well as the OpBigTiffReader (even better...)

moved the tests for `OpBigTiffReader` to `tests/test_lazyflow/test_operators/test_ioOperators/testOpTiffReader.py` 

closes #2798


btbest (approved):
Yay for removing unneeded stuff 🎉

Issue #2797, opened 2024-01-17
tiff reader cache blocksize should be im sync with axisorder


k-dominik:
In https://github.com/ilastik/ilastik/pull/2737 we introduced a blocked cache for tiff files that accelerated reading speeds for 3D data by having a larger cache size in z-direction.

With tiff files [we still do some guessing of axes](https://github.com/ilastik/ilastik/blob/380af1b1425e0f8ec7f8c4116425655cfd3baca3/lazyflow/operators/ioOperators/opTiffReader.py#L80-L81) which means we can be wrong sometimes. Our users can correct that in the UI. The cache, however, does not get updated upon such changes. Currently our guessing order is `"tzc"` for `"I"` (index) axes in tiff files. When changing from `"t"` to `"z"` manually, the access pattern will be very slow for the views orthogonal to the tiff page shapes.

In general we could investigate if
1. whether the time axes also benefits from a larger caching window
2. how to change block-shape in line with the set user axes

PR #2803, opened 2024-01-23
Cache output from all types of readers

k-dominik:
We've been kind of inefficient with reading some file formats, e.g. multipage tiff z-stacks. As a remedy [we added a cache](https://github.com/ilastik/ilastik/pull/2737). When axes are not recognized correctly, this cache could still be off (e.g. recognizing the axis first as "t", then the user changing it to "z", the viewing would still suffer, as this change cannot be communicated back to the cache efficiently.

This PR will adds a cache to all output formats that respects the readers `ideal_blockshape` attribute, with the exception for the "z" dimension, where a minimum cache size of `32` is enforced, that guarantees better interactivity for the orthoviews.

This might also be beneficial for the web-formats (but need to check if `ideal_blockshape` there is set up correctly.

## Checklist

- [ ] Check this does not impact performance with formats that might already do some form of internal caching (what about h5 files that have a chunk size of 1 in z, would we be faster there, or does the hdf5 library already do some magic caching there)
- [ ] check _all_ readers if they set `ideal_blockshape` in a sensible way. Maybe enforce it for readers.
- [ ] Add tests.

fixes https://github.com/ilastik/ilastik/issues/2797

### DIFF
--- a/ilastik/applets/dataSelection/opDataSelection.py
+++ b/ilastik/applets/dataSelection/opDataSelection.py
@@ -46,10 +46,16 @@ from ilastik.utility import OpMultiLaneWrapper
 from ilastik.workflow import Workflow
 from lazyflow.utility.pathHelpers import splitPath, globH5N5, globNpz, PathComponents
 from lazyflow.utility.helpers import get_default_axisordering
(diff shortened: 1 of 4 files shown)


k-dominik:
abandoning this for now as not good enough ;)

PR #2804, opened 2024-01-23
Tests: Fix mock table and assert combobox attempts to change

btbest:
The test broke because of https://github.com/ilastik/ilastik/pull/2793, but on CI it still passed because the pop up assert is off on CI, and that left only an assert for a mock not being called. Which still passes when the test does nothing.

### DIFF
--- a/tests/test_ilastik/test_applets/dataSelection/testDataSelectionGui.py
+++ b/tests/test_ilastik/test_applets/dataSelection/testDataSelectionGui.py
@@ -1,12 +1,11 @@
 # Data selection is already covered in workflow tests (e.g. testPixelClassificationGui.py)
 # Additional tests here should be workflow-agnostic.
 import os
-import platform
 from unittest import mock
 
 import pytest
 from PyQt5.QtGui import QStandardItemModel, QStandardItem
-from PyQt5.QtWidgets import QComboBox, QMessageBox
+from PyQt5.QtWidgets import QComboBox, QMessageBox, QApplication
 
 from ilastik.applets.dataSelection.datasetDetailedInfoTableModel import DatasetColumn
 from ilastik.applets.dataSelection.datasetDetailedInfoTableView import DatasetDetailedInfoTableView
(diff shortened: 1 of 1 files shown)


tests/test_ilastik/test_applets/dataSelection/testDataSelectionGui.py
```diff
@@ -80,9 +80,12 @@ def test_scale_select_exists_and_triggers_gui_event(dataset_table, mock_gui):
 def test_locked_scale_select_does_not_trigger_gui_and_informs_user(dataset_table, mock_gui, intercept_info_popup):
     dataset_table.scaleSelected.connect(mock_gui.handleScaleSelected)
     scale_cell_multiscale_locked = dataset_table.model().index(2, DatasetColumn.Scale)
-    editor_locked = dataset_table.indexWidget(scale_cell_multiscale_locked)
+    editor = dataset_table.indexWidget(scale_cell_multiscale_locked)
```
k-dominik, line 83: at the start of the file
```
from PyQt5.QtWidgets import QComboBox, QMessageBox, QApplication
```
plus
```suggestion
    editor = dataset_table.indexWidget(scale_cell_multiscale_locked)
    QApplication.processEvents()
```

makes it work for me

k-dominik (commented):
okay now on osx the test fails with `AttributeError: 'NoneType' object has no attribute 'currentIndexChanged'` which is referenced in the `skipif(...)`.

I debugged into it and when debugging the object suddenly exists, so I thought it might be a timing issue. `QApplication.processEvents()` does the trick for me...

side note, `editor_locked` is now called `editor` which is not reflected in the `skipif`

tests/test_ilastik/test_applets/dataSelection/testDataSelectionGui.py
btbest: Nice, thanks. Let's see if it works.

btbest: Looks promising, all green on the first try. I restarted macos-ilastik just to give it another run, since the problem before was flaky, not 100% fail (though tbh, it was failing more often than not).

PR #2805, opened 2024-01-25
Reimplement urldatasetinfo as legacy

btbest:
Here's what legacy UrlDatasetInfo compatibility could look like if we want to "upgrade in memory".

When attempting to load an old HBP-style project, the user gets a warning to please make a copy: 
image

If the user loads, we use a reimplemented UrlDatasetInfo class that provides the additional fields to upgrade the deserialized dict to a MultiscaleUrlDatasetInfo-equivalent. If the user then saves, we store the upgraded dict claiming to be a "MultiscaleUrlDatasetInfo", which makes older ilastik versions unable to load the newly saved project file.

This is an alternative solution to https://github.com/ilastik/ilastik/pull/2795

## Checklist

- [x] Add test to ensure UrlDatasetInfo.from_h5_group outputs an equivalent dict to MultiscaleUrlDatasetInfo.from_h5_group.

### DIFF
--- a/ilastik/applets/dataSelection/dataSelectionSerializer.py
+++ b/ilastik/applets/dataSelection/dataSelectionSerializer.py
@@ -26,6 +26,7 @@ import vigra
 
 import ilastik.utility.globals
 from ilastik.applets.base.appletSerializer import AppletSerializer, getOrCreateGroup, deleteIfPresent
+from ilastik.exceptions import UserAbort
 from ilastik.utility import bind
 from lazyflow.utility import PathComponents
 from lazyflow.utility.pathHelpers import getPathVariants, isUrl, isRelative
@@ -260,7 +261,7 @@ class DataSelectionSerializer(AppletSerializer):
 
         if info_class == UrlDatasetInfo:
             if not _confirmLoadLegacyProject():
-                raise RuntimeError("Loading aborted by user.")
+                raise UserAbort("User aborted loading of legacy project file.")
             dirty = True
 
         try:
--- a/ilastik/applets/dataSelection/dataSelectionSerializer.py
+++ b/ilastik/applets/dataSelection/dataSelectionSerializer.py
@@ -1,5 +1,3 @@
-from __future__ import absolute_import
-
 ###############################################################################
 #   ilastik: interactive learning and segmentation toolkit
 #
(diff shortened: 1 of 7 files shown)


ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -563,6 +558,42 @@ def _nickname_from_url(url: str) -> str:
...
+
+    @classmethod
+    def from_h5_group(cls, group: h5py.Group, params: Dict = None):
+        """
+        Returns a deserialized DatasetInfo dict equivalent to a dict from a MultiscaleUrlDatasetInfo
+        """
+        return super().from_h5_group(
+            group,
+            {
+                "url": group["filePath"][()].decode(),
```
btbest, line 590: Also this line should go since super().from_5_group handles that.
```suggestion
```

ilastik/applets/dataSelection/dataSelectionSerializer.py
```diff
@@ -260,6 +257,12 @@ def _readDatasetInfo(self, infoGroup, projectFilePath, headless):
                 info_class = PreloadedArrayDatasetInfo
 
         dirty = False
+
+        if info_class == UrlDatasetInfo:
+            if not _confirmLoadLegacyProject():
+                raise RuntimeError("Loading aborted by user.")
```
k-dominik, line 263: Just a small comment, but from a user perspective I find it a bit odd to first get a popup that implies a choice (load project or not) and if I go to abort, I'll get hit with a `RuntimeError` + messagebox:


### 1
image

### 2
image

Maybe some other error class could be raised here and to prevent the popup in `ilastikShell.py` https://github.com/ilastik/ilastik/blob/9058943988a8257b932c1d9f907ce9f3f9c29d71/ilastik/shell/gui/ilastikShell.py#L1671

k-dominik (approved):
:+1: lgtm, tried it out and works :)

emilmelnikov:
Hmm, wouldn't it be better to save a project and make a backup copy somewhere right away after a user clicks OK here?

btbest:
> Hmm, wouldn't it be better to save a project and make a backup copy somewhere right away after a user clicks OK here?

The project file could be any ridiculous size, so definitely not without asking the user. The dialog could of course give three options, "Make backup", "Continue without backup", "Cancel". Triggering a Save here creates a circular dependency though because an applet's serializer would have to go back calling Shell functions instead of the other way round. Not that it would be technically difficult to do, but our Save functions are all over the place and I'd want to do some major refactoring before calling any of them from further down the hierarchy.

k-dominik:
I agree in general that it would be safer to do a backup first, but,
* the instances of `UrlDatasetInfo` should be super rare, if any
* there is a fair warning
* I would want to avoid spending time _now_ to rewrite the logic so this would be feasible...

PR #2806, opened 2024-01-25
Tests: Deflake scale combobox in test on MacOS

btbest:
`QApplication.processEvents` didn't actually deflake the test unfortunately. I've messed around a bit with a `QTimer` to try whether a hard wait would fix it, and what amount of time would be required.

~Sadly, this seems to work, but only with 200ms or more wait time. It's not a small amount of time to sacrifice for a single test, so I'm not sure if we want to do this or if it's better to skip this test on MacOS...~ This didn't work because QTimer.singleShot of course dispatches the timer on a separate thread, it doesn't make the test thread wait :facepalm: Would have had to use time.sleep or sth.


k-dominik (approved):
awesome - definitely works for me and looks more deliberate than some random `processEvents` :)

Issue #2796, opened 2024-01-05
Export feature table without image


oanegros:
I only need my locations, predictions, and features once i'm running many files, and currently i need to create a lot of h5 files of masks of big datasets that i then subsequently have to delete again. 
It would be nice to be able to only export feature tables

k-dominik:
Unfortunately exporting those things separately is not really possible (for the table, image can be exported if feature table is not configured).
There still is a point to be made for exporting the image as well - there is no additional computational cost, and exporting the table with the object id image is a potential nice combo. If you choose "compressed hdf5" as output format, it should really not take a loot of disk space.

oanegros:
Ah okay, i didn't notice the compressed hdf5 option, maybe this is a useful default in general?

PR #2812, opened 2024-02-15
Change default export format to compressed hdf5

k-dominik:
most ilastik outputs compress very well - compressed hdf5 should be made the default format.
Furthermore the default format can now be customized in the config.

accompanying volumina request to change the default order: https://github.com/ilastik/volumina/pull/313

CC: @oanegros (https://github.com/ilastik/ilastik/issues/2796#issuecomment-1878606777)


k-dominik:
this change has @akreshuk's blessing (pending me checking whether our compressed data would open in MoBiE and our plugin)

btbest (approved):
I agree it's a sensible change, and looks fine on my system 👍 25%-ish file size reduction even on a tiny test image.

PR #2814, opened 2024-02-16
Ensure all object features get exported to table

k-dominik:
Everytime I look at `exportFile.py` I want to rewrite it. In lieu of time, another hack...

Previously, only features computed in the 1st timeframe would get a column in the export table. One would expect that with this bug there would be none ;). Global features are always computed -> produce an (empty) table in every frame, but local ones, only if there are objects.

This is more a bandaid then a proper fix. It would probably be better to go from `selected` features (from the dialog) for 
column names, but currently the plugin information is stripped from those :(

CC: @oanegros (thank you for pointing it out and helping to debug)

### DIFF
--- a/ilastik/utility/exportFile.py
+++ b/ilastik/utility/exportFile.py
@@ -83,7 +83,24 @@ def flatten_ilastik_feature_table(table, selection, signal):
     feature_channels = []
     feature_types = []
 
-    for plugin_name, feature_dict in computed_feature[0].items():
+    # HACK: local features are not computed if there are no objects.
+    # As a result, the feature table will not contain these columns in time those
+    # time frames.
+    # Ideally we would go by `selection` here. However, selection is
+    # a flat list, only with the feature names, without plugin association.
+    # So we got through the whole table and and check whether there are object
+    # from which we can get the column names
+    computed_features = {}  # {plugin_name: {feature_name: (feature_shape, feature_dtype)}}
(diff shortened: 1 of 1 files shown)


btbest (changes requested):
Oof, this one's a doozie... Let's say this function is hard to follow, and the naming is all over the place, but I don't feel like the addition in this PR makes it much worse either.

I also couldn't actually try out whether it fixes what it is supposed to fix. Is there a straightforward way to get a dataset where some object features are only computed from time point >0 onwards? I managed to coerce a dataset so that it has no objects in some time points, but that seems to cause errors much earlier in the workflow (see comment below).

Primarily, I would really like `computed_features` renamed, because it is super confusing to have it existing (and being important to the function) next to `computed_feature`. I suggest `features_by_time`. Besides that, basically just typos in the comment...

ilastik/utility/exportFile.py
```diff
@@ -105,8 +122,8 @@ def flatten_ilastik_feature_table(table, selection, signal):
                 feature_long_names.append(long_name)
                 feature_short_names.append(feat_name)
                 feature_plugins.append(plugin_name)
-                feature_channels.append((feat_array.shape[1]))
-                feature_types.append(feat_array.dtype)
+                feature_channels.append((feat_array_ch))
```
btbest, line 125: IntelliJ says "Remove redundant parentheses"

PR #2816, opened 2024-02-19
Multiscale: Support project files containing an integer working_scale

btbest:
All project files created with ilastik1.4.1b13 break otherwise due to the switch to string-indexed scale handling.

It may be worth looking into version compatibility testing. It should be possible on CI; could be costly though in terms of pipeline time.


lazyflow/utility/io_util/RESTfulPrecomputedChunkedVolume.py
```diff
@@ -21,6 +21,7 @@
 ###############################################################################
 import json
 import logging
+from typing import Union, Literal
```
k-dominik, line 24:
```suggestion
```

unused...

k-dominik (approved):
Glad this is fixed.

PR #2818, opened 2024-02-20
Added note that data is copied to project file

k-dominik:
image stack can be quite big. Users should be made aware that data will be duplicated.

xref: https://github.com/ilastik/ilastik.github.io/pull/264

### DIFF
--- a/ilastik/widgets/stackFileSelectionWidget.ui
+++ b/ilastik/widgets/stackFileSelectionWidget.ui
@@ -111,6 +111,13 @@ Examples:
       </widget>
      </item>
      <item>
+      <widget class="QLabel" name="import_copy_info">
+       <property name="text">
+        <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;&lt;span style=&quot; font-style:italic;&quot;&gt;Note: ilastik will copy the data into the project (.ilp) file.&lt;/span&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+       </property>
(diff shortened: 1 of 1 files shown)


ilastik/widgets/stackFileSelectionWidget.ui
```diff
@@ -110,6 +110,13 @@ Examples:
        </layout>
       </widget>
      </item>
+     <item>
+      <widget class="QLabel" name="import_copy_info">
+       <property name="text">
+        <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;&lt;span style=&quot; font-style:italic;&quot;&gt;Note: ilastik will copy the data into the project (.ilp) file.&lt;/span&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+       </property>
+      </widget>
```
btbest, line 118: Italicising the font can be done without parsing a whole HTML doc; is there a benefit to doing it that way?
```suggestion
      <widget class="QLabel" name="import_copy_info">
       <property name="text">
        <string>Note: ilastik will copy the data into the project (.ilp) file.</string>
       </property>
       <property name="font">
        <font><italic>true</italic></font>
       </property>
      </widget>
```

btbest (approved):

ilastik/widgets/stackFileSelectionWidget.ui
k-dominik: that's what QtDesigner produced 🤷 , but sure if there is a simpler way to do it...

PR #2820, opened 2024-02-22
Improve web loading gui

btbest:
A couple of small improvements to the web-source selector dialog: Ensuring the URL contains "http", and a history.

`        RECENT_URLS_KEY = "Recent URLs"` <- is this an acceptable key to use for inside the preferences?


btbest:
I've lowercased the preferences key after looking at my preferences.json - looks like lowercase keys are more common.

k-dominik (approved):
cool to have a history!

I noticed though, no playing around with it, that the scale seems not to be recovered after loading (add data, select features, save) -> https://github.com/ilastik/ilastik/issues/2824

PR #2831, opened 2024-03-08
Fix layerviewer update after shape change

btbest:
Fixing and test-covering a couple of subtle bugs I found in the layerViewerGui while implementing OME support

### DIFF
--- a/tests/test_ilastik/test_applets/layerViewer/testLayerViewerGui.py
+++ b/tests/test_ilastik/test_applets/layerViewer/testLayerViewerGui.py
@@ -1,4 +1,5 @@
 import numpy
+import pytest
 import vigra
 
 from ilastik.applets.layerViewer.layerViewerGui import LayerViewerGui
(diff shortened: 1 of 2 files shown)


tests/test_ilastik/test_applets/layerViewer/testLayerViewerGui.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def make_simple_op():
+    op = OpPassthrough(parent=Operator(graph=Graph()))
+    op.Image.setValue(numpy.array([[1, 1], [1, 1]]))
+    op.Output.meta.axistags = vigra.defaultAxistags("xy")
+    return op
+
+
+def test_layer_update_respects_2d_default():
```
k-dominik, line 31: for completeness, this test could be parametrized to also test the `prefer_2d=False` case.

tests/test_ilastik/test_applets/layerViewer/testLayerViewerGui.py
```diff
@@ -0,0 +1,75 @@
...
+def test_layer_update_does_not_update_viewer_with_unready_layers(monkeypatch):
+    """
+    This reproduces a complex race condition encountered when switching scales in the dev-version OME implementation.
+    * LayerViewerGui.updateAllLayers detects shape change, sets self.editor.dataShape = newShape
+    * Qt triggers drawBackground on the editor's imageScenes (I think this happens after `dataShape.setter` exits)
+    * _refreshTile calls `LazyflowSource.request` (tileprovider.py:277 `ims_req = ims.request(dataRect, stack_id[1])`)
+    * dispatches the req to the threadpool (tileprovider.py:287 `fetch_fn = partial(self._fetch_layer_tile, ...)`)
+    * `ims.request` calls op.Output[] and catches SlotNotReadyErrors, but at this time the slot is still ready
+    * the delayed _fetch_layer_tile awaits the request, but now the slot is not ready anymore
+    * _fetch_layer_tile does not catch SlotNotReadyErrors, so it crashes
```
k-dominik, line 50: I'm sure I'm missing something, but I cannot follow the timeline here fully:
`LayerViewerGui.updateAllLayers` uses `LayerViewerGui.determineDatashape` to determine the new shape from it's observed slots. From there, if I follow the timeline, volumina would still access some old slots somehow? So the ones that trigger the shape change become unready?

btbest: In any case, I don't think this description should be in the test actually... I'll document it here instead.

Here are the parts I do understand, even though I could never quite figure out how exactly the specific race condition happens.

Root cause: `_refreshTile` in volumina.tileprovider is called while a layer's dataSlot is ready, so it successfully creates a request. By the time the request is executed, the slot is unready, so `_fetch_layer_tile` raises and crashes the volumina thread.

Event chain as far as I can gather:

* DataSelectionGui maintains one LayerViewerGui per slot in opDataSelection.DatasetGroup, which does not change at any point in the context of this bug, so the LayerViewerGui is one persistent instance throughout.
* Change in opDataSelection input slot (ActiveScale) sets up a new opInputDataReader.
* `opDataSelection.Image` is connected to the new reader's output slot.
* slot.notifyReady (or possibly .notifyUnready; both are bound but I forget which is the trigger here) -> LayerViewerGui.setNeedUpdate -> LayerViewerGui.updateAllLayers
* updateAllLayers detects shape change, sets self.editor.dataShape = newShape
* Qt triggers drawBackground on the editor's imageScenes (I think this happens after `dataShape.setter` exits)
* _refreshTile calls `LazyflowSource.request` (tileprovider.py:277 `ims_req = ims.request(dataRect, stack_id[1])`)
* dispatches the req to the threadpool (tileprovider.py:287 `fetch_fn = partial(self._fetch_layer_tile, ...)`)
* `ims.request` calls op.Output[] and catches SlotNotReadyErrors, but at this time the slot is still ready
* the delayed _fetch_layer_tile awaits the request, but now the slot is not ready anymore
* _fetch_layer_tile does not catch SlotNotReadyErrors, so it crashes

The datasources' dataSlot in question is `opDataSelection.Image`, so the slot that is switched over to the new opInputDataReader's output. 

I suspected that somehow the problem is that the shape update is at the wrong place, because it can trigger requests at a time when layers may be unready. That generally shouldn't happen, hence the test. And it does fix the issue in the OME implementation.

But I'm not sure how the race condition happens. When halting during debug, of course the problem doesn't occur, hence why I'm pretty sure it's a race condition. But the shape change implies that the setup is already finished - otherwise the Image slot wouldn't have its new shape yet. So then what makes the slot unready again afterwards?

k-dominik (approved):
Lgtm, thank you for the clarification in the comments!

PR #2832, opened 2024-03-08
Precomputed handling

btbest:
A couple of smaller changes from my OME branch that we can independently integrate into main:
- Better feedback for connection errors in the URL dialog
- Moving stripping the "precomputed://" prefix from URLs into the Precomputed adapter to reduce duplications of the string constant across data loading classes (and because knowing about the prefix should be none of their concern)

### DIFF
--- a/lazyflow/operators/ioOperators/opInputDataReader.py
+++ b/lazyflow/operators/ioOperators/opInputDataReader.py
@@ -289,10 +289,9 @@ class OpInputDataReader(Operator):
     def _attemptOpenAsRESTfulPrecomputedChunkedVolume(self, filePath):
         if not filePath.lower().startswith("precomputed://"):
             return ([], None)
-        url = filePath.lstrip("precomputed://")
         reader = OpRESTfulPrecomputedChunkedVolumeReader(parent=self)
         reader.Scale.connect(self.ActiveScale)
-        reader.BaseUrl.setValue(url)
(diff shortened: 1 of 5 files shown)


lazyflow/operators/ioOperators/opInputDataReader.py
```diff
@@ -289,10 +289,9 @@ def _attemptOpenAsUfmf(self, filePath):
     def _attemptOpenAsRESTfulPrecomputedChunkedVolume(self, filePath):
         if not filePath.lower().startswith("precomputed://"):
             return ([], None)
-        url = filePath.lstrip("precomputed://")
         reader = OpRESTfulPrecomputedChunkedVolumeReader(parent=self)
```
k-dominik, line 292: probably not worth a change, but `OpRESTfulPrecomputedChunkedVolumeReader` seems to take `baseUrl` also as a parameter in `__init__`.

btbest: It doesn't look like it takes it as a parameter, but worse - it compares `self.BaseUrl.value` (full input url) to `self._volume_object.volume_url` (prefix stripped). I guess the web adapter does need to keep a copy of the full url before stripping.

btbest: This, and (more importantly) determining that it's a Precomputed URL in OpInputDataReader

Issue #2836, opened 2024-04-03
Feature Selection constraint prevents adding dataset, crashes ilastik


btbest:
When adding a dataset to a project that already has active downstream applets, these may reject the new dataset and raise a DatasetConstraintError. The user flow for adding a dataset responds by opening a dialog that allows the user to e.g. reorder axes of the dataset to be added. In this case, the constraint is from Feature Selection ("Some of your selected scales are too large for your dataset"). Cancelling the Dataset Properties dialog that pops up in this case does not properly reset the GUI state, and the downstream applet remains active for the user to navigate to. Doing so crashes ilastik because the Feature Selection applet looks for a single-lane GUI, but no lanes and hence no GUIs exist.

## Expected behavior
Naively: Cancelling the Dataset Properties dialog should correctly reset the GUI and deactivate downstream applets.

When thinking about it: Applets like Feature Selection can continue to impose constraints on data selection even when all lanes are removed, so they should remain active if their backend is instantiated, and should be able to handle lane count = 0. Or there should be some other way for the user to remove the constraint (e.g. by resetting the applet completely).

## To Reproduce
Within Pixel Classification
1. Add some data in the data selection applet (e.g. some screenshot png)
2. Go to Feature Selection, add features to the largest scale the dataset can handle
3. Go back to data selection, remove the dataset -- Feature Selection applet becomes deactivated
4. Click Add New, and attempt to add a much smaller dataset (e.g. a tiny crop from the screenshot)
5. Feature Selection should raise a DatasetConstraintError about some scales being too large
6. Ok the error, Dataset Properties should pop up
7. Cancel Dataset Properties
8. Feature Selection applet remains active, click it to provoke the error below

## Error message/traceback

```pytb
ERROR 2024-04-03 17:53:23,115 excepthooks 12296 20896 Unhandled exception in thread: 'MainThread'
ERROR 2024-04-03 17:53:23,134 excepthooks 12296 20896 Traceback (most recent call last):
  File "C:\Program Files\ilastik-1.4.1b15\lib\site-packages\ilastik\shell\gui\ilastikShell.py", line 1225, in handleAppletBarItemExpanded
    self.setSelectedAppletDrawer(modelIndex)
  File "C:\Program Files\ilastik-1.4.1b15\lib\site-packages\ilastik\shell\gui\ilastikShell.py", line 1244, in setSelectedAppletDrawer
    self.showCentralWidget(applet_index)
[...]

PR #2837, opened 2024-04-03
Multi-GUI adapter: Stop throwing if no lanes exist

btbest:
This is a band-aid for #2836 that at least stops downstream applets from throwing errors because they are looking for GUIs of lanes that do not exist.

The user then ends up on a Feature Selection applet with an empty viewer, which is less weird than seeing the data selection gui. In fact, this allows them to change the selected scales, so they can even fix the project without necessarily having to close it...
image


k-dominik (commented):
You'll also need to modify `ilastik/dev/environment-dev.yml` - circleci and appveyor build from that one.

k-dominik (approved):

PR #2838, opened 2024-04-05
Refactor applet serializer

k-dominik:
I was in the process of modifying appleSerializer not to use pickle anymore and found it quite annoying that it requires so much jumping around in the file. This is a simple refactor of the `appletSerializer` module in a few submodules, specifically `appletSerializer.appletSerializer`, `appletSerializer.slotSerializer`, and `appletSerializer.serializerUtils`. I exported everything that is used in ilastik, so imports don't change.

No functional change; small additional changes:
* removed some unused serializer imports.
* updated the black version in the pre-commit config ; in CI it seems we're always using the latest one (didn't find an option to pin it for the darker action).

Todos:
- [x] needs #2837 for tests to pass


btbest (approved):
lgtm :) always good to clean up some unused imports

k-dominik:
thank you, @btbest

PR #2841, opened 2024-04-12
Debug logging

btbest:
I've been wondering for a while why starting ilastik in debug mode (`--debug`) doesn't change the logging behaviour, even though judging by the line in `app.py`, it seemed like it should.

This PR doesn't quite fix that, because it turns out we have a mechanism that overwrites the current logging config with the contents of `logging_config.json` every 10 seconds. Unless we remove that mechanism, this config file has the final say, and it contains `level=INFO` for basically everything (root level and all top-level modules we develop).

With this PR
- The blanket overwrites that set loglevel to INFO on root level and all top-level modules are removed
- Starting ilastik with the --debug flag actually sets the loglevel to DEBUG by default on root level
- A whole bunch of submodule-specific overwrites setting level=INFO remain. I didn't check all, but some of them do suppress submodules that can be a bit spammy on DEBUG, which one might not want if not specifically working on that. Removing the setting from the config file lets the DEBUG default (in --debug mode) bubble down to them from the root level.


ilastik/ilastik_logging/default_config.py
```diff
@@ -132,85 +128,20 @@ def get_default_config(
...
-            "volumina.utility.shortcutManager": {"level": "INFO"},
-            # Python doesn't provide a trace log level, so we use a workaround.
-            # By convention, trace loggers have the same hierarchy as the regular loggers, but are prefixed with 'TRACE' and always emit DEBUG messages
+            "OpenGL": {"level": "INFO"},
+            "yapsy": {"level": "INFO"},
+            # Loglevels for our own modules (ilastik, lazyflow, volumina) are in ./logging_config.json.
+            # Note regarding TRACE settings: Python doesn't provide a trace log level, so we use a workaround.
+            # By convention, trace loggers have the same hierarchy as the regular loggers,
+            # but are prefixed with 'TRACE' and always emit DEBUG messages.
             # To enable trace messages, change one or more of these to use level DEBUG
```
k-dominik, line 144: Whatever this TRACE setting is supposed to do, is maybe better documented in the file where it's used (`logging_config.json`).

k-dominik (commented):
Changes look good to me.

The debug mode was always more  on the debated side in the team. It's main use seems really to hide certain functionality we do not deem ready for user adoption. It is good you fixed the general logging functionality in debug mode.

I've never used this changing the config at runtime feature - do you think it works? I tried but logging settings would be only picked up if changed before startup. In any case I think this functionality can be removed, as it's pretty non-standard, and noone would expect it (maybe in a future pr).

ilastik/ilastik_logging/default_config.py
btbest: I was going to move this comment to the json file but json doesn't support comments. I'll move it over anyway I guess and just make it a text field (as [suggested here](https://stackoverflow.com/questions/244777/can-comments-be-used-in-json)).

btbest:
> I've never used this changing the config at runtime feature - do you think it works? I tried but logging settings would be only picked up if changed before startup. In any case I think this functionality can be removed, as it's pretty non-standard, and noone would expect it (maybe in a future pr).

It does work, but now I realise it's only on when you're _not_ in debug mode (app.py#295). Maybe this was added as a backdoor to modify logging on a user's machine when giving support? Though it's not like modifying the json file or the .py file right next to it would make any difference in that setting. If you agree, I think I would just remove it along with the json file and put the config back into the python code where 118d1a5e removed it.

ilastik/ilastik_logging/default_config.py
btbest: Looks like in our case this is only actually used in a handful modules. I left the corresponding entries and removed all the other ones; plus the various traceLogger instantiations that weren't used across our codebase.

btbest:
Noticed I had overlooked another problem - `logging.dictConfig()` disables all loggers not mentioned in the new config 🤦 

Fixed and will merge once CI is through

PR #2842, opened 2024-04-16
Update open recent project list when saving as a project

### LINKED ISSUE #2830: Save Project As does not add project to recent projects in ilastik splash screen
Just a slight annoyance, that if i 'Save Project As...' i wil have to look up this project in my file tree again. This also leads to me accidentally working in the wrong project as i select the most recent project, assuming this is where i left off.

 
Related, if people are looking into this: can we have a few more recent projects listed? Or even a scrollable list?

thodkatz:
Hey, I am Theo! This is my first PR for the ilastik :) 

I am attempting to close #2830 . With this change the menu option `Save Project As` will refresh the opened recent files, in the exact same way that is happening when a project is loaded.

I wanted to add tests for this change, but I realized that there aren't any for the `ilastik/shell/gui/ilastikShell.py`. If this needs to be addressed on this PR please let me know  :)

### DIFF
--- a/ilastik/shell/gui/ilastikShell.py
+++ b/ilastik/shell/gui/ilastikShell.py
@@ -158,7 +158,7 @@ class MemoryWidget(QWidget):
         self.setMemoryBytes(0)
 
     def setMemoryBytes(self, bytes):
-        self.label.setText("Cached Data: %1.1f MB" % (bytes / (1024.0 ** 2.0)))
+        self.label.setText("Cached Data: %1.1f MB" % (bytes / (1024.0**2.0)))
 
 
 # ===----------------------------------------------------------------------------------------------------------------===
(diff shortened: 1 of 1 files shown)


ilastik/shell/gui/ilastikShell.py
```diff
@@ -1832,6 +1834,8 @@ def saveAs():
 
                 try:
                     self.projectManager.saveProjectAs(newPath)
+                    workflowDisplayName = self.projectManager.workflow.workflowDisplayName
+                    self._setRecentlyOpenedList(newPath, workflowDisplayName)
```
k-dominik, line 1838: a bit of a nitpick, but technically the try-except block only makes sense for `self.projectManager.saveProjectAs(newPath)` (as only `ProjectManager.SaveError` is caught here). Anything else will raise. So I'd put the code after the except block.

k-dominik (commented):
Hello @thodkatz,

thank you so much for your contribution! I checked out your changes. Factoring out setting the recently opened project list makes sense :). I also gave these changes a go (in lieu of tests, manually) and it works nicely!

Thanks again!

ilastik/shell/gui/ilastikShell.py
thodkatz: Thank you very much for the review! My pleasure contributing :)

Initially I also put it after the except block, and then I changed it, because I thought that if there is a save error, the list shouldn't be updated. Please correct me if I am wrong.

k-dominik: yes, you're totally right :)

k-dominik (approved):

k-dominik:
Thank you very much for your first contribution @thodkatz!

PR #2846, opened 2024-04-17
Use pydantic for .ilp deserialization in experimental/api

k-dominik:
Summary:
* public api unchanged
* internal representations of "applets" based on pydantic models - currently limited to the data that is needed for processing (will add more along with the tool for data relocation)
* added `_h5helpers` module in `ilastik.experimental`, which could also live in `ilastik.applet.base.appletSerializer.serializerUtils`, but it's not needed there
* some internals have changed -> tests needed to be adapted

## Checklist

- not needed, with pydantic classes mostly documents themselves ;D Add docstrings and comments.
- [x] Add tests.
- n/a (api unchanged) Update [user documentation](https://github.com/ilastik/ilastik.github.io).
- n/a Reference relevant issues and other pull requests.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -47,6 +47,7 @@ outputs:
         # need to update our packages to be compatible with pandas 2 API
         - pandas
         - psutil
+        - pydantic 2.*
         - pyopengl
         - pyqt 5.15.*
         # previous versions would set thread limits globally with side effects
--- a/ilastik/experimental/parser/types.py
+++ b/ilastik/experimental/parser/types.py
@@ -72,9 +72,8 @@ class InputData(BaseModel):
     @property
     def axis_order(self) -> str:
         """Returns the axis keys of the last lane as a string"""
-        # should usually be 'Raw Data'
         last_lane = self._last_lane
-        base_role_name = self.role_names[0]
+        base_role_name = self.role_names[0]  # should usually be 'Raw Data'
         base_info = last_lane[base_role_name]
         assert base_info
         return "".join(ax.key for ax in base_info.axistags)
@@ -83,7 +82,7 @@ class InputData(BaseModel):
     def num_channels(self) -> int:
         """Returns the number for channels in the last lane"""
         last_lane = self._last_lane
-        base_role_name = self.role_names[0]
+        base_role_name = self.role_names[0]  # should usually be 'Raw Data'
         base_info = last_lane[base_role_name]
         assert base_info
         n_channels = 1
--- a/ilastik/experimental/parser/types.py
+++ b/ilastik/experimental/parser/types.py
@@ -1,7 +1,7 @@
 ###############################################################################
 #   ilastik: interactive learning and segmentation toolkit
 #
-#       Copyright (C) 2011-2023, the ilastik developers
+#       Copyright (C) 2011-2024, the ilastik developers
 #                                <team@ilastik.org>
 #
 # This program is free software; you can redistribute it and/or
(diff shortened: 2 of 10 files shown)


lazyflow/classifiers/sklearnLazyflowClassifier.py
```diff
@@ -69,7 +69,7 @@ def __ne__(self, other):
 
 class SklearnLazyflowClassifier(LazyflowVectorwiseClassifierABC):
 
-    VERSION = 2  # Used for pickling compatibility
+    VERSION = 555  # Used for pickling compatibility
```
btbest, line 72: What's up with this random number?

ilastik/experimental/parser/types.py
```diff
@@ -19,139 +19,113 @@
...
+    def _last_lane(self):
+        last_lane_name = next(reversed(self.infos))
+        return self.infos[last_lane_name]
+
+    @property
+    def axis_order(self) -> str:
+        """Returns the axis keys of the last lane as a string"""
+        # should usually be 'Raw Data'
+        last_lane = self._last_lane
+        base_role_name = self.role_names[0]
```
btbest, line 77:
```suggestion
        last_lane = self._last_lane
        base_role_name = self.role_names[0]  # should usually be 'Raw Data'
```

btbest (approved):
Looks good I think. There are some subtle changes to the loading behaviour, so it's not strictly only a migration, but I think these changes are improvements and I don't see any risk.

- Shoving the actual deserialization into BeforeValidators reads like a hack, but going by the docs, it looks like this is pretty much the intended usage.
- I liked the key enums in the old implementation a bit more than the current field aliases. The naming made it really obvious that these are the key strings inside the ilp file. It was also visually slightly nicer than having some aliases at the end of a line and some broken over onto the next. If there's some easy way to get this back, it would be nice, but otherwise not worth it.
- Similarly, `model_validate` is less expressive than `from_ilp_file`, but at least this adds a terminological distinction between API classes and serialization interface classes.

Overall this solution is a lot more declarative, which is a great improvement for a serialization interface like this. Also some nice smaller improvements like having the lane naming pattern explicitly in there.

As a side note, the example usage from the PixelClassificationPipeline docstring doesn't work because `imageio.imread` gives a numpy array ("NotImplementedError: Data type '<class 'numpy.ndarray'>' not supported, use `vigra.VigraArray` or `xarray.DataArray`."). Easy to solve with something like `img = vigra.VigraArray(img, axistags=vigra.defaultAxistags("xyc")` of course. I wouldn't complain if we tack the small change to the docstring on here :grin:

lazyflow/classifiers/sklearnLazyflowClassifier.py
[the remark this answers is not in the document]
k-dominik: ah sorry, I fixed this in the morning, but forgot to push :/

lazyflow/operators/classifierOperators.py
```diff
@@ -94,6 +94,7 @@ def setupOutputs(self):
         else:
             raise Exception("Unknown classifier factory type: {}".format(type(classifier_factory)))
 
+        print(f"{new_mode=}")
```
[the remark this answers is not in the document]
k-dominik: ah sorry, I fixed this in the morning, but forgot to push :/

conda-recipe/meta.yaml
```diff
@@ -47,6 +47,7 @@ outputs:
         # need to update our packages to be compatible with pandas 2 API
         - pandas
         - psutil
+        - pydantic 2.*
```
FynnBe, line 50: I'd consider a pin to the minor version. pydantic is still noticibly changing

FynnBe (approved):

conda-recipe/meta.yaml
k-dominik: after discussion with @FynnBe, we'll keep the pin as is. The `pydantic` features used so far in ilastik are _very_ basic. Also pinning it further would require some coordination with `bioimageio` dependencies.

Issue #2207, opened 2020-03-20
Usage of slots in lazyflow


m-novikov:
Can we drop slots from lazyflow in favor of function arguments (or constructor arguments).
Currently, slots are:
* Represent computation graph
* Help with incremental initialization of operator, especially important for top-level operators that have a lot of slots
* Allow dirty propagation

cc @Tomaz-Vieira

btbest:
We've removed the Transaction class that presumably came out of this discussion back (see #2848), since it was unused and unnecessarily bloated stack traces. I did the removal in a single commit (4580a98571d8792f7dde7e1441b4d5edf6b4efbc) so that it is easy to revert. If we want to continue work on this, we can restore it by merging https://github.com/ilastik/ilastik/tree/restore-transaction.

PR #2848, opened 2024-04-19
Remove transaction

btbest:
This removes a Transaction mechanism that was implemented in 2019 and then never used.

I am guessing that this was intended to allow us to remove a TransactionSlot that is used in a bunch of ops as a workaround, and judging by https://github.com/ilastik/ilastik/issues/2207, some "disconnect slot X first before we do this other thing so that we don't end up inconsistent in the meantime". Possibly also as an alternative solution to the current SetupDepthContext / `call_when_setup_finished` queuing mechanism ~~that buffers `setupOutputs` calls~~ correction: the SetupDepthContext is only used to prevent `LayerViewerGui.updateAllLayers` from running while the graph is being changed through `Slot.connect, disconnect, setValue` or `setDirty`. (That, plus a few notifyDirty with defer=True in countingGui and edgeTrainingGui.)

In the meantime, the additional piping through functions to check if a transaction is going on has been bloating stack traces unnecessarily since then.

We could keep some sort of archive branch with a revert of this commit in case we ever want to seriously implement it...

Current:
Screenshot 2024-04-19 154008

With this PR:
Screenshot 2024-04-19 154046


k-dominik (approved):
Heya,

part of me is saddened that this never took off, but implementing transactions on that level made the stack traces indeed more complicated than they should be - something noone considered back then, and I think we just got used to it in the end. Your example is striking. I almost look forward to future exceptions during setup ;).

We need some sort of more natural transaction mechanism, than the once you can currently find in the code, so +1 on your suggestions to "archive" this transaction approach in a branch (maybe with an accompanying issue (could also be a comment in #2207), so it's a bit more findable).

btbest:
The branch with the revert is here: https://github.com/ilastik/ilastik/tree/restore-transaction

Will crosspost on the other issue

PR #2850, opened 2024-04-25
Lazyflow logging and fix default operator naming

btbest:
Now that debug logging actually works, this introduces a bunch of new debug statements that allow tracing what lazyflow is doing. Also:

- Fix default naming of operators. This was only happening in `_after_init`, which is too late. We want the default name to already be set at the start of `__init__` so that slots instantiated during `__init__` can put it in their logger name.
- Add utf-8 encoding to log config to override cp1252 on windows and make it possible to log slot.meta (which can contain σ,  throws in cp1252)

At this point, the changes I make when I need to debug lazyflow are:

2. Set `write_logs=True` in `Slot.__init__` instead of False to set up slot-wise logging
3. Change `INFO` to `DEBUG` in logging_config.json for lazyflow.signals, lazyflow.graph and lazyflow.op_debug
4. Exchange `module` for `name` in the `verbose` format in `default_config.py` -- This is also the format for the log file in prod, otherwise I would commit this change.

### DIFF
--- a/lazyflow/operator.py
+++ b/lazyflow/operator.py
@@ -205,6 +205,8 @@ class Operator(metaclass=OperatorMetaClass):
         :param parent: the parent operator; if None the instance is a
         root operator
         :param graph: a Graph instance
+        :param write_logs: Debugging feature. The operator will write debug logs if True.
+        Make sure the `lazyflow.op_debug` logger has level=DEBUG in the logging config.
 
         """
         if not (parent is None or isinstance(parent, Operator)):
--- a/lazyflow/operator.py
+++ b/lazyflow/operator.py
@@ -232,7 +232,7 @@ class Operator(metaclass=OperatorMetaClass):
         if write_logs:
             self._debug_logger = logging.getLogger(f"lazyflow.op_debug.{self.name}")
             self._debug_logger.debug(
-                f"Instantiated {self.name} {id(self)} with parent={self.parent.name if self.parent else ''}"
+                f"Instantiated {self.name} {id(self)} with parent={self.parent.name if self.parent is not None else ''}"
             )
 
         self._initialized = False
--- a/lazyflow/operator.py
+++ b/lazyflow/operator.py
@@ -228,7 +228,7 @@ class Operator(metaclass=OperatorMetaClass):
         if parent is not None:
             parent._add_child(self)
 
-        self.logger = logging.getLogger(f"lazyflow.op_debug.{type(self).__name__}")
+        self.logger = logging.getLogger(f"lazyflow.op_debug.{self.name}")
         self.logger.debug(f"Instantiated {self.name} {id(self)} with parent={self.parent.name if self.parent else ''}")
 
         self._initialized = False
--- a/lazyflow/operator.py
+++ b/lazyflow/operator.py
@@ -197,7 +197,7 @@ class Operator(metaclass=OperatorMetaClass):
         obj.outputs = OutputDict(obj)
         return obj
 
-    def __init__(self, parent=None, graph=None):
+    def __init__(self, parent=None, graph=None, write_logs=False):
         """
         Either parent or graph have to be given. If both are given
         parent.graph has to be identical with graph.
@@ -228,8 +228,12 @@ class Operator(metaclass=OperatorMetaClass):
         if parent is not None:
             parent._add_child(self)
 
-        self.logger = logging.getLogger(f"lazyflow.op_debug.{self.name}")
-        self.logger.debug(f"Instantiated {self.name} {id(self)} with parent={self.parent.name if self.parent else ''}")
+        self._debug_logger = None
+        if write_logs:
+            self._debug_logger = logging.getLogger(f"lazyflow.op_debug.{self.name}")
+            self._debug_logger.debug(
+                f"Instantiated {self.name} {id(self)} with parent={self.parent.name if self.parent else ''}"
+            )
 
         self._initialized = False
 
@@ -393,7 +397,8 @@ class Operator(metaclass=OperatorMetaClass):
         if self._parent is not None:
             del self._parent._children[self]
 
-        self.logger.debug(f"Cleaning up {self.name} {id(self)}")
+        if self._debug_logger:
+            self._debug_logger.debug(f"Cleaning up {self.name} {id(self)}")
         # Disconnect ourselves and all children
         self._disconnect()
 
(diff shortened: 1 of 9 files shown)


lazyflow/slot.py
```diff
@@ -108,6 +105,7 @@ def __init__(
         allow_mask=False,
         subindex=(),
         top_level_slot=None,
+        write_logs=False,
```
k-dominik, line 108: This is kinda neat, to enable per slot debug messages which would otherwise be overkill via a config. Seeing this I wonder if this would not also be nice to have selective per instance logging for only some operators.

k-dominik (approved):
lgtm

It does add quite some noise if one doesn't need to debug the whole graph with all ops, but I guess that will be a tooling question.

lazyflow/slot.py
btbest: I actually thought it was a bit ugly, adding a parameter and what feels like too much logic purely for logging. Plus, it's effectively logging config outside of the logging_config. But in a pixel classification example, this prevents bloating the already existing 469 loggers up to 901 loggers, so I figured it might be worth it...

k-dominik: Yeah, I had the same thoughts first, but then I tried it out and found it convenient :)

lazyflow/utility/orderedSignal.py
```diff
@@ -22,6 +20,9 @@
 # 		   http://ilastik.org/license/
 ###############################################################################
 from collections import OrderedDict
+import logging
+
+logger = logging.getLogger("lazyflow.signals")
```
k-dominik, line 25: using this naming convention here is probably fine as `lazyflow.signals` will be findable.

k-dominik (approved):
It's been probably too long since I looked at this PR the last time, so I don't quite remember if it was already the case: My feeling for logger names was that we had a convention that was basically based on `__name__` + "Class name". Here this is being stretched a bit introducing special `lazyflow.slot.<OperatorName>.<Slot>`. This introduces possible ambiguity wrt to logger names with operators that have the same `name` attr, but live in different modules. Also it breaks the previous naming logic. In principle I can live with this, but I'd favor to solve it similar to what you've done with `op_debug` - to have a special logger for `slot_debug` as well, and maybe do `lazyflow.slot_debug.<OperatorName>.<Slot>`...

lazyflow/operator.py
```diff
@@ -301,6 +300,10 @@ def _after_init(self):
             islot.notifyUnready(self.handleInputBecameUnready)
 
         self._initialized = True
+
+        self.logger = logging.getLogger(f"lazyflow.op_debug.{type(self).__name__}")
```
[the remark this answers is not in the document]
btbest: I think in a scenario where somehow having identically named operators in different modules seemed like a good idea, both of them writing to the same logger most likely isn't an issue... Would you rather use `f"lazyflow.op_debug.{type(self).__module__}.{type(self).__name__}"` and get e.g. `'lazyflow.op_debug.ilastik.applets.dataSelection.opDataSelection.OpMultiLaneDataSelectionGroup'`?

lazyflow/operatorWrapper.py
```diff
@@ -90,7 +90,7 @@ def __init__(
 
         self._initialized = False
 
-        if operatorClass.name == "":
+        if operatorClass.name == Operator.name:
             self._name = "Wrapped " + operatorClass.__name__
         else:
             self._name = "Wrapped " + operatorClass.name
```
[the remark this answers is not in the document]
btbest: The default naming of operators gives their _instances_ the class name as their default `self.name`, but the _class_ remains with `cls.name == "Operator (base class)`. This branching runs into both cases a lot because we have many with and many without name in the class definition. Now that I think about it... we already have a custom OperatorMetaClass. This is exactly the kind of stuff metaclasses are for :smile: I will see if I can move this into the meta class, as I think that will allow us to basically remove all the other default renaming.

The wrapper has `._name` because `.name` is a property, which it is so that the setter can track whether it's been called using `._customName`, which exists so that the wrapper can take the `instance.name` from the wrapped operator once it creates instances of it.

The switch in `wrapper.__init__` is necessary as long as operator default naming only happens to instances, but I don't see the point of renaming the wrapper when it creates instances of the wrapped op. The only benefit of this extra logic is to catch if the custom `self.name` is set inside `MyOp.__init__`. Which none of our operators do. I will remove the renaming on creating wrapped instances, together with `_customName`, `_name` and the `name` getter/setter.

Overall the value of all this logic is pretty limited anyway, we have maybe 5 or 6 custom names in our entire code base. Those few cases where it really seems to help could also be covered with creating an empty subclass with a unique name...

btbest:
> I'd favor to solve it similar to what you've done with `op_debug` - to have a special logger for `slot_debug` as well, and maybe do `lazyflow.slot_debug.<OperatorName>.<Slot>`...

Done. I've also moved the per-operator loggers behind a `write_logs` flag the same way as the slot loggers. There's a catch compared to slots because we don't subclass `Slot`, while op subclasses sometimes set their own `cls.logger` in line with the naming scheme you mentioned. This means those particular classes will not have an `op_debug` logger, and their `self.logger` will print logs from the base class when turned on. I suppose the workaround would be to use a more obscure variable name like `self._debug_logger`, but will have to decide what to do with `self.logger` in classifierOperators.py tomorrow...

btbest:
Oh dear, I figured I should actually _try_ the instance-level debugging (which I hadn't actually used myself...) and found another gotcha. `Slot.__init__` works differently way depending on whether it is called during an op class construction, or during an op instantiation. I found the place to pass on the parameter, so now you can get an individual slot logging everything it's doing like:

```
class MyOp(Operator):
  Image = InputSlot(write_logs=True)
```

On the operator level, the flag has to be passed to the instantiation either like `MyOp(write_logs=True)` if the subclass pipes kwargs through to super, otherwise directly into the super call like `super().__init__(write_logs=True)`.

lazyflow/operatorWrapper.py
```diff
@@ -90,12 +84,9 @@ def __init__(
...
 
-        if operatorClass.name == "":
-            self._name = "Wrapped " + operatorClass.__name__
-        else:
-            self._name = "Wrapped " + operatorClass.name
-
-        self._customName = False
+        self.name = "Wrapped " + operatorClass.name
+        if self._debug_logger:
+            self._debug_logger.debug(f"Wrapper {id(self)} name={self.name}")
```
k-dominik, line 89: I don't think there is a way currently to get the operatorwrapper a debug logger (_write_logs not exposed). Same goes for any other operator class that overrides `__init__` (e.g. `OpReorderAxes`). And in general I think more operators should use init to set initial values. So currently the following operators will not be debuggable like this (because they override `__init__` without passing `**kwargs` up:

* `OperatorWrapper`
* `OpPixelOperator`
* `OpReorderAxes`
* `OpObjectFeatures`
* `Workflow`
* `OpManualTracking`
* `OpDifference`
* `OpZeroBasedConsecutiveIndexRelabeling`
* `OpRelabeledMergerFeatureExtraction`
* `OpConservationTracking`
* `OpTrackingFeatureExtraction`
* `OpAnnotations`
* `OpCarving`

k-dominik: I see you mention this, too, in your last comment.

PR #2851, opened 2024-04-26
Web-loading GUI: Update OK button

btbest:
Just a tiny improvement to the dialog for choosing a URL. Currently, the user first has to "Check URL", and if the URL gives a successful response, the Ok button is activated. If the user then changes the URL, the button remains activated and the user can add any arbitrary string to the project (which will attempt to fetch from there and fail).

### DIFF
--- a/ilastik/applets/dataSelection/precomputedVolumeBrowser.py
+++ b/ilastik/applets/dataSelection/precomputedVolumeBrowser.py
@@ -7,6 +7,7 @@ Todos:
   - check whether can me somehow merged with dvidDataSelctionBrowser
 
 """
+
 import logging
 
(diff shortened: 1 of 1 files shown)


k-dominik (approved):
Works!

Behavior on enter is interesting:
* if URL is not checked, enter performs check url. if URL has been successfully checked, next enter accepts.
* editing (e.g. adding a character), enter (unsuccessful check) -> the `Unknown error while trying to connect to this address....` shows up, which is correct.
* editing (removing character from before), `Add to project` becomes enabled immediately, text field doesn't change (still the Unknown error) - one enter here directly adds the data.

it's a tiny bit inconsistent.

btbest:
Thanks for noticing 🤦

PR #2854, opened 2024-04-30
Add Object Classification from label image workflow

### LINKED ISSUE #2416: Enable pre-indexed superpixels as input for object classification
## Description

When using a workflow where the user inputs their own segmentation image, (e.g. object classification), if the objects in the segmentation mask pre-indexed (i.e. have intensities provided), the workflow should use these as the object IDs rather than re-indexing the objects. 

## Scenario

We've a utilized another program to segment, label, and extract data on objects. These objects and their corresponding data are indexed according to the object IDs (intensity values) from the other program. We want to use Ilastik to classify these objects, and are using the labeled segmentation images as inputs in the object classification workflow. By default, Ilastik re-indexes these objects using a different labeling algorithm (diamond vs box), which leads to object splitting, shape changes, and indexing issues in the Ilastik output data. Currently we're getting around the indexing issues by uploading the segmentation as an atlas image, but we don't have a workaround for the shape changes.

This is also relevant for time-course experiments. For instance, let's say we are analyzing time course images, but only want to compare the initial and final frames to save on processing. Under the current build, if the objects have moved between the initial and final frames, Ilastik's default re-indexing could change the object identities, leading to incorrect class assignment. Having the option to use the input's intensity values as object identities would bypass this issue.

## Rationale
By default, Ilastk relabels the objects based on a connected components labeling algorithm (diamond), which splits diagonally-connected superpixels into separate objects. This can lead to identity and indexing issues if, say, the user labeled the objects using a different algorithm (e.g. box).

## Scope
As above. Enabling this feature would help to integrate Ilastik into other workflows and prevent indexing/identity issues from relabeling superpixels.

## Possible implementation

Should be pretty easy, just use the intensity values from the user input if it is not a binary mask. In addition, allowing the user to choose the connected components labeling function given a binary mask could bypass the splitting issue if they are using Ilastik for segmentation. 

-->

k-dominik:
This is wip - but already kinda works ;)

Inputs such as label images from cellpose or stardist are treated a bit poorly  in ilastik - connected components is run again (with possibly different connectivity) resulting in different objects in ilastik after loading such label image data.

This PR introduces a new workflow that internally doesn't do CCA anymore but merely relabels the label image to ensure consecutive labels (required by feature extraction).

see discussion in #2416

fixes #2416

## Checklist

- [ ] **Add original labels to export table.** 
- [ ] simplify `OpObjectExtraction` such that `setupOutputs` can also go to the base class (labelingop should expose a slot for cache pre-loading).
- [ ] Add tests.

### DIFF
--- a/benchmarks/objectExtractionRunningTime.py
+++ b/benchmarks/objectExtractionRunningTime.py
@@ -56,6 +56,7 @@ FEATURES = {
     }
 }
 
+
 # Cleanup functions (used in vigra_objfeats.py)
 def cleanup_key(k):
     return k.replace(" ", "")
@@ -207,7 +208,7 @@ class ObjectExtractionTimeComparison(object):
         # Object extraction
         self.opObjectExtraction = OpObjectExtraction(graph=g)
         self.opObjectExtraction.RawImage.connect(self.op5Raw.Output)
-        self.opObjectExtraction.BinaryImage.connect(self.op5Binary.Output)
+        self.opObjectExtraction.SegmentationImage.connect(self.op5Binary.Output)
         self.opObjectExtraction.Features.setValue(FEATURES)
 
         # Simplified object features operator (No overhead)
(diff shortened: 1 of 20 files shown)


k-dominik:
closed in favor of  #2921

PR #2856, opened 2024-05-06
Python 3.10 compatiblity fixes

### LINKED ISSUE #2852: Test failure for gh actions on MacOS
since recently, tests on MacOS seem to segfault: [example build](https://github.com/ilastik/ilastik/actions/runs/8849996198/job/24307555126)

Likely in this test file:
`tests/test_ilastik/widgets/test_datasetInfoEditorWidget.py::test_modify_axistags_in_stack`.

k-dominik:
One of our test configurations picked up Python 3.10 (which is fine for the "lirbary" usage). This resulted in some minor errors:

1. Threadings `condition.notifyAll` has been marked for deprecation in favor of `condition.notify_all`. In `TestCompatibilityChecks.test_access_opaque_slot_value_should_not_warn_or_raise` we check that no warning is raised. This was of course broken by the deprecation. Rather than adapting the test I preferred keeping up with Python
2. Python stopped implicitly converting types in function calls -> results in errors/segfaults in pyqt. Explicitly converted types or made sure e.g. integer value division is used (e.g. in draw methods of `featureTableWidget`).

fixes #2852

### DIFF
--- a/ilastik/applets/counting/countingGui.py
+++ b/ilastik/applets/counting/countingGui.py
@@ -56,7 +56,7 @@ def _countingColorTable():
     colortable = matplotlib_to_qt4_colortable("jet", N=n_gradual, asLong=False)
     weights = numpy.tanh(numpy.linspace(0, 1.0, n_gradual))
     for w, c in zip(weights, colortable):
-        c.setAlpha(w * 255)
+        c.setAlpha(int(w * 255))
     colortable = [c.rgba() for c in colortable]
 
     return [*colortable, *[colortable[-1]] * (n_total - n_gradual)]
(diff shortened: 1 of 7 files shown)


btbest (approved):
Nicely tracked down 🫡 This means these bits of code actually weren't python3.10 compatible this whole time..?

There are three `drawEllipse` calls in the codebase that use old_div and could hence pass floats to the Qt function (which I assume now only works with int); featureTableWidget.py:174 and two in `preView.py`.

I suppose to be absolutely sure, one would have to explicitly `int()` the division results as well, since `6//2.0` is `2.0`, but I think we'll be fine assuming that we're only passing `int`s into those divisions into the first place.

k-dominik:
> There are three `drawEllipse` calls in the codebase that use old_div and could hence pass floats to the Qt function (which I assume now only works with int); featureTableWidget.py:174 and two in `preView.py`.

ok - I will make those `old_div` -> `//` as well for better consistency, too, thank you for pointing to that.

The type annotations for pyqt help a lot with the `int` issue - so in most cases pyright tells me it's allright without casting. But yes, to be super sure we could cast.

PR #2862, opened 2024-05-22
omg Zarr

btbest:
- Load OME-Zarr datasets from https:// and file:// URLs
- Viewer defaulted to hide orthoviews because many public datasets seem to come in chunk sizes of "one full xy slice at a time" (hence orthoview = download full dataset)
- Switch scales
- Export to already supported file formats without botched default file names

Road map: Export to Zarr...

## Checklist

- [x] Add tests.
- [x] Update [user documentation](https://github.com/ilastik/ilastik.github.io) -- [PR](https://github.com/ilastik/ilastik.github.io/pull/272)

### DIFF
--- a/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
+++ b/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
@@ -23,7 +23,6 @@ from typing import List, Dict
 from PyQt5.QtCore import Qt, QAbstractItemModel, QModelIndex
 from ilastik.utility import bind
 from ilastik.utility.gui import ThreadRouter, threadRouted
-from lazyflow.utility.io_util.multiscaleStore import Multiscale
 from .opDataSelection import DatasetInfo
 from .dataLaneSummaryTableModel import rowOfButtonsProxy
 
@@ -38,12 +37,12 @@ class DatasetColumn:
     NumColumns = 6
 
 
-def _scale_to_display_string(scale: Multiscale, axiskeys: str) -> str:
+def _dims_to_display_string(dimensions: List[int], axiskeys: str) -> str:
     """Generate labels to put into the scale combobox.
-    scale.resolution must be in xyz and will be reordered to match axiskeys."""
-    input_axes = dict(zip("xyz", scale.dimensions))
-    reordered_resolution = [input_axes[axis] for axis in axiskeys if axis in input_axes]
-    return ", ".join(str(size) for size in reordered_resolution)
+    Scale dimensions must be in xyz and will be reordered to match axiskeys."""
+    input_axes = dict(zip("xyz", dimensions))
+    reordered_dimensions = [input_axes[axis] for axis in axiskeys if axis in input_axes]
+    return ", ".join(str(size) for size in reordered_dimensions)
 
 
 @rowOfButtonsProxy
@@ -210,7 +209,7 @@ class DatasetDetailedInfoTableModel(QAbstractItemModel):
             return str(datasetInfo.drange or "")
         if index.column() == DatasetColumn.Scale:
             if datasetInfo.scales:
-                return _scale_to_display_string(datasetInfo.scales[datasetInfo.working_scale], datasetInfo.axiskeys)
+                return _dims_to_display_string(datasetInfo.scales[datasetInfo.working_scale], datasetInfo.axiskeys)
             return UninitializedDisplayData[index.column()]
 
         raise NotImplementedError(f"Unknown column: row={index.row()}, column={index.column()}")
@@ -225,7 +224,7 @@ class DatasetDetailedInfoTableModel(QAbstractItemModel):
         datasetInfo = datasetSlot.value
         if not datasetInfo.scales:
             return {}
-        return {key: _scale_to_display_string(scale, datasetInfo.axiskeys) for key, scale in datasetInfo.scales.items()}
+        return {key: _dims_to_display_string(dims, datasetInfo.axiskeys) for key, dims in datasetInfo.scales.items()}
 
     def is_scale_locked(self, laneIndex) -> bool:
         datasetSlot = self._op.DatasetGroup[laneIndex][self._roleIndex]
--- a/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
+++ b/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
@@ -23,7 +23,7 @@ from typing import List, Dict
 from PyQt5.QtCore import Qt, QAbstractItemModel, QModelIndex
 from ilastik.utility import bind
 from ilastik.utility.gui import ThreadRouter, threadRouted
-from lazyflow.utility.io_util.multiscaleWebStore import Multiscale
+from lazyflow.utility.io_util.multiscaleStore import Multiscale
 from .opDataSelection import DatasetInfo
 from .dataLaneSummaryTableModel import rowOfButtonsProxy
 
--- a/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
+++ b/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
@@ -1,7 +1,7 @@
 ###############################################################################
 #   ilastik: interactive learning and segmentation toolkit
 #
-#       Copyright (C) 2011-2014, the ilastik developers
+#       Copyright (C) 2011-2024, the ilastik developers
 #                                <team@ilastik.org>
 #
 # This program is free software; you can redistribute it and/or
--- a/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
+++ b/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
@@ -41,9 +41,7 @@ class DatasetColumn:
 def _scale_to_display_string(scale: Multiscale, axiskeys: str) -> str:
     """Generate labels to put into the scale combobox.
     scale.resolution must be in xyz and will be reordered to match axiskeys."""
-    if scale.resolution is None:
-        return scale.key
-    input_axes = dict(zip("xyz", scale.resolution))
+    input_axes = dict(zip("xyz", scale.dimensions))
     reordered_resolution = [input_axes[axis] for axis in axiskeys if axis in input_axes]
     return ", ".join(str(size) for size in reordered_resolution)
 
--- a/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
+++ b/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
@@ -23,6 +23,7 @@ from typing import List, Dict
 from PyQt5.QtCore import Qt, QAbstractItemModel, QModelIndex
 from ilastik.utility import bind
 from ilastik.utility.gui import ThreadRouter, threadRouted
+from lazyflow.utility.io_util.multiscaleWebStore import Multiscale
 from .opDataSelection import DatasetInfo
 from .dataLaneSummaryTableModel import rowOfButtonsProxy
 
@@ -37,15 +38,12 @@ class DatasetColumn:
     NumColumns = 6
 
 
-def _resolution_to_display_string(resolution: List[int], axiskeys: str) -> str:
-    """
-    In Precomputed format, resolution is in xyz order, which we want to display in the same order as axiskeys.
-    To support formats other than Precomputed, the tableModel would need to be able to obtain
-    the order of the resolution axes (or generally, how to transform resolution to a display string)
-    from the datasetSlot.
-    """
-    assert len(resolution) == 3, "Expected resolution to be in xyz order, please report this on http://image.sc"
-    input_axes = dict(zip("xyz", resolution))
+def _scale_to_display_string(scale: Multiscale, axiskeys: str) -> str:
+    """Generate labels to put into the scale combobox.
+    scale.resolution must be in xyz and will be reordered to match axiskeys."""
+    if scale.resolution is None:
+        return scale.key
+    input_axes = dict(zip("xyz", scale.resolution))
     reordered_resolution = [input_axes[axis] for axis in axiskeys if axis in input_axes]
     return ", ".join(str(size) for size in reordered_resolution)
 
@@ -214,9 +212,7 @@ class DatasetDetailedInfoTableModel(QAbstractItemModel):
             return str(datasetInfo.drange or "")
         if index.column() == DatasetColumn.Scale:
             if datasetInfo.scales:
-                return _resolution_to_display_string(
-                    datasetInfo.scales[datasetInfo.working_scale]["resolution"], datasetInfo.axiskeys
-                )
+                return _scale_to_display_string(datasetInfo.scales[datasetInfo.working_scale], datasetInfo.axiskeys)
             return UninitializedDisplayData[index.column()]
 
         raise NotImplementedError(f"Unknown column: row={index.row()}, column={index.column()}")
@@ -231,10 +227,7 @@ class DatasetDetailedInfoTableModel(QAbstractItemModel):
         datasetInfo = datasetSlot.value
         if not datasetInfo.scales:
             return {}
-        return {
-            key: _resolution_to_display_string(scale["resolution"], datasetInfo.axiskeys)
-            for key, scale in datasetInfo.scales.items()
-        }
+        return {key: _scale_to_display_string(scale, datasetInfo.axiskeys) for key, scale in datasetInfo.scales.items()}
 
     def is_scale_locked(self, laneIndex) -> bool:
         datasetSlot = self._op.DatasetGroup[laneIndex][self._roleIndex]
--- a/ilastik/applets/dataSelection/datasetDetailedInfoTableView.py
+++ b/ilastik/applets/dataSelection/datasetDetailedInfoTableView.py
@@ -224,7 +224,7 @@ class ScaleComboBoxDelegate(QStyledItemDelegate):
         if model.is_scale_locked(index.row()):
             message = (
                 "You have already continued in the project with this dataset at the selected scale. "
-                'To inspect another scale, please use "Add New" and add the same URL as '
+                'To inspect another scale, please use "Add New" and add the same URL or path as '
                 "another dataset, or create a new project."
             )
             QMessageBox.information(self.parent(), "Scale locked", message)
--- a/ilastik/applets/dataSelection/datasetDetailedInfoTableView.py
+++ b/ilastik/applets/dataSelection/datasetDetailedInfoTableView.py
@@ -1,7 +1,7 @@
 ###############################################################################
 #   ilastik: interactive learning and segmentation toolkit
 #
-#       Copyright (C) 2011-2014, the ilastik developers
+#       Copyright (C) 2011-2024, the ilastik developers
 #                                <team@ilastik.org>
 #
 # This program is free software; you can redistribute it and/or
--- a/ilastik/applets/dataSelection/datasetDetailedInfoTableView.py
+++ b/ilastik/applets/dataSelection/datasetDetailedInfoTableView.py
@@ -175,7 +175,7 @@ class InlineAddButtonDelegate(QItemDelegate):
                 button = AddFileButton(parent_view, index=index)
                 button.addFilesRequested.connect(partial(parent_view.handleCellAddFilesEvent, button))
                 button.addStackRequested.connect(partial(parent_view.handleCellAddStackEvent, button))
-                button.addPrecomputedVolumeRequested.connect(
+                button.addDatasetByUrlRequested.connect(
                     partial(parent_view.handleCellAddPrecomputedVolumeEvent, button)
                 )
                 button.addDvidVolumeRequested.connect(partial(parent_view.handleCellAddDvidVolumeEvent, button))
@@ -251,7 +251,7 @@ class DatasetDetailedInfoTableView(QTableView):
 
     addFilesRequested = pyqtSignal(int)  # Signature: (lane_index)
     addStackRequested = pyqtSignal(int)  # Signature: (lane_index)
-    addPrecomputedVolumeRequested = pyqtSignal(int)  # Signature: (lane_index)
+    addRemoteSourceRequested = pyqtSignal(int)  # Signature: (lane_index)
     addDvidVolumeRequested = pyqtSignal(int)  # Signature: (lane_index)
     addFilesRequestedDrop = pyqtSignal(object, int)  # Signature: (filepath_list, lane_index)
 
@@ -297,7 +297,7 @@ class DatasetDetailedInfoTableView(QTableView):
 
     @pyqtSlot(int)
     def handleCellAddPrecomputedVolumeEvent(self, button):
-        self.addPrecomputedVolumeRequested.emit(button.index.row())
+        self.addRemoteSourceRequested.emit(button.index.row())
 
     def wheelEvent(self, event):
         """
@@ -359,7 +359,7 @@ class DatasetDetailedInfoTableView(QTableView):
         button.addFilesRequested.connect(partial(self.addFilesRequested.emit, -1))
         button.addStackRequested.connect(partial(self.addStackRequested.emit, -1))
         button.addDvidVolumeRequested.connect(partial(self.addDvidVolumeRequested.emit, -1))
-        button.addPrecomputedVolumeRequested.connect(partial(self.addPrecomputedVolumeRequested.emit, -1))
+        button.addDatasetByUrlRequested.connect(partial(self.addRemoteSourceRequested.emit, -1))
         layout.setContentsMargins(0, 0, 0, 0)
         layout.addWidget(button)
         layout.addStretch()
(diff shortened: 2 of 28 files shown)


ilastik/applets/dataSelection/multiscaleDatasetBrowser.py
```diff
@@ -91,22 +114,27 @@ def update_ok_button(text):
...
+            # Ask each store type if it likes the URL to avoid web requests during instantiation attempts.
+            if OMEZarrStore.is_url_compatible(url):
+                rv = OMEZarrStore(url)
+            elif RESTfulPrecomputedChunkedVolume.is_url_compatible(url):
+                rv = RESTfulPrecomputedChunkedVolume(volume_url=url)
+            else:
+                store_types = [OMEZarrStore, RESTfulPrecomputedChunkedVolume]
+                supported_formats = "\n".join([f"{s.NAME} ({s.URL_HINT})" for s in store_types])
+                msg = f"Address does not look like any supported format.\n\nSupported formats:\n{supported_formats}"
+                self.result_text_box.setText(msg)
```
emilmelnikov, line 136:
```suggestion
                supported_formats = "".join(f"<li>{s.NAME} ({s.URL_HINT})</li>" for s in store_types)
                self.result_text_box.setHtml(
                    f"<p>Address does not look like any supported format.</p>"
                    f"<p>Supported formats:</p>"
                    f"<ul>{supported_formats}</ul>"
                )
```

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -550,7 +550,13 @@ def create_data_reader(self, parent: Optional[Operator] = None, graph: Optional[
 
     @property
     def display_string(self):
-        return "Remote: " + self.url
+        return "URI: " + self.url
+
+    @property
+    def default_output_dir(self) -> Path:
+        if self.url.startswith("file://"):  # Might be // or ///
+            return Path(self.url.lstrip("file://").lstrip("/")).absolute().parent
```
emilmelnikov, line 558:
```suggestion
            return Path(self.url.removeprefix("file:").lstrip("/")).resolve().parent
```

`Path.resolve` also resolves symlinks.

lazyflow/utility/io_util/OMEZarrStore.py
```diff
@@ -0,0 +1,181 @@
...
+from zarr.storage import FSStore, LRUStoreCache
+
+from lazyflow import rtype
+from lazyflow.utility import Timer
+from lazyflow.utility.io_util.multiscaleStore import MultiscaleStore, Multiscale, DEFAULT_LOWEST_SCALE_KEY
+
+logger = logging.getLogger(__name__)
+
+OME_ZARR_V_0_4_ARGS = {
+    "dimension_separator": "/",
```
emilmelnikov, line 37: This is just a preference, but I personally like to use `dict(k1=v1, k2=v2, ...)` instead of dict literals, but only when all keys are function kwargs: this makes it easier to change code in case you need to inline these kwargs later (no need to remove quotes and replace `:` with `=`).

Of course, feel free to ignore this, it's not really important.

lazyflow/utility/io_util/OMEZarrStore.py
```diff
@@ -0,0 +1,181 @@
...
+    https://github.com/ome/ome-zarr-py/blob/master/ome_zarr/format.py#L69
+    Checks the metadata dict for a version
+
+    Returns the version of the first object found in the metadata,
+    checking for 'multiscales', 'plate', 'well' etc
+    """
+    multiscales = metadata.get("multiscales", [])
+    if multiscales:
+        dataset = multiscales[0]
+        return dataset.get("version", None)
```
emilmelnikov, line 56:
```suggestion
        return dataset.get("version")
```

tests/test_ilastik/test_applets/dataSelection/testOpDataSelection.py
```diff
@@ -864,17 +861,119 @@ class TestOpDataSelection_PrecomputedChunks:
...
+    def op(self, graph, monkeypatch, datasetInfo) -> OpDataSelection:
+        op = OpDataSelection(graph=graph)
+        op.WorkingDirectory.setValue(os.getcwd())
+        op.ActiveScale.setValue(DEFAULT_LOWEST_SCALE_KEY)
+        op.Dataset.setValue(datasetInfo)
+        return op
+
+    def test_load_precomputed_chunks_over_http(self, op):
+        loaded_scale0 = op.Image[:].wait()
+        assert numpy.allclose(loaded_scale0, self.IMAGE_SCALED)
```
emilmelnikov, line 878:
```suggestion
        numpy.testing.assert_allclose(loaded_scale0, self.IMAGE_SCALED)
```

I think this produces better error messages.

ilastik/applets/dataSelection/opDataSelection.py
btbest: Would you recommend to generally replace all `absolute` usages with `resolve` or are there pitfalls that I'm overlooking?

tests/test_ilastik/test_applets/dataSelection/testOpDataSelection.py
btbest: Looks like it behaves slightly differently. numpy.testing.allclose fails the test: `AssertionError:  Not equal to tolerance rtol=1e-07, atol=0 (shapes (1, 1, 1, 10, 12), (1, 10, 12) mismatch)`, so numpy.allclose has been ignoring a mismatch of singleton axes all this time

ilastik/applets/dataSelection/opDataSelection.py
emilmelnikov: I think it's generally better to work with "canonical" paths if possible (no `.`/`..`, no symlinks, `/` instead of `\\` on Windows etc.) but maybe sometimes you have to keep `\\` because you are interfacing with other code, or you need to handle symlinks in a special way, so just doing `s/absolute/resolve/g` everywhere is probably not a good idea.

emilmelnikov: By the way, is `lstrip("/")` really what do we want here? I've seen somewhere that `file:///my/file` (three slashes) gets mapped to absolute `/my/file`, but `file://my/file` (two slashes) gets mapped to relative `my/file`.

tests/test_ilastik/test_applets/dataSelection/testOpDataSelection.py
emilmelnikov: Wow, did not expect that...

Excerpt from NumPy docs:

> Returns True if two arrays are **element-wise equal** within a tolerance.

ilastik/applets/dataSelection/opDataSelection.py
btbest: I suppose overall while I thought this would be a neat feature, it introduces a lot of complexity because it's the first time we need to convert a URI to a `pathlib.Path`. It might be better to roll this back and let the default export folder be the project file's location.

k-dominik: I'd not use `.resolve` for user paths/files. Usually there is a reason one symlinks to ones files.

lazyflow/utility/io_util/OMEZarrStore.py
```diff
@@ -0,0 +1,181 @@
...
+        except jsonschema.ValidationError:
+            raise ValueError(f"Metadata for this dataset did not match OME-Zarr spec.\nReceived:\n{self.ome_spec}.")
+        multiscale_spec = self.ome_spec["multiscales"][0]
+        axistags = _get_axistags_from_spec(multiscale_spec)
+        datasets = multiscale_spec["datasets"]
+        assert len(datasets) > 0, "The OME-Zarr store contains no datasets."
+        dtype = None
+        gui_scale_metadata = {}  # Becomes slot metadata -> must be serializable (no ZarrArray allowed)
+        self._scale_data = {}
+        for scale in reversed(datasets):
```
k-dominik, line 141: Using `OrderedDict` from `collections` would slightly improve understanding of the code where normally I don't think users rely much on the insertion order to dicts, and (at least it looks like it for me, since you reverse the sequence before inserting) you might.

Second point here concerns the reversing of order. I checked with both precomputed and ngff, and those seem to have the ordering from highest resolution (largest image shape) to lowest (smallest image shape). You seem to reverse this order, without documenting why (other than that maybe the dropdown menu starts with the lowest res, which would be the cheapest to load?!). If there is no good reason to reverse then I think sticking to the order that the majority of the data formats exhibit would be preferable...

lazyflow/utility/io_util/OMEZarrStore.py
```diff
@@ -0,0 +1,181 @@
...
+                dtype = zarray.dtype.type
+                gui_scale_metadata[scale["path"]] = Multiscale(
+                    key=scale["path"], resolution=list(zarray.shape[-1:-4:-1])  # xyz
+                )
+                self._scale_data[scale["path"]] = {
+                    "zarray": zarray,
+                    "chunks": zarray.chunks,
+                    "shape": zarray.shape,
+                }
+                logger.info(f"Initializing scale {scale['path']} took {timer.seconds()*1000} ms.")
```
k-dominik, line 153: shouldn't seconds -> milliseconds require division by 1000?

```suggestion
                logger.info(f"Initializing scale {scale['path']} took {timer.seconds()/1000} ms.")
```

in case I'm not mistaken here, this would also be applicable to other occurrences in this file

lazyflow/operators/ioOperators/opInputDataReader.py
```diff
@@ -286,6 +288,18 @@ def _attemptOpenAsUfmf(self, filePath):
         else:
             return ([], None)
 
+    def _attemptOpenAsOmeZarrMultiscaleFileset(self, filePath):
```
k-dominik, line 291: Does the `FileSet` bit in the name communicate some important information?

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -566,9 +572,16 @@ def from_h5_group(cls, group: h5py.Group, params: Dict = None):
...
+        """
+        Take the part after the last /, make it safe for use as a file name,
+        remove anything that looks like an extension ('.zarr') and replace remaining dots
+        to ensure exporting logic does not mistake them for file extensions.
+        """
         last_url_component = url.rstrip("/").rpartition("/")[2]
         filename_safe = re.sub(r"[^a-zA-Z0-9_.-]", "_", last_url_component)
-        return filename_safe
+        extensionless = os.path.splitext(filename_safe)[0]
+        exporter_safe = extensionless.replace(".", "_")
```
k-dominik, line 583: I don't yet understand this one - Why is this only an issue with URL-type datasets? I tried with an image with lots of dots in the name and found no odd behavior on export.

I'm only asking because for users it might not be obvious behavior - for some datatypes the filename stem is preserved, for others not. Could you give an example filename that wouldn't work?

k-dominik (commented):
Finally had a look!

Good that you're being nice and friendly, and really try to read the URIs that the user enters. 

Also nice to see the (for us relevant) changes to the ngff spec summarized in a compact piece of code (`OMEZarrStore.py`)

Naming-wise it appears that the terms "scale" and "resolution" are used interchangeably (e.g. `DEFAULT_LOWEST_SCALE_KEY` and `lowest_resolution_key`) which could be more uniform (unless it makes sense beyond what I got when reading it). I found that at least for precomputed chunks and ngff seem not to agree on the lingo either (precomputed has mutliple scales of different resolutions, wheras ngff has multiple resolution levels that could be scaled differently 😮‍💨  - if we call this stuff multi-scale then we should probably stick to that, although). It would be cool to adopt maybe one of the definitions if possible, or give up 🤷 and maybe state that resolution and scale are often meant to be the same.

For me the biggest itch is probably the reversing of scales - as it adds one more transformation one has to keep in mind when going through the code. Do you think it's possible to keep the order as in the data?

lazyflow/utility/io_util/OMEZarrStore.py
k-dominik: awww :/ sorry I was wrong :)

ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
```diff
@@ -37,15 +38,12 @@ class DatasetColumn:
...
-    the order of the resolution axes (or generally, how to transform resolution to a display string)
-    from the datasetSlot.
-    """
-    assert len(resolution) == 3, "Expected resolution to be in xyz order, please report this on http://image.sc"
-    input_axes = dict(zip("xyz", resolution))
+def _scale_to_display_string(scale: Multiscale, axiskeys: str) -> str:
+    """Generate labels to put into the scale combobox.
+    scale.resolution must be in xyz and will be reordered to match axiskeys."""
+    if scale.resolution is None:
+        return scale.key
```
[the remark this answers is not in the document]
btbest: It's leftover from an earlier state before I decided to pass the shape as resolution to the GUI in the case of OME-Zarr. We can remove the case and make `.resolution` mandatory. Maybe should rename this to `dimensions` since it's kind of just "Whatever info related to the dimensions at that scale level for the GUI".

lazyflow/utility/io_util/multiscaleStore.py
```diff
@@ -0,0 +1,100 @@
...
+        ...
+
+    @abstractmethod
+    def get_chunk_size(self, scale_key: str) -> Tuple[int]:
+        """Preferred download/computation chunk size of the dataset at the given scale."""
+        ...
+
+    @property
+    @abstractmethod
+    def NAME(self) -> str:
```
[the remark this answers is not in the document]
btbest: They're hard-coded string constants. As said in the other comment, the reason the ABC has them as properties is because this is the only way I'm aware of in Python to force subclasses to define class members. Will update the docstring to make it clear.

lazyflow/utility/pathHelpers.py
```diff
@@ -416,3 +417,26 @@ def globNpz(path: str, globString: str):
 def globList(listOfPaths, globString):
     matches = [x for x in listOfPaths if fnmatch.fnmatch(x, globString)]
     return matches
+
+
+def uri_to_Path(uri: str) -> pathlib.Path:
+    """Copy of the Python 3.13 implementation of pathlib.Path.from_uri"""
```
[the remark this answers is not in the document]
btbest: You're right, looks like we need to include the copyright notice and a short summary of any changes made with the code, the Python license in a separate file, and a reference to it in the readme ([link](https://discuss.python.org/t/how-do-i-properly-adhere-to-licenses-when-copying-code-from-cpython/5288/3)). I think it's worth it though.

tests/test_ilastik/test_applets/dataSelection/testOpDataSelection.py
```diff
@@ -897,40 +1017,65 @@ class TestOpDataSelection_DatasetInfo:
         "type": "image",
         "data_type": "uint16",
         "num_channels": 1,
-        "scales": [{"key": "foo", "chunk_sizes": [[1, 1, 1]], "resolution": [""], "size": [1, 1, 1]}],
+        "scales": [{"key": "foo", "chunk_sizes": [[1, 1, 1]], "resolution": [0], "size": [1, 1, 1]}],
```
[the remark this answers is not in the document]
btbest: Looks like it isn't; in this test I wanted to supply the minimum technically necessary inside ilastik. It looks like our precomputed adapter doesn't validate that part of the spec, though it kinda should...

ilastik/applets/dataSelection/datasetDetailedInfoTableView.py
```diff
@@ -224,7 +224,7 @@ def on_combo_selected(self, index):
         if model.is_scale_locked(index.row()):
             message = (
                 "You have already continued in the project with this dataset at the selected scale. "
-                'To inspect another scale, please use "Add New" and add the same remote source as '
+                'To inspect another scale, please use "Add New" and add the same URL as '
```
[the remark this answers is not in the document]
btbest: I looked up the terms and it looks like the URI/URL distinction has [nothing to do with the scheme](https://datatracker.ietf.org/doc/html/rfc3986#section-1.1.3). Going by blogs/posts/comments, it seems to be mainly a matter of taste where "URL" is considered obsolete technical terminology. Googling around for "URI" on the other hand, the term really doesn't seem to exist outside of the depths of technical language. I don't think it's helpful in user-facing text.

The most sensible conclusion to me would be to use `uri` in code and "URL" in GUI strings, and I think that's mostly what I did (will go through again and make sure). Thanks to the already broad usage of "url" in our code though, among others in the project file interface (`UrlDatasetInfo`), we're stuck with both flying around.

Agree on "or path" either way.

lazyflow/utility/io_util/multiscaleStore.py
```diff
@@ -0,0 +1,100 @@
...
+        """Human-readable name of the store, e.g. "OME-Zarr" or "Precomputed"."""
```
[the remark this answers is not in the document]
btbest: If there is any other/nicer way to enforce that subclasses define a class property, I'm all ears. Will add this to the docstring!

lazyflow/utility/io_util/OMEZarrStore.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def get_shape(self, scale_key=DEFAULT_LOWEST_SCALE_KEY):
+        scale_key = scale_key if scale_key else self.lowest_resolution_key
+        return self._scale_data[scale_key]["shape"]
+
+    def request(self, roi: rtype.Roi, scale_key=DEFAULT_LOWEST_SCALE_KEY):
+        scale_key = scale_key if scale_key else self.lowest_resolution_key
+        with Timer() as timer:
+            data = self._scale_data[scale_key]["zarray"][roi.toSlice()]
+            logger.info(f"Requesting roi {roi} from scale {scale_key} took {timer.seconds()*1000} ms.")
```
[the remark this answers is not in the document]
btbest: It's every volumina tile, which I think isn't necessarily identical to chunks. It's indeed a bit spammy, will remove.

lazyflow/utility/io_util/OMEZarrStore.py
btbest: Reversing is indeed for the dropdown box, will move this to the gui and keep everything as ordered in the dataset till then

lazyflow/utility/io_util/OMEZarrStore.py
btbest: I'm also still confused by this 🤷

lazyflow/operators/ioOperators/opInputDataReader.py
btbest: Guess not? 😅 Some of the other methods are named -Volume and some -Fileset, I guess I preferred the latter. Will remove

ilastik/applets/dataSelection/opDataSelection.py
btbest: image

Here's an example: The file was called `photo.jpg.jpg`. The nickname became `photo.jpg`, and the default export location trims that second "extension" including the `_Probabilities` part that is supposed to be in the default export file name. So in fact not fixing this would have been consistent with existing behaviour...

btbest: So should I revert or also fix it for the other `DatasetInfo` subclasses? 🙃

btbest: In general though, we remove one file extension to create the nickname, so removing `.zarr` is canon, replacing `.` with `_` is a new feature.

btbest:
Alright, that was a bunch... some of the most significant:

- We now have a LICENSE.Python file in the repo root because we are redistributing Python code
- I realised that `Multiscale.key` was a relic from an earlier version, and in the current implementation we don't even need a dataclass for this purpose, so I removed it.
- Made the scale dict an OrderedDict and commented regarding the ordering where it seemed sensible. I don't see any way around the ordering as long as we want to start the user off at the smallest scale (which I think we should).
- Naming should be more consistent now and mentions about the logic have been added

and...

- Default export file names: Instead of protecting OME-Zarr dataset nicknames by replacing full stops, I changed opExportSlot to stop removing "file extensions" from nicknames. The exporter should not modify whatever the user might have entered.

Sadly, it looks like this is a bigger issue than expected; names seem to work differently for headless. Will revert this and leave OME-Zarr with the same bug, then move the bug fixing to another PR.

lazyflow/utility/io_util/OMEZarrStore.py
```diff
@@ -0,0 +1,180 @@
...
+from lazyflow.utility.io_util.multiscaleStore import MultiscaleStore, DEFAULT_SCALE_KEY
+
+logger = logging.getLogger(__name__)
+
+OME_ZARR_V_0_4_KWARGS = dict(dimension_separator="/", normalize_keys=False)
+OME_ZARR_V_0_1_KWARGS = dict(dimension_separator=".")
+
+
+def _get_axistags_from_spec(ome_spec: Dict) -> vigra.AxisTags:
+    if "axes" not in ome_spec:
```
Tomaz-Vieira, line 44: if you are covering both cases of the `if` statement, couldn't you phrase them to not use negative logic? That is

```python
if "axes" in ome_spec:  # instead of if NOT
    # positive case
else:
    # negative case
```

lazyflow/utility/io_util/OMEZarrStore.py
```diff
@@ -0,0 +1,180 @@
...
+        # v0.1 and v0.2 did not allow variable axes
+        axis_keys = ["t", "c", "z", "y", "x"]
+    else:
+        ome_axes = ome_spec["axes"]
+        assert isinstance(ome_axes, list), f"Expected axis information to be a list, received: {ome_axes}."
```
Tomaz-Vieira, line 49: Assertions are really there to pick up on bugs in our application, but here you're checking for a condition on a file that could have been malformed. We should foresee that we could encounter malformed files in the wild, and react appropriately to them.

lazyflow/utility/io_util/OMEZarrStore.py
```diff
@@ -0,0 +1,180 @@
...
+        if "name" not in ome_axes[0]:
```
Tomaz-Vieira, line 50: We shouldn't trust the stuff we get out of external files to be well-formed. Even if the spec says it should either be a `List[str]` or a `List[Dict]`, there's no telling what we might actually find there.

So the absolute safest thing to do would be something like:

```python
raw_axes = ome_spec.get("axes")
if not isinstance(raw_axes, list):
    # raise Exception, return Error, whatever
if all(isinstance(a, str) and a in "tczyx" for a in raw_axes):
   parsed_axes: List[str] = raw_axes
else:
    parsed_axes: List[str] = []
    for axis_obj in raw_axes:
        if not isinstance(axis_obj, dict):
            # raise eception, return err, whatever
        axis_name = axis_obj.get("name")
        if not isinstance(axis_name, str) or axis_name not in "xyztc":
            # raise, whatever
        parsed_axes.append(axis_name) 
```

Tomaz-Vieira (commented):
It's great we'll finally support Zarr =)
I've added some "needs more paranoia" comments for now, but I'll get back to have a look later

lazyflow/utility/io_util/OMEZarrStore.py
btbest: This happened because I implemented the branches incrementally as I traced the changes in spec between versions. I like the fact that this ends up ordering the function logic by spec version. But I'm perfectly happy inverting the order by inverting both if-branchings 👍

lazyflow/utility/io_util/OMEZarrStore.py
btbest: The input here is already `jsonschema.validate`d; I guess it would be more consistent to just remove the assert altogether.

lazyflow/utility/io_util/OMEZarrStore.py
btbest: Validating the input isn't this function's job; that's up to `jsonschema.validate`. I think it's worth pointing this out in a comment, but given the scope here, I wouldn't go fully pydantic (or custom types to reflect validated JSON contents)...

k-dominik (approved):
Had another look - all works, can look at 2TB volumes now from my tiny laptop :clap: -  didn't check the code in depth though.

It would be great if you could resolve all open discussions before merging (indicating whether you agreed/disagreed/done something about it).

I gave also gave it another spin and, at least with a "normal" internet connection encountered two issues, that would be nice if they could be avoided (in the future):

* I had _a lot_ of timeout errors bubbling up - here I would wish the client (ilastik) would retry.
* switching scales while there are still requests queued can take ages...

ilastik/applets/dataSelection/opDataSelection.py
btbest: Stuck with `.absolute`, for above reason and because `.resolve` breaks some paths on Windows

lazyflow/utility/io_util/OMEZarrStore.py
btbest: I've left it at `validate` + comments for now, though I suspect I'll be pydantifying this when I implement export to OME-Zarr, which will interact more with the metadata.

btbest:
> It would be great if you could resolve all open discussions before merging (indicating whether you agreed/disagreed/done something about it).

Done

> I gave also gave it another spin and, at least with a "normal" internet connection encountered two issues, that would be nice if they could be avoided (in the future):
> 
>     * I had _a lot_ of timeout errors bubbling up - here I would wish the client (ilastik) would retry.
> 
>     * switching scales while there are still requests queued can take ages...

I've encountered this too, left it for later on the assumption that you wouldn't want to work on a huge dataset on a slow connection anyway. At the latest, by the time you export, you'll have a bad time. But it would be nice to enable earlier steps in the workflow while you're on the go, and leave exporting for the high-speed connection (or maybe to be run headless on the server).

Issue #2174, opened 2020-01-31
Autocontext, second round display problems with float images


k-dominik:
## Describe the bug
When using a `float32` image as input, it will be displayed in black in the second round.

## Expected behavior
Image should be displayed properly (as in first round, too)

## To Reproduce

you can produce some data:

```python
import numpy
data = numpy.random.random((100, 200))
numpy.save("testdata.npy", data)
```

Within the autocontext workflow
1. add the testdata
2. select some features
3. add some annotations (random)
4. go to the next feature selection applet - here the image will be all black.

## Desktop (please complete the following information):
 - ilastik version: 1.3.3pos2
 - OS: Ubuntu 18.04

sdamrich:
I also experienced this issue with
ilastik version: 1.3.3.post3
OS: Ubuntu 18.04
and with `.tif` images whose data type was decoded to `UINT16`, giving the warming
`Warning: TIFFDecoder: no TIFFTAG_SAMPLEFORMAT or TIFFTAG_DATATYPE, guessing pixeltype 'UINT16'`

The issue seems to be fixed in ilastik version 1.4.0b15.

Issue #1410, opened 2017-01-25
Make stage 2 RGB again!


Steffen-Wolf:
Setting: auto context with RGB input data

Problem: In the second stage of the auto-context worklfow, the input data channel dimension becomes larger than 3 and can therefore not be displayed in RGB anymore. This makes it hard to label images, where the colour is the distinguishing feature.

Solution suggestion: Make the original input image layer available in second stage viewer (in addition to the augmented input data)

jjerphan:
Related to #2174.

k-dominik:
closed via #2866

PR #2866, opened 2024-06-04
Autocontext: Connect raw data directly in 2nd (or later) rounds

### LINKED ISSUE #2567: Autocontext workflow on multichannel data
## Describe the bug
I created an Auto-context workflow using multichannel fluorescence microscopy 3D data. After a first successful training round, the second training applet shows a data window with only the first channel.  

## Expected behavior
All raw data channels should be visible in the second training round.

## To Reproduce
Within the ... workflow
1. Add some [3d] multichannel data in the data selection applet, the raw data is visible with RGB lookup tables 
2. Train first pixel classifier
3. Go to second training applet
4. Click on Raw Input-->Adjust Thresholds-->only one channel is visible

## Error message/traceback

If applicable, please add tracebacks with formatting, such as

NA

## Screenshots
If applicable, please add screenshots to help explain your problem.

 - ilastik version: [e.g. 1.4.0]
 - OS: [Windows 10]

k-dominik:
Forgotten to push this...

This connects the raw data to the "overlay" layer in 2nd round of autocontext. This enables seeing the raw-data as is (all channels), as opposed to the raw data (that has preds stacked on top) and thus might not be interpreted correctly anymore.

Especially useful when viewing rgb like data.

fixes #2567 #2174  #1410

### DIFF
--- a/ilastik/applets/pixelClassification/opPixelClassification.py
+++ b/ilastik/applets/pixelClassification/opPixelClassification.py
@@ -71,6 +71,9 @@ class OpPixelClassification(Operator):
     PredictionMasks = InputSlot(
         level=1, optional=True
     )  # Routed to OpClassifierPredict.PredictionMask.  See there for details.
+    AutocontextInput = InputSlot(
+        level=1, optional=True
+    )  # Connected if in "autocontext mode", stacked input + preds from previous round - for display only
 
     LabelInputs = InputSlot(optional=True, level=1)  # Input for providing label data from an external source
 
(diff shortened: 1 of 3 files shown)


btbest:
Can we call this "Original Image"? I pulled the branch, ran, looked for which layer is the new one, and it took me a bit to figure out because overlay is quite a generic term in a stack of image layers. Otherwise very nice 🚀

k-dominik:
Hey @btbest. Thank you for having a look! It made me realize that it was all wrong to add an "Overlay" there for the semantics of the "InputImages" in Pixel Classification - which is supposed to be the raw data ;). So I went on to actually connect it and add a special optional slot for the autocontext input, which is then added to the UI only if connected. This also has the advantage that the "i" shortcut will still work without having to hack around that (modifying the layersstack, :yuk:).

btbest (approved):
I love it!

UX-wise, it would be helpful if we can somehow tell the user what the channels on this layer are. From the comment and what I see, I gather it's the raw data channels + prediction channels. Getting this into the GUI would be overkill I think, but a note in the docs would be nice.

PR #2875, opened 2024-07-03
Oc plugins via entrypoints

k-dominik:
ilastik object features can be added via a (`yapsy`-based) plugin system. The usual way would be to put them into the user plugins folder. However, some plugins will have requirements wrt dependencies. In order to better track those, they should be added as conda package, configured with the appropriate dependencies. This will enable us to include them in the distribution, along with all necessary dependencies.

- [x] add example plugin
- [ ] Add tests.
- n/a Update [user documentation](https://github.com/ilastik/ilastik.github.io).

CC: @oanegros

### DIFF
--- a/ilastik/applets/objectClassification/objectClassificationGui.py
+++ b/ilastik/applets/objectClassification/objectClassificationGui.py
@@ -39,7 +39,7 @@ from functools import partial
 from ilastik.config import cfg as ilastik_config
 from ilastik.utility import bind
 from ilastik.utility.gui import ThreadRouter, threadRouted
-from ilastik.plugins import pluginManager
+from ilastik.plugins.manager import pluginManager
 
 from lazyflow.request import Request, RequestPool
 
--- a/ilastik/applets/objectExtraction/objectExtractionGui.py
+++ b/ilastik/applets/objectExtraction/objectExtractionGui.py
@@ -38,7 +38,7 @@ from functools import partial
 from itertools import chain
 from ilastik.applets.objectExtraction.opObjectExtraction import max_margin
 
-from ilastik.plugins import pluginManager
+from ilastik.plugins.manager import pluginManager
 from ilastik.utility.gui import threadRouted
 from ilastik.utility import log_exception
 from ilastik.config import cfg as ilastik_config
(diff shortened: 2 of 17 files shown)


oanegros:
Will this implement also some of #2744 ?

k-dominik:
Hey @oanegros,

this doesn't really address #2774 - as this would take more time. This is basically the shortest term solution I could come up with to enable conda-based plugins. That doesn't mean that the interface will not change - on the contrary - while doing that I got pretty annoyed with the interface as it is (strong overlap with your gripes).

What I can offer in supporting object feature development, currently would be:

* a cookiecutter for object feature plugins: https://github.com/k-dominik/ilastik-oc-feats-cookiecutter
* which will create barebones plugin, with implementation examples for both `compute_global` and `compute_local` (see here for the repo directly after creation: https://github.com/k-dominik/example_feature_plugin)
* an example implementation with sklearn regionprops (doesn't require `compute_local`):  https://github.com/k-dominik/skimage_ilastik_oc_features . This one you can even test with `conda create -n idev-oc-sckimage -c conda-forge -c pytorch -c kdominik -c ilastik-forge skimage_ilastik_oc_features ilastik` (which should install ilastik, also), then after activating: `ilastik`...

btbest (approved):
Nice! :D I just tried setting up the cookiecutter example, `pip install -e .`, first try failed because `ilastik` is in the requirements, and pip only discovers `ilastik-core`, not `ilastik`... But after removing that requirement, the example Just Works and pops up in the object feature selection.

btbest:
We should add a tutorial for this in the docs...

k-dominik:
> Nice! :D I just tried setting up the cookiecutter example, pip install -e ., first try failed because ilastik is in the requirements, and pip only discovers ilastik-core, not ilastik

yeah, the development setup here assumes, that the environment was create with the supplied `environment.yaml` which includes `ilastik` as a package (currently would have to be sourced from `kdominik` channel, as this PR is not merged yet / no release out with the required code). I would hope "normal" developers wouldn't have to.

:+1: on adding docs :)

oanegros:
On documentation, this workflow should also be highlighted in the contributing guidelines

PR #2876, opened 2024-07-03
Fix circleci

k-dominik:
CircleCI tests started hanging. Root cause could not be determined, but something must have changed that made `xvfb-run` hang. Switched to starting Xvfb manually now and waiting until it's ready in hopes of more reliable runs in the future.

- [x] check what's up with coverage, why 10% less suddenly?! --> accidentally removed UI tests :)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,7 +38,7 @@ jobs:
       TEST_ENV_NAME: test-env
       VOLUMINA_SHOW_3D_WIDGET: 0
     docker:
-    - image: condaforge/mambaforge
+    - image: condaforge/miniforge3
     steps:
     - restore_cache:
         name: restoring dependency cache
@@ -57,8 +57,8 @@ jobs:
         # due to an incompatibility of mamba with conda 4.13, one currently cannot update conda first
         command: >
             conda config --set always_yes yes --set changeps1 no --set channel_priority strict &&
-            mamba update -n base -c conda-forge --all &&
-            mamba install -n base -c conda-forge conda-build boa
+            conda update -n base -c conda-forge --all &&
+            conda install -n base -c conda-forge conda-build boa
     - run:
         name: install apt dependencies
         command: >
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -68,22 +68,32 @@ jobs:
...
-            mamba install --name ${TEST_ENV_NAME} --freeze-installed -c ilastik-forge -c conda-forge mpi4py pytest-cov volumina &&
-            mamba run -n ${TEST_ENV_NAME} pip install -e ilastik
+            conda env create --name ${TEST_ENV_NAME} --file ilastik/dev/environment-dev.yml &&
+            conda install --name ${TEST_ENV_NAME} --freeze-installed -c ilastik-forge -c conda-forge mpi4py pytest-cov volumina &&
+            conda run -n ${TEST_ENV_NAME} pip install -e ilastik
 
+    - run:
+        name: start xvfb
+        command: >
+          Xvfb :42 -screen 0 1024x768x24
```
btbest, line 78: Is 42 a magic number here (and below for `DISPLAY`) or arbitrary?

.circleci/config.yml
```diff
@@ -68,22 +68,32 @@ jobs:
...
+        command: while [ ! -e /tmp/.X11-unix/X42 ]; do sleep 0.1; done
     - run:
         name: run ilastik tests
+        environment:
+            DISPLAY: :42
         command: >
             cd ${ILASTIK_ROOT}/ilastik-meta/ilastik &&
             mkdir test-results &&
-            xvfb-run --server-args="-screen 0 1024x768x24" mamba run -n ${TEST_ENV_NAME} pytest --cov-report=xml --cov=lazyflow --cov=ilastik --run-legacy-gui --junitxml=test-results/junit.xml &&
+            conda run -n ${TEST_ENV_NAME} pytest -v --run-legacy-gui --cov-report=xml --cov=lazyflow --cov=ilastik --junitxml=test-results/junit.xml &&
```
btbest, line 90: Is the leftover `-v` intentional here? Weirdly, I can't actually see a difference in verbosity between the new and some old logs... 🤷

btbest (approved):

.circleci/config.yml
k-dominik: it's random

.circleci/config.yml
k-dominik: I'd want to keep -v as the it's easier to follow test progress.

PR #2878, opened 2024-07-04
Fix circleCI pt2: try to get pytest output immediately

k-dominik:
Apparently the previous "fix" was coincidence. Root cause is using `conda run` to run pytest. This hides the output per default until the process completes. Adding `--live-stream` to `conda run` will print things immediately. That way the test run can also run more than 10 minutes 🥇 .

In the process I separated the steps a bit more, such that coverage upload happens separately.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,14 +81,18 @@ jobs:
         name: wait for xvfb to be ready
         command: while [ ! -e /tmp/.X11-unix/X42 ]; do sleep 0.1; done
     - run:
-        name: run ilastik tests
+        name: setup test paths
+        command: mkdir test-results
+    - run:
+        name: run ${ILASTIK_ROOT}/ilastik-meta/ilastik/test-results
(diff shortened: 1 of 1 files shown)


btbest (approved):
I looked at https://codecov.io/bash out of curiosity. It was a mistake.

k-dominik:
okay, looks like there is another problem. Tests hang after `tests/test_lazyflow/test_request/testRequestPool.py::test_basic`, which I cannot reproduce on MAC. Will try linux box at home.

k-dominik:
update: I ran the test in various combinations at home on my linux box but couldn't reproduce the problem. During debugging I tried to add `-s` flag to pytest in order to see some debug prints in circleci and the tests started passing. I don't want to spend more time on this right now, so I left in the `-s` flag. If this test starts failing again I suggest to remove the timeout logic (apparently it's off on windows and we never saw any hangs).

Issue #2717, opened 2023-06-28
add "clear log" functionality


k-dominik:
Since we never delete the log it can grow to unfortunate sizes. So when I ask users for logs, I usually describe the following procedure: 

1. Start up ilastik
2. go to Settings → Open Log Folder
3. delete the log.txt file
4. restart ilastik
5. open the project in question and look at the log.txt file.

which is a bit cumbersome. I'd suggest to have a new menu entry to clear the log file, no restart required.

emilmelnikov:
Does it make sense to delete old logs automatically? For example, if they are at least 6 months old and occupy more than 100MB.

emilmelnikov:
It just seems to me that clearing log files is not something users should do themselves, and it should be done automatically (obviously, with some reasonable caveats, like always keep recent logs). If someone really wants/needs to do this, they can _Open log file directory_ and delete them as usual files.

k-dominik:
> It just seems to me that clearing log files is not something users should do themselves, and it should be done automatically (obviously, with some reasonable caveats, like always keep recent logs). 

why?

again, I want to shortcut the process from 5 steps to 2 steps.

emilmelnikov:
Sure, we could do that, but I think it would be even better to have 0 steps, not 2 for the common case.

k-dominik:
there will always be one step :D

k-dominik:
not needed anymore  with session logs introduced in #2879 :)

PR #2879, opened 2024-07-05
Easier problem reporting

btbest:
This adds a "Report issue by email" menu action to the Help menu. It's intended to give us something shorter to tell people on the forum than the old "Please find the log with `Settings > Open Log folder`, delete it, reproduce the error, and then email us the log or upload it here on your forum post".

Now we would have to tell them, "Please go to `Help > Report issue by email` and attach the log as described there."

What the button actually does, is open the default email client with a blueprint email that includes version, OS, active workflow and applet, as well as open the log folder. The hope being that this makes it easier for the user to provide the basic info we need. (The reason it opens the log folder instead of pre-attaching the log file, is because that's not supported in `mailto:` links.)

What do you think @k-dominik? Can you think of something better we can do here? Or suggestions for better phrasing of the email template?

Next step could be to add crash detection, with a prompt on the startup after a crash that offers to email the devs about the crash.

### DIFF
--- a/ilastik/ilastik_logging/default_config.py
+++ b/ilastik/ilastik_logging/default_config.py
@@ -1,7 +1,7 @@
 ###############################################################################
 #   ilastik: interactive learning and segmentation toolkit
 #
-#       Copyright (C) 2011-2014, the ilastik developers
+#       Copyright (C) 2011-2024, the ilastik developers
 #                                <team@ilastik.org>
 #
 # This program is free software; you can redistribute it and/or
@@ -20,21 +20,27 @@
 ###############################################################################
 
 import os
+import re
 import logging.config
 import warnings
 from datetime import datetime
 from typing import Optional
+from pathlib import Path
 
 import appdirs
 from . import loggingHelpers
 from ilastik.config import cfg as ilastik_config
 
 SESSION_FILEHANDLER_NAME = "session_file"
+SESSION_LOGFILE_NAME = "log_%Y%m%d_%H%M%S.txt"
+SESSION_LOGFILE_NAME_PATTERN = r"log_\d{8}_\d{6}\.txt"  # For deleting old ones, must correspond to _NAME
 SESSION_LOGFILE_PATH = os.path.join(
-    appdirs.user_log_dir(appname="ilastik", appauthor=False), datetime.now().strftime("log_%Y%m%d_%H%M%S.txt")
+    appdirs.user_log_dir(appname="ilastik", appauthor=False), datetime.now().strftime(SESSION_LOGFILE_NAME)
 )
+KEEP_SESSION_LOGS = 10
 DEFAULT_FILEHANDLER_NAME = "rotating_file"
-DEFAULT_LOGFILE_PATH = os.path.join(appdirs.user_log_dir(appname="ilastik", appauthor=False), "log.txt")
+DEFAULT_LOG_NAME = "log.txt"
+DEFAULT_LOGFILE_PATH = os.path.join(appdirs.user_log_dir(appname="ilastik", appauthor=False), DEFAULT_LOG_NAME)
 
 
 class OutputMode(object):
@@ -61,6 +67,16 @@ def get_session_logfile_path() -> Optional[str]:
         if isinstance(handler, logging.FileHandler) and handler.name == SESSION_FILEHANDLER_NAME:
             return handler.baseFilename
     return None
+
+
+def _delete_old_session_logs(logfile_path: str):
+    """Delete all but the last KEEP_SESSION_LOGS log files."""
+    log_dir = Path(logfile_path).parent
+    if not log_dir.exists():
+        return
+    log_files = sorted([f for f in os.listdir(log_dir) if re.match(SESSION_LOGFILE_NAME_PATTERN, f)], reverse=True)
+    for log_file in log_files[KEEP_SESSION_LOGS:]:
+        os.remove(str(log_dir / log_file))
 
 
 def get_default_config(
@@ -196,6 +212,7 @@ def init(format_prefix="", output_mode=OutputMode.LOGFILE_WITH_CONSOLE_ERRORS, l
 
     if output_mode != OutputMode.CONSOLE:
         os.makedirs(os.path.dirname(logfile_path), exist_ok=True)
+        _delete_old_session_logs(logfile_path)
 
     # Preserve pre-existing handlers
     original_root_handlers = list(logging.getLogger().handlers)
--- a/ilastik/ilastik_logging/default_config.py
+++ b/ilastik/ilastik_logging/default_config.py
@@ -195,7 +195,7 @@ def init(format_prefix="", output_mode=OutputMode.LOGFILE_WITH_CONSOLE_ERRORS, l
         output_mode = OutputMode.CONSOLE
 
     if output_mode != OutputMode.CONSOLE:
-        os.makedirs(os.path.dirname(DEFAULT_LOGFILE_PATH), exist_ok=True)
+        os.makedirs(os.path.dirname(logfile_path), exist_ok=True)
 
     # Preserve pre-existing handlers
     original_root_handlers = list(logging.getLogger().handlers)
--- a/ilastik/ilastik_logging/default_config.py
+++ b/ilastik/ilastik_logging/default_config.py
@@ -18,22 +18,21 @@
 # on the ilastik web site at:
 # 		   http://ilastik.org/license.html
 ###############################################################################
-
+import appdirs
 import os
 import re
 import logging.config
 import warnings
+
 from datetime import datetime
 from typing import Optional
 from pathlib import Path
-
-import appdirs
 from . import loggingHelpers
 from ilastik.config import cfg as ilastik_config
 
 SESSION_FILEHANDLER_NAME = "session_file"
 SESSION_LOGFILE_NAME = "log_%Y%m%d_%H%M%S.txt"
-SESSION_LOGFILE_NAME_PATTERN = r"log_\d{8}_\d{6}\.txt"  # For deleting old ones, must correspond to _NAME
+SESSION_LOGFILE_NAME_PATTERN = re.compile(r"^log_\d{8}_\d{6}\.txt$")  # For deleting old ones, must correspond to _NAME
 SESSION_LOGFILE_PATH = os.path.join(
     appdirs.user_log_dir(appname="ilastik", appauthor=False), datetime.now().strftime(SESSION_LOGFILE_NAME)
 )
@@ -69,14 +68,15 @@ def get_session_logfile_path() -> Optional[str]:
     return None
 
 
-def _delete_old_session_logs(logfile_path: str):
-    """Delete all but the last KEEP_SESSION_LOGS log files."""
-    log_dir = Path(logfile_path).parent
+def _delete_old_session_logs(*, log_dir: Path):
+    """Delete all but the last n=KEEP_SESSION_LOGS log files in log_dir."""
     if not log_dir.exists():
         return
-    log_files = sorted([f for f in os.listdir(log_dir) if re.match(SESSION_LOGFILE_NAME_PATTERN, f)], reverse=True)
+    log_files = sorted(
+        [child for child in log_dir.iterdir() if re.match(SESSION_LOGFILE_NAME_PATTERN, child.name)], reverse=True
+    )
     for log_file in log_files[KEEP_SESSION_LOGS:]:
-        os.remove(str(log_dir / log_file))
+        log_file.unlink()
 
 
 def get_default_config(
@@ -212,7 +212,7 @@ def init(format_prefix="", output_mode=OutputMode.LOGFILE_WITH_CONSOLE_ERRORS, l
 
     if output_mode != OutputMode.CONSOLE:
         os.makedirs(os.path.dirname(logfile_path), exist_ok=True)
-        _delete_old_session_logs(logfile_path)
+        _delete_old_session_logs(log_dir=Path(logfile_path).parent)
 
     # Preserve pre-existing handlers
     original_root_handlers = list(logging.getLogger().handlers)
--- a/ilastik/ilastik_logging/default_config.py
+++ b/ilastik/ilastik_logging/default_config.py
@@ -22,11 +22,18 @@
 import os
 import logging.config
 import warnings
+from datetime import datetime
+from typing import Optional
 
 import appdirs
 from . import loggingHelpers
 from ilastik.config import cfg as ilastik_config
 
+SESSION_FILEHANDLER_NAME = "session_file"
+SESSION_LOGFILE_PATH = os.path.join(
+    appdirs.user_log_dir(appname="ilastik", appauthor=False), datetime.now().strftime("log_%Y%m%d_%H%M%S.txt")
+)
+DEFAULT_FILEHANDLER_NAME = "rotating_file"
 DEFAULT_LOGFILE_PATH = os.path.join(appdirs.user_log_dir(appname="ilastik", appauthor=False), "log.txt")
 
 
@@ -37,10 +44,21 @@ class OutputMode(object):
     LOGFILE_WITH_CONSOLE_ERRORS = 3
 
 
-def get_logfile_path():
+def get_logfile_path() -> Optional[str]:
+    root_handlers = logging.getLogger().handlers
+    file_handlers = [h for h in root_handlers if isinstance(h, logging.FileHandler)]
+    for handler in file_handlers:
+        if handler.name == DEFAULT_FILEHANDLER_NAME:
+            return handler.baseFilename
+    if file_handlers:
+        return file_handlers[0].baseFilename
+    return None
+
+
+def get_session_logfile_path() -> Optional[str]:
     root_handlers = logging.getLogger().handlers
     for handler in root_handlers:
-        if isinstance(handler, logging.FileHandler):
+        if isinstance(handler, logging.FileHandler) and handler.name == SESSION_FILEHANDLER_NAME:
             return handler.baseFilename
     return None
 
@@ -58,16 +76,25 @@ def get_default_config(
         warnings_module_handlers = ["console_warnings_module"]
 
     if output_mode == OutputMode.LOGFILE:
-        root_handlers = ["rotating_file"]
-        warnings_module_handlers = ["rotating_file"]
+        root_handlers = [DEFAULT_FILEHANDLER_NAME]
+        warnings_module_handlers = [DEFAULT_FILEHANDLER_NAME]
+        if logfile_path == DEFAULT_LOGFILE_PATH:
+            root_handlers.append(SESSION_FILEHANDLER_NAME)
+            warnings_module_handlers.append(SESSION_FILEHANDLER_NAME)
 
     if output_mode == OutputMode.BOTH:
-        root_handlers = ["console", "console_warn", "rotating_file"]
-        warnings_module_handlers = ["console_warnings_module", "rotating_file"]
+        root_handlers = ["console", "console_warn", DEFAULT_FILEHANDLER_NAME]
+        warnings_module_handlers = ["console_warnings_module", DEFAULT_FILEHANDLER_NAME]
+        if logfile_path == DEFAULT_LOGFILE_PATH:
+            root_handlers.append(SESSION_FILEHANDLER_NAME)
+            warnings_module_handlers.append(SESSION_FILEHANDLER_NAME)
 
     if output_mode == OutputMode.LOGFILE_WITH_CONSOLE_ERRORS:
-        root_handlers = ["rotating_file", "console_errors_only"]
-        warnings_module_handlers = ["rotating_file"]
+        root_handlers = [DEFAULT_FILEHANDLER_NAME, "console_errors_only"]
+        warnings_module_handlers = [DEFAULT_FILEHANDLER_NAME]
+        if logfile_path == DEFAULT_LOGFILE_PATH:
+            root_handlers.append(SESSION_FILEHANDLER_NAME)
+            warnings_module_handlers.append(SESSION_FILEHANDLER_NAME)
 
     default_log_config = {
         "version": 1,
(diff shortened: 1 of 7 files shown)


k-dominik:
in general I like it. I see two problems though:

* If I understand correctly, the user is asked to attach the complete log file. I recently deleted mine, and it's already 3Mb. Normal users never delete it and it can grow to significant sizes. Also there could be lots of private information that is not related to the issue at all. I don't care to learn a users full directory structure by reading all file names that might get logged in ilastik
* This might be biased, but most people I know didn't configure an email client and to webmail (me included on all my private machines). So having it like this would be annoying for me :)

Alternatively, what could happen is we prepopulate the text pretty much as you've done (but maybe more according to our issue template), then add the _relevant_ part of the project file (maybe even stripping some paths there) to it as a code block and ask users to open an issue, or post on image.sc.

With the config file we could at least find the portion since last startup (`INFO YYY-MM-DD H... app ... Starting ilastik from ....`).

btbest:
Custom report-an-issue dialog it is then I suppose...

k-dominik:
idk, maybe it's just me not seeing the light with the email - might be a good one to discuss at the meeting.

btbest:
Custom report-an-issue dialog here we go 🎉

ilastik/shell/gui/reportIssueDialog.py
```diff
@@ -0,0 +1,94 @@
...
+        button_layout.addWidget(close_button)
+        main_layout.addLayout(button_layout)
+
+        # Set the layout to the dialog
+        self.setLayout(main_layout)
+        self.show()
+
+    def copy_to_clipboard(self):
+        clipboard = QApplication.clipboard()
+        clipboard.setText(self.report_text.toPlainText())
```
emilmelnikov, line 84: Wouldn't it make sense to compress the log text first with zip or gz? Should help a lot with the size issue, and scrolling through 1MB of text in an email is (probably) not what you want. I think it can be done with [`QClipboard.setMimeData`](https://doc.qt.io/qt-6/qclipboard.html#setMimeData).

However, a user has to paste the compressed log as an attachment themselves in this case. Although [some mail clients support attachments in `mailto:`](https://news.ycombinator.com/item?id=24196625), this is not universally supported (tried with GMail, didn't work), and it is sketchy security-wise anyway.

ilastik/shell/gui/reportIssueDialog.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def open_in_email_app(self):
+        body = (
+            "Please describe the issue you are experiencing and adapt the email subject.\n"
+            "\n\n--\nAuto-generated report:\n"
+        )
+        body += self.report_text.toPlainText()
+        subject = "ilastik issue report"
+        mailto_url = f"mailto:team@ilastik.org?subject={subject}&body={quote(body)}"
+        webbrowser.open(mailto_url)
```
emilmelnikov, line 94: What if someone uses a desktop mail client?

ilastik/shell/gui/reportIssueDialog.py
```diff
@@ -0,0 +1,94 @@
...
+        clipboard = QApplication.clipboard()
+        clipboard.setText(self.report_text.toPlainText())
+
+    def open_in_email_app(self):
+        body = (
+            "Please describe the issue you are experiencing and adapt the email subject.\n"
+            "\n\n--\nAuto-generated report:\n"
+        )
+        body += self.report_text.toPlainText()
+        subject = "ilastik issue report"
```
emilmelnikov, line 92: It's probably a good idea to prepend special tag(s) in the subject line that explicitly marks this email as an auto-generated report (e.g. `[auto] [bugreport]`). Filtering will be easier later.

ilastik/shell/gui/reportIssueDialog.py
```diff
@@ -0,0 +1,94 @@
...
+        self.setFixedSize(600, 300)
+
+        main_layout = QVBoxLayout()
+
+        # Create and add the instruction label
+        instruction_label = QLabel()
+        instruction_label.setText(
+            "<p>Sorry you ran into a problem! Please copy the information below, and include it when you "
+            "report your issue.</p>"
+            '<p>The best place to get help is the community forum, using the "ilastik" tag '
```
emilmelnikov, line 44: Maybe encourage user to post their issue to `image.sc` first? With motivational wording like "this way you will help other people and make everyone happy".

ilastik/shell/gui/reportIssueDialog.py
btbest: As Dominik commented above, we're trying to make this easier for the user, and asking them to manually handle email attachments is probably too much to ask.

We could reduce it down to 500KB and add a phrasing asking users to please shorten the log to the relevant parts if they can identify them. But I'd be willing to risk it and just see what our users do with it.

ilastik/shell/gui/reportIssueDialog.py
btbest: On my system, this opens Thunderbird's Compose. On another laptop without a desktop client, it opens Firefox, which then asks me to choose between the email websites I'm logged in to. Actually that's already more than I thought I can hope for.

ilastik/shell/gui/reportIssueDialog.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    log_path = ilastik.ilastik_logging.default_config.get_logfile_path()
+    marker = b"Starting ilastik"
+    text_size_limit = 1048576  # 1 MB
+    file_size = int(os.path.getsize(log_path))
+    with open(log_path, "rb") as log_file:
+        position = max(0, file_size - text_size_limit)
+        log_file.seek(position)
+        log_text = log_file.read()
+    if marker in log_text:
```
Tomaz-Vieira, line 24: Always returning `b"".join(log_text.rpartition(marker)[1:])` would produce the same results of branching on `marker in logtext`

ilastik/shell/gui/reportIssueDialog.py
```diff
@@ -0,0 +1,94 @@
...
+from urllib.parse import quote
+
+
+def _get_log_since_last_startup() -> str:
+    """
+    Reads the log up to maximum 1 MB from the end and returns the text since the last startup, or the entire 1 MB.
+    We don't want people emailing us more than 1MB of log...
+    """
+    log_path = ilastik.ilastik_logging.default_config.get_logfile_path()
+    marker = b"Starting ilastik"
```
Tomaz-Vieira, line 17: This is dangerous; It attributes logical significance to a throwaway `log` statement elsewhere in the code, which could also be disabled via logging configuration. You could make this official, by putting the `"Starting ilastik"` string in a constant somewhere, commenting heavily that such string is not to be touched, and then reference such constant both in the print statement as well as in this function.

Still there's the problem that this string isn't supper unique; Something logging a phrase like `[info] Starting ilastik with such and such parameters will make it do so and so...` would throw this logic off.

And finally, `marker` is looking for the ASCII representation of the string `"Starting ilastik"`. Are we sure windows isn't writing that string in some `UTF16-like` encoding that will escape detection when you do `marker in log_text` ? If that turns out to be the case, it'd be safer (though slower) to first decode the bytes, then search within the string instead of the raw bytes.

Maybe we could try something safer, like recording the initial size of the log file on startup and then just grab everything that comes after that? Though I haven't thought deeply about how this could fail

btbest:
New state:
image

ilastik/shell/gui/reportIssueDialog.py
btbest: +1 for string constant. I am on Windows and the encoding was no issue, but +1 for decoding before comparison. I don't think decoding 500kib of text is worth optimising away here.

I'd much rather risk us screwing up the signal at some point in the future than tie this dialog (which I'm trying not to over-engineer) to a global state defined at startup...

ilastik/shell/gui/reportIssueDialog.py
btbest: New wording is up, screenshot below or see diff :)

ilastik/shell/gui/reportIssueDialog.py
```diff
@@ -0,0 +1,94 @@
...
+    text_size_limit = 1048576  # 1 MB
+    file_size = int(os.path.getsize(log_path))
+    with open(log_path, "rb") as log_file:
+        position = max(0, file_size - text_size_limit)
+        log_file.seek(position)
+        log_text = log_file.read()
```
emilmelnikov: Maybe skip to the next whole line?

```suggestion
        log_text = log_file.read()
        log_text = log_text[log_text.find("\n") + 1:]
```

btbest: I don't really want to cut off the startup line if the log file was empty before start up...

ilastik/shell/gui/reportIssueDialog.py
emilmelnikov: Theoretically we could also gzip the log text and base64 it for transmitting via email. But I'm not a fan of this myself, as it could look weird for users.

Another idea is to just filter out lines with loglevel below INFO.

ilastik/shell/gui/reportIssueDialog.py
emilmelnikov: Good point. Then, if a startup marker is used anyway, it seems that, instead of reading the last 1MB, a better approach overall is to find that marker in `log_file.readlines()` from the end, and only include those in the email. This is of course slower, but emails will be smaller and more precise.

emilmelnikov:
Maybe you've already discussed this possibility, but what if instead of trying to extract the last run from the `log.txt`, just modify logging config so that it also writes to another `lastlog.txt`, which is recreated every time on every session (`mode="w"` instead of default `mode="a"`)?

btbest:
>  another `lastlog.txt`, which is recreated every time on every session (`mode="w"` instead of default `mode="a"`)?

I'd be happy just in general writing one logfile per session...

ilastik/shell/gui/reportIssueDialog.py
Tomaz-Vieira: > than tie this dialog [...] to a global state defined at startup

A file that is being written elsewhere **is** global state. Even more so when there's the expectation of a marker showing up on it.

Another issue is that we don't know how much buffering the logging system might be using. For all we know at this point in the code, it might just be using `io.DEFAULT_BUFFER_SIZE`, so we might miss the last `~8192` bytes of logging that just haven't been flushed down to the file yet.

I was thinking  about the crash detection idea from last meeting. If the temporary file that is created for crash detection is also that run's log file, we could just take the entire contents of that file; no reading from the end, no looking for markers :thinking:

btbest: I don't see a problem here: If the marker doesn't show up, worst case we read 64kib of text. If the logger hasn't flushed in a while, worst case we miss some log. Should I add an explicitly flush of the RotatingFileHandler before we read the log file?

The only real way I see to make anything here more elegant would be one-log-per-session, but I would leave that for another time.

ilastik/shell/gui/reportIssueDialog.py
```diff
@@ -0,0 +1,114 @@
...
+    Reads the log up to maximum `text_size_limit` from the end (64 KiB).
+    Returns the text since the last startup plus `pre_marker_padding`, or the entire text if the marker is not found.
+    We don't want people posting or emailing us huge amounts of log...
+    """
+    log_path = ilastik_logging.default_config.get_logfile_path()
+    file_size = int(os.path.getsize(log_path))
+    with open(log_path, "rb") as log_file:
+        position = max(0, file_size - text_size_limit)
+        log_file.seek(position)
+        log_text = log_file.read().decode()
```
Tomaz-Vieira, line 44: Because seek/read/etc deals in bytes, this can break if `position` ends up splitting a unicode codepoint:

```python
f = open("/tmp/blas.txt", "w+")
f.write("á")
f.close()

fr = open("/tmp/blas.txt", "rb")
import io
fr.seek(-1, io.SEEK_END)
fr.read().decode()
#UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa1 in position 0: invalid start byte
```

Even if we never write non-ascii in our strings, surely some file paths in the user's machine might have them.

ilastik/shell/gui/reportIssueDialog.py
Tomaz-Vieira: > Should I add an explicitly flush of the RotatingFileHandler [...]?

That would help. Still, the log could have recently rotated, which would cause other problems.

The way I see it, our bug reporter can't be itself buggy. If we end up deciding that we can't trust the submitted logs and would always ask the user to send the entire thing to us, then we waste the effort of this feature. Some still standing problems with this approach are:

- The log rotated and chopped the log;
- The extra stuff we get before the marker contains another marker which confuses our human parsing;
- someone somewhere logs something that is detected as a marker;
- reconfiguring the logger invalidates our assumptions;
- The clipped log at the start could hinder our anonymizing efforts by making strings less path-like

Some of these could be ameliorated by reading the entire things as a string instead of bytes (I've pointed byte encoding issue elsewhere) , and maybe doing something more line-oriented. The log rotation and reconfiguration could still get us, though.

btbest: I think these should all be addressed now with the session log. The downside is that the only way to implement this without breaking whatever people might already be doing with our logs, is to also keep the old config and double log everything. I'd be happy to go more aggressive on the log file limit for the session log (i.e. rotate earlier, keep fewer chunks), since the config for the main file (20MB, 5 chunks) guarantees it can't grow above 100MB in total, while for session logs there is no limit to their number. This would make https://github.com/ilastik/ilastik/issues/2717 a bit more urgent.

> The log rotated and chopped the log;

The most recent chunk retains the "main" log file name, so that gets read in this case. Not sure how this could be an issue.

> The extra stuff we get before the marker contains another marker which confuses our human parsing;

Gone

> someone somewhere logs something that is detected as a marker;

Gone

> reconfiguring the logger invalidates our assumptions;

The user gets to put their own log path through the `logfile` cmdline parameter. If that's the case, no session log. I think that's the most sensible since the user is more likely to be running things on the files in that folder if they are using a custom one. The dialog falls back to not including any log in that case, and I added an "open log folder" button instead. If they are somehow messing with the logging config beyond that, they're advanced enough not to need our help (or handholding on how to contact us)...

> The clipped log at the start could hinder our anonymizing efforts by making strings less path-like

Gone. I figured this would also be a potential issue if logs can be rotated mid-message, but as expected `logging` is smart enough. If you set the rotation size to 1, it creates one file per message.

ilastik/shell/gui/reportIssueDialog.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def _get_current_session_log_or_hint() -> str:
+    log_path = ilastik_logging.get_session_logfile_path()
+    if not log_path:
+        return 'Could not find session log file. Please use the "Open log folder" button to locate the log.'
+    with open(log_path, "r") as log_file:
+        log_text = log_file.read()
+    if len(log_text) > LOG_LENGTH_THRESHOLD:
+        return 'Session log is too long to include here. Please use the "Open log folder" button to locate the log.'
```
k-dominik, line 45:
```suggestion
        return 'Session log is too long to include here. Please use the "Open log folder" button to locate the log file and attach it to your forum thread or email.'
```

ilastik/shell/gui/reportIssueDialog.py
```diff
@@ -0,0 +1,120 @@
...
+        webbrowser.open(FORUM_URI)
+
+    def open_in_email_app(self):
+        subject = "[Bugreport] --Summarise your issue here--"
+        body = (
+            "Please describe the issue you are experiencing and adapt the email subject.\n"
+            "\n\n--\nAuto-generated report:\n"
+        )
+        body += self.report_text.toPlainText()
+        mailto_url = f"mailto:{TEAM_EMAIL}?subject={subject}&body={quote(body)}"
```
k-dominik, line 119:
```suggestion
        mailto_url = f"mailto:{TEAM_EMAIL}?subject={subject}&body={body}"
```

for me removing the `quote` call would open the message in the email app as expected. But I guess it's probably important to do in order to encode as a url....

ilastik/ilastik_logging/default_config.py
```diff
@@ -129,6 +156,15 @@ def get_default_config(
...
             },
+            SESSION_FILEHANDLER_NAME: {
+                "level": "DEBUG",
+                "class": "logging.handlers.RotatingFileHandler",
+                "filename": SESSION_LOGFILE_PATH,
+                "maxBytes": 50e6,  # 20 MB
+                "backupCount": 5,
+                "formatter": "verbose",
+                "encoding": "utf-8",
+            },
```
k-dominik, line 167: naive question: is the intention to keep all session logs?

ilastik/shell/gui/reportIssueDialog.py
```diff
@@ -0,0 +1,120 @@
...
+    if not log_path:
+        return 'Could not find session log file. Please use the "Open log folder" button to locate the log.'
+    with open(log_path, "r") as log_file:
+        log_text = log_file.read()
+    if len(log_text) > LOG_LENGTH_THRESHOLD:
+        return 'Session log is too long to include here. Please use the "Open log folder" button to locate the log.'
+    return log_text
+
+
+class ReportIssueDialog(QDialog):
```
k-dominik, line 49: I'd prefer to have this dialog modal - without it, the user can have the dialog open, do more in ilastik and send off an outdated log...

k-dominik (commented):
I like it :)

fyi, with apple mail I currently get this if I click on open in email app:

```
Please%20describe%20the%20issue%20you%20are%20experiencing%20and%20adapt%20the%20email%20subject.%0A%0A%0A--%0AAuto-generated%20report%3A%0Ailastik%20version%3A%201.4.1b19.dev6%2Bg0d9e9325c.d20240730%0AOS%3A%20macOS-14.5-arm64-arm-64bit%0ASession%20log%20%28please%20shorten%20to%20relevant%20parts%20if%20possible%29%3A%0A%60%60%60%0AINFO%202024-07-30%2013%3A55%3A07%2C626%20app%2046885%208637090816%20Using%20tiktorch%20executable%3A%20%5B%27%28file%20path%20masked%29%20%27-m%27%2C%20%27tiktorch.server%27%5D%0AINFO%202024-07-30%2013%3A55%3A07%2C626%20app%2046885%208637090816%20config%20file%20location%3A%20%28file%20path%20masked%29%0AINFO%202024-07-30%2013%3A55%3A20%2C019%20app%2046885%208637090816%20Starting%20ilastik%20from%20%28file%20path%20masked%29%0AINFO%202024-07-30%2013%3A55%3A20%2C380%20app%2046885%208637090816%20Resetting%20lazyflow%20thread%20pool%20with%2010%20threads.%0AWARNING%202024-07-30%2013%3A55%3A23%2C823%20opConservationTracking%2046885%208637090816%20Could%20not%20find%20any%20ILP%20solver%0AWARNING%202024-07-30%2013%3A55%3A23%2C828%20opStructuredTracking%2046885%208637090816%20Could%20not%20find%20any%20ILP%20solver%0AWARNING%202024-07-30%2013%3A55%3A23%2C828%20structuredTrackingWorkflow%2046885%208637090816%20Could%20not%20find%20any%20learning%20solver.%20Tracking%20will%20use%20flow-based%20solver%20%28DPCT%29.%20Learning%20for%20tracking%20will%20be%20disabled%21%0A%0A%60%60%60 
```

not ideal. For the forum it works perfectly. Also copy-pasting that into the email app works.

Anonymization of paths is cool and while I probably would like to see filenames/extensions sometimes, it has too many drawbacks (as we discussed already in person).

ilastik/shell/gui/reportIssueDialog.py
btbest: Can you check again? I suspect it actually wasn't too much, but rather not enough quoting.

ilastik/ilastik_logging/default_config.py
btbest: I was going to leave it for this PR and deal with it in the context of https://github.com/ilastik/ilastik/issues/2717 soon-ish

ilastik/shell/gui/reportIssueDialog.py
btbest: I chose to have it as a floating dialog because I wanted to be able to use the "show log folder" menu option. Now that there's a button for that here, I'd be ok making it a modal.

ilastik/ilastik_logging/default_config.py
k-dominik: I'm asking as I already have 10+ session logs...

ilastik/shell/gui/reportIssueDialog.py
k-dominik: works now 😻

k-dominik:
open log folder probably should not be the default button:

image

I also found that you cannot escape keyboard focus on the textedit.

ilastik/ilastik_logging/default_config.py
btbest: How many session logs do you reckon we should keep? 10?

btbest:
> I also found that you cannot escape keyboard focus on the textedit.

Managed to fix this one

> open log folder probably should not be the default button:
> Edit: now looking at it ... and I know that's super minor and not a problem really, but the alignment of the buttons is a tiny bit unsettling. I mean it does make logical sense that the button aligns with the textedit outer edge. Still having it there, slightly off the button below, and also at a different width...

On Win11 the default button doesn't get a special style, and the misalignment also doesn't exist, so I can't tell if I fixed it or not...

Also now deleting all but the last 10 session logs on startup. I wrote a test for the deleting-function, and then removed the test because creating test files etc. turned out almost more complex than the code it's testing... can push that if wanted.

btbest:
Alright, feedback on session log deleting should be done. If I can get another quick check how the buttons look on Mac now (@k-dominik), I'd be happy to merge 🚀

k-dominik (approved):
On mac the buttons are now also aligned, the logs get properly deleted - lgtm :)

btbest:
For the record, one piece of information that I would've liked to include but couldn't find a sensible way to extract: Is the user running a "-gpu" version? This would have been auto-reported inside the "starting ilastik from..." log, but masking that file path is the right thing to do for privacy reasons. Unfortunately I couldn't find a straightforward way to get that information, but adding this should be easy if I just overlooked something.

PR #2880, opened 2024-07-05
Skip requestLock test on CircleCI

btbest:
does removing the timeout actually fix it?

edit: It doesn't, guess we'll have to skip the test for now


k-dominik (approved):
sad, but it's too annoying, I agree.

PR #2881, opened 2024-07-08
Fix deploy pipeline

btbest:
may need to replace mamba install with conda install

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,9 +1,8 @@
 name: deploy
 
-on:  
-  push:
-    tags:
-      - '1.*'
+on:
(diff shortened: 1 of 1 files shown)


btbest:
Great, so on a PR the pipeline is completely fine. I diff-ed the logs of the failing `deploy` on `1.4.1b16` and the one on this PR, and they are 100% identical up to the point where the one on `1.4.1b16` just fails with `Currently, only install, create, list, search, run, info, clean, remove, update, repoquery, activate and deactivate are supported through mamba.`, while the run here just continues fine. Absolutely useless. I guess I'll just have to try replacing mamba with conda directly on main.

PR #2882, opened 2024-07-08
Add wait cursor when tiktorch server is started

### LINKED ISSUE #2721: Indicate startup of tiktorch server in local workflow
Especially after increasing the timeout to 60s in #2720 we should give some sort of feedback/indication about what is going on whilst starting the workflow - currently the user is forced to stare at the startup screen - on a machine where this took more than a few seconds I was concerned that I had done something wrong...

thodkatz:
Closes #2721

I did some refactoring as well, as I was navigating through the code.

I could potentially add a little bit more logic to make the cursor busy only for the neural network workflows, where the tiktorch server is needed, but I skipped it, since it is a very lightweight process, and it won't cause harm even for the simpler, faster workflows that don't introduce latency (shoud maybe do it though?).

Manually tested by introducing latency at the creation of the project manager with `time.sleep()`:

```python
 import time
 time.sleep(2)
 self.projectManager = ProjectManager(self, workflow_class, self._workflow_cmdline_args, project_creation_args)
```

### DIFF
--- a/ilastik/shell/gui/ilastikShell.py
+++ b/ilastik/shell/gui/ilastikShell.py
@@ -29,6 +29,7 @@ import numbers
 import platform
 import threading
 import warnings
+from typing import List, Type, Optional
 
 # SciPy
 import numpy
@@ -74,7 +75,7 @@ from volumina.utility import preferences, ShortcutManagerDlg, ShortcutManager
 
 # ilastik
 import ilastik.ilastik_logging.default_config
-from ilastik.workflow import getAvailableWorkflows, getWorkflowFromName
+from ilastik.workflow import getAvailableWorkflows, getWorkflowFromName, Workflow
 from ilastik.utility import bind, log_exception
 from ilastik.utility.gui import ThunkEventHandler, ThreadRouter, threadRouted
 from ilastik.exceptions import UserAbort
@@ -507,8 +508,7 @@ class IlastikShell(QMainWindow):
     def loadWorkflow(self, workflow_class):
         self.onNewProjectActionTriggered(workflow_class)
 
-    def getWorkflow(self, w=None):
-
+    def getWorkflow(self, w: Optional[str] = None) -> Type[Workflow]:
         listOfItems = [workflowDisplayName for _, __, workflowDisplayName in getAvailableWorkflows()]
         if w is not None and w in listOfItems:
             cur = listOfItems.index(w)
@@ -578,7 +578,6 @@ class IlastikShell(QMainWindow):
         return (menu, shellActions)
 
     def setupOpenFileButtons(self):
-
         for b in self.openFileButtons:
             b.close()
             b.deleteLater()
@@ -1638,7 +1637,6 @@ class IlastikShell(QMainWindow):
                 workflow_cmdline_args=self._workflow_cmdline_args,
                 project_creation_args=project_creation_args,
             )
-
         except Exception as e:
             msg = "Could not load project file.\n" + str(e)
             log_exception(logger, msg)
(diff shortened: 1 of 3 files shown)


ilastik/shell/gui/ilastikShell.py
```diff
@@ -1764,7 +1856,11 @@ def ensureNoCurrentProject(self, assertClean=False):
                 message += "(Unsaved changes in: {})".format(", ".join(dirtyApplets))
                 buttons = QMessageBox.Yes | QMessageBox.Cancel
                 response = QMessageBox.warning(
-                    self, "Discard unsaved changes?", message, buttons, defaultButton=QMessageBox.Cancel
+                    self,
+                    "Discard unsaved changes?",
+                    message,
+                    buttons,
+                    defaultButton=QMessageBox.Cancel,
```
k-dominik, line 1863: not sure this really comes from the pre-commit hook. At least for me black would leave the formatting alone in the pre-commit hook. Might be good to revisit your settings there (I guess key here is the trailing comma, do you have any other plugin that might have added it automagically?)

k-dominik (commented):
I gave the changes a spin on windows and it now I feel more comfortable starting the NN workflow :). Also a nice and simple change.

Additionally

* great that you simplified classmethods to staticmethods where its applicable.
* Better and more type annotations are always appreciated.
* readability is also improved by removing the try-else construct (imo)

the changes were a bit noisy due to some additional formatting changes (mainly something seems to be adding trailing commas and black will rearrange), which would be great if you could double check (revert, if this indeed is not default black behavior of the version we're currently using in pre-commit) before we merge this.

ilastik/shell/gui/ilastikShell.py
thodkatz: yep, I am sorry for the formatting, I messed up a bit, but now I don't seem to have this issue. I will revert it back as well.

ilastik/shell/projectManager.py
```diff
@@ -88,13 +86,42 @@ class SaveError(RuntimeError):
 
         pass
 
-    #########################
-    ## Class methods
-    #########################
+    def __init__(
```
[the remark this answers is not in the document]
thodkatz: as suggested I have moved the property [_applets](https://github.com/ilastik/ilastik/pull/2882/commits/36d04f5fe5698b3b9d0159103b75e64383ed1640#diff-f0d57b4cfd246289ff1bacc690e409bb1240a98982bf890cacdd983021ecb420L425) on top of the class, followed by the `__init__` :)

ilastik/workflow.py
```diff
@@ -82,6 +78,10 @@ def workflowDescription(self):
     def defaultAppletIndex(self):
         return 0
 
+    @property
```
thodkatz, line 81: same as https://github.com/ilastik/ilastik/pull/2882/commits/36d04f5fe5698b3b9d0159103b75e64383ed1640#r1695468302

thodkatz (commented):
Thanks a lot @k-dominik for the review, and sorry for the formatting inconvenience :)

I have attempted to resolve the conflicts, by removing the incorrect formating, and following the principle of keeping properties before the __init__ method.

btbest:
I'm a bit confused. The clip below shows what loading a project looks like for me in ilastik-1.4.1b18. I don't see any change in the cursor in this PR's version, and it wouldn't surprise me if `Qt.WaitCursor` is exactly this loading-circle cursor.
https://github.com/user-attachments/assets/2a46889f-91e5-4bf2-9c18-df35bf2c292f

I'm still happy with the refactoring, and of course happy with the cursor change if this helps on some other systems, but I guess on Win11 it's not strictly necessary 😅

k-dominik:
I don't get the wait cursor on windows without this PR 🤷

k-dominik (approved):
lgtm

PR #2883, opened 2024-07-09
Fixes for 1.4.1b17 release

btbest:
This PR originally was about fixing a SyntaxWarning message (see below), which I was going to just leave here until b18. Then discovered that `zarr.storage.FSStore` can't connect to remote servers in b17 on Linux and Mac due to issues with the SSL cert path, so b17 is unreleasable. Pushing the fix for that here instead and we'll just take the SyntaxWarning fixes along.

Original description: 
image

This is not a great UX... (not to mention one of these was clearly a typo)

### DIFF
--- a/ilastik_scripts/ilastik_startup.py
+++ b/ilastik_scripts/ilastik_startup.py
@@ -1,11 +1,16 @@
 import ctypes.util
+import logging
 import os
 import pathlib
 import platform
 import shlex
+import ssl
 import sys
 import warnings
-from typing import Iterable, Mapping, Sequence, Tuple, Union
+from typing import Iterable, Mapping, Sequence, Tuple, Union, Optional
+
+
+logger = logging.getLogger(__name__)
 
 
 def _env_list(name: str, sep: str = os.pathsep) -> Iterable[str]:
@@ -102,7 +107,7 @@ def path_setup():
             _clean_paths(ilastik_root)
         else:
             warnings.warn(
-                f"--clean_paths argument only supported on windows and should only be used in the binary distribution. Skipping path cleanup."
+                "--clean_paths argument only supported on windows and should only be used in the binary distribution. Skipping path cleanup."
             )
 
     # Allow to start-up by double-clicking a project file.
@@ -138,9 +143,59 @@ def fix_macos() -> None:
     ctypes.util.find_library = find_library
 
 
+def fix_ssl() -> None:
+    """
+    Paths to cert.pem and the CA certificates dir are hard-coded into openssl during the build process
+    of our installer packages.
+    It has some special logic for Windows, and there the paths don't seem to be an issue.
+    On Linux and macOS, the build machine's path is written (/opt/conda/envs/ilastik-release/ssl/),
+    which does not exist on the end user's machine.
+    Being unable to find the cert files means aiohttp cannot connect via https. Currently only
+    zarr.storage.FSStore depends on this.
+    We override the broken paths using the env vars that openssl listens to. Better would be if we can
+    find a way to write the correct paths during build.
+    """
+    if platform.system() == "Windows":
+        return
+
+    ssl_paths = ssl.get_default_verify_paths()
+    if os.path.isfile(ssl_paths.openssl_cafile):
+        # All good then I guess
+        return
+
+    def check_parent_for_dir_recursive(path: pathlib.Path, dir_name: str) -> Optional[pathlib.Path]:
+        if path.is_dir() and any(p.name == dir_name for p in path.iterdir()):
+            return path
+        if path.parent == path:
+            raise RecursionError(f"Could not find {dir_name} directory")
+        return check_parent_for_dir_recursive(path.parent, dir_name)
+
+    ssl_parent = None
+    try:
+        ssl_parent = check_parent_for_dir_recursive(pathlib.Path(__file__).parent, "ssl")
+    except RecursionError:
+        try:
+            # Try from PWD and OLDPWD env variables
+            # on macOS: os.environ["PWD"] = {app_root}/Contents/MacOS; os.environ["OLDPWD"] = {app_root}/Contents
+            # on Linux: os.environ["PWD"] = {tar_root}; os.environ["OLDPWD"] = {tar_root}/bin
+            app_root_or_subdir = pathlib.Path(os.environ.get("PWD", "") or os.environ.get("OLDPWD", ""))
+            ssl_parent = check_parent_for_dir_recursive(app_root_or_subdir, "ssl")
+        except RecursionError:
+            logger.warning(
+                "Could not find SSL certificate directory, may not be able to connect to web servers via https. "
+                "If you know it, you can explicitly set the environment variables SSL_CERT_FILE and SSL_CERT_DIR. "
+                "SSL_CERT_FILE should point to /your/path/ssl/cert.pem and SSL_CERT_DIR to /your/path/ssl/certs"
+            )
+    if ssl_parent:
+        logger.info(f"Found SSL dir in {ssl_parent}, writing to env vars SSL_CERT_FILE / SSL_CERT_DIR")
+        os.environ["SSL_CERT_FILE"] = str(ssl_parent / "ssl" / "cert.pem")
+        os.environ["SSL_CERT_DIR"] = str(ssl_parent / "ssl" / "certs")
+
+
 def main():
     path_setup()
     fix_macos()
+    fix_ssl()
 
     from ilastik.__main__ import main
 
--- a/ilastik_scripts/ilastik_startup.py
+++ b/ilastik_scripts/ilastik_startup.py
@@ -7,7 +7,7 @@ import shlex
 import ssl
 import sys
 import warnings
-from typing import Iterable, Mapping, Sequence, Tuple, Union, Optional
+from typing import Iterable, Mapping, Sequence, Tuple, Union
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


ilastik_scripts/ilastik_startup.py
```diff
@@ -138,9 +143,59 @@ def find_library(name):
...
+
+    ssl_parent = None
+    try:
+        ssl_parent = check_parent_for_dir_recursive(pathlib.Path(__file__).parent, "ssl")
+    except RecursionError:
+        try:
+            # Try from PWD and OLDPWD env variables
+            # on macOS: os.environ["PWD"] = {app_root}/Contents/MacOS; os.environ["OLDPWD"] = {app_root}/Contents
+            # on Linux: os.environ["PWD"] = {tar_root}; os.environ["OLDPWD"] = {tar_root}/bin
+            app_root_or_subdir = pathlib.Path(os.environ.get("PWD", "") or os.environ.get("OLDPWD", ""))
```
Tomaz-Vieira, line 181: why not use `pathlib.Path.cwd()` ?

Also, do we have much to gain from looking from `CWD`? Will we ever find anything there? Shouldn't we look in the places where the certs could actually be, like `/etc/pki/tls/cert.pem` or something like it?

ilastik_scripts/ilastik_startup.py
```diff
@@ -138,9 +143,59 @@ def find_library(name):
...
+            ssl_parent = check_parent_for_dir_recursive(app_root_or_subdir, "ssl")
+        except RecursionError:
+            logger.warning(
+                "Could not find SSL certificate directory, may not be able to connect to web servers via https. "
+                "If you know it, you can explicitly set the environment variables SSL_CERT_FILE and SSL_CERT_DIR. "
+                "SSL_CERT_FILE should point to /your/path/ssl/cert.pem and SSL_CERT_DIR to /your/path/ssl/certs"
+            )
+    if ssl_parent:
+        logger.info(f"Found SSL dir in {ssl_parent}, writing to env vars SSL_CERT_FILE / SSL_CERT_DIR")
+        os.environ["SSL_CERT_FILE"] = str(ssl_parent / "ssl" / "cert.pem")
```
Tomaz-Vieira, line 191: Shouldn't we check that those files are actually there inside the `ssl` folder instead of just assuming they're there?

tests/test_startup.py
```diff
@@ -0,0 +1,11 @@
...
+import platform
+import pytest
+
+from ilastik_scripts.ilastik_startup import fix_ssl
+
+
+@pytest.mark.skipif(platform.system() == "Windows", reason="Windows does not need this fix")
+def test_startup_ssl_paths():
+    fix_ssl()
+    assert os.path.isfile(os.environ["SSL_CERT_FILE"])
```
Tomaz-Vieira: Maybe you could check first that `ssl.get_default_verify_paths()` isn't finding the certs and that it _does_ find it after calling `fix_ssl()`

ilastik_scripts/ilastik_startup.py
```diff
@@ -138,9 +143,59 @@ def find_library(name):
...
+    ssl_paths = ssl.get_default_verify_paths()
+    if os.path.isfile(ssl_paths.openssl_cafile):
+        # All good then I guess
+        return
+
+    def check_parent_for_dir_recursive(path: pathlib.Path, dir_name: str) -> Optional[pathlib.Path]:
+        if path.is_dir() and any(p.name == dir_name for p in path.iterdir()):
+            return path
+        if path.parent == path:
+            raise RecursionError(f"Could not find {dir_name} directory")
```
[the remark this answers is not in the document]
btbest: This short-cuts to the actual RecursionError that I was getting before adding this if-clause to exit the recursion. In that state, the function just keeps recursing because when `path` is the FS root, `path.parent` is also the FS root.

I had it returning None in the previous commit and found the code unreadable that way; the try-except phrasing seemed more expressive.

tests/test_startup.py
btbest: It still won't find it after setting SSL_CERT_FILE; the default_verify_paths remain unchanged. I should 1) add a check that SSL_CERT_FILE isn't already set in `ssl_fix`, and 2) for the test I've added the alternative permitted outcomes (so hopefully they should pass CI now)

ilastik_scripts/ilastik_startup.py
btbest: Honestly didn't think of `cwd`, I was just collecting all env variables I have on Mac and Linux and seeing where I can get a path to the package root 😅 Will have to check what `cwd` actually returns on Linux and Mac when running from the package.

Instead of guessing where on the user's system their system certs might be, I was going for the certs that the conda env comes with. In the Linux tarball, they're in `$tar_root/ssl` and on Mac they're in `$app_root/Contents/ssl` (iirc, will have to check again)

ilastik_scripts/ilastik_startup.py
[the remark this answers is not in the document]
emilmelnikov: Wouldn't something like this be easier to work with instead of recursion and nested exception handling?

```python
# ... explain why PWD/OLDPWD is here ...
candidate_ssl_parents = [
    *pathlib.Path(__file__).parents,
    os.environ.get("PWD"),
    os.environ.get("OLDPWD"),
]
candidate_ssl_parents = list(map(pathlib.Path, filter(None, parents)))

for parent in parents:
    if any(
        (p / "ssl/cert.pem").is_file() and (p / "ssl/certs").is_dir()
        for p in parent.iterdir()
    ):
        logger.info(...)
        ...
        break
else:
    logger.warning(...)
```

btbest:
We are depending on certifi and it's not helping :/

emilmelnikov:
> We are depending on certifi and it's not helping :/

I mean just set `os.environ["SSL_CERT_FILE"] = certifi.where()` explicitly.

btbest:
> > We are depending on certifi and it's not helping :/
> 
> I mean just set `os.environ["SSL_CERT_FILE"] = certifi.where()` explicitly.

Right, so we can get a pem file this way. Not sure if we also need the certs dir or not. I also noticed that we have certifi somewhere upstream, not directly in ilastik so far. I don't see any problem putting it as a direct dep though.

emilmelnikov:
> > I mean just set `os.environ["SSL_CERT_FILE"] = certifi.where()` explicitly.
> 
> Not sure if we also need the certs dir or not..?

`certifi` comes with all root certs concatenated into a single file, it is probably enough (but don't quote me on that :)

Tomaz-Vieira:
> I mean just set `os.environ["SSL_CERT_FILE"] = certifi.where()` explicitly.

That's interesting... it would make it so that `aiohttp` behaves like `requests`, which is nice. Though come to think of it, what about the opposite, that is, forcing both `aiohttp` and `requests` to always use the system's certificates? I wonder if an old version of ilastik could eventually start breaking because it's relying on old certificates from our release, whether they are provided by `ssl` or by `certifi`

emilmelnikov:
> I wonder if an old version of ilastik could eventually start breaking because it's relying on old certificates from our release, whether they are provided by `ssl` or by `certifi`

Unfortunately, lots of things can break in TLS anyway (old unsupported OS, expired certs on the server, revoked root cert, a hypothetical future TLS v2.0 that drops backwards compatibility). Still, perhaps it's a good idea to set `SSL_CERT_FILE` only if it's not already set, just in case if a user wants to use their own certs.

Whether to use system vs. app-provided certs is a long-standing debate, and as far as I know without any definite answers (at least for desktop apps).

ilastik_scripts/ilastik_startup.py
```diff
@@ -138,9 +143,110 @@ def find_library(name):
     ctypes.util.find_library = find_library
 
 
+def check_parent_for_dir_recursive(path: pathlib.Path, dir_name: str) -> Optional[pathlib.Path]:
+    if path.is_dir() and any(p.name == dir_name for p in path.iterdir()):
+        return path
```
Tomaz-Vieira, line 148: This would also return a `Path` if it came across a file (not a dir) named `dir_name`.

ilastik_scripts/ilastik_startup.py
```diff
@@ -138,9 +143,110 @@ def find_library(name):
     ctypes.util.find_library = find_library
 
 
+def check_parent_for_dir_recursive(path: pathlib.Path, dir_name: str) -> Optional[pathlib.Path]:
```
Tomaz-Vieira, line 146: Would it not be clearer if this function produced the target directory instead of the _parent_ of such directory? Then if the caller wants the parent, they cal always call `.parent` on the result. To that point, because this function produces the parent and not the target dir itself, the "ssl" component needs to be added back to the result to produce the correct paths to the certificates.

ilastik_scripts/ilastik_startup.py
btbest: Decided to remove the cwd attempt altogether because on Linux cwd/PWD could literally be anything (wherever the user called run_ilastik.sh from); on Mac it would be quite likely to be the app directory, but I would end up hard-coding all sorts of relative path shenanigans.
I'm verifying now that `__file__` actually is inside `ilastik_release` on Mac, i.e. that we will actually discover `ilastik_release/ssl` by traversing up the tree

ilastik_scripts/ilastik_startup.py
btbest: In fact I think not having any function for this is clearer, since apparently it can be solved with a simple iteration :)

btbest:
Alright, I think we should be ready to merge at this point. I'll let CI run just to be sure, and then build b18 once it's merged.

PR #2889, opened 2024-07-22
Pixel Classification: Fix pop error

### LINKED ISSUE #1244: pop index out of range error in pixelclassification workflow
There is an issue with dropping the channel axis during prediction export in the pixelclassification applet.

How to replicate:
During prediction export if the axis order in the image export options is set from e.g. yxc to yx (dropping the channel) after closing the options menu saving the project leads to an "pop index out of range IndexError". If the project is closed in that condition it can't be opened again.

The only work around to save the labels was to manually add an empty channel axis to the hdf5 project file. After that the project can be opened again.

Image of the error:
error

Cheers, Adrian

btbest:
The default selection is "Probabilities". When the user switches to "Simple Segmentation", the OpMultiArraySlicer2 created for Probabilities is still connected to `opLane.ImageToExport`. If the user now changes export axis order to remove "c", this change bubbles down to the orphaned OpMultiArraySlicer2, which cannot handle it as it has two channels. 

Solution: Keep a ref to the op and delete it

Reproduce/Manual test:
1. Pixel Classification project with any image (mitocheck works), export applet
2. Change output source to Simple Segmentation
3. Open Image export settings
4. Check Reorder axes
5. Change `yxc` to `yx`
6. Click OK
7. main crashes; PR works

Fixes #1244

## Checklist

- [ ? ] Add tests --- Not sure? Generally of course nice to have regression tests, but this is Gui code...

### DIFF
--- a/ilastik/applets/pixelClassification/pixelClassificationDataExportGui.py
+++ b/ilastik/applets/pixelClassification/pixelClassificationDataExportGui.py
@@ -1,7 +1,7 @@
 ###############################################################################
 #   ilastik: interactive learning and segmentation toolkit
 #
-#       Copyright (C) 2011-2014, the ilastik developers
+#       Copyright (C) 2011-2024, the ilastik developers
 #                                <team@ilastik.org>
 #
 # This program is free software; you can redistribute it and/or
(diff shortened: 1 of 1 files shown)


ilastik/applets/pixelClassification/pixelClassificationDataExportGui.py
```diff
@@ -48,8 +41,15 @@ def __init__(self, *args, **kwargs):
         super(PixelClassificationResultsViewer, self).__init__(*args, **kwargs)
         self.topLevelOperatorView.PmapColors.notifyDirty(bind(self.updateAllLayers))
         self.topLevelOperatorView.LabelNames.notifyDirty(bind(self.updateAllLayers))
+        self._childOps = []
```
k-dominik, line 44: looks like there will always be only one of those `childops`? Why would it be a list?

ilastik/applets/pixelClassification/pixelClassificationDataExportGui.py
```diff
@@ -48,8 +41,15 @@ def __init__(self, *args, **kwargs):
...
 
     def setupLayers(self):
+        for op in self._childOps:
+            # self._initPredictionLayers may have created an op for our Gui purposes.
+            # Clean up before potentially creating another one.
+            op.Input.disconnect()
+            op.cleanUp()
```
k-dominik, line 51: is there a reason not to remove the reference once cleaned up so that it can be gc'ed?

k-dominik (commented):
great! 8 year old bug, that is super cool :)

btbest:
Very sensible feedback 😅 All done and should be ready now

PR #2890, opened 2024-07-24
Add sphericaltexture_plugin to default object feature plugins

k-dominik:
CC: @oanegros

this is ready now. There were some complications on windows which I didn't follow to the root cause, but, some versions of `pyshtools` (a dependency of `sphericaltextures`) would clash with `pytorch`. It looks like these issues have been resolved with pytorch 2.4.*.

Also removed `inferno` (resulted in conflicts, but since it's not used in our code it was a good reminder to actually remove it).

The Object Classification tests now also calculate the spherical texture features, so the reference data needed regenerating.

- [ ] Add tests.


conda-recipe/meta.yaml
```diff
@@ -67,9 +67,12 @@ outputs:
         - z5py
         - zarr 2.*
       run_constrained:
-        - mkl !=2024.1.0
+        - mkl <2024.1.0
```
btbest, line 70: :(

tests/test_ilastik/test_workflows/testObjectClassificationGui.py
```diff
@@ -253,6 +252,13 @@ def impl():
                 for plugin in features
             }
 
+            # in order to also calculate neighborhood features, we need to set one
+            # margin to non-zero
+            for plugin in features:
+                for _, feature_descr in features[plugin].items():
+                    if "margin" in feature_descr:
+                        feature_descr["margin"] = (3, 3, 3)
```
btbest, line 260: I'm not familiar with what the properties of the feature description mean, but why do we need to overwrite a value if it exists?

btbest (approved):

conda-recipe/meta.yaml
k-dominik: yeah, hope this is resolved soon (related: https://github.com/pytorch/pytorch/issues/129310)

tests/test_ilastik/test_workflows/testObjectClassificationGui.py
k-dominik: yes, this setting is to be overridden by the UI (hence, manually in test). The existence of the "margin" key indicates that a feature is computed on the object + some neighbourhood around it. The logic will hopefully soon change, but for now it's like that - and for the tests to trigger the computation of those neighbourhood features, there needs to be a non-zero value (all "neighbourhood features" have a 0 there as default).

PR #2892, opened 2024-07-30
Object classification: Restore default feature selection from previous verisons

k-dominik:
with the inclusion of the sphericaltexture plugin, the "All excl. location" button behavior changed - it would also include
those new features. Here we change the behavior of the button to include only the default features, as they were present in previous versions.

Timing-wise the sphericaltexture features are about a factor of 10x
slower to compute than our other features - so these should be selected
only if useful for the task (tbd by users).

## Checklist

- [ ] Add tests.

### DIFF
--- a/ilastik/applets/objectExtraction/featureSelectionWithHelp.ui
+++ b/ilastik/applets/objectExtraction/featureSelectionWithHelp.ui
@@ -104,7 +104,10 @@
      <item>
       <widget class="QPushButton" name="allButLocationButton">
        <property name="text">
-        <string>All excl. Location</string>
+        <string>Default selection</string>
+       </property>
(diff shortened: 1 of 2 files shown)


btbest (approved):
lgtm

I noticed that the "incomplete check" mechanism doesn't respond correctly; i.e. the "default selection" button turns the check boxes on the top-level into "all checked" marks, even though some sub-items are unchecked. (It's supposed to be a horizontal dash, i.e. a "partially checked" mark). But it turns out the "All excl location" button actually behaves the same in 1.4.1b18 :)
image

k-dominik:
thx Bene! For some reasone I missed the non-partial checkboxes! (https://github.com/ilastik/ilastik/issues/2894)

Issue #1568, opened 2017-07-31
hdf5 subvolume selection, once a subvolume has been selected, it will be the default one


k-dominik:
When selecting `Add separate images` in the `Input Data` applet with a h5 file containing multiple subvolumes, a dialog pops up that asks the user to select one of the subvolumes to open. So far, so good.
However, if one has selected the wrong one by accident, the choice cannot be undone. After removing the first image, and, again clicking on `Add separate images` with the same file, the previously selected subvolume will automatically be loaded.

verma014:
If you double click on file in the "Input Data" tab, it will open up the image tab. Change the subvolume in the "Internal Dataset Name:"

k-dominik:
Thank you for this workaround!

However I will leave this issue open, because it's not especially intuitive

PR #2895, opened 2024-07-31
OME-Zarr in headless and API

### LINKED ISSUE #2888: Headless scale selection for ome zarr
Is there a parameter to select scale for ome-zarr images when used in headles mode?

CC: @tischi @btbest

btbest:
* Headless processing OME-Zarr datasets with `/direct/path/to.zarr/scale` or `https://uri.com/to.zarr/scale`
* Expose `OMEZarrStore` internal ZarrArrays directly for use in API
* Also allows the user to add a single image from an arbitrary Zarr store dataset through the GUI by choosing the `.zgroup` file in the file picker - going through the H5N5 dataset picker dialog.
* And that dialog had a default behaviour of locking the choice of internal path per store until ilastik is restarted. I've added a checkbox to make that optional at least when coming through `DataSelectionGui`. Could change this to always display the check box and just have the checked-status on or off based on the new parameter, but then every usage needs to actually handle the box's status when it's accepted, so this would touch a few more files.

The backend (i.e. `OpInputDataReader`) now uses two different OME-Zarr readers depending on what file path is passed:
* path on disc goes through `z5py.ZarrFile` (Pro: can handle arbitrary subpaths inside the store, and we have internal handling for globs. Con: Errors are typically cryptic, so diagnosing issues with a particular store can be hard; can't handle URIs)
* path is a URI (http* or file scheme):  goes through our custom `OMEZarrStore` (Pro: Handling issues with the store is up to us; handles URIs. Con: Can only handle subpaths of `OME-Spec.multiscales[0]`; no globbing)

Since z5py doesn't support web, we could alternatively recognise disc paths with `.zarr` and pipe them through OMEZarrStore too. But this limits some freedom in handling hierarchical stores that the disc-path flow supports.

Fixes #2888 

## Checklist

- [x] Add tests.
- [ ? ] Update [user documentation](https://github.com/ilastik/ilastik.github.io). ---- not sure if we want to add examples maybe


btbest:
Headless tests are expensive, so I've kept it to a minimum:

- Headless pixel classification with a `file:` URI and an OME-Zarr store in the test dir
- Headless pixel classification with an `http:` URI and a server under localhost serving the OME-Zarr store

The second one in particular takes very long, which seems to directly depend on the number of requests to the server. The loader needs one request per chunk that the test dataset is stored in, which is why I went down to 4 chunks. We could even say we don't care about testing chunking here and go down to 1.

btbest:
Last commit is subject to drop or change depending on what we want. This locks down reading Zarr from disc to Zarr stores with valid OME multiscales metadata. Pro: The feature-set is better defined, tested, and more consistent between supplying a file-path vs a URI to a Zarr store. Con: We throw away reading arbitrary Zarr stores when supplying a file-path, where we could alternatively go with a "hope for the best" approach when looking for or guessing axes.

lazyflow/operators/ioOperators/opStreamingH5N5Reader.py
```diff
@@ -22,21 +22,68 @@
...
 
 from lazyflow.graph import Operator, InputSlot, OutputSlot
 from lazyflow.utility import Timer
 from lazyflow.utility.helpers import get_default_axisordering, bigintprod
+from lazyflow.utility.io_util.OMEZarrStore import get_axistags_from_spec as get_ome_zarr_axistags
 
 logger = logging.getLogger(__name__)
 
 
+def _find_or_infer_axistags(file: Union[h5py.File, z5py.N5File, z5py.ZarrFile], internalPath: str) -> vigra.AxisTags:
```
k-dominik, line 40: great to have single clear place for this. In principle one could differentiate by file type, and not try h5/n5 files for ome metadata, and vice versa...

tests/test_ilastik/test_workflows/test_HeadlessPixelClassificationWorkflow.py
```diff
@@ -224,3 +231,128 @@ def test_distributed_results_are_identical_to_single_process_results(
...
+@pytest.fixture
+def ome_zarr_store_on_disc(tmp_path) -> Dict[str, numpy.ndarray]:
+    """Sets up a zarr store of a random image at raw scale and a downscale.
+    Returns a dictionary of the expected datasets (scale_key: numpy.ndarray)"""
+    zarr_dir = tmp_path / "some.zarr"
+    zarr_dir.mkdir(parents=True, exist_ok=True)
+
+    dataset_shape = [3, 100, 100]  # cyx - to match the 2d3c project
+    scaled_shape = [3, 50, 50]
+    chunk_size = [3, 64, 64]
```
k-dominik, line 245: is it valid to have a chunk size bigger than, say the scaled shape?

tests/test_ilastik/test_workflows/test_HeadlessPixelClassificationWorkflow.py
```diff
@@ -224,3 +231,128 @@ def test_distributed_results_are_identical_to_single_process_results(
...
+
+def wait_for_server(host_and_port: Tuple[str, int], timeout=5):
+    end_time = time.time() + timeout
+    while time.time() < end_time:
+        try:
+            with socket.create_connection(host_and_port, timeout=1):
+                return True
+        except OSError:
+            time.sleep(0.1)
+    raise RuntimeError(f"Server on {':'.join(host_and_port)} didn't start within {timeout} seconds")
```
k-dominik, line 317: for fun I ran something on `localhost:8889`. At least on mac, this `wait_for_server` doesn't error out. It would be cool if this test wouldn't rely on a specific port being open (maybe random, check if available first?).

tests/test_lazyflow/test_operators/test_ioOperators/testOpStreamingH5N5Reader.py
```diff
@@ -21,77 +19,142 @@
...
+                    ],
+                }
+            ],
+            ["t", "c", "x", "y", "z"],
+        ),
+        ([{"datasets": [{"path": "volume/ome_data"}], "axes": ["t", "c", "z", "y", "x"]}], ["t", "c", "z", "y", "x"]),
+        (
+            [  # Multiple multiscales tested on v0.3 because it's more compact :)
+                {"datasets": [{"path": "somewhere/else"}], "axes": ["t", "c", "z", "y", "x"]},
+                {"datasets": [{"path": "volume/ome_data"}], "axes": ["t", "c", "y", "x", "z"]},
```
k-dominik, line 114: :+1:

k-dominik (approved):
Awesome :) I also gave this a try and all feels very natural and convenient :)

ilastik/experimental/api/_pipelines.py
```diff
@@ -52,6 +52,21 @@ class PixelClassificationPipeline:
...
+    ```Python
+    import xarray
+    from ilastik.experimental.api import PixelClassificationPipeline
+    from lazyflow.utility.io_util.OMEZarrStore import OMEZarrStore
+
+    store = OMEZarrStore("<https://example.com/data.zarr>")
+    zarray = store.get_zarr_array("<scale name>")
+    dims = tuple(store.axistags.keys())
+
+    img = xarray.DataArray(zarray, dims=dims)  # downloads the entire image
```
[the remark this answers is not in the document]
btbest: The zarray is of course capable of doing things lazily - but this is up to the API user's implementation. I could extend the docstring to mention that you can do e.g. `da = dask.array.from_zarr(zarray)` and then do your thing with `da` if you want to run lazy computations?

Or, you know, we could point them to the lazyflow docs and suggest setting up their own computation graph 😝

ilastik/applets/dataSelection/dataSelectionGui.py
```diff
@@ -542,20 +542,24 @@ def _get_dataset_full_path(self, filePath: Path, roleIndex: int) -> Path:
...
-                dlg = SubvolumeSelectionDlg(datasetNames, self)
+                dlg = SubvolumeSelectionDlg(datasetNames, self, offer_remember_dataset=True)
                 if dlg.exec_() != QDialog.Accepted:
                     raise DataSelectionGui.UserCancelledError()
                 selected_index = dlg.combo.currentIndex()
+                keep_selected_as_default = dlg.checkbox.isChecked()
                 selected_dataset = str(datasetNames[selected_index])
-        self._add_default_inner_path(roleIndex=roleIndex, inner_path=selected_dataset)
+        if keep_selected_as_default:
+            self._add_default_inner_path(roleIndex=roleIndex, inner_path=selected_dataset)
```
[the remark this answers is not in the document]
btbest: We should be grateful to the existing implementation for that 😅

lazyflow/utility/io_util/OMEZarrStore.py
```diff
@@ -222,3 +222,13 @@ def request(self, roi: rtype.Roi, scale_key=DEFAULT_SCALE_KEY):
...
         data = self._scale_data[scale_key]["zarray"][roi.toSlice()]
         return data
+
+    def get_zarr_array(self, scale_key: str):
+        """
+        Intended exclusively for use through ilastik-API.
+        Internally, ilastik and lazyflow should never directly access the ZarrArray.
+        """
+        if scale_key not in self._scale_data:
+            msg = f"No scale named {scale_key} in this store. Please inspect `.multiscales` to see available scales."
```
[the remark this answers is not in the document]
btbest: I had this hint in the docstring before, and then realised I can keep the docstring more concise and make the usage easier at the same time by putting it here.

tests/test_ilastik/test_workflows/test_HeadlessPixelClassificationWorkflow.py
btbest: Yes, it just means the entire dataset is stored in one file

lazyflow/operators/ioOperators/opStreamingH5N5Reader.py
btbest: It is a bit weird that this makes OME-H5 and OME-N5 a practical reality 🙃

It would probably be sensible to put the OME checking behind a ZarrFile condition and the N5 check behind an N5File condition. But I would keep the preference for vigra AxisTags common to all.

tests/test_ilastik/test_workflows/test_HeadlessPixelClassificationWorkflow.py
btbest: True, trying random and checking availability would be sensible

emilmelnikov: If you bind to port 0, then a random, free port will be selected automatically.

ilastik/experimental/api/_pipelines.py
```diff
@@ -52,6 +52,21 @@ class PixelClassificationPipeline:
     prob_maps = pipeline.get_probabilities(img)
     ```
 
+    Example using an OME-Zarr dataset stored remotely:
+    ```Python
+    import xarray
+    from ilastik.experimental.api import PixelClassificationPipeline
+    from lazyflow.utility.io_util.OMEZarrStore import OMEZarrStore
+
+    store = OMEZarrStore("<https://example.com/data.zarr>")
```
emilmelnikov, line 61: This is of course very subjective, but some people could interpret `<` and `>` in example strings as literal, required characters (maybe some formatting delimiters like we have in dataset nicknames).

k-dominik:
fixes https://github.com/ilastik/ilastik/issues/1568

ilastik/experimental/api/_pipelines.py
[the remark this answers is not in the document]
btbest: I was a bit off-point with the above comment; the limitation to laziness here is `get_probabilities`. It only accepts `VigraArray` or `xarray.DataArray`, neither of which support lazy accessing as far as I can tell.

So not a topic for this PR from that point of view, though it would be a great feature.

lazyflow/operators/ioOperators/opStreamingH5N5Reader.py
```diff
@@ -22,21 +22,68 @@
...
+
+    try:
+        # Look for OME-Zarr metadata (found at store root, not in dataset)
+        # OME-Zarr stores with more than one multiscale don't exist in public, but the spec allows it
+        multiscale_index = None
+        for i, scale in enumerate(file.attrs["multiscales"]):
+            if any(d.get("path", "") == internalPath.lstrip("/") for d in scale.get("datasets", [])):
+                multiscale_index = i
+        if multiscale_index is None:
+            raise KeyError("no spec for dataset path")
```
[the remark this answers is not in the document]
btbest: TIL about Python's `for` statements having an else clause 😅 Googling suggests that it's not just me who finds this quite obscure, so I'd rather avoid using it (particularly in a context where it can be easily misread as belonging to an `if` clause).

ilastik/experimental/api/_pipelines.py
btbest: I see the point and was about to change it. Then saw that all other examples in the API docstrings use the same notation, which must be why I used it. So I'll leave it :)

btbest:
Ready to merge unless there are more comments. The CI failure is due to pytorch 2.4.1 (see https://github.com/ilastik/ilastik/pull/2904)

PR #2896, opened 2024-08-09
H5N5 writing: Refactor test - readability, disentangle test cases

btbest:
Minor PR for refactoring a test. I rewrote the test because I was going to add zarr as another test case, then decided OME-Zarr warrants its own operator and won't go here.

I think the refactoring is still helpful in its own right. It makes the difference between the test cases clearer, and makes sure the h5 and n5 tests don't depend on each other (plus it's a bit shorter :smile: ).

The actual test cases covered and how they're implemented could be improved I think. I'm also not sure about the value of the third test using `.Output.meta.ram_usage_per_requested_pixel = 1000000.0` to force writing in tiny blocks. This test takes 0.5s to run, which is a long time for a test case that seems unnecessary to cover to me. But I intentionally purely refactored here to make review easier.

Note: Squash-merge when re-integrating to main

### DIFF
--- a/tests/test_lazyflow/test_operators/test_ioOperators/testOpH5N5WriterBigDataset.py
+++ b/tests/test_lazyflow/test_operators/test_ioOperators/testOpH5N5WriterBigDataset.py
@@ -1,9 +1,7 @@
-from builtins import object
-
 ###############################################################################
 #   lazyflow: data flow based lazy parallel computation framework
 #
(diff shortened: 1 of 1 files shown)


k-dominik (approved):
Nice refactor :), lgtm

I personally don't mind to keep the 0.5 seconds for the tiny request-shape test.

Issue #2886, opened 2024-07-17
Spammy gui warnings


btbest:
While testing 1.4.1b17 I noticed that we have various sources of random spam in the console window next to the GUI on Windows and Linux. On Mac there doesn't seem to be an extra console window (?), so not an issue. On Windows and Linux it's a bit irritating though.

###  QFont::fromString: Invalid description '(empty)'
On Windows, from at least 1.4.1b15 onwards. Spammed in the data selection applet when moving the horizontal separator bar between the input data table and the viewer below it. I'm also getting this printed once when switching back from another applet to data selection.

### QOpenGLFramebufferObject: Framebuffer incomplete attachment.
On Windows, possibly only introduced in 1.4.1b17. `QOpenGLFramebufferObject: Framebuffer incomplete attachment.` and `QOpenGLFramebufferObject: Framebuffer incomplete, missing attachment.` printed ~6 times when opening a project. I'm not seeing this in 1.4.1b15.

## Screenshots
![image](https://github.com/ilastik/ilastik/assets/63287233/7e4a8ccb-0f32-4ae5-b154-375142c2e6e5)

k-dominik:
can't reproduce the

```
run_ilastik.sh: 6: [[: not found
run_ilastik.sh: 16: [[: not found
run_ilastik.sh: 25: [[: not found
run_ilastik.sh: 25: [[: not found
run_ilastik.sh: 32: [[: not found
run_ilastik.sh: 32: [[: not found
run_ilastik.sh: 44: [[: not found
run_ilastik.sh: 49: [[: not found
```

bit. I tried bash and zsh on my linux machine. Would you have a little more detail on how you got these?

emilmelnikov:
> can't reproduce the
> 
> ```
> run_ilastik.sh: 6: [[: not found
> run_ilastik.sh: 16: [[: not found
> run_ilastik.sh: 25: [[: not found
> run_ilastik.sh: 25: [[: not found
> run_ilastik.sh: 32: [[: not found
> run_ilastik.sh: 32: [[: not found
> run_ilastik.sh: 44: [[: not found
> run_ilastik.sh: 49: [[: not found
> ```
> 
> bit. I tried bash and zsh on my linux machine. Would you have a little more detail on how you got these?

This is probably coming from `dash` or a similar POSIX-compatible shell that doesn't support ["bashisms"](https://mywiki.wooledge.org/Bashism).

Tomaz-Vieira:
[...]

PR #2898, opened 2024-08-19
Fix dark mode (for mac)

k-dominik:
Wow, how much time can one spend on choosing colors for a button...

Motivation (gallery of horror): I personally don't use dark mode, but almost every user I support (on osx) does. And it was really ugly:

<details><summary>About dialog</summary>
Screenshot 2024-08-19 at 16 41 00
</details>

<details><summary>Feature Selection</summary>
Screenshot 2024-08-19 at 16 41 09
</details>

<details><summary>Pixel Classification Training</summary>
Screenshot 2024-08-19 at 16 41 21
</details>

Dealing with dark mode in qt5 seems also to be challenging to do properly (see [here](https://stackoverflow.com/questions/75457687/detect-dark-application-style)). This PR adds some band aids to at least make ilastik usable in  dark mode:

<details><summary>About dialog</summary>
Screenshot 2024-08-19 at 16 42 05
</details>

<details><summary>Feature Selection</summary>
Screenshot 2024-08-19 at 16 42 12
</details>

<details><summary>Pixel Classification Training</summary>
Screenshot 2024-08-19 at 16 42 21

</details>

_Some_ of these changes adapt dynamically to switching the theme (not all unfortunately).

### Updates:

* 🚀 now also partially addresses  #2886: `QFont::fromString: Invalid description '(empty)'` crap is gone!
* apparently, in data selection, for roles that have no data associated with a lane yet, we would print the string `<empty>` below the in-line "Add" button. This was barely visible in light mode, but in dark mode, the string would be printed on top of the button :shrug:
<details><summary>light mode zoomed in (before)</summary> 

 </details>
   <details><summary>dark mode (before)</summary>
image
</details>
   <details><summary>dark mode (after)</summary> 
image
 </details>

Todo:
- [x] check windows (thx at @btbest, see https://github.com/ilastik/ilastik/pull/2898#issuecomment-2298234191)
- [x] Fix dark mode in volumina export dialog https://github.com/ilastik/volumina/pull/314

### DIFF
--- a/ilastik/widgets/featureTableWidget.py
+++ b/ilastik/widgets/featureTableWidget.py
@@ -22,6 +22,7 @@ import enum
 from PyQt5.QtGui import QBrush, QColor, QFont, QIcon, QPainter, QPen, QPixmap, QPolygon, QKeyEvent
 from PyQt5.QtWidgets import (
     QAbstractItemView,
+    QApplication,
     QHeaderView,
     QItemDelegate,
     QStyle,
@@ -30,6 +31,8 @@ from PyQt5.QtWidgets import (
     QTableWidgetSelectionRange,
 )
 from PyQt5.QtCore import QPoint, QRect, QSize, Qt
+
+from ilastik.utility.gui import is_qt_dark_mode
 
 
 class FeatureEntry(object):
@@ -62,6 +65,44 @@ class FeatureTableWidgetVSigmaHeader(QTableWidgetItem):
         self.setText(text)
 
 
+class _FeatureSelectionColors:
+    """
+    Class that holds some color "constants". Colors change according to
+    used system theme (on windows and MacOS).
+    """
+
+    @staticmethod
+    def foreground():
+        return QApplication.palette().windowText().color()
+
+    @staticmethod
+    def green():
+        if is_qt_dark_mode():
+            # dark mode
+            return QColor("#00ad6b")
+        else:
+            # light mode
+            return QColor("#00fa9a")
+
+    @staticmethod
+    def light_gray():
+        if is_qt_dark_mode():
+            # dark mode
+            return QColor("#4d4d4d")
+        else:
+            # light mode
+            return QColor("#dcdcdc")
+
+    @staticmethod
+    def dark_gray():
+        if is_qt_dark_mode():
+            # dark mode
+            return QColor("#dcdcdc")
+        else:
+            # light mode
+            return QColor("#4d4d4d")
+
+
 # ==============================================================================
 # FeatureTableWidgetVHeader
 # ==============================================================================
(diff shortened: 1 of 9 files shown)


ilastik/widgets/featureTableWidget.py
```diff
@@ -62,6 +65,44 @@ def __init__(self, text="Sigma"):
         self.setText(text)
 
 
+class _FeatureSelectionColors:
+    """
+    Class that holds some color "constants". Colors change according to
+    used system theme (on windows and MacOS).
+    """
+
+    @staticmethod
```
emilmelnikov, line 74: Turn these methods into properties? Unfortunately, `staticmethod(property(meth))` still doesn't work even in Python 3.12, but `classmethod(property(meth))` works just fine.

Also could drop `_color` suffix since it's obvious from the class name.

k-dominik: Heya, thanks @emilmelnikov for checking this out! I also wanted to have properties (I wanted something as close to an enum as I could get, I guess, and did have chained classmethod with properties), but it turns out [these are deprecated in py11](https://docs.python.org/3/whatsnew/3.11.html#language-builtins).
Dropping the `_color` makes sense :)

btbest:
I wanted to give it a try, but not sure how :)

I tried `Personalization > Colors > Choose your mode > Dark`, but ilastik happily ignores that and continues being light. Same for `Personalization > Themes > (choose dark theme)`.

`Accessibility > Contrast themes > Aquatic` is I suppose some form of dark mode? Though let's say this doesn't look great even in Firefox. Besides the main screen, it's pretty ok in ilastik actually.
<details><summary>Firefox</summary>

image

</details>

<details><summary>ilastik main screen</summary>

image
</details>
<details><summary>about</summary>

image
</details>
<details><summary>feature selection</summary>

image
</details>
<details><summary>training</summary>

image
</details>

k-dominik:
Hey @btbest, thanks so much for trying! I realized that I don't even have access to a win11 machine :D. As for the main screen - is it possible that you switched themes while ilastik was running? Unfortunately not all widgets will  adapt at runtime (this would be quite some more effort) - but after a restart even the startup screen should look nice. In the end I hope users don't switch themes often.

Looks mostly great (well, let's ignore the yellow in the title bars...)  except for the about dialog :/

btbest:
Ah, my bad... So yes as you say, the main screen is fine, but About looks as in the screenshot.

ilastik/widgets/featureTableWidget.py
emilmelnikov: Thanks, didn't know this was deprecated!

In this case, it's trivial to roll up a custom descriptor/decorator that does the same thing:

```python
class staticproperty:
    def __init__(self, getter):
        self._getter = getter

    def __get__(self, _instance, _owner=None):
        return self._getter()
```

Probably could be useful in other places too.

k-dominik: Hi @emilmelnikov. Thank you again for the suggestion! I guess "trivial" is subjective.

In general I am against introducing syntactic sugar that goes beyond standard Python. I'll keep in in mind for a situation where I feel the benefit is worth introducing this concept that is very likely absent in most other Python code-bases for the sake of removing `()` for function calls. In a way, having an explicit function call here makes it obvious that this value is computed at runtime.

PR #2903, opened 2024-08-28
Use vigra.analysis.unique instead of numpy.unique

### LINKED ISSUE #1200: Remove all calls to numpy.bincount()
I didn't realize this until recently, but `numpy.bincount()` casts the data as `int64` internally, copying it if necessary.  That means `bincount()` has a RAM overhead of 2x the input (assuming the input is `uint32`), instead of 0x overhead as expected.

In contrast, vigra can do the job with nearly zero RAM overhead, so we should use this instead.

```python
def vigra_bincount(labels):
    labels = labels.reshape((1,-1), order='A')
    # We don't care what the 'image' parameter is, but we have to give something
    image = labels.view(np.float32)
    counts = vigra.analysis.extractRegionFeatures(image, labels, ['Count'])['Count']
    return counts.astype(np.int64)
```

There are only a few places where we're using `bincount()`, but they're in bottleneck functions related to connected components filtering.

We should remove all instances of `bincount()` in our code, and consider monkey-patching it in our test suites to `assert` that we never use it.

k-dominik:
vigra.analysis.unique seems to still be 2x faster than numpy unique. Also replaced the last remaining numpy.bincounts.
Some additional black formatting...

Closes #1200

## Checklist

- [ ] Add tests.

### DIFF
--- a/bin/train_headless.py
+++ b/bin/train_headless.py
@@ -29,10 +29,11 @@ Note: This script does not make any attempt to be efficient with RAM usage.
       (The entire label volume is loaded at once.)  As a result, each image volume you
       train with must be significantly smaller than the available RAM on your machine.
 """
-from __future__ import print_function
-from builtins import range
+
 import os
 import numpy as np
+import vigra
+
 
 def main():
     # Cmd-line args to this script.
@@ -204,7 +205,7 @@ def generate_trained_project_file(
             label_volume = label_volume[..., None]
 
         # Auto-calculate the max label value
-        label_classes.update(np.unique(label_volume))
+        label_classes.update(vigra.analysis.unique(label_volume))
 
         print("Applying label volume to lane #{}".format(lane))
         entire_volume_slicing = roiToSlice(*roiFromShape(label_volume.shape))
--- a/ilastik/applets/labeling/labelingImport.py
+++ b/ilastik/applets/labeling/labelingImport.py
@@ -135,7 +135,7 @@ def import_labeling_layer(labelLayer, labelingSlots, parent_widget=None):
             readData = req.result
 
             # Can't use return_counts feature because that requires numpy >= 1.9
-            # unique_read_labels, readLabelCounts = numpy.unique(readData, return_counts=True)
+            # unique_read_labels, readLabelCounts = vigra.analysis.unique(readData, return_counts=True)
 
             # This does the same as the above, albeit slower, and probably with more ram.
             bincounts = chunked_bincount(readData)
@@ -551,14 +551,14 @@ class LabelImportOptionsDlg(QDialog):
     def _updatePosition(self):
         writeAxes = self._writeSeedsSlot.meta.getAxisKeys()
 
-        for (k, v) in list(self._insert_position_boxes.items()):
+        for k, v in list(self._insert_position_boxes.items()):
             insertBox, _ = v
             self.imageOffsets[writeAxes.index(k)] = insertBox.value()
 
     def _updateMappingEnabled(self):
         max_labels, _ = self._labelInfo
 
-        for (k, v) in list(self._insert_mapping_boxes.items()):
+        for k, v in list(self._insert_mapping_boxes.items()):
             enabledBox, mapToBox = v
             enabled = enabledBox.isChecked()
             if enabled:
@@ -580,6 +580,6 @@ class LabelImportOptionsDlg(QDialog):
         self.buttonBox.button(QDialogButtonBox.Ok).setEnabled(enableOk)
 
     def _updateMapping(self):
-        for (k, v) in list(self._insert_mapping_boxes.items()):
+        for k, v in list(self._insert_mapping_boxes.items()):
             _, mapToBox = v
             self.labelMapping[k] = mapToBox.value()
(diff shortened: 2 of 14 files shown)


stuarteberg:
For better or worse, `pandas` is already a dependency of ilastik, so you might consider `pd.unique()`. On my machine, I don't observe a 2x speedup for vigra, but I do observe a 9x speedup for pandas:

```python
import numpy as np
import pandas as pd
import vigra

def pandas_unique(a):
    return np.sort(pd.unique(a.reshape(-1)))

def bincount_unique(a):
    return np.bincount(a.reshape(-1)).nonzero()[0]

data = np.random.randint(0, 256, (1000, 500, 250), dtype="uint32")

u1 = np.unique(data)
u2 = vigra.analysis.unique(data)
u3 = pandas_unique(data)
u4 = bincount_unique(data)

# Sanity check
assert u1.tolist() == u2.tolist() == u3.tolist() == u4.tolist()
```

```python
In [52]: %timeit np.unique(data)
    ...: %timeit vigra.analysis.unique(data)
    ...: %timeit pandas_unique(data)
    ...: %timeit bincount_unique(data)
4.97 s ± 272 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
3.89 s ± 81.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
531 ms ± 21.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
847 ms ± 15 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```


There's a similar story for `bincount()`.  The equivalent function in pandas is `value_counts()`, but one has to add some additional manipulation to mimic the `bincount()` output:

```python
import vigra
import pandas as pd

def vigra_bincount(labels):
    """
    A RAM-efficient implementation of numpy.bincount() when you're dealing with uint32 labels.
    If your data isn't int64, numpy.bincount() will copy it internally -- a huge RAM overhead.
    (This implementation may also need to make a copy, but it prefers uint32, not int64.)
    """
    labels = labels.astype(np.uint32, copy=False)
    labels = np.ravel(labels, order="K").reshape((-1, 1), order="A")
    # We don't care what the 'image' parameter is, but we have to give something
    image = labels.view(np.float32)
    counts = vigra.analysis.extractRegionFeatures(image, labels, ["Count"])["Count"]
    return counts.astype(np.int64)

def pandas_bincount(labels):
    labels = np.ravel(labels, order="K")
    return pd.Series(labels).value_counts().reindex(range(data.max()+1), fill_value=0).values

data = np.random.randint(0, 256, (1000, 500, 250), dtype="uint32")

# Sanity check
assert np.bincount(data.reshape(-1)).tolist() == vigra_bincount(data).tolist() == pandas_bincount(data).tolist()
```

```python
In [69]: %timeit np.bincount(data.reshape(-1))
    ...: %timeit vigra_bincount(data)
    ...: %timeit pandas_bincount(data)
863 ms ± 18.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
1.76 s ± 114 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
594 ms ± 23.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

(Also, it seems that `np.bincount()` is not too slow, so maybe it has been improved in recent years.  It still loses to pandas, though.)

k-dominik:
Thanks Stuart :) This is lovely. The original motivation for the alternative bincount were memory considerations - I'll check if these still hold.

k-dominik:
Hey @stuarteberg,

I'm trying to get a more complete picture here and, without over engineering came up with this benchmark, where I included your different methods. Maybe you can also run it on your machine and add your results: https://github.com/ilastik/unique-bincount-benchmark

Cheers
Dominik

stuarteberg:
Since I am procrastinating at my real job, I took a look at the places where [`bincount` (and variants of it) is being used](https://github.com/search?q=repo%3Ailastik%2Filastik%20bincount&type=code).  I suspect all of those use-cases could be reimplemented to use `pd.Series(a).value_counts()`, which would probably be more efficient.  The `pandas_bincount()` function above takes pains to behave exactly like `np.bincount`, but that's probably not necessary for the actual use-cases in ilastik.

Looking in particular at [`computeSupervoxelLabels()`](https://github.com/ilastik/ilastik/blob/07e107876f211983be788f3ac9bfd68b53c002c6/ilastik/workflows/voxelSegmentation/opVoxelSegmentation.py#L537), it looks like there's a lot of room for improvement using pandas.

For instance, maybe something like the following?  (Note: Requires [this function](https://github.com/janelia-flyem/neuclease/blob/master/neuclease/util/segmentation.py#L604) as a dependency, so you'd have to copy that into the code, too.)

```python
def computeSupervoxelLabels(self, slice_=None):
    """
    For each supervoxel ID in self.SupervoxelSegmentation,
    return the label ID with the largest overlap from self.Labels.

    Returns:
        dict {sv: label}
    """
    supervoxel_mask = self.SupervoxelSegmentation.value[..., 0]
    labels = self.Labels.value[:]

    # Overlapping pairs of (sv, label), ordered by size (largest first).
    ct = (
        contingency_table(supervoxel_mask, labels)
        .rename_axis(['sv', 'label'])
        .reset_index()
    )

    # Select the largest non-zero label for each SV, unless the
    # SV has no non-zero label at all, in which case select 0.
    ct = pd.concat((ct.query('label != 0'), ct.query('label == 0')))
    ct = ct.drop_duplicates('sv')
    return dict(ct[['sv', 'label']].values)
```

PR #2904, opened 2024-09-06
Stop using mamba

btbest:
mamba is officially deprecated since July, and now it throws ugly warnings when you use it.

More places where this should be replaced, and contributing.md to be updated, but already pushing to test CI

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ The following text equips you with knowledge that makes contributing to ilastik 
 
 ## Development Environment
 
-1. Download and install [Git][git], [GitHub CLI][github-cli], and _the latest 64-bit_ [Mambaforge][mambaforge].
+1. Download and install [Git][git], [GitHub CLI][github-cli], and _the latest 64-bit_ [Miniforge][miniforge].
    - Windows: all following commands should be executed in the _Anaconda Prompt_ from the Start Menu.
    - Linux: Git might be already preinstalled, GitHub CLI might be available in your system package manager.
    - macOS: Git is already preinstalled, GitHub CLI is available in [Homebrew][homebrew].
@@ -29,10 +29,12 @@ The following text equips you with knowledge that makes contributing to ilastik 
    gh repo clone ilastik
    ```
 
-1. If you have an existing [Miniconda][miniconda] installation, you can use it for ilastik development, but install [Mamba][mamba] and configure conda to use [strict channel priority][strict-channel-priority]:
+1. If you have an existing [Miniconda][miniconda] installation, you can use it for ilastik development, but configure conda to use [strict channel priority][strict-channel-priority]. If your conda version is older than 23.10.0, make sure install `mamba`, and for all following steps, run the commands with "`mamba`" instead of "`conda`" (or consider updating conda instead):
 
    ```
    conda config --set channel_priority strict
+
+   conda --version  # If below 23.10.0, install mamba, and use mamba instead of conda for all further steps:
    conda install --name base --channel conda-forge mamba
    ```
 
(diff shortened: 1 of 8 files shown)


btbest:
Not sure why `pytorch=2.*` for `test`, but `pytorch>=1.6` for `run` in `meta.yaml`

k-dominik:
`pytorch>1.6` in meta to allow people to try if they can get stuff to run. It does still work, but I don't want to test all possible combinations.

CONTRIBUTING.md
```diff
@@ -29,20 +29,22 @@ The following text equips you with knowledge that makes contributing to ilastik
    gh repo clone ilastik
    ```
 
-1. If you have an existing [Miniconda][miniconda] installation, you can use it for ilastik development, but install [Mamba][mamba] and configure conda to use [strict channel priority][strict-channel-priority]:
+1. If you have an existing [Miniconda][miniconda] installation, you can use it for ilastik development, but configure conda to use [strict channel priority][strict-channel-priority]. If your conda version is older than 23.10.0, make sure install `mamba`, and for all following steps, run the commands with "`mamba`" instead of "`conda`" (or consider updating conda instead):
```
k-dominik, line 32: And I'm not sure about the mixing of miniconda, miniforge. I would ditch all mention of miniconda.

I wouldn't suggest that. I don't think there are guarantees for older versions of conda to keep up with the ecosystem. I think we _can_ be opinionated here.

```suggestion
1. If you have an existing [Miniconda][miniconda] installation, you can use it for ilastik development, but configure conda to use [strict channel priority][strict-channel-priority].
```

k-dominik (commented):
The sunsetting of mamba had to be done at some point, great that this is now done.

Do you have maybe some details on why pytorch doesn't run on windows?

btbest:
I reproduced the CI problem by running `test_nn_local_headless_prediction` in a fresh env set up from `environment-dev.yml` (without `pytorch!=2.4.1`, which served me with `pytorch 2.4.1 py3.9_cpu_0`):

```pytb
  File "C:\Users\me\.conda\envs\ilastik\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\me\.conda\envs\ilastik\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\me\.conda\envs\ilastik\lib\site-packages\tiktorch\server\__main__.py", line 1, in <module>
    from .base import main
  File "C:\Users\me\.conda\envs\ilastik\lib\site-packages\tiktorch\server\base.py", line 6, in <module>
    from torch import multiprocessing as mp
  File "C:\Users\me\.conda\envs\ilastik\lib\site-packages\torch\__init__.py", line 137, in <module>
    raise err
OSError: [WinError 182] The operating system cannot run %1. Error loading "C:\Users\me\.conda\envs\ilastik\lib\site-packages\torch\lib\fbgemm.dll" or one of its dependencies.
```

Based on [forum thread](https://discuss.pytorch.org/t/failed-to-import-pytorch-fbgemm-dll-or-one-of-its-dependencies-is-missing/201969), the problem has been around for many people already with 2.4.0 and was supposed to be [_fixed_](https://github.com/pytorch/pytorch/issues/131662#issuecomment-2268984561) by 2.4.1. I don't know why it's exactly the other way around for CI and me, maybe related to cuda version. The manual fix seems to be an additional libomp140.x86_64.dll that one could get from a Visual Code install. My impression is that pytorch doesn't quite have its dependencies straight on Win with 2.4.* so far. 2.4.0 seems to be fine with tiktorch, so instead of digging for extra dependencies, I would rather just forbid 2.4.1.

emilmelnikov:
What about `micromamba`? They say in the docs that

> micromamba is especially well fitted for the CI use-case but not limited to that!

k-dominik:
the problem with pytorch is a different one - it's a dependency clash we have with `pyshtools` or more specifically one of it's dependencies (gmt). I've uploaded patched versions of pyshtools to `ilastik-forge/label/patched`, so I guess the solution would be to add this channel to the ci. Could you try doing this instead of the `!=2.4.1` pin?


xref: https://github.com/conda-forge/pyshtools-feedstock/issues/63

btbest:
Freaky, very nice catch... The test runs on my machine, let's see CI.

btbest:
Unfortunately that doesn't seem to fix it... [this run](https://github.com/ilastik/ilastik/actions/runs/10882613777/job/30194161988) used pyshtools 4.12.2 from /labels/patched, still fails

k-dominik (approved):

PR #2911, opened 2024-09-26
Fix flaky ObjectClassificationGui test

k-dominik:
this test was flaky due to so many features in training - fixing the seed should fix this.

Edit: triggered the test runs 3 times and _that particular_ test didn't fail.

## Checklist

- [ ] Add tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,11 +1,8 @@
 from pathlib import Path
 import os
-import shutil
 import tempfile
 import threading
-import time
 import queue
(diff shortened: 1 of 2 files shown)


btbest (approved):
What was causing the flakiness here, since the test only asserts that prediction layers exist at all..?

k-dominik:
the flaky test would be the next one that compares the exported output. The RF is not retrained for that. So fixing the seed for the training fixes that test in the end.

Issue #2131, opened 2019-09-30
Carving: No mesh export on 2D data


k-dominik:
## Describe the bug
Currently it is possible to click `export all meshes` in Carving when using 2D data, but it results in an error message.

## Expected behavior
I am not sure about the use case of 2D meshes, so we should probably just disable mesh-export for 2D data. Alternatively, of course we could hunt for a marching squares implementation (e.g. from [skimage](https://scikit-image.org/docs/dev/api/skimage.measure.html#find-contours)).

## To Reproduce
Within the Carving workflow
1. Add some 2d data in the data selection applet
2. Go to the preprocessing applet a applet, and do the superpixel computation
3. Go to Labeling applet, carve some objects (don't forget to save them)
4. Click on _Export all meshes_ -> :boom: 

## Error message/traceback

If applicable, please add tracebacks with formatting, such as

```pytb
ERROR 2019-09-30 11:21:00,651 excepthooks 18886 140685849134848 Unhandled exception in thread: 'Worker #1'
ERROR 2019-09-30 11:21:00,653 excepthooks 18886 140685849134848 Traceback (most recent call last):
  File ".../ilastik-1.3.3rc2-Linux/ilastik-meta/lazyflow/lazyflow/request/request.py", line 347, in _execute
    self._result = self.fn()
  File ".../ilastik-1.3.3rc2-Linux/ilastik-meta/ilastik/ilastik/workflows/carving/carvingGui.py", line 560, in exportMeshes
    _, mesh_data = list(labeling_to_mesh(object_volume, [1]))[0]
  File ".../ilastik-1.3.3rc2-Linux/ilastik-meta/volumina/volumina/view3d/meshgenerator.py", line 62, in labeling_to_mesh
    vertices, normals, faces = march(where(labeling == label, 2, 0).astype(int).T, 4)
ValueError: vertex or face count are zero: terminating marching cubes
```

## Screenshots
![bug](https://user-images.githubusercontent.com/24434157/65866848-d1ad8700-e375-11e9-84b7-b11c39a6de33.png)

## Desktop (please complete the following information):
 - ilastik version: 1.3.2rc2
 - OS: Ubuntu 18.04

PR #2912, opened 2024-10-01
Carving: Disable mesh export for 2D and improve responsiveness

btbest:
While testing if Carving works with OME-Zarr import/export, I ran into an error which turns out to be known (https://github.com/ilastik/ilastik/issues/2131), because I had chosen a 2D dataset.

I decided to implement the quick-fix (disabling the mesh export button for 2D), and also fixed a couple of responsiveness issues along the way, where the Preprocessing applet wouldn't let me progress sometimes after loading a project, removing the lane and adding a new one.

## Checklist

- x Add docstrings and comments.
- x Add tests :)
- x Update [user documentation](https://github.com/ilastik/ilastik.github.io).
- Rebase commits into a logical sequence --- I've made the commits extremely granular now, would squash this into just two commits ("Carving: Disable mesh export button for 2D inputs" and "Carving: Collapse and simplify writeprotect mechanisms")


k-dominik (commented):
Any PR with such a ratio of removed/added lines has to be good ;).

The simplification of the write-protect logic is great!

It would be great to also remove the option to export the mesh from the right click menu!

And just a comment, I noticed that after saving a project, loading it and then changing the preprocessing the objects are not deleted (and this can result in errors too). To be fair that was the case before, too :). Maybe we should not allow to change the preprocessing then :D haha.

btbest:
Thanks for pointing out the right-click option, I wasn't aware of that... I wanted to leave the option disabled with an informative tooltip. It looks like Qt doesn't show tooltips on disabled menu options though, so I just made the text informative instead.

> And just a comment, I noticed that after saving a project, loading it and then changing the preprocessing the objects are not deleted (and this can result in errors too). To be fair that was the case before, too :). Maybe we should not allow to change the preprocessing then :D haha.

Can you elaborate on this..? I've tried saving three objects, then saving the project, reopening, and whichever Preprocessing option I change, I always get the warning and if I click Yes and then go to carving/"Labelling", the saved objects from before are gone. I was doing this with one time point from the mitocheck dataset.

k-dominik:
I don't think that the kind of data is relevant, but for the record I was on 2dcells.

I did not save the project or anything. I did

1.) load the data
2.) filter (step edges)
3.) do one object (incl. saving it)
4.) filter (change one random parameter)
5.) go to carving, activate the complete segments layer

image

k-dominik (commented):
Cool, the complete segments layer is now clear - as it should be after changing the preprocessing!

When I create a project, or load a project and click on segment without doing any annotations I now get

```pytb
ERROR 2025-01-28 16:38:46,179 excepthooks 12177 8382724160 Traceback (most recent call last):
  File ".../ilastik/ilastik/workflows/carving/carvingGui.py", line 228, in onSegmentButton
    if not self.topLevelOperatorView.CanRunSegmentation.value:
  File ".../ilastik/lazyflow/slot.py", line 1043, in value
    temp = self[:].wait()
  File ".../ilastik/lazyflow/slot.py", line 987, in __getitem__
    assert self.meta.shape is not None, (
AssertionError: Can't ask for slices of this slot yet: self.meta.shape is None! (operator Wrapped Carving [self=<ilastik.utility.opMultiLaneWrapper.OpMultiLaneWrapper object at 0x33a592790>] slot: CanRunSegmentation, key=slice(None, None, None)
```

whereas, when I have already saved an object (also no annotations then) I get a somewhat more useful error message:

image

I think it would be better to have consistent behavior there.

btbest:
Looks like Carving still has a bunch of issues to sort out... this problem also exists on main. And I've also managed to hang (not error or crash) Carving by Ctrl+Z undoing a brush stroke while reproducing this one.

The `CanRunSegmentation` one should be a quick one and I'll fix it here, but the Ctrl+Z will have to wait for another time.

btbest:
@k-dominik quick look at the last commit would be nice, otherwise this should be ready to merge

k-dominik:
sure! Could you rebase this on the current main just to see the checkmarks? :)

k-dominik (approved):
Works as expected 🚀 😻

PR #2919, opened 2024-10-15
Update ilastik with the new bioimageio v0.5 spec

thodkatz:
This is a work in progress to update ilastik with the spec v5.

It uses a non released version of tiktorch. You can find it on this PR https://github.com/ilastik/tiktorch/pull/222.

To test this, you need to manually install the PR version of tiktorch.

The only use case that I have currently tested and it works with our first v5 model is the following:
- Local Neural Network workflow
- Load a 2d image
- affable-shark

Known issues:
- https://github.com/bioimage-io/spec-bioimage-io/issues/638

TODO:
- [x] 3d
- [ ] Fix progress bar


k-dominik:
Note, there's something weird going on with the GH actions on windows - the post build takes ages (not sure about the root cause, didn't have time to investigate, but I guessed this: https://github.com/conda-incubator/setup-miniconda/issues/380).

tests/test_ilastik/test_utility/test_bioimageio_utils.py
```diff
@@ -0,0 +1,118 @@
...
+        [
+            BatchAxis(),
+            SpaceInputAxis(size=SizeReference(tensor_id=TensorId("ref"), axis_id=AxisId("x")), id=AxisId("x")),
+            SpaceInputAxis(size=11, id=AxisId("y")),
+        ],
+    )
+    ut = ilastik.utility.bioimageio_utils.InputAxisUtils([descr1, descr2])  # type: ignore
+    tile_shape = ut.get_best_tile_shape("test")
+
+    assert tile_shape == OrderedDict({"x": MOCK_MIN_SIZE_2D, "y": 11})
```
thodkatz, line 118:
```suggestion
@pytest.mark.parametrize(
    "axes, expected_dict",
    (
        (
            [SpaceInputAxis(size=10, id=AxisId("x"))],
            OrderedDict({"x": 10, "y": 11}),
        ),
        (
            [SpaceInputAxis(size=ParameterizedSize(min=10, step=2), id=AxisId("x"))],
            OrderedDict({"x": MOCK_MIN_SIZE_2D, "y": 11}),
        ),
    ),
    ids=["Fixed-2D", "Mixed-2D",],
)
def test_get_best_tile_shape_size_ref(mock_min_tile_sizes, axes, expected_dict):
    descr1 = MockTensorDescr("ref", axes=axes)
    descr2 = MockTensorDescr(
        "test",
        [
            BatchAxis(),
            SpaceInputAxis(size=SizeReference(tensor_id=TensorId("ref"), axis_id=AxisId("x")), id=AxisId("x")),
            SpaceInputAxis(size=11, id=AxisId("y")),
        ],
    )
    ut = ilastik.utility.bioimageio_utils.InputAxisUtils([descr1, descr2])  # type: ignore
    tile_shape = ut.get_best_tile_shape("test")

    assert tile_shape == expected_dict
```

We could also maybe include the explicit size case ?

ilastik/utility/bioimageio_utils.py
```diff
@@ -0,0 +1,278 @@
...
+                factors.append(factor)
+            elif isinstance(axis_size, int):
+                size_funcs[axis_name] = partial(const_size, axis_size)
+            else:
+                raise NotImplementedError
+
+        # if all sizes are fixed, factors might be empty
+        min_factor = min(factors) if factors else None
+        for k, f in size_funcs.items():
+            sized_axes[k] = f(min_factor)
```
thodkatz, line 209: Very cool idea with the partials!

k-dominik: I had concerns about long term maintainability of this code (alternative would be to iterate twice), but if you like it I'll keep it.

Tomaz-Vieira (commented):
The `*Utils` and `*Validator` classes make me think that maybe we shouldn't be that tied to the spec in runtime after all. We could, say, just grab the spec and produce something like a `IlastikPredictionPipeline` instance that has all the logic and semantics it needs, without needing to keep going back to the spec. As a bonus, this would also work for spec v4 and a potential spec v6 in the future, as they might express the same concepts in different ways; all we'd need is another factory method to read the v4/v5 spec and then we'd be back to only the runtime class.

ilastik/utility/bioimageio_utils.py
```diff
@@ -0,0 +1,278 @@
...
+
+TaggedShape = OrderedDict[str, int]
+AnyAxis = Union[InputAxis, OutputAxis]
+
+VIGRA_TO_SPEC = {"b": "batch", "t": "time", "y": "y", "x": "x", "z": "z", "c": "channel"}
+SPEC_TO_VIGRA = {spec: vigra for vigra, spec in VIGRA_TO_SPEC.items()}
+
+
+def get_model_descr_from_model_bytes(model_bytes: bytes) -> ModelDescr:
+    with tempfile.NamedTemporaryFile(suffix=".zip", delete=False) as _tmp_file:
```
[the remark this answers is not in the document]
k-dominik: will do so once we update to core 0.7 :hide: - for this PR I'll keep 0.6 and roll with that code.

PR #2923, opened 2024-10-24
Multiscale: Preview scale file size in dropdown

btbest:
I like it in FIJI so I put it in ilastik :) should try to figure out some time how to change the default col width

current:

image

new:

image

expanded:

image

### DIFF
--- a/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
+++ b/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
@@ -23,6 +23,7 @@ from typing import Dict
 from PyQt5.QtCore import Qt, QAbstractItemModel, QModelIndex
 from ilastik.utility import bind
 from ilastik.utility.gui import ThreadRouter, threadRouted
+from lazyflow.utility.helpers import bigintprod
 from .opDataSelection import DatasetInfo
 from .dataLaneSummaryTableModel import rowOfButtonsProxy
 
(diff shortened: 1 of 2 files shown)


btbest:
I thought the same thing, and there's two options:

1. Remove the space between the number and the unit :) Being one word, Qt will keep them together (but it will hide the opening bracked, or parts in front of a full stop).
2. Fully custom option drawing, as I've committed now. I couldn't find any way to have Qt doing "the default styling for options but put this text instead please"

btbest:
Another aspect is that, while this is useful information, the goal should be to end up with an implementation where it really doesn't matter for the user how big the dataset is as a whole. In that context, emphasising the dataset size seems like a counterproductive thing to do.

In practice, and in particular as long as export-to-web is not possible, dataset size of course remains quite important, so I think this is a neat thing to add.

btbest:
image

😬

btbest:
Ok, it's clear there's no good way to do this with Qt. Here are the options. It's a pretty direct trade-off between code length/complexity/ugliness vs. UI prettiness. Whatever we do, I would set the default column width to a value that can accommodate 3D with 3-digit shape numbers and 3-digit file size (240 seems to suffice).

- Option 1: Default text eliding, see original screenshots. Could switch from the default ElideMiddle to ElideLeft so that file size is always visible. Or... simply put file size at the start of the string like "`240 MB (500 / 240 / 1022)`")
- Option 2: Custom word eliding only for the dropdown options, not for the dropdown box itself. Difficulty is it's not just the text that is customised, the entire option is custom painted. Reproducing the default styling seems reasonably simple at least on Windows, but would need to be checked on Linux and Mac how this looks.

image

- Option 3: Custom word eliding for both dropdown options and the box. This kind of sucks, because it's not really feasible to custom reproduce the default style, e.g. on Windows, mouseover triggers a fade to highlight-blue. I tried to minimise the awkwardness by default painting the box, and then re-paint only the inside content where the text is painted.

image

- Option 4: Do the word eliding in `DatasetDetailedInfoTableModel` instead of the View. This violates Qt's OOP design and dependency hierarchy, because then the model has to know about the view (column width) to decide whether and how to elide, which it shouldn't. It also fixes the text on instantiation, so the elision doesn't update on column width change. I haven't tried to get this to work, but I think it will be extremely awkward (reinstantiating the combobox on mouseup or sth like that I guess?). But maybe this doesn't matter too much, maybe people won't really try to resize it anyway, but I don't like it.

btbest:
@k-dominik Which of the four options above do you prefer? I think I tend towards 1, with changing the string as described there. The branch in its current state implements option 3.

k-dominik:
Hey @btbest, great dive into this :). I agree with using option 1 being probably the best compromise, also easier for long-term maintenance. Since these datasets are predominantly used in the context of larger data I'd go for sizing it to at least 4 digit shapes :)

btbest:
Alright, I think I'm happy with this now. @k-dominik please have a look on Mac whether the default width works and fits reasonably (link for convenience: https://s3.embl.de/i2k-2020/platy-raw.ome.zarr)

Here's the final look on Windows:
image

k-dominik:
Definitely readable on mac!

image

k-dominik (approved):
lgtm!

Issue #2922, opened 2024-10-21
Investigate problems with pytorch on windows for 2.4.0, 2.4.1, 2.5.0


k-dominik:
Unfortunately we had to disallow those versions because of CI errors on Windows.

PR #2924, opened 2024-10-24
OME-Zarr export: Fix export when source dataset is OME-Zarr v0.1

btbest:
Oops

### DIFF
--- a/lazyflow/utility/io_util/OMEZarrStore.py
+++ b/lazyflow/utility/io_util/OMEZarrStore.py
@@ -107,7 +107,7 @@ def _validate_transforms(
     """
     if coordinate_transformations is None:
         return None
-    if not isinstance(coordinate_transformations, list):
+    if not isinstance(coordinate_transformations, list) or not coordinate_transformations:
         return InvalidTransformationError()
     scale_transform = translation_transform = None
     for t in coordinate_transformations:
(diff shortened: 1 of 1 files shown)


btbest:
Windows CI failure see https://github.com/ilastik/ilastik/issues/2922

k-dominik (commented):
I guess that slipped through previous reviews, but the return type of `_validate_transforms` is not accurately annotated, but the docstring mentions all types. Would be nice to have those in sync.

Otherwise, let's ignore the pytorch problem on win in this PR for now and discuss in #2927

btbest:
> I guess that slipped through previous reviews, but the return type of `_validate_transforms` is not accurately annotated, but the docstring mentions all types. Would be nice to have those in sync.

... I'm not seeing where it's wrong 😅  Would the typing be clearer if we alias it like below?

```
# tuple(scale_transform, Optional[translation_transform])
ValidTransformations = Tuple[OMEZarrCoordinateTransformation, Optional[OMEZarrCoordinateTransformation]]


def _validate_transforms(
    coordinate_transformations: Optional[List[Dict[str, Union[str, List[float]]]]],
) -> Optional[Union[ValidTransformations, InvalidTransformationError]]:
```

k-dominik:
oh wow, yeah sorry, I think I'm still very tired. The `Or` completely escaped me :[

k-dominik (approved):

PR #2927, opened 2024-11-07
Fix pytorch fbgemm.dll dependency issue

btbest:
Issue we first saw in https://github.com/ilastik/ilastik/pull/2904 with pytorch 2.4.1

E.g. https://github.com/ilastik/ilastik/actions/runs/11721367647/job/32654563102

```
  File "C:\Users\runneradmin\miniconda3\conda-bld\ilastik-gpu_1730976400239\test_tmp\run_test.py", line 23, in <module>
    import torch
  File "C:\Users\runneradmin\miniconda3\conda-bld\ilastik-gpu_1730976400239\_test_env\lib\site-packages\torch\__init__.py", line 262, in <module>
    _load_dll_libraries()
  File "C:\Users\runneradmin\miniconda3\conda-bld\ilastik-gpu_1730976400239\_test_env\lib\site-packages\torch\__init__.py", line 245, in _load_dll_libraries
    raise err
OSError: [WinError 182] The operating system cannot run %1. Error loading "C:\Users\runneradmin\miniconda3\conda-bld\ilastik-gpu_1730976400239\_test_env\lib\site-packages\torch\lib\fbgemm.dll" or one of its dependencies.
```

Conclusion:

I've put the `pytorch <2.4.1` constraint in meta.yaml, and also environment-dev.yml, so we should effectively be on 2.4.0 now.

The error we see on Windows CI with 2.4.1 and up may look identical to an error widely described for pytorch 2.4.0 (https://github.com/pytorch/pytorch/issues/131662), but I think they are separate issues. The error looks identical because in both cases, a dll related to openmp is missing/broken.

Evidently, unlike a huge number of people as documented by the pytorch issue, ilastik seems to be fine with pytorch 2.4.0. I suspect everyone else was having openmp ruined by `torchaudio`, which we don't pull.

In #2904, @k-dominik describes a clash between pytorch 2.4.1 and `pyshtools` due to its dependency `gmt` overwriting openmp. An attempted workaround with a modified pyshtools package, uploaded to the conda channel `ilastik-forge/label/patched`, may have slimmed the dependencies but didn't solve the problem.

In my attempts here, `pyshtools` was also the source of the problem; but instead of `gmt`, it looks like openmp was being overwritten by `libflang`. Specifically, it was pulling openmp and libflang both at versions 5, which are rather old. I tried if pytorch 2.5.1 was able to coexist with other openmp versions in an env, and looks like `openmp=8.0.1` would have been fine. `libflang=11.*` pulled `llvm-openmp` instead of `openmp`, with the overwriting issue persisting, but `libflang=17.*` no longer pulled any openmp and was fine as well.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -116,7 +116,7 @@ outputs:
       run:
         - python 3.9.*
         - ilastik-core {{ setup_py_data.version }}
-        - pytorch >=1.6,!=2.4.1
+        - pytorch >=1.6,<2.4.1
         - tiktorch
         - torchvision
         - volumina
@@ -168,7 +168,7 @@ outputs:
       run:
         - python 3.9.*
         - ilastik-core {{ setup_py_data.version }}
-        - pytorch >=1.6,!=2.4.1
+        - pytorch >=1.6,<2.4.1
         - tiktorch
         - torchvision
         - pytorch-cuda
--- a/dev/environment-dev.yml
+++ b/dev/environment-dev.yml
@@ -49,26 +49,22 @@ dependencies:
   - aiohttp
   - fsspec
 
-  # Neural Network Workflow dependencies
+  # Deep learning dependencies (neural net workflow, trainable domain adaptation)
   # can be changed to request gpu versions
-  # pytorch 2.4.* will work for all platforms except osx-64, where currently the
-  # latest version is 2.2.2
-  - pytorch 2.4.*,!=2.4.1
+  - pytorch <2.4.1  # >=2.4 can clash with pyshtools on win; 2.2.2 newest available on osx
   - cpuonly  # comment out for pytorch with cuda
+  # - pytorch-cuda=11.3  # uncomment for pytorch with cuda, specify cuda version
+  - tiktorch 24.9.0*
   # mamba does not "respect" track_features, which is the idea behind cpuonly,
   # with ilastik-pytorch-version-helper-cpu we help mamba on linux and windows
   # to install a cpu-build
   - ilastik-pytorch-version-helper-cpu  # comment out for pytorch with cuda
-  # - pytorch-cuda=11.3  # uncomment for pytorch with cuda, specify cuda version
-
-  - tiktorch 24.9.0*
 
 
   # Third-party object feature plugins
   - sphericaltexture_plugin
   - sphericaltexture >=0.0.4
-  # pyshtools 4.10 conflicts with pytorch (via openmp) on windows
-  - pyshtools >=4.12
+  - pyshtools >=4.12  # If >=4.13 becomes available for win, might be able to unpin pytorch
 
   # dev-only dependencies
   - conda-build


k-dominik:
do you have a link to a source for "because libomp140 was no longer shipped in a VS redistributable.", or did you diff different VS redistributables yourself?

btbest:
Progress: After verifying that a barebones pytorch install works fine, I've been able to find the culprit package that we install and with which it breaks: `sphericaltexture`...

k-dominik:
do you install the patched version of pyshtools from `ilastik-forge/label/patched` ?

btbest:
Found the culprit; the patched pyshtools didn't fix it. It's because sphericaltexture depends on openmp 5.0.0. Installing 5.0.0 is sufficient to cause the issue even with pytorch 2.5.1. Installing 8.0.0 doesn't cause the issue.

btbest:
```
(torchtest) >conda-tree whoneeds openmp
libflang

(torchtest) >conda-tree whoneeds libflang
pyshtools

(torchtest) >conda-tree depends pyshtools
astropy
cartopy
ducc0
fftw
libblas
libflang
liblapack
libpgmath
matplotlib-base
numpy
pooch
python
python_abi
requests
scipy
ucrt
vc
vc14_runtime
xarray
```

But `libflang` doesn't show up in the shtools repo

k-dominik:
I had a look of at such an environment and I see at least two openmp libs, `openmp` and `intel-openmp`... if you dig into these packages a bit you can find that both of them come with a version of `libiomp5md.dll` (note the different sizes, hashes):

```
# intel-openmp
      {
        "_path": "Library/bin/libiomp5md.dll",
        "path_type": "hardlink",
        "sha256": "3e69ef3e52deff22ba0e44c36e1198dcae1657098e63a2d550842b717eb59175",
        "sha256_in_prefix": "3e69ef3e52deff22ba0e44c36e1198dcae1657098e63a2d550842b717eb59175",
        "size_in_bytes": 1310600
      },


# openmp:
      {
        "_path": "Library/bin/libiomp5md.dll",
        "path_type": "hardlink",
        "sha256": "cf43b5e78b8664c734366ad99bbe7a367fe50d3844f267c920f1e6280d3c56be",
        "sha256_in_prefix": "cf43b5e78b8664c734366ad99bbe7a367fe50d3844f267c920f1e6280d3c56be",
        "size_in_bytes": 612864
      },

```

not sure why conda doesn't issue a clobber warning.

My guess would be this causes the problem. So the pulling in of openmp per se is not the problem - but clobbering is.

versions of openmp aside, you can also "fix" this with `conda install -c conda-forge intel-openmp --force-reinstall` which just overwrites `libiomp5md.dll` and friends with an ABI compatible version that pytorch was build with.

btbest:
> I had a look of at such an environment and I see at least two openmp libs, `openmp` and `intel-openmp`... if you dig into these packages a bit you can find that both of them come with a version of `libiomp5md.dll` (note the different sizes, hashes)

🤦 I hadn't noticed that. Not just with openmp btw; in an attempt with a newer `libflang` (18.1.1), `llvm-openmp` clobbers it the same way.

> not sure why conda doesn't issue a clobber warning.

but it's happy to issue dozens of them on our CI, none of them seemingly an issue :( 

> My guess would be this causes the problem. So the pulling in of openmp per se is not the problem - but clobbering is.
> 
> versions of openmp aside, you can also "fix" this with `conda install -c conda-forge intel-openmp --force-reinstall` which just overwrites `libiomp5md.dll` and friends with an ABI compatible version that pytorch was build with.

This does seem to work; so I think we can choose to pin pytorch 2.3, or unpin pytorch and do this force-reinstall instead. Do you have a preference for either? Neither is really clean. The force-reinstall would extend CI duration. We could even pin for CI and force-reinstall for release, though CI durations aren't enough of a bottleneck to warrant that extra complexity I think.

k-dominik:
strong preference for pinning `pytorch<2.4` - this allows at least for _some_ flexibility when creating ilastik envs - in release we pin to `2.3` currently.

The reinstall "hack" will break something else - I'm sure :)

btbest:
> do you have a link to a source for "because libomp140 was no longer shipped in a VS redistributable.", or did you diff different VS redistributables yourself?

For reference, I had read that [over here](https://github.com/kohya-ss/sd-scripts/issues/1597#issuecomment-2350806417):

> fbgemm.dll seems to require libomp140.x86_64.dll, and that one was removed from the Windows 11 C++ redistributables, as it seems

But as it turns out isn't related to the issue we have here.

btbest:
Fixes #2922

btbest:
I've updated the original post to summarise my understanding of what's going on now for future reference

k-dominik (approved):

PR #2928, opened 2024-11-19
Add logging widget to status bar

k-dominik:
Log messages that are emitted using the log level 23 (or `lazyflow.USER_LOGLEVEL`) will show up in the status bar. Double clicking that area will open the log history.

image

Currently this will show the number of labels and oob in pixel classification, object classification (and related workflows - tracking...) as well as multicut.

- [x] Add tests.
- [couldn't find one] Reference relevant issues and other pull requests.

### DIFF
--- a/ilastik/shell/gui/ilastikShell.py
+++ b/ilastik/shell/gui/ilastikShell.py
@@ -29,7 +29,7 @@ import numbers
 import platform
 import threading
 import warnings
-from typing import List, Type, Optional
+from typing import Optional, Type, Union
 
 # SciPy
 import numpy
@@ -151,8 +151,8 @@ class MemoryWidget(QWidget):
         self.label.setText("Cached Data: %1.1f MB" % (bytes / (1024.0**2.0)))
 
 
-class LogWindow(QTextEdit):
-    """Window/Widget that shows the log history"""
+class NotificationsWindow(QTextEdit):
+    """Window/Widget that shows a history of notifications to users"""
 
     def __init__(self, data, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -180,15 +180,19 @@ class LogWindow(QTextEdit):
         self._jump_to_end()
 
 
-class LogBar(QLabel):
-    """Log widget for the status bar"""
+class NotificationsBar(QLabel):
+    """Notifications widget for the status bar
+
+    The notification status bar / NotificationsWindow combo shows messages that
+    are emitted with `lazyflow.USER_LOGLEVEL`.
+    """
 
     textAdded = pyqtSignal(str)
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._history = []
-        self.log_widget: LogWindow | None = None
+        self.notifications_widget: Union[NotificationsWindow, None] = None
         self.setToolTip("Double click for history.")
         self._setup_logging()
 
@@ -209,18 +213,18 @@ class LogBar(QLabel):
         r.addHandler(s)
 
     def mouseDoubleClickEvent(self, event):
-        if self.log_widget is None:
-            self.log_widget = LogWindow(self._history)
-            self.textAdded.connect(self.log_widget.appendMessage)
+        if self.notifications_widget is None:
+            self.notifications_widget = NotificationsWindow(self._history)
+            self.textAdded.connect(self.notifications_widget.appendMessage)
 
-        self.log_widget.show()
-        self.log_widget.raise_()
+        self.notifications_widget.show()
+        self.notifications_widget.raise_()
 
     def cleanUp(self):
         self._history = []
         self.setText("")
-        if self.log_widget is not None:
-            self.log_widget.clear()
+        if self.notifications_widget is not None:
+            self.notifications_widget.clear()
 
     def write(self, text):
         text = text.rstrip("\n")
@@ -257,7 +261,7 @@ class ProgressDisplayManager(QObject):
         self.statusBar.addWidget(self.progressBar)
         self.progressBar.setHidden(True)
 
-        self.logBar = LogBar()
+        self.logBar = NotificationsBar()
         self.statusBar.addPermanentWidget(self.logBar)
 
         self.requestStatus = QLabel()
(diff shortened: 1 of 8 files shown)


ilastik/shell/gui/ilastikShell.py
```diff
@@ -160,9 +151,88 @@ def setMemoryBytes(self, bytes):
...
+
+class LogBar(QLabel):
+    """Log widget for the status bar"""
+
+    textAdded = pyqtSignal(str)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._history = []
+        self.log_widget: LogWindow | None = None
```
btbest, line 191: Invalid typing syntax for Python 3.9

I'm not sure what the interpreter actually does here, since it's not throwing or warning either...

ilastik/shell/gui/ilastikShell.py
```diff
@@ -160,9 +151,88 @@ def setMemoryBytes(self, bytes):
...
+    def showEvent(self, event):
+        super().showEvent(event)
+        self._jump_to_end()
+
+    def appendMessage(self, msg: str):
+        self.append(msg)
+        self._jump_to_end()
+
+
+class LogBar(QLabel):
```
btbest, line 183: I think the UI feature this is trying to provide is notifications, so that's what I'd all it, even if it's currently implemented using a special log level

ilastik/shell/gui/ilastikShell.py
```diff
@@ -160,9 +151,88 @@ def setMemoryBytes(self, bytes):
...
+
+    def __init__(self, data, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._setup_ui(data)
+
+    def _jump_to_end(self):
+        self.verticalScrollBar().setValue(self.verticalScrollBar().maximum())
+
+    def _setup_ui(self, data):
+        self.setWindowTitle("Log")
```
btbest, line 165: I can foresee scenarios where users discover this feature, but never the "real" log. Then we ask them to provide logs on the forum and they think we're asking them for the contents of this box...

I'd call it "Notifications" (see below)

lazyflow/__init__.py
```diff
@@ -25,6 +23,11 @@
 import json
 import numpy as np
 
+# This level can be used to issue user-facing log messages
+# log messages at this level get special treatment in ilastik:
+# those are shown in the status bar
+USER_LOGLEVEL = 23
```
btbest, line 29: I get that it's 23 to position it between INFO and WARN, and I think that's a good place. But look at the full message in the log:

`Level 23 2024-12-02 18:15:34,494 parallelVigraRfLazyflowClassifier 4464 21876 Training took, 0.007 seconds`

Such a good setup for an easter egg number. "Level 9001"

tests/test_ilastik/test_shell/test_status_log.py
```diff
@@ -0,0 +1,93 @@
...
+
+    some_text = "ilastik logbar"
+    logger.log(USER_LOGLEVEL, some_text)
+    assert log_bar.text() == some_text
+
+
+def test_double_click_opens_window(log_bar: LogBar, qtbot):
+    qtbot.mouseDClick(log_bar, Qt.MouseButton.LeftButton)
+
+    with qtbot.waitExposed(log_bar.log_widget):
```
btbest, line 55: What's up with the `with...pass` instead of just calling the function? (Seems to work fine for me, is it just because my machine is fast enough?)

btbest (commented):
I really like it.

What do you think about clearing the bar when switching applets (in `ilastikShell.setSelectedAppletDrawer` I guess)? To me it feels a bit odd/unresponsive for the feedback to hang around there for the rest of the workflow; or respectively in Autocontext to hang around and then get replaced in stage 2.

lazyflow/__init__.py
emilmelnikov: An alternative way to implement this is to prepend a special well-known string prefix to a log message (could be something that still reads well in isolation, e.g. `NOTICE: ` or `USER INFO: `). This way, one could show WARN/ERROR messages in the status bar too if needed (maybe even in a red-ish color or something).

It is unfortunate that Python's default logging doesn't support attribute metadata (e.g. like the `structlog` package does).

k-dominik: @emilmelnikov: Good point. I considered such a pattern but thought for some reason that concept would be even more alien than doing the custom log level. But since you also thought of it :)...

@btbest: had to google 9001 and am still not sure I get the reference :D...

tests/test_ilastik/test_shell/test_status_log.py
k-dominik: hm - don't remember tbh. works fine without also on ci :) good call, never good if such patterns creep in without reason.

lazyflow/__init__.py
emilmelnikov: Come to think of it, a more conventional way to do this would be to implement an ad-hoc structured logging: optionally prepend `{attr1=val1,attr2=val2} ` to a message, then try to find and parse it into a dict:

```python
def parse_log_attrs(line: str) -> tuple[dict[str, str], str]:
    prefix, sep, msg = line.partition(" ")
    if not (sep and len(prefix) >= 2 and prefix[0] == "{" and prefix[-1] == "}"):
        return {}, line
    attrs = {}
    for kv in prefix[1:-1].split(","):
        key, sep, val = kv.partition("=")
        if not (key and sep and val):
            return {}, line
        attrs[key] = val
    return attrs, msg
```

This is not as flexible as a full-blown proper JSON structured logging system, but it should be enough to smuggle a bunch of simple tags, and everything else should work as before.

k-dominik: I like the idea and if we see that we use that "feature" more I think we should consider something like this. For now I prefer to keep it conceptually simple. But whenever we find we need more fine-grained control, we'll have  this idea in the backs of our heads.

btbest: > @btbest: had to google 9001 and am still not sure I get the reference :D...

It would be [over 9000](https://en.wikipedia.org/wiki/It%27s_Over_9000!) 😅

btbest (approved):
❤️

k-dominik:
thank you @btbest and @emilmelnikov for the discussions/reviews!

ilastik/shell/gui/ilastikShell.py
btbest: Oops, I think the window title slipped through still saying "Log" 😅

k-dominik: yes yes I realized directly after clicking merge  ;).

PR #2935, opened 2025-01-20
Switch to .conda package format (with latest conda build)

k-dominik:
`.conda` is the default package format since conda build `25.1.0`


btbest (approved):
lgtm

just wondering how CI was green in the previous runs - I guess there's still backwards compatibility that looks for tarballs as a fallback

k-dominik:
There hasn't been a build with the new conda-build before. Today `.tar.bz2` would fail.

PR #2937, opened 2025-01-21
Fix persisting NotificationsWindow during test runs

k-dominik:
* since the tests that involve both `NotificationsBar` and its `NotificationsWindow`, it needs to be added to `qtbot` in order to get closed automatically. In order to allow for this the `NotificationsWindow` is constructed when `NotificationsBar` is constructed.
* Fixed parent-child relationship in `NotificationsBar`.
* also changed all `qtbot.mouseDoubleClickEvents` to explicit invocation of the `NotificationsBar.mouseDoubleClickEvent` as it's more reliable according to pytest-qt docs.
* cleaned up constructors a bit.
* fixed a typo when logging from classifier training.
* fixed naming from `log_whatever` to `notifications_whatever`.
* rename test file for consistency.

cleanup of #2928

Note: separate commits to ease reviewing - Squash and merge...


tests/test_ilastik/test_shell/test_status_notification.py
```diff
@@ -0,0 +1,91 @@
...
+        notifications_bar.mouseDoubleClickEvent(_event=None)
+
+    assert id(notifications_bar.notifications_widget) == original_id
+
+
+def test_clear(notifications_bar: NotificationsBar, qtbot, enable_log):
+    assert notifications_bar.text() == ""
+
+    for i in range(42):
+        logger.log(USER_LOGLEVEL, f"msg {i}")
```
btbest, line 78: This logger call triggers
```
--- Logging error ---
  File "~\.conda\envs\domilastik\lib\logging\__init__.py", line 1086, in emit
    stream.write(msg + self.terminator)
  File "~\ilastik\ilastik\shell\gui\ilastikShell.py", line 237, in write
    self.setText(short_log)
RuntimeError: wrapped C/C++ object of type NotificationsBar has been deleted
Call stack:
  File ... (bunch of pytest stack)
  File "~\ilastik\tests\test_ilastik\test_shell\test_status_notification.py", line 79, in test_clear
    logger.log(USER_LOGLEVEL, f"msg {i}")
Message: 'msg 0'
Arguments: ()
```
(times 42 due to the loop)

I tried moving this inside the `with qtbot.waitExposed` block, but that didn't fix it.

Test still passes though...

k-dominik: this unfortunately happens if you call your tests with `pytest -s`, didn't really investigate as I rarely do it - but it must be something related to log streams being hijacked by pytest.

k-dominik: https://github.com/ilastik/ilastik/pull/2937/commits/56aaca6ccbe91d6ec0287c50091e03b02bf149ef is the best I can do without digging super deep - skipping those tests if log is not captured. But definitely better than before!

btbest: Since it seems to be specific to my IDE, I've been trial-and-erroring:

When I run only `test_clear`: no errors ✅ 

When I run "all tests in this file", errors in `test_clear` as I described above 🔴 

When I run all ilastik tests (the top-level `tests` module): no errors ✅ (in `test_clear`)

So getting these errors inside `test_clear` is specific to running "all tests in this file".

However, while running all ilastik tests, I instead found the same errors this time from `tests/test_lazyflow/test_classifiers/testParallelVigraRfLazyflowClassifier.py::TestParallelVigraRfLazyflowClassifier::test_basic` and `test_with_feature_importance` - with their respective log messages (e.g. `Message: 'Training complete. Label counts: (40, 60). Average OOB: 0.055'`).

But I don't get those errors when I run only `TestParallelVigraRfLazyflowClassifier`, or all of `test_classifiers`, or all of `test_lazyflow`. They only happen when I run all `tests`. And best of all, they even happen when I run on this PR's base commit instead of its head (but still only when running all `tests`).

In summary, I have no idea anymore and maybe it's best to ignore this. Probably the skip switch for `-s` is sensible though, even if it's not really relevant for either of us..?

k-dominik: Ouch, I really dislike having this side effect on testing because we use logging for the notifications...

btbest: fixed 😍

btbest (approved):
100% worth it 🥳

PR #2939, opened 2025-01-23
OME-Zarr: Report v0.5 not supported

btbest:
Supporting OME-Zarr v0.5 will require zarr v3, which we currently can't have in our env because python-elf restricts zarr<3.

But we can show a nicer error message than `400 Bad Request` (which is what happens e.g. on `https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.5/idr0066/ExpD_chicken_embryo_MIP.ome.zarr` due to the "try parent folders" mechanism - https://uk1s3.embassy.ebi.ac.uk/.zattrs gives 400 instead of 404)


k-dominik (approved):
lgtm, tested - works - very nice :)

PR #2940, opened 2025-01-24
OME-Zarr: Support reading from URIs with s3: scheme

btbest:
This enables reading OME-Zarr from URIs pointing to AWS S3 and S3-compatible objects.

## AWS S3

You can try it live with a public bucket I set up for testing: `s3://btbest-public/img.zarr/`

While it's not super user friendly, this also at least enables users to access S3 buckets that require authentication _at all_, by setting up a credentials file in the default location (`~/.aws/credentials`). Alternatively, one can invoke ilastik with the credentials as env variables (`AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`).

You cannot try it with the private bucket that I also set up to try the authenticated access :) (at least not without some private communication)

Setting up your own private bucket for testing is quite a bit of work. You have to:

1. Create an AWS account (which requires payment information)
2. Set up billing rules to make sure you get at least an alert if something shady happens and you incur charges
3. Create a private S3 bucket
4. Upload some OME-Zarr folder to it
5. Create a new user in your IAM, and a new user group to assign it into, and an access policy for the group that gives readonly for S3
6. Create an access key for the user
7. Store the [access key and secret in a file](https://github.com/aws/aws-cli/?tab=readme-ov-file#configuration) called `~/.aws/credentials`

This file has to be present _before_ you run ilastik, botocore seems to only pick it up on first load/import.

Considering all the work involved in just setting this up, maybe we don't even need to make authentication more convenient than that. People who want to use S3 might already be accustomed to authenticating this way and it will "just work" for them.

## S3-compatible

All of the public OME-Zarr datasets on IDR are actually hosted on S3-compatible storage, so the publicly accessible ones have been supported since day 1 (e.g. https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457227.zarr)

Access requiring authentication apparently works according to email with collaborators. I admittedly haven't tried it with s3.embl.de.

### DIFF
--- a/lazyflow/operators/ioOperators/opInputDataReader.py
+++ b/lazyflow/operators/ioOperators/opInputDataReader.py
@@ -100,6 +100,9 @@ class OpInputDataReader(Operator):
         h5_n5_Exts + n5Selection + npyExts + npzExts + rawExts + vigraImpexExts + blockwiseExts + videoExts + klbExts
     )
 
+    supported_ome_zarr_schemes = ["http:", "https:", "file:", "s3:"]
+    precomputed_protocol = "precomputed://"
+
     if _supports_dvid:
         dvidExts = ["dvidvol"]
         SupportedExtensions += dvidExts
@@ -301,7 +304,7 @@ class OpInputDataReader(Operator):
             return ([], None)
         if not isUrl(filePath):
             filePath = Path(filePath).as_uri()
-        if not any(filePath.startswith(scheme) for scheme in ["http", "file", "s3"]):
+        if not any(filePath.startswith(scheme) for scheme in self.supported_ome_zarr_schemes):
             return ([], None)
         # DatasetInfo instantiates a standalone OpInputDataReader to obtain laneShape and dtype.
         # We pass this down to the loader so that it can avoid loading scale metadata unnecessarily.
@@ -311,7 +314,7 @@ class OpInputDataReader(Operator):
         return [reader], reader.Output
 
     def _attemptOpenAsRESTfulPrecomputedChunkedVolume(self, filePath):
-        if not filePath.lower().startswith("precomputed://"):
+        if not filePath.lower().startswith(self.precomputed_protocol):
             return ([], None)
         reader = OpRESTfulPrecomputedChunkedVolumeReader(parent=self)
         reader.Scale.connect(self.ActiveScale)
--- a/lazyflow/operators/ioOperators/opInputDataReader.py
+++ b/lazyflow/operators/ioOperators/opInputDataReader.py
@@ -301,7 +301,7 @@ class OpInputDataReader(Operator):
             return ([], None)
         if not isUrl(filePath):
             filePath = Path(filePath).as_uri()
-        if not (filePath.startswith("http") or filePath.startswith("file")):
+        if not any(filePath.startswith(scheme) for scheme in ["http", "file", "s3"]):
             return ([], None)
         # DatasetInfo instantiates a standalone OpInputDataReader to obtain laneShape and dtype.
         # We pass this down to the loader so that it can avoid loading scale metadata unnecessarily.
--- a/lazyflow/utility/io_util/OMEZarrStore.py
+++ b/lazyflow/utility/io_util/OMEZarrStore.py
@@ -31,7 +31,7 @@ import jsonschema
 import s3fs
 import vigra
 from aiohttp import ClientConnectorError, ClientResponseError
-from botocore.exceptions import NoCredentialsError
+from botocore.exceptions import NoCredentialsError, EndpointConnectionError
 from zarr.core import Array as ZarrArray
 from zarr.errors import ArrayNotFoundError
 from zarr.storage import FSStore, LRUStoreCache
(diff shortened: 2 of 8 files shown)


k-dominik:
`s3fs` (and `fsspec`) `2022.8.2` work for me - could be  a good candidate for a minimal pin.

lazyflow/operators/ioOperators/opInputDataReader.py
```diff
@@ -301,7 +301,7 @@ def _attemptOpenAsOmeZarrUri(self, filePath):
             return ([], None)
         if not isUrl(filePath):
             filePath = Path(filePath).as_uri()
-        if not (filePath.startswith("http") or filePath.startswith("file")):
+        if not any(filePath.startswith(scheme) for scheme in ["http", "file", "s3"]):
```
k-dominik, line 304: it would be a bit more in tune with the rest of the file to declare a class variable in `OpInputDataReader` that lists these (same for precomputed scheme). Also is it will the scheme not always include `://` as well?

lazyflow/utility/io_util/OMEZarrStore.py
```diff
@@ -256,13 +257,42 @@ def _get_zarr_cache_max_size() -> int:
...
     if uri.startswith("file:"):
         # Zarr's FSStore implementation doesn't unescape file URLs before piping them to
         # the file system. We do it here the same way as in pathHelpers.uri_to_Path.
         # Primarily this is to deal with spaces in Windows paths (encoded as %20).
         uri = os.fsdecode(unquote_to_bytes(uri))
-    return FSStore(uri, mode=mode, **kwargs)
+    store = FSStore(uri, mode=mode, **kwargs)
+    try:
+        store["test"]
+        # Any error not handled here is either a successful connection (404 for "test"), or an unknown problem
```
k-dominik, line 269: not sure I understand the motivation here. So you try to access the key "test" in the store to test if you have access to the store in general? What if the object behind "test" is a 10GB file? "test" is probably not the most uncommon random key for stuff (I at least at some point had large random files called "test" around to ... test stuff).

lazyflow/utility/io_util/OMEZarrStore.py
```diff
@@ -256,13 +257,42 @@ def _get_zarr_cache_max_size() -> int:
...
+    except KeyError as e:
+        # FSStore wraps some errors in KeyError (FileNotFoundError even double-wrapped)
+        if isinstance(e.__context__, ClientConnectorError):
+            raise ConnectionError(f"Could not connect to {e.__context__.host}:{e.__context__.port}.") from e
+    except ClientResponseError as e:
+        if not (e.status == 403 and uri.startswith("https://")):
+            raise
+        # 403 indicates we need authentication; FSStore maps https: to HTTPFileSystem, which cannot authenticate
```
k-dominik, line 277: it would be nice to give some sort of feedback to the user that you are trying to authenticate (and how) in the logs at least

lazyflow/utility/io_util/OMEZarrStore.py
```diff
@@ -256,13 +257,42 @@ def _get_zarr_cache_max_size() -> int:
...
+        sub_uri = split_bucket[4]
+        fs = s3fs.S3FileSystem(anon=False, endpoint_url=base_uri_inc_bucket)
+        store = FSStore(sub_uri, fs=fs, **kwargs)
+        try:
+            store["test"]
+        except KeyError as ke:
+            if isinstance(ke.__context__, PermissionError):
+                # This probably is an S3-compatible store, but auth rejected/not found.
+                raise ConnectionError(
+                    f"Server refused permission to read {uri}.\nCheck your S3 credentials setup."
```
k-dominik, line 293: Same here, a hint on the s3 credential setup might be good (or is it so heterogeneous that it is not possible to hint at something?)

k-dominik (commented):
Hey this looks good in principle. I couldn't test it with the public bucket you are providing (getting `Could not connect to the endpoint URL: "http://169.254.169.254/latest/api/token`). Nor do I have access to any ome-zarr images on private buckets. But since @constantinpape et al have tried it (if those are the collaborators you mention) I'm good.

lazyflow/utility/io_util/OMEZarrStore.py
btbest: There are three different ways they might be authenticating with their bucket (creds file, env variables, token). It's a bit too impractical for ilastik to determine which way it is here just for a nicer error. Plus, all of those would have been manually set up by the user, so they should know how they are doing it (or who set it up for them)...

lazyflow/utility/io_util/OMEZarrStore.py
btbest: Yeah should go for a key with higher chance to go 404...

lazyflow/operators/ioOperators/opInputDataReader.py
btbest: The `:` would be guaranteed, but file URIs don't necessarily continue with a / after that.

btbest:
The connection problem is worrying - it's something about not reaching an AWS bucket metadata endpoint and I can't reproduce it even with s3fs=2022.8.2... it's possible that it's related to the DNS setup on the router you were behind (in which case, nothing we can do from our code), so we should confirm that it works if we are on the same LAN. Hopefully it's not a Mac issue.

k-dominik (approved):
lgtm

PR #2941, opened 2025-01-28
bump vigra to 1.11.1 build 1038

k-dominik:
build 1038 is the latest version of vigra built against boost 1.78.0 (which is the latest boost version we have builds of our packages for, that depend on it).

Note for windows devs (@btbest): It's now necessary to include `ilastik-forge/label/patched` if you want to create a development env close to the release env (where the patched pyshtools is already used).

### DIFF
--- a/.github/workflows/core-conda-bld.yml
+++ b/.github/workflows/core-conda-bld.yml
@@ -134,7 +134,7 @@ jobs:
         run: |
           set VOLUMINA_SHOW_3D_WIDGET=0
           conda build --test --override-channels ^
-            -c ./pkgs -c pytorch -c nvidia -c ilastik-forge -c conda-forge ^
+            -c ./pkgs -c pytorch -c nvidia -c ilastik-forge/label/patched -c ilastik-forge -c conda-forge ^
             ./pkgs/noarch/%ILASTIK_PACKAGE_NAME%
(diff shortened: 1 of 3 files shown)


btbest (approved):
Env created without issues and ilastik looks to be running fine 👍

btbest:
Github Actions are red atm (on main) due to a numpy update to 1.26.4 (and our meta.yaml only says `numpy > 1.12`). One issue is a deprecated import via `numpy.testing.utils`, which no longer exists.

But after fixing that, the remaining issue is in `OpLazyConnectedComponents.growRegion`, where `vigra.analysis.unique(extendingLabels)` triggers `NumpyArray::setupArrayView(): only singleton axes may have zero stride.`

I see that meta.yaml doesn't pin vigra, so the actions pull `1.11.2`. What do you think should be the approach here? Pin vigra also in meta.yaml (in which case it would fit in this PR?), or does the error sounds fixable on our side?

k-dominik:
I cool that you already debugged it all. I'll add a pin to the recipe as well and merge right away.

k-dominik:
Okay 😞 it looks like there are problems with vigra and numpy>1.22 - I'll add an upper pin for numpy for now (1.22, since we know everything works with that version).

PR #2948, opened 2025-02-05
Added drag-n-drop support for projects to startup page

k-dominik:
Going via the "open file" dialog always is a pain.

Also since I'm already touching files here: replaced twitter link with link to bluesky ❤️ 

Todos:

- [x] Add tests

### DIFF
--- a/ilastik/shell/gui/ilastikShell.py
+++ b/ilastik/shell/gui/ilastikShell.py
@@ -1,7 +1,7 @@
 # ##############################################################################
 #   ilastik: interactive learning and segmentation toolkit
 #
-#       Copyright (C) 2011-2014, the ilastik developers
+#       Copyright (C) 2011-2025, the ilastik developers
 #                                <team@ilastik.org>
 #
 # This program is free software; you can redistribute it and/or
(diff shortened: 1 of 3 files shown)


ilastik/shell/gui/ilastikShell.py
```diff
@@ -396,6 +396,51 @@ def handleAppletProgressImpl(self, index: int, percentage: int, cancelled: bool)
...
+            return
+        mimedata = a0.mimeData()
+        if mimedata is None:
+            return
+        # Only accept drag-and-drop events that consist of urls to local .ilp files.
+        if not mimedata.hasUrls():
+            return
+        urls = mimedata.urls()
+
+        if len(urls) != 1:
```
emilmelnikov, line 436: Wouldn't it be more convenient/forgiving to allow dropping multiple items, and just pick the first `.ilp` from the list? Maybe even allow to drop a folder, and search for a project file inside.

I could see this happening if e.g. someone likes to drag-select things in a file manager instead of just clicking.

Obviously, the tradeoff here is a potential confusion "What project did I open?.."

k-dominik: Hello @emilmelnikov, first of all thank you for looking into this!!

After reading your comment I thought about it and I would prefer this to be very predictable. If you drop a folder with multiple project files what would you expect? Would you follow the full tree?

The question is whether something more helpful should happen than what is implemented here (nothing, will not accept unsuitable drops). One could still take the drop and inform the user that only a single `.ilp` file can be dropped?

emilmelnikov: Yes, this is probably the best solution, since ilastik is used to analyze "serious" data. What if someone has two similar projects nearby and opens the wrong one?..

I've suggested this because I've seen some desktop apps allow you to DnD any files into a resource manager window, and adds only those that it recognizes.

k-dominik: Yes, I thought about different scenarios, too, the one you mention included. But this I expect only from applications that somehow have the concept of many files. For the softwares that have the concept of "projects" I wouldn't expect them to handle them. But yes, I'll warn if there were more than one project file :)

Issue #2925, opened 2024-10-28
MultiscaleDatasetBrowser accepts URIs that the backend cannot handle


btbest:
(Reported by @lorenzocerrone)

The problem stems from path handling. The multiscaleDatasetBrowser takes the full URI as given, and looks for `uri/.zattrs`. If that file contains a valid multiscale spec, it accepts it and shows the scales.

The backend instead splits the uri into external/internal path based on the `.zarr` extension, and checks for `externalPath/.zattrs`.

If `uri != externalPath`, this can lead to the GUI finding URIs valid that the backend then cannot handle. Most commonly, examples like `https://host.com/path/store.zarr/labels/nuclei`, where `/nuclei/.zattrs` is a valid multiscale spec (of the nuclei labels), but `/store.zarr/.zattrs` is e.g. a plate spec or just doesn't exist.

## Fix

Quick patch: split paths at extension in`OMEZarrStore` the same way `OpInputDataReader` does

Real fix: Support multiscales with arbitrary store roots

btbest:
This specific issue is fixed, so for now the front- and backend are in sync regarding what URIs they report as compatible.

In the current state, it's possible though for `OMEZarrStore` to become able to handle URIs with different schemes than the ones `OpInputDataReader` knows, which would lead to the same issue of the dialog reporting that it's fine, and then erroring when trying to add it to the project. Unlikely to happen (through inattentiveness, or backend package updates), but not impossible.

PR #2956, opened 2025-02-24
OME-Zarr: Sync front- and backend check for compatibility

btbest:
Tiny change that should have 0 functional impact.

This is to ensure that we don't get in a situation like https://github.com/ilastik/ilastik/issues/2925 again, where the multiscale URI selector dialog checks and confirms compatibility for a dataset, and then once the user clicks "Add to project", they get an error because the backend checks the compatibility differently.

By using the same function to check compatibility (`OMEZarrStore.is_uri_compatible`), whatever passes the dialog should also pass the backend.


k-dominik (approved):

btbest:
Actually should adjust the `URI_HINT` to reflect the supported schemes.

PR #2958, opened 2025-02-28
Quickfix: Same object size limits in tracking and oc

### LINKED ISSUE #2810: Tracking: Size filter max needs better default value
## Description

Undersegmented objects can get quite large. Our default max size filter setting in tracking (_Tracking_ applet) has a value of `10000`. There is also the one in the _Thresholding_ step, which has a higher default value. It would be better if these values would be in sync, at least in the beginning. With the current setup, users might _see_ objects (as in having a mask) that get filtered out during tracking (with a setting that is buried quite low in the ui, needs scrolling to on most screens).

k-dominik:
This commit sets the size filter initial values to the same ones than as in thresholding. Also same initial value in all tracking flavors.

Tracking reuses some elements from Object Classification. E.g. the thresholding applet. For the actual tracking there is another size-filter step, as there is one in thresholding, too. These don't necessarily need to be in sync, but it's better to start off that way.

fixes #2810

Ideally these values would come from a unique source to keep them in sync also in the future, but every time I look at tracking I think it's either complete rewrite or minimal fix...

### DIFF
--- a/ilastik/applets/thresholdTwoLevels/drawer.ui
+++ b/ilastik/applets/thresholdTwoLevels/drawer.ui
@@ -393,7 +393,7 @@
             </sizepolicy>
            </property>
            <property name="toolTip">
-            <string>Min object area in pixels. Cannot be higher than the max value of a 32-bit integer (2147483647).</string>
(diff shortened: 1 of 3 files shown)


btbest (approved):
lgtm

PR #2960, opened 2025-03-04
Fix show model name for models from ilp

k-dominik:
Tiny fix for a problem that I found just now when poking the workflow a bit more...

Previously, if a model loaded from ilp was not in the list of models from the zoo, the field would say "remove model". This commit fixes it by always setting the current text.


btbest (approved):
Wasn't sure how to test this, but apparently at the moment downloading the pytorch state dict of affable-shark directly from the zoo and loading the zip here works :D

PR #2961, opened 2025-03-04
Add cpu fallback for tiktorch server on macos silicon

k-dominik:
currently, pytorch does not support all operators, especially in 3D on apple silicon (mps) devices. Setting the env variable "PYTORCH_ENABLE_MPS_FALLBACK". for the tiktorch server process allows running these networks on CPU instead of failing.

Looks like this will still be an issue for a while. With this env variable set, pytorch will decide to fall back to CPU if the network cannot be run on MPS devices.

### DIFF
--- a/ilastik/workflows/neuralNetwork/_localLauncher.py
+++ b/ilastik/workflows/neuralNetwork/_localLauncher.py
@@ -21,6 +21,7 @@
 import json
 import logging
 import os
+import platform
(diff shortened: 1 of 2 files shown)


btbest (approved):
Not like I have a machine to actually try this out, but looks good and tests run fine :)

k-dominik:
thank you @btbest!

Issue #2621, opened 2022-10-11
Future of Pixel + Object Classification Workflow


k-dominik:
As far as I can remember we discourage usage of this, combined, workflow. Because of thresholding accessing the whole image ram requirements usually explode and result of ilastik being force-closed by the system. Even with some form of lazy connected components this danger would still depend on how objects cover multiple tiles.

Furthermore, this workflow has a number of issues that would take a long time to fix.

With @akreshuk we decided to remove this workflow. I propose we start deprecating this workflow actively in `1.4.1` and remove it completely for `1.5.0`.

PR #2962, opened 2025-03-04
Add deprecation notice for PC+OC workflow

k-dominik:
In https://github.com/ilastik/ilastik/issues/2621 we agreed to deprecate and ultimately remove this workflow due to the many issues it introduces (huge memory consumption, unreliability...).

I added "deprecated" to the workflow name, and also added a note on the data selection applet, as well as a logger warning to inform headless users, too.

I added a "welcome message" option some time in the past, but it was slashed ;) https://github.com/ilastik/ilastik/pull/2333. Could be revived for more aggressive deprecation...

Edit: somewhere the console/log warning is filtered out 🙄

### DIFF
--- a/ilastik/workflows/objectClassification/objectClassificationWorkflow.py
+++ b/ilastik/workflows/objectClassification/objectClassificationWorkflow.py
@@ -601,7 +601,7 @@ class ObjectClassificationWorkflow(Workflow):
 
 class ObjectClassificationWorkflowPixel(ObjectClassificationWorkflow):
     workflowName = "Object Classification (from pixel classification)"
-    workflowDisplayName = "Pixel Classification + Object Classification"
+    workflowDisplayName = "Pixel Classification + Object Classification (deprecated)"
 
     @property
     def ExportNames(self):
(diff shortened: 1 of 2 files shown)


btbest:
While we're at it, how about removing the (beta) from tda, given that it's no longer in beta now? :)

k-dominik:
Hey @btbest thank you for the reminder about the tda beta qualifier.

Note: i changed the warning category from `DeprecationWarning` to `UserWarning` as deprecationswarnings are [apparently ignored if not in python dev mode](https://docs.python.org/3/library/exceptions.html#DeprecationWarning).

btbest (approved):
I'm ok with it like this, though I'd consider upgrading the `warnings.warn` to a `logger.error` so that it shows up in the console on Windows and Linux rather than just the logfile (warnings.warn gets filtered out because the default logging mode is `output_mode == OutputMode.LOGFILE_WITH_CONSOLE_ERRORS`).

```
Starting ilastik from C:\Users\me\ilastik
ERROR 2025-03-05 16:06:04,899 objectClassificationWorkflow 20252 8308 The Pixel Classification + Object Classification workflow is deprecated and will be removed. Please use Pixel Classification and Object Classification as separate workflows.
```

meh, also not that great with it saying ERROR 🙃

k-dominik:
I thought it's okay to show the warning in console only in headless mode (there it'll show up in both, console and log file), whereas in gui mode, the warning is there in text in the UI.

PR #2963, opened 2025-03-06
Fix: forgot to add tda without beta to list of workflows :[

k-dominik:
## Checklist

- [x] Add tests.

### DIFF
--- a/ilastik/shell/gui/ilastikShell.py
+++ b/ilastik/shell/gui/ilastikShell.py
@@ -612,14 +612,18 @@ class IlastikShell(QMainWindow):
         self.onNewProjectActionTriggered(workflow_class)
 
     def getWorkflow(self, w: Optional[str] = None) -> Type[Workflow]:
-        listOfItems = [workflowDisplayName for _, __, workflowDisplayName in getAvailableWorkflows()]
+        listOfItems = [
+            workflowDisplayName
+            for wf_class, __, workflowDisplayName in getAvailableWorkflows()
+            if getattr(wf_class, "show_in_startup_menu", True)
+        ]
         if w is not None and w in listOfItems:
             cur = listOfItems.index(w)
         else:
             cur = 0
 
         res, ok = QInputDialog.getItem(
-            self, "Workflow Selection", "Select a workflow which should open the file.", listOfItems, cur, False
+            self, "Workflow Selection", "Select a workflow which should open the file.", sorted(listOfItems), cur, False
         )
 
         if ok:


btbest:
We also forgot that this breaks existing projects with this workflow... 
image

k-dominik:
Bummer, thanks for checking this @btbest

k-dominik:
Hey @btbest, could you please take another look? I couldn't bring myself to keep the `workflowName` as `Trainable Domain Adaptation (Local) (beta)`. Instead I added a shallow Legacy class that enables backwards compatibility. Tiny plus is that it is also visible in the recent project list even after opening them...

image

btbest:
I love it, that's actually a really smooth solution imo 👍

Maybe even a naggy `workflowDisplayName = 'Trainable Domain Adaptation (Local) (beta - please upgrade this project file with "Import Project")'` :grin: (not sure if I'm serious about this suggestion, Import Project is a pretty obscure feature afterall...)

k-dominik:
> Jokes aside, considering that it does still pop up in the Other Workflows list, maybe the displayName should be "(beta / deprecated)". Or alternatively exclude it from the Other Workflows.

I think it should just work - workflow not listed if the `auto_register` thing is false. So I decluttered the workflow-listing logic a bit. In short our workflow listing is now simplified:

All workflows that are imported and have `show_in_startup_menu == True` will show up in the menu. Whether it's in a specific group, or other depends on `ilastik.shell.ilastikShell.IlastikShell_populateWorkflows`.

Before there were multiple levels with `ilastik.workflows.AVAILABLE_WORKFLOWS` that could implicitly override that property. I removed that global variable.

btbest:
Beautiful refactoring, love that this also simplifies some tests.

There's always more of course. The drop-down on Import Project still lists the beta variant, because it goes through `getAvailableWorkflows`, which collects all `Workflow` subclasses 🤦 But I think I would cut off here and accept that as it is, unless you think simply filtering that list by the new `show_in_startup_menu` property is sufficient. A user who wants to upgrade a "(beta)" to the non-beta variant already knows what they want to choose here after all.

image

ilastik/workflow.py
```diff
@@ -38,8 +36,8 @@ class Workflow(Operator):
 
     name = "Workflow (base class)"
     workflowDisplayName = None  # override in your own workflow if you need it different from name
-    #: Should workflow be automatically added to start widget
-    auto_register = True
+    #: Should workflow added to start widget
+    show_in_startup_menu = True
```
emilmelnikov, line 40: Hmm, if workflows in the startup menu are explicitly listed and grouped in `_populateWorkflows()` anyway, is this flag really necessary?

k-dominik: yes, this is currently the case only for the "old" name of the trainable domain adaptation workflow, but in future, we can remove the PC+OC workflow that way and still enable headless running... Which I think is a nice property.

PR #2965, opened 2025-03-10
Filter out unsupported extensions

### LINKED ISSUE #2916: Widget offers unsupported export format options `xv` and `exr`
image

and

image

Same for `exr` and `exr sequence`

Culprit:

```
    _2d_exts = vigra.impex.listExtensions().split()
    _3d_sequence_formats = [FormatInfo(ext + " sequence", ext, 3, 3) for ext in _2d_exts]
```

Evidently ilastik doesn't actually support all extensions that vigra supports; either manually filter `_2d_exts` or add support in ilastik

AryanK37:
Fixes #2916

This PR filters out unsupported in by updating the list comprehension.
The change ensures that these formats are not included in the file format options, which should prevent them from appearing in the drop-down list in ilastik. (I verified it by running ilastik)

## Checklist

### DIFF
--- a/lazyflow/operators/ioOperators/opExportSlot.py
+++ b/lazyflow/operators/ioOperators/opExportSlot.py
@@ -76,7 +76,8 @@ class OpExportSlot(Operator):
     ExportPath = OutputSlot()
     FormatSelectionErrorMsg = OutputSlot()
 
-    _2d_exts = vigra.impex.listExtensions().split()
+    # Filter out formats that are not supported by vigra
+    _2d_exts = [ext for ext in vigra.impex.listExtensions().split() if ext not in {"xv", "exr"}]
 
     # List all supported formats
     # Only FormatInfo.name is used (to generate help text for a cmd parameter, DataExportApplet)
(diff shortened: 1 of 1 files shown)


btbest (commented):
I was considering whether we shouldn't aim to actually support these formats (maybe all it takes is to define them correctly in... I've concluded that excluding the formats should be fine, since users haven't been able to actually use them in the past anyway.

XV seems to be rather outdated anyway. EXR is at least still in active development, so we could explore if it's worth supporting. Though it's not like anyone has ever asked or complained that it doesn't work...

AryanK37:
Hi @btbest, thank you for the suggestion. While someone might misinterpret the comment, the code itself makes the intent clearer. 

I agree that since users haven’t been able to actually use these formats and there has been no feedback about missing functionality, excluding them makes sense for now.

However, in Python 3, list comprehensions have their own local scope. This means that when the list comprehension is executed at the class level, it does not have access to variables defined in the outer class scope (like unsupported_vigra_formats). To avoid this issue, I made changes accordingly.

Issue #1956, opened 2019-02-13
Incompatible datasets still appear in table after error


paulhfu:
## Describe the bug
When adding an incompatible probability map to an existing raw-dataset, the dataset is still loaded into the gui dataset table after an error message appeared.

## Expected behavior
The dataset should not be loaded.

## To Reproduce
1. For example in Multicut add some raw datasets
2. Then try adding some probability maps with non-matching size or dimension to the respective raw data

## Desktop :
 - OS: Ubuntu 18.04

ilastik ilastik/ilastik@7a31ec0395e38136da6f81427d32402327e51bab
lazyflow ilastik/lazyflow@8c102eae96b07954a46914ebf41400199fe4f5e9
volumina ilastik/volumina@c500bc07f9dde45aad2d54464738e5af1cbfa776

k-dominik:
fixed in #1955

k-dominik:
sadly this seems to have resurfaced (table entries still filled after dataset was rejected from loading, in multicut)

PR #2968, opened 2025-03-14
Add `DatasetOut` slot to OpDataSelection, Fix segfaults

### LINKED ISSUE #992: Importing raw data and prediction map while reordering axes crashes
**Steps to reproduce:**
- start a new Object Classification or Tracking from Prediction Map workflow
- load some raw data and change the axes order (I changed from xyzc to tyxc)
- load the prediction map, reorder axes as well
- **crash** with the following log: https://gist.github.com/chaubold/e3486acc53cb3a167ce6

@christophdecker experienced similar issues, but apparently got rid of them by pulling ilastik and lazyflow master. This does not fix it for me though.

### LINKED ISSUE #2967: Segfault when adding incompatible data
## Describe the bug

Adding an incompatible secondary input and aborting segfaults (test on OSX, Linux). Confirmed affected workflows: Object classification (all flavors), Multicut.

## Expected behavior
No segfault :)

## To Reproduce
Within the Object classification, or Multicut workflow
1. Add some 2dcells as raw data in the data selection applet
2. Add something incompatible, e.g. mitocheck as probabilities
3. In the pop up dialog, cancel -> crash

## Desktop:
 - ilastik version: 1.4.1rc2
 

maybe related to #992

k-dominik:
This slot exposed the same information as the input `OpDataSelection.Dataset`, but it is only connected if there are no data constraints violated. The slot is intended to be connected to UI elements of the data selection table. Ui then can rely on this data being valid when connected.

`DatasetDetailedInfoTable` now listens to this slots, instead of `Dataset` so no update gets triggered if data is not valid. Before, invalid states could be already added to the table and had to be rectified which I think could lead to objects being destroyed prematurely. Now the "correct axistags" loop runs before any data is added to the table.

Not sure if I want to merge this before 1.4.1. On the one hand -> segfaults that we know of are avoided, on the other hand, I would test drive this setup for a few weeks in case there are corner UI cases that I haven't reached yet.

In any case this should stay a single commit for easy revertability.

Before we were experiencing segfaults that were hard to debug. So the issue of segfaults is likely related to the item delegates for datasets.

Fixes #2967
Fixes #992
Fixes https://github.com/ilastik/ilastik/issues/1956

## Checklist

- [ ] Add tests.

### DIFF
--- a/ilastik/applets/dataSelection/dataLaneSummaryTableModel.py
+++ b/ilastik/applets/dataSelection/dataLaneSummaryTableModel.py
@@ -1,7 +1,7 @@
 ###############################################################################
 #   ilastik: interactive learning and segmentation toolkit
 #
-#       Copyright (C) 2011-2024, the ilastik developers
+#       Copyright (C) 2011-2025, the ilastik developers
 #                                <team@ilastik.org>
 #
 # This program is free software; you can redistribute it and/or
@@ -88,7 +88,7 @@ class DataLaneSummaryTableModel(QAbstractItemModel):
         self._op = topLevelOperator
 
         def handleNewLane(multislot, laneIndex):
-            assert multislot is self._op.DatasetGroup
+            assert multislot is self._op.DatasetGroupOut
             self.beginInsertRows(QModelIndex(), laneIndex, laneIndex)
             self.endInsertRows()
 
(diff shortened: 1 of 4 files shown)


k-dominik:
> Still haven't got round to looking in detail / running it, but an idea already: Can we make the InputSlot "write-only"? To make sure only the validated output is used.

I think I know what you mean, but even in putting it into words doesn't really make sense: "An InputSlot that is writeOnly and then read...."

Since validation is happening in `setupOutputs` that's more or less what I would define an Output to be. Having said that, the PR is a draft for a reason. I'm not happy with it yet, but wanted CI to run. Ideally I want to move out the validation logic to the workflow level where it could be used in UI/headless before setting any inputslot.

btbest:
Yeah... I guess this would require `InputSlot.get` to assert that its caller is `self.op`, and I don't think we want to go to callstack inspection for any reason tbh. Sensible positioning in the code + comment should suffice to communicate the intention...

k-dominik:
Hey @btbest,

I slimmed this PR down a bit leaving all bits related to workflows intact and adding this out slot only for the UI of dataselection. Would be cool to get some feedback...

k-dominik:
thank you @btbest !

PR #2974, opened 2025-03-19
Optimize bioimage model downloads, update tiktoch/bioimageio

### LINKED ISSUE #2678: neural network workflow: too many channels will cause model to silently fail
When checking for incompatible shapes, the current code

https://github.com/ilastik/ilastik/blob/e093db405cda4a4eeb529aebebbc6f55f339a5a5/ilastik/applets/neuralNetwork/nnClassGui.py#L767-L770

only checks if data extend is at least the same as the  minimal model shape. This works for the spacial axes `x`, `y`, `z`. However, in the channel axes, this usually doesn't make sense - those are not treated with padding, or tiling to accommodate for larger sizes.

example model: [affable-shark](https://bioimage.io/#/?tags=affable-shark&id=10.5281%2Fzenodo.5764892). There the training data is multi-channel, but for training it was flattened to a single channel. Then trying to predict on the "original" training/testing/validation set silently fails in ilastik :/

k-dominik:
Updated tiktorch to `25.4.0` which depends on the current bioimageio versions for spec (`0.5.4.1`) and core (`0.8.0`)

Also:
* only get one set of weights.
* removed tensorflow from supported models, since it is currently not supported on our side
* Don't add model if data channels don't match model channels (#2678)

Fixes #2678

### DIFF
--- a/ilastik/utility/bioimageio_utils.py
+++ b/ilastik/utility/bioimageio_utils.py
@@ -343,5 +343,11 @@ class InputValidator:
             ), f"Axis {axis_id} not an expected bioimage io axis string {SPEC_TO_VIGRA.keys()}."
 
             target_axis_size = spec_shape[axis_id]
-            if target_axis_size < min_size:
-                raise ValueError(f"Incompatible axis {axis}: {target_axis_size} < {min_size}")
+            if isinstance(axis, ChannelAxis):
+                if target_axis_size != min_size:
+                    raise ValueError(
+                        f"Incompatible axis {axis}: Number of channels in data: {target_axis_size} != Number of channels expected by model: {min_size}"
+                    )
+            else:
+                if target_axis_size < min_size:
+                    raise ValueError(f"Incompatible axis {axis}: {target_axis_size} < {min_size}")
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -69,7 +69,7 @@ outputs:
         - zarr 2.*
       run_constrained:
         - mkl <2024.1.0
-        - tiktorch 24.12.0
+        - tiktorch 25.4.0
         - volumina >=1.3.10
         # TODO: sphericaltexture 0.1.1 currently times out in the tests
         - sphericaltexture 0.0.4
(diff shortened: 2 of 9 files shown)


ilastik/utility/bioimageio_utils.py
```diff
@@ -343,5 +343,11 @@ def check_single_image_min_shape(self, tensor_id: str, target_vigra_shape: Tagge
             ), f"Axis {axis_id} not an expected bioimage io axis string {SPEC_TO_VIGRA.keys()}."
 
             target_axis_size = spec_shape[axis_id]
-            if target_axis_size < min_size:
-                raise ValueError(f"Incompatible axis {axis}: {target_axis_size} < {min_size}")
+            if isinstance(axis, ChannelAxis):
+                if target_axis_size != min_size:
+                    raise ValueError(
+                        f"Incompatible axis {axis}: Number of channels in data: {target_axis_size} != Number of channels expected by model: {min_size}"
```
btbest, line 349:
```suggestion
                    raise ValueError(
                        f"Selected input image has {target_axis_size} channels, but model expects {min_size}"
```

image

btbest (changes requested):
I can't run affable-shark :(

```pytb
ERROR 2025-04-02 12:31:30,181 classifier 4008 14472 Predict call failed with exception.
  File "..\ilastik\lazyflow\operators\tiktorch\classifier.py", line 256, in predict
    inference_pb2.PredictRequest(
AttributeError: module 'tiktorch.proto.inference_pb2' has no attribute 'PredictRequest'
```

And I assume this co-error is a consequence of the above:
```pytb
ERROR 2025-04-02 12:31:30,183 operators 4008 14472 Prediction failed with exception.
  File "..\ilastik\lazyflow\operators\tiktorch\operators.py", line 272, in execute
    result[...] = session.predict([input_data], [predictions_roi], [axistags])[0]
TypeError: 'int' object is not subscriptable
```

(tried saving the project after this error and loading it in 1.4.1rc2, and there it runs...)

ilastik/utility/bioimageio_utils.py
k-dominik: Sounds reasonable coming with a two channel image?

k-dominik: but yeah the other one.. .lets see.

k-dominik: in any case, tests were not passing for similar reasons, some changes in the tiktorch api. Fixed now

btbest (approved):
Works now. I still find the error message far too wordy, but in the end the information is there, if the user is willing to actually read the text.

Maybe in a future version, we can extend the validation to act retrospectively when a model is already loaded and the input data is changed afterwards.

Issue #2867, opened 2024-06-16
Add "open export location" button to export applet


k-dominik:
## Description

In our export applet:

<img width="1042" alt="image" src="https://github.com/ilastik/ilastik/assets/24434157/38bdcd5c-426c-4903-bf33-8dba3683c0dd">

it would be  cool to have a button to go to the export location. This could probably best be added to the table above the image view (as export locations could differ for different files)

tanishac25:
@k-dominik I've implemented the UI for the button and taken a screenshot. Now, I'd like to integrate its functionality to open the file explorer at the export location. As this is my first contribution to ilastik, I’d appreciate some guidance on the best approach. Would using **QDesktopServices.openUrl(QUrl.fromLocalFile(path))** be appropriate, or is there an existing utility in ilastik for handling file paths that I should use instead?

![Image](https://github.com/user-attachments/assets/64f673c6-0775-4c02-b2fa-ae3904f93b04)

k-dominik:
Hey @tanishac25,  awesome that you're looking into it and providing screenshot is really valuable! I think after seeing it I'd prefer to have a right click menu for that. While the functionality is less discoverable it's also not so important to devote so much space to it (in the end this is for convenience, users usually know well how to navigate the file system with system tools.

`QDesktopServices` should be fine, but it should point to the parent of the file (the folder/directory the file will be located in).

tanishac25:
@k-dominik Appreciate the feedback! I’ve updated it to use a right-click menu and made sure QDesktopServices opens the parent directory instead of the file itself. Also took a look at pathHelpers.py to properly handle hierarchical formats like HDF5. The PR is linked—would love to hear your thoughts!

PR #2981, opened 2025-03-22
data-export: Add 'Open Export Location' feature to export applet

tanishac25:
This PR enhances the data export applet by adding a right-click context menu option to open the export folder.

Fixes [#2867](https://github.com/ilastik/ilastik/issues/2867)

Below is a screenshot to show how the UI looks after this change.

Screenshot from 2025-03-22 19-51-10

More changes and improvements are welcome! Feel free to suggest enhancements or modifications.

## Checklist


ilastik/applets/dataExport/dataExportGui.py
```diff
@@ -223,11 +227,86 @@ def initCentralUic(self):
...
+        """Opens the parent folder of the exported file, handling all supported formats."""
+        # Get the export path from the selected row
+        exportPath = self.batchOutputTableWidget.item(row, Column.ExportLocation).text()
+
+        if not exportPath:
+            # If there's no export path, let the user know something's wrong
+            QMessageBox.warning(self, "Empty Path", "No export path available.")
+            return
+
+        print("Export Path:", exportPath)
```
k-dominik, line 280: please remove all `print` statements for production code.

ilastik/applets/dataExport/dataExportGui.py
```diff
@@ -223,11 +227,86 @@ def initCentralUic(self):
...
+        ".png",
+        ".jpg",
+        ".jpeg",
+        ".bmp",
+        ".npy",
+        ".csv",
+        ".tsv",
+        ".xlsx",
+        ".hdf5",
+    }
```
k-dominik, line 252: How did you generate this list? it is not complete and includes values that ilastik does not export to. Also comment wise, please don't add comments where it is super obvious. With a variable name `VALID_EXPORT_EXTENSIONS` it is clear enough.

ilastik/applets/dataExport/dataExportGui.py
```diff
@@ -223,11 +227,86 @@ def initCentralUic(self):
...
+
+        # Convert to a proper Path object and normalize it
+        exportPath = Path(exportPath).resolve()
+        print("Normalized Export Path:", exportPath)
+
+        # Split the path into parts (directories and filenames)
+        parts = list(exportPath.parts)
+
+        # Find where in the path a valid export file appears
+        export_file_indices = [i for i, part in enumerate(parts) if Path(part).suffix in self.VALID_EXPORT_EXTENSIONS]
```
k-dominik, line 290: we already have code for handling image paths with possible internal ones in `ilastik/lazyflow/utility/pathHelpers.py`. This is our central class to handle these paths and it makes sense to reuse, such that we can have path handling uniform across our code base.

ilastik/applets/dataExport/dataExportGui.py
```diff
@@ -223,11 +227,86 @@ def initCentralUic(self):
...
+    def showContextMenu(self, pos):
+        """Handles right-click menu in the export file list."""
+        menu = QMenu(self)
+
+        # Determine which cell in the table was clicked
+        row = self.batchOutputTableWidget.rowAt(pos.y())
+        col = self.batchOutputTableWidget.columnAt(pos.x())
+
+        # Only show the option if the user clicked in the 'Export Location' column
+        if col == Column.ExportLocation:
```
k-dominik, line 263: I find this a little to restrictive, I don't mind which column is right clicked, it should always show the context menu.

ilastik/applets/dataExport/dataExportGui.py
```diff
@@ -223,11 +227,86 @@ def initCentralUic(self):
...
+
+        # Show the context menu at the correct position
+        menu.exec_(self.batchOutputTableWidget.mapToGlobal(pos))
+
+    def openExportFolder(self, row):
+        """Opens the parent folder of the exported file, handling all supported formats."""
+        # Get the export path from the selected row
+        exportPath = self.batchOutputTableWidget.item(row, Column.ExportLocation).text()
+
+        if not exportPath:
```
k-dominik, line 275: Maybe rather show the context menu with the "Open Export Folder" option grayed out. This would require to get the data already in `showContextMenu`.

k-dominik (changes requested):
Dear @tanishac25, thank you so much for opening this PR and adding this cool feature to ilastik. I had some comments that would need to be addressed before it can be merged.

ilastik/applets/dataExport/dataExportGui.py
```diff
@@ -223,11 +227,78 @@ def initCentralUic(self):
...
+        if not exportPathItem:
+            QMessageBox.warning(self, "Empty Path", "No export path available.")
+            return
+
+        exportPath = PathComponents(exportPathItem.text()).externalPath
+        exportPath = Path(exportPath).resolve()
+
+        if not exportPath:
+            QMessageBox.warning(self, "Invalid Path", "Could not determine a valid export path.")
+            return
```
btbest, line 291: Since this validation logic is duplicated in the other method (with different case handling), it would be nice to extract it into a separate method used by the other two. Or - to me it seems like it shouldn't be necessary in the first place? Does the item or the `exportPath` ever evaluate to False?

btbest (changes requested):

ilastik/applets/dataExport/dataExportGui.py
tanishac25: Since the same logic is repeated, extracting it into a helper method makes sense. It might seem unnecessary at first, but these checks serve a purpose:`exportPathItem` can be None if no export path is set for the row & `exportPath` can be empty or invalid if something goes wrong while processing it.
Both can evaluate to False, these checks make sure we're only dealing with valid paths, avoiding potential errors.

ilastik/applets/dataExport/dataExportGui.py
```diff
@@ -223,11 +227,78 @@ def initCentralUic(self):
...
         self.batchOutputTableWidget.itemSelectionChanged.connect(self.handleTableSelectionChange)
+        # Handle the right-click event
+        self.batchOutputTableWidget.setContextMenuPolicy(Qt.CustomContextMenu)
+        self.batchOutputTableWidget.customContextMenuRequested.connect(self.showContextMenu)
 
         # Set up the viewer area
         self.initViewerStack()
         self.splitter.setSizes([150, 850])
 
+    VALID_EXPORT_EXTENSIONS = {
```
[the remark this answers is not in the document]
tanishac25: I initially added it to check file extensions, but since the goal was just to add a right-click menu to open the export folder, it's not needed. So I'll remove it.

ilastik/applets/dataExport/dataExportGui.py
btbest: I get that these are possible in theory, but do these cases actually happen in ilastik? Case-handling code that exists suggests that the cases it handles actually happen. In the future, it can be hard to tell what parts of the code are actually important and which parts are "just in case" or simply unnecessary, and this leads to bloat...

ilastik/applets/dataExport/dataExportGui.py
```diff
@@ -223,11 +227,51 @@ def initCentralUic(self):
...
+
+        if exportPath:
+            exportPath = PathComponents(exportPath).externalPath
+            exportPath = Path(exportPath).resolve() if exportPath else None
+
+        isValidPath = bool(exportPath) and exportPath.exists()
+
+        openFolderAction = menu.addAction("Open Export Folder")
+        openFolderAction.triggered.connect(lambda: self.openExportFolder(row))
+        openFolderAction.setEnabled(isValidPath)
```
btbest, line 254: Atm this will only enable if the export file exists, but I think it would also be fine to enable it if the parent directory exists, even if there's no files in it.

Most of the logic between the two functions still looks like it could be extracted into a common function like `export_dir = self._get_export_dir(row)`. Or does `showContextMenu` need the checks but `openContextFolder` doesn't?

PR #2983, opened 2025-03-23
Removed "Add New" Button from Data Selection Applet's Second Tab

Palak1725:
<!-- Thank you very much for opening a PR!
This update removes the redundant "Add New" button from the second tab of the Data Selection applet, which previously allowed adding prediction/segmentation files independently. The new behavior restricts adding these files only to existing raw data entries, reducing user confusion and aligning with intended functionality. Corresponding changes have been implemented in the GUI logic, ensuring seamless integration. Tests have been updated to reflect this change.

### DIFF
--- a/docs/ilastik/Makefile
+++ b/docs/ilastik/Makefile
@@ -1,4 +1,4 @@
-# Makefile for Sphinx documentation
+# Make file for Sphinx documentation
 #
 
 # You can set these variables from the command line.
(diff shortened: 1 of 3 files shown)


btbest:
Hi @Palak1725, as mentioned in the other PR, there should be no changes to the makefile if this is about spelling errors.

If you're still working on this, it would be nice if you could put it mark it as draft (top-right corner under Reviewers).

PR #2986, opened 2025-03-25
Scaling operator

btbest:
Still in draft stage, but feel free to have a look at the preview.

Todo:

- Convert this op into a base class and split off subclasses that implement appropriate scaling methods for different data types. Current one handles image data and possibly probabilities, but definitely need at least one other implementation to handle label images.
- Use it to implement multiscale export.


btbest:
Integrated in https://github.com/ilastik/ilastik/pull/3017

PR #2991, opened 2025-03-31
DataSelection don't show viewer for every added file

k-dominik:
When loading multiple files, ilastik used to instantiate a viewer for every file added, which was shown only very briefly. With this commit, the viewer is only instantiated for the first added file.

I've seen 3-5 times speedup in loading files with this patch.

## Checklist

- [ ] Add tests.

### DIFF
--- a/ilastik/applets/dataSelection/dataSelectionGui.py
+++ b/ilastik/applets/dataSelection/dataSelectionGui.py
@@ -18,8 +18,6 @@
 # on the ilastik web site at:
 # 		   http://ilastik.org/license.html
 ###############################################################################
-# Python
-import itertools
 import logging
 import os
 import threading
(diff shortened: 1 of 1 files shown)


ilastik/applets/dataSelection/dataSelectionGui.py
```diff
@@ -408,22 +407,23 @@ def addFileNames(self, paths: Optional[List[Path]], startingLaneNum: Optional[in
         if paths is None:
             paths = []
```
btbest, line 408:
```suggestion
        if paths is None:
            return
```

Now that `addLanes` is executed even when `paths is None`, I'd rather this just exits right here instead of continuing with an empty list.

As far as I can trace, it looks like the downstream code is ok with empty lists, but I didn't check whether all `handleNewLanesAdded` implementations are ok with being triggered when no actual new lanes have been added. I also don't really see the point in continuing with an empty input when the function knows there's nothing to be done.

btbest (commented):
Side note: This changes the behaviour. Previously, after adding multiple files, the last one added would be shown; now it's the first one.

Judging by code comments, the new behaviour was intended before anyway, so I'd classify that as a free bugfix.

ilastik/applets/dataSelection/dataSelectionGui.py
btbest: I changed the condition to `not paths` because cancelling the ImageFileDialog gives `paths = []`

k-dominik: a right ;D

btbest (approved):

PR #2992, opened 2025-04-03
Update environment for next development cycle

k-dominik:
In progress...

* Removed deprecated `pytorch` and `nvidia` channels in favor of `conda-forge`

Major dependency updates:

* `python`: `3.9` -> `3.11`
* `vigra`: `1.11.1` (with annoying build pinning) -> `1.12.1`
* `pytorch`: `2.4` (from _pytorch_ channel) -> `2.6` (from _conda-forge_ channel)

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -20,12 +20,12 @@ outputs:
         - python -m pip install . --no-deps --ignore-installed --no-cache-dir -vvv
     requirements:
       host:
-        - python >=3.7
+        - python >=3.9
         - pip
         - setuptools >=40.0
         - setuptools_scm
       run:
-        - python >=3.7
+        - python >=3.9
         - numpy >1.12,<1.23
         # aiohttp to enable zarr.storage.FSStore
         - aiohttp
@@ -118,7 +118,7 @@ outputs:
       run:
         - python 3.9.*
         - {{ pin_subpackage("ilastik-core", exact=True) }}
-        - pytorch >=1.6,<2.4.1
+        - pytorch >=1.6
         - tiktorch
         - torchvision
         - volumina
@@ -135,8 +135,6 @@ outputs:
       requires:
         - pytest >4
         - pytest-qt
-        # need to help mamba here a bit
-        - ilastik-pytorch-version-helper-cpu
         - pytorch 2.*
 
       imports:
@@ -170,7 +168,7 @@ outputs:
       run:
         - python 3.9.*
         - {{ pin_subpackage("ilastik-core", exact=True) }}
-        - pytorch >=1.6,<2.4.1
+        - pytorch >=1.6
         - tiktorch
         - torchvision
         - pytorch-cuda
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -68,7 +68,6 @@ outputs:
         - z5py
         - zarr 2.*
       run_constrained:
-        - mkl <2024.1.0
         - tiktorch 25.4.0
         - volumina >=1.3.10
         # TODO: sphericaltexture 0.1.1 currently times out in the tests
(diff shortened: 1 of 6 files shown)


conda-recipe/meta.yaml
```diff
@@ -63,12 +63,11 @@ outputs:
         - tifffile >=2022
         # build 1.11.1=*_1028 on cf is the first to be compatible with numpy>1.19
         # need to bump this manually until there is a true version bump in vigra
```
btbest, line 65:
```suggestion
```
sync with dev/env.yml

dev/environment-dev.yml
```diff
@@ -56,32 +49,18 @@ dependencies:
 
   # Deep learning dependencies (neural net workflow, trainable domain adaptation)
   # can be changed to request gpu versions
-  - pytorch <2.4.1  # >=2.4 can clash with pyshtools on win; 2.2.2 newest available on osx
-  - cpuonly  # comment out for pytorch with cuda
-  # - pytorch-cuda=11.3  # uncomment for pytorch with cuda, specify cuda version
+  - pytorch 2.7.*  # >=2.4 can clash with pyshtools on win; 2.2.2 newest available on osx
```
btbest, line 52:
```suggestion
  - pytorch 2.7.*
```
outdated comment

dev/environment-dev.yml
```diff
@@ -56,32 +49,18 @@ dependencies:
...
-  # to install a cpu-build
-  - ilastik-pytorch-version-helper-cpu  # comment out for pytorch with cuda
-
 
   # Third-party object feature plugins
   - sphericaltexture_plugin
   # TODO: sphericaltexture 0.1.1 currently times out in the tests
   - sphericaltexture 0.0.4
-  - pyshtools >=4.12  # If >=4.13 becomes available for win, might be able to unpin pytorch
+  - pyshtools >=4.13
```
btbest, line 59: 🎉

dev/environment-dev.yml
```diff
@@ -56,32 +49,18 @@ dependencies:
 
   # Deep learning dependencies (neural net workflow, trainable domain adaptation)
   # can be changed to request gpu versions
```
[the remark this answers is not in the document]
k-dominik: you'd want to request something like `pytorch=*=*cuda*`, although in principle conda figures this out now for you.

So in principle one could even get away without `ilastik-gpu` as `cuda` is really the only thing added explicitly there. That package is still useful in the process of building binaries to identify the GPU variant an make sure to ship cuda even on a machine that doesn't have it - so for now I'd like to keep it.

k-dominik:
Hey @btbest, I had to do a small detour to remove our last remaining yield-style test, but now everything seems to be passing 🥳

btbest (approved):
`1453 passed, 72 skipped, 4 xfailed, 1126 warnings in 471.67s (0:07:51)` on my machine 🚀

btbest:
Wait a second, `UserWarning: UserWarning('Workflow startup menu: Could not find Trainable Domain Adaptation (Local)')` doesn't sound right...

k-dominik:
> Re-ran with a new env and the workflow warning no longer appears, so all good 😅

it should still show up - maybe you ran without gui tests?

The tests involving tiktorch are quite defensive - it's basically only required for these. The workflow listing of the shell (instantiated by any gui test) expects certain workflows to be there, but since the tiktorch exectable is unset - it complains.

I see two options:
1. Add a autouse fixture to all tests, that sets `ilastik.config.runtime_cfg.tiktorch_executable` - all tests will fail if tiktorch is not present for some reason
2.  Ignore the warning in gui tests - acceptable since we don't actually test this workflow...

btbest:
> maybe you ran without gui tests?

I ran `pytest --run-legacy-gui`, but apparently also just didn't look close enough. The warning is still there.

Issue #1875, opened 2018-10-29
Delete All button should also delete the exported Feature Table


wolny:
After exporting Object Predictions together with Object Feature Table, when the user clicks 'Delete All' only the Object Predictions file is removed, but the Feature Table file is not.

PR #2996, opened 2025-04-07
Add CSV file deletion to deleteAllResults method

iCarlosVega:
Add CSV file deletion to deleteAllResults method" -m "Previously, the deleteAllResults method only deleted ilastik files, leaving associated CSV table files behind. This enhancement ensures complete cleanup of all exported files.

The changes include:
- Access table exporter from parent applet
- Extract CSV file path from export settings
- Handle \"_table.csv\" suffix in file paths
- Add proper error handling and path validation

This ensures that both ilastik files and their associated CSV files are properly deleted when the user clicks \"Delete All\".

Resolves: #1875

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -69,7 +69,7 @@ outputs:
         - zarr 2.*
       run_constrained:
         - mkl <2024.1.0
-        - tiktorch 24.12.0
+        - tiktorch 25.4.0
         - volumina >=1.3.10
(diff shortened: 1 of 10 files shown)


btbest:
Hi @iCarlosVega, thank you for the contribution! It looks like you accidentally included a couple of `patch` files, and committed the changes with a commit message from an existing commit? Please amend the commit with a more descriptive message and removing the patch files (you then have to force-push the branch to update this PR).

PR #3008, opened 2025-04-27
Fix relative logfile path handling

emilmelnikov:
If logfile path is relative (e.g. from `--logfile my-log.txt`), then `os.path.dirname(logfile_path)` returned an empty string, which broke on `os.makedirs`.

Now, logdir is wrapped in `Path` that correctly handles relative dirs.


k-dominik (approved):
Thank you so much for fixing this @emilmelnikov :)

PR #3017, opened 2025-05-07
OME-Zarr multiscale export

btbest:
This introduces multi-scale export in OME-Zarr.

* New OpResize - a lazyflow implementation of [`skimage.transform.resize`](https://scikit-image.org/docs/stable/api/skimage.transform.html#skimage.transform.resize). This is used for multiscale export for now, but I intend to implement cross-scale workflows with this soon ™️.
  * The operator handles blockwise resizing of arbitrary-to-arbitrary input and output image shapes, at arbitrary block size.
  * The implementation loops over channels. In my tests this was comparably fast as not looping, but reduced RAM usage by ~25%.
  * Anti-aliasing uses smaller sigmas than the skimage reference implementation when downscaling by factors > 2. The inspiration for this came from [Robert Haase's notebook](https://haesleinhuepf.github.io/BioImageAnalysisNotebooks/19_spatial_transforms/downsampling_with_denoising.html). I played around a bit with it; the smaller sigma seemed fine, and can save a bit of data cost.
  * Tests ensure that outputs are identical to `skimage.transform.resize` - within a margin because of the different anti-aliasing.
  * The operator offers nearest-neighbor and linear interpolation for now.
* New "multi-scale OME-Zarr" export format. Accompanying volumina PR tbd.
  * The format comes with no additional options for the user for now.
  * Scaling is done by 2x until the image fits in one chunk. If the source was already multiscale, the output scales match the input scales.
  * Linear interpolation is default. Exports where this is not appropriate should be flagged to use nearest-neighbor interpolation instead; the PR contains an example of how this could be done in Pixel Classification. Other workflows' exports would need to still be done. I'm not sure this is the best way to go about this.

image

## Checklist

- [ x ] Format code and imports.
- [ x ] Add docstrings and comments.
- [ x ] Add tests.
- [ x ] Rebase commits into a logical sequence.


tests/test_lazyflow/test_operators/test_opResize.py
```diff
@@ -0,0 +1,99 @@
+from unittest import mock
+
+import numpy
+import pytest
+import vigra
+from skimage.transform import resize as sk_resize
+
+from lazyflow.operators.opResize import OpResize
+from lazyflow.operators.opSplitRequestsBlockwise import OpSplitRequestsBlockwise
+from lazyflow.utility import Timer
```
k-dominik:
```suggestion
```
unused

lazyflow/operators/opResize.py
```diff
@@ -0,0 +1,225 @@
...
+
+class OpResize(Operator):
+    """
+    Reimplements scikit-image.transform.resize as a lazyflow operator,
+    handling in particular the halo for antialiasing in the context of subregions.
+
+    The gaussian filter for antialiasing is run with custom sigmas that are negligibly larger
+    than the scikit-image default at scaling factors <2, and smaller at scaling factors >2.
+    This should reduce the required halo sizes and possibly maintain more information
+    (remains to be tested).
```
k-dominik: did you have a chance to look into this? What measure would you use to compare information content? Maybe it's worth discussing it in an issue.

lazyflow/operators/opResize.py
```diff
@@ -0,0 +1,225 @@
...
+        n_channels = n_source_pixels[channel_axis]
+        n_source_pixels[channel_axis] = 1  # Will be 1 after splitting
+        source_coords_starts[channel_axis] = 1
+        source_coords_stops[channel_axis] = 1
+        scaled_source_grid = self._roi_to_coord_grid(source_coords_starts, source_coords_stops, n_source_pixels)
+        filtered_split = np.split(filtered, n_channels, axis=channel_axis)
+        for c in range(n_channels):
+            channel_slicing = [slice(None)] * len(factors)
+            channel_slicing[channel_axis] = slice(c, c + 1)
+            result[channel_slicing] = scipy_ndimage.map_coordinates(
```
k-dominik:
```suggestion
            channel_slicing = [slice(None)] * len(factors)
            channel_slicing[channel_axis] = slice(c, c + 1)
            channel_slicing = tuple(channel_slicing)
            result[channel_slicing] = scipy_ndimage.map_coordinates(
```

otherwise it errors out for me with numpy 1.26

lazyflow/operators/opResize.py
```diff
@@ -0,0 +1,225 @@
...
+        interpolation_order = self.InterpolationOrder.value
+        required_min_padding = {0: 0, 1: 2}
+        # If higher-order interpolation is desired, need to figure out required padding
+        assert interpolation_order in required_min_padding, "supports only order 0 (nearest-neighbor) or 1 (linear)"
+        axes_to_pad = np.not_equal(factors, 1)
+        # Antialiasing only for downscale (f > 1), not identity or upscale
+        if interpolation_order > 0:
+            antialiasing_sigmas = np.array([f / 4 if f > 1 else 0 for f in factors])
+        else:
+            antialiasing_sigmas = np.zeros_like(factors)
```
k-dominik: Mini optimization: Can be moved to setupoutputs and stored as a variable (so it's not calculated for _every_ request).

lazyflow/operators/opResize.py
```diff
@@ -0,0 +1,225 @@
...
+            sigma=antialiasing_sigmas,
+            enlarge_axes=axes_to_pad,
+        )
+        raw_roi_interpolation_halo = self._extend_halo_to_minimum(
+            raw_roi_antialiasing_halo, required_min_padding[interpolation_order], axes_to_pad, raw_roi
+        )
+        raw_roi_final_halo = np.clip(ceil_roi(raw_roi_interpolation_halo), 0, self.RawImage.meta.shape)
+
+        raw = self.RawImage[roiToSlice(*raw_roi_final_halo)].wait().astype(np.float64)
+        filtered = scipy_ndimage.gaussian_filter(raw, antialiasing_sigmas, mode="mirror")
```
k-dominik, line 171: once you go into making this code faster this would be one of the obvious places. Does the scipy implementation even release the gil?

lazyflow/operators/opResize.py
```diff
@@ -0,0 +1,225 @@
...
+
+        n_source_pixels = roi.stop - roi.start
```
k-dominik: very minor: maybe a name like `source_shape`, or `roi_shape` would be more descriptive, at least I had to go back to see what `n_source_pixels` is about and was a little surprised to find something else than an integer.

lazyflow/operators/opResize.py
```diff
@@ -0,0 +1,225 @@
...
+        )
+
+    @staticmethod
+    def _extend_halo_to_minimum(
+        roi_with_halo: NDArray, minimum: int, axes_to_extend: NDArray, original_roi: NDArray
+    ) -> NDArray:
+        """
+        Extend an existing halo on `roi_with_halo` to `minimum` along axes in `axes_to_extend`.
+        `original_roi` required to determine how much halo was already present.
+        """
```
k-dominik, line 226:
```suggestion
    def _extend_halo_to_minimum(
        roi_with_halo: NDArray, minimum: int, axes_to_extend: NDArray, original_roi: NDArray
    ) -> NDArray:
        """
        Extend an existing halo on `roi_with_halo` to `minimum` along axes in `axes_to_extend`.
        `original_roi` required to determine how much halo was already present.
        """
        if minimum == 0:
            return roi_with_halo
```

I see this function is very general and could work for many values, but out of the two currently intended ones, only one is actually doing anything. At least the fast path for zero min should go in imo.

lazyflow/operators/opResize.py
```diff
@@ -0,0 +1,225 @@
...
+        Meshgrid returns a list of one coordinate grid per axis.
+        """
+        assert len(starts) == len(stops) == len(steps), "Dimensions must match"
+        return np.meshgrid(
+            *(np.linspace(start, stop, n) for start, stop, n in zip(starts, stops, steps)),
+            indexing="ij",
+        )
+
+    @staticmethod
+    def _extend_halo_to_minimum(
```
k-dominik, line 220: I'd suggest a different name here to communicate the intend better. To me it looks like `_ensure_minimum_halo`.

Edit: or maybe not :). I'll leave the comment in as I stumbled here.

tests/test_lazyflow/test_operators/test_opResize.py
```diff
@@ -0,0 +1,99 @@
...
+    with mock.patch.object(OpResize, "execute", wraps=op.execute) as mock_execute:
+        op_resized_block = split_op.Output[:].wait()
+        assert mock_execute.call_count == expected_n_blocks, "blocks not split as expected"
+
+    sk_resized = sk_resize(data, scaled_shape, anti_aliasing=True, preserve_range=True)
+    op_resized = op.ResizedImage[:].wait()
+
+    # Different antialiasing, see test_resize_matches_skimage - with float image need some tolerance
+    numpy.testing.assert_allclose(op_resized, sk_resized, rtol=0.07)
+    numpy.testing.assert_allclose(op_resized_block, op_resized)  # Block-splitting artifacts not tolerated
```
k-dominik: `allclose` still has a default of `rtol=1e-7`... in fact on my machine the test would still pass for `rtol=1e-15`...

lazyflow/operators/opResize.py
```diff
@@ -0,0 +1,225 @@
...
+    This should reduce the required halo sizes and possibly maintain more information
+    (remains to be tested).
+
+    Cannot resize along channel axis (would be nonsense).
+    Time is treated like space axes, so resize along t at your own risk.
+    """
+
+    RawImage = InputSlot()
+    TargetShape = InputSlot()  # Tuple[int, ...]
+    InterpolationOrder = InputSlot(value=1)  # 0 (nearest-neighbor) and 1 (linear) supported
```
k-dominik: minor: Would maybe be cool to have an intEnum there... I'd find `Interpolation.NEAREST` and `Interpolation.LINEAR` more readable throughout the code.

lazyflow/operators/ioOperators/opExportSlot.py
```diff
@@ -176,6 +191,18 @@ def _executeExportPath(self, result):
         result[0] = formatted_path
         return result
 
+    def _executeTargetScales(self, result):
+        # Generate default scaling: scale down by factor of 2 in all spatial dimensions,
```
k-dominik: since it's matching target scales it's not necessarily factor 2?

lazyflow/operators/ioOperators/opExportSlot.py
```diff
@@ -415,7 +445,17 @@ def _export_ome_zarr(self):
...
         try:
-            write_ome_zarr(self.ExportPath.value, self.Input, offset_meta, self.progressSignal)
+            write_ome_zarr(self.ExportPath.value, self.Input, self.progressSignal, offset_meta)
+        finally:
+            self.progressSignal(100)
+
+    def _export_ome_zarr_multiscale(self):
+        assert self.TargetScales.ready(), "export target scales must be configured for multi-scale export"
+        self.progressSignal(0)
+        target_scales = self.TargetScales.value
```
k-dominik: I think it would be nicer _not_ to go via the slot here and have `_executeTargetScales(self)` return the scales (no reason to do a request, really). In `execute` you could still put it into the `result` if you wanted to.

tests/test_lazyflow/test_operators/test_ioOperators/testOpExportSlot.py
```diff
@@ -115,6 +117,57 @@ def test_ome_zarr(self):
         finally:
             opRead.cleanUp()
 
+    def test_ome_zarr_multi_scale(self):
+        """Ensure multi-scale export generates one downscale for 400x400."""
+        # Chunk size is 358x357 for square 2D, so 400x400 is larger, and 200x200 is one chunk.
```
k-dominik: I assume this is because of big request streamer logic... Could mention that here.

tests/test_lazyflow/test_operators/test_ioOperators/testOpExportSlot.py
```diff
@@ -115,6 +117,57 @@ def test_ome_zarr(self):
...
+        )
+        assert opExport.TargetScales.value == expected_scales
+
+        opExport.run_export()
+
+        opRead = OpInputDataReader(graph=graph)
+        try:
+            for i, scale in enumerate(["s0", "s1"]):
+                opRead.FilePath.setValue(str(expected_export_path / scale))
+                expected_data = data.view(numpy.ndarray).reshape((1, 1, 1) + data.shape)  # OME-Zarr always tczyx
```
k-dominik: `vigra` is actually very nice for this kind of stuff.
```suggestion
                expected_data = data.withAxes("tczyx")  # OME-Zarr always tczyx
```

tests/test_lazyflow/test_utility/test_io_util/test_write_ome_zarr.py
```diff
@@ -8,31 +8,53 @@
...
+            (21, 23, 3),
+            "yxc",
+            OrderedDict({"s0": tagged_shape("tczyx", (1, 3, 1, 10, 12))}),
+        ),
     ],
 )
-def test_metadata_integrity(tmp_path, graph, shape, axes):
+def test_metadata_match_ome_zarr_spec(tmp_path, graph, shape, axes, target_scales):
     data_array = vigra.VigraArray(shape, axistags=vigra.defaultAxistags(axes))
     data_array[...] = numpy.indices(shape).sum(0)
```
k-dominik: out of curiosity - the data in the `data_array` is never compared or anything, but still has this peculiar content (as compared to, say, `numpy.random.random`... is there a reason?
```suggestion
    data_array = vigra.taggedView(numpy.random.randint(0, 255, shape, dtype="uint8"), axes)
```

tests/test_lazyflow/test_utility/test_io_util/test_write_ome_zarr.py
```diff
@@ -43,12 +65,14 @@ def test_metadata_integrity(tmp_path, graph, shape, axes):
     assert all([value is not None for value in written_meta.values()])  # Should not write None anywhere
```
k-dominik:
```suggestion
    assert all(written_meta.values())
```
unless you want to allow other values that would evaluate to `False`.

lazyflow/utility/io_util/write_ome_zarr.py
```diff
@@ -51,14 +52,67 @@
...
+            scaled_shape = ODict([(a, int(size / scale_factors[a])) for a, size in export_shape.items() if a != "c"])
+            target_shape = _reorder(scaled_shape, OME_ZARR_AXES, 1)
+            if "c" in export_shape:
+                target_shape["c"] = export_shape["c"]
+            target_item = (scale_key, target_shape)
+            scales_items.append(target_item)
+    return ODict(scales_items)
+
+
+def generate_default_target_scales(unscaled_shape: TaggedShape, dtype) -> Multiscales:
```
k-dominik, line 102: Here maybe just for completeness a comment/docstring how the default scales are calculated (2x).

lazyflow/utility/io_util/write_ome_zarr.py
```diff
@@ -51,14 +52,67 @@
...
+            break
+    return ODict(scales_items)
 
 
-@dataclasses.dataclass
-class ImageMetadata:
-    path: str
-    scaling: OrderedScaling
-    translation: OrderedTranslation
+def _reorder(shape: Dict, axes: Iterable[Axiskey], default_value: Union[int, float]) -> TaggedShape:
```
k-dominik: should this silently drop axes in `shape`, or could the typing be more narrow here like

`ShapeDict = Dict[Axiskey, int]`...

lazyflow/utility/io_util/write_ome_zarr.py
```diff
@@ -74,173 +128,47 @@ def _get_chunk_shape(tagged_image_shape: TaggedShape, dtype) -> Shape:
...
-) -> List[OrderedScaling]:
+    output_axiskeys: Iterable[Axiskey],
+) -> ScalingsByScaleKey:
     """Multiscales and base_shape may have arbitrary axes.
     Output are scaling factors relative to base_shape, with axes output_axiskeys.
     Scale factor 1.0 for axes not present in scale or base shape, and for channel."""
-    scalings = []
+    factors = []
     for scale_shape in multiscales.values():
         common_axes = [a for a in scale_shape.keys() if a in base_shape.keys()]
```
k-dominik: no need to call keys.

```suggestion
        common_axes = [a for a in scale_shape if a in base_shape]
```

lazyflow/utility/io_util/write_ome_zarr.py
```diff
@@ -262,71 +190,131 @@ def _get_axes_meta(export_axiskeys, input_ome_meta):
     return axes
 
 
-def _get_total_offset(
-    absolute_scaling: OrderedScaling,
-    image: ImageMetadata,
+def _get_input_transforms_reordered(
+    input_ome_meta: Optional[OMEZarrMultiscaleMeta], scale_key: str, axes: Iterable[Axiskey]
+) -> Tuple[Optional[OrderedScaling], Optional[OrderedTranslation]]:
```
k-dominik: Super minior: Total matter of taste, but I find `Optional` nice to read for arguments, but for outputs `Union[Whatever, None]` is more expressive imo.

lazyflow/utility/io_util/write_ome_zarr.py
```diff
@@ -51,14 +52,67 @@
 OrderedScaling = OrderedTranslation = OrderedDict[Axiskey, float]  # { axis: scaling }
```
k-dominik, line 53: my typechecker complains that `OrderedScaling` is a mere variable and cannot be used as a type. How is it for you?

lazyflow/utility/io_util/write_ome_zarr.py
```diff
@@ -74,173 +128,47 @@ def _get_chunk_shape(tagged_image_shape: TaggedShape, dtype) -> Shape:
...
-        factors = [ODict([(a, 1.0) for a in export_shape.keys()])]
-        scalings_relative_to_export = ODict(zip([f"s{i}" for i in range(len(factors))], factors))
-        scalings_relative_to_raw = None
-    return scalings_relative_to_export, scalings_relative_to_raw
+        factors.append(axes_matched_factors)
+    return ODict(zip(multiscales.keys(), factors))
 
 
-def _create_empty_zarrays(
+def _create_empty_zarray(
```
k-dominik, line 174: is there some deeper knowledge here, like this will actually create the files?

lazyflow/utility/io_util/write_ome_zarr.py
```diff
@@ -364,36 +352,59 @@ def _get_multiscale_transformations(
...
+        za = zarr.Array(store, path=path)
         _write_to_dataset_attrs(ilastik_meta, za)
 
 
 def write_ome_zarr(
-    export_path: str, image_source_slot: Slot, export_offset: Optional[Shape], progress_signal: OrderedSignal
+    export_path: str,
+    image_source_slot: Slot,
+    progress_signal: OrderedSignal,
+    export_offset: Optional[Shape],
```
k-dominik: minor (taste): I stumble over something annotated as Optional without a default value (although Python docs clearly state that `Optional` does not imply a default value). Still I would probably want to express this as `Union[Shape, None]`.

lazyflow/utility/io_util/write_ome_zarr.py
```diff
@@ -411,36 +422,78 @@ def write_ome_zarr(
...
-        zarrays, export_meta = _create_empty_zarrays(
-            abs_export_path, export_dtype, chunk_shape, export_shape, export_scalings
-        )
 
-        requester = BigRequestStreamer(reordered_source, roiFromShape(reordered_source.meta.shape))
-        requester.resultSignal.subscribe(partial(_scale_and_write_block, export_scalings, zarrays))
-        requester.progressSignal.subscribe(progress_signal)
-        requester.execute()
+        export_scalings = _multiscales_to_scalings(target_scales, export_shape, export_shape.keys())
+        combined_scaling_mag = {key: numpy.prod(list(scale.values())) for key, scale in export_scalings.items()}
```
k-dominik, line 471: The way the mind works... let's say I want to
* scale in x and y by factor of 2 (downscaling)
* but also scale in z by 0.1 (say, to achieve isotropy, who knows) (upscaling)

then I'd land in the (v =  downscale code path, despite one axis being upscaled. Does this pose any problems?

lazyflow/utility/io_util/write_ome_zarr.py
```diff
@@ -51,14 +52,67 @@
...
-class ImageMetadata:
-    path: str
-    scaling: OrderedScaling
-    translation: OrderedTranslation
+def _reorder(shape: Dict, axes: Iterable[Axiskey], default_value: Union[int, float]) -> TaggedShape:
+    """Reorder a tagged shape to `axes`, using `default_value` for axes missing in `shape`."""
+    return ODict([(a, shape[a] if a in shape else default_value) for a in axes])
 
 
 def _get_chunk_shape(tagged_image_shape: TaggedShape, dtype) -> Shape:
```
k-dominik, line 135: I wonder if the chunk shape should not be independent of the data type. My expectation would be that I get the same scales if I convert some raw data (8bit) and their corresponding prob maps (that might be float32). This is, however, not the case.

k-dominik (commented):
Hello @btbest,

In general I found this super nice to review 🚀  :), although I didn't even make use at all of the extensive commit history. I left some comments, many of them minor (which I indicated and please don't feel inclined to act there, only if it makes sense, to you, too.).

So for the issue of where to put the information about the interpolation order... Instant hunch: Something like semantic channel types. I discussed this with @tischi and in mobie there are conceptually two types: "intensity" and "label" which would map to your interpolation order as 1 and 0. We could default to something like "intensities" and only change the relevant to "label".

lazyflow/operators/opResize.py
btbest: Plot just to illustrate the difference (x: scaling factor; y: sigma; blue: scikit-image; red: current implementation).

The most common thing people do is of course 2x, in which case the sigmas are identical. I haven't really looked into what practical difference the smaller sigmas make at larger scaling factors, but if Robert Haase uses these smaller sigmas, I decided to go by example if it means potentially fewer pixels and cheaper computations :)

Happy to open an issue as a reminder that we should look into this if larger scaling steps start becoming a regular use case (#3037)

image

lazyflow/operators/opResize.py
btbest: Yes, the two costly operations in this `execute` are the gaussian filter and, even worse, map_coordinates (third in line is np.meshgrid). If there are faster implementations for any of those, that would be nice. I remember looking at vigra's gaussian but iirc it doesn't support enough dimensions. I liked having the technical possibility of scaling along t, and making this op a genuine re-implementation of `skimage.transform.resize`, but the technical possibility for t isn't worth taking a performance penalty, and I'm happy to abandon the reference implementation if ours can be better :) I'll keep a note to look into this for a follow-up PR?

btbest: Looks to me like gaussian_filter can run in parallel with `threading.Thread`, which would imply it doesn't hold on to the gil? Although the execution time for each run is a bit longer, so there might be some waiting.

```
sync gaussian
Gaussian 0 started at 0.00053
Gaussian 0 finished at 0.01568
Gaussian 0 duration: 0.01515
sync gaussian done
# threaded execution
Gaussian 1 started at 0.01764
Gaussian 2 started at 0.01789
Gaussian 1 finished at 0.03493
Gaussian 1 duration: 0.01730
Gaussian 2 finished at 0.03563
Gaussian 2 duration: 0.01774
```
<details>
    <summary>Code</summary>

  ```
    import numpy as np
    import threading
    from time import perf_counter
    from scipy.ndimage import gaussian_filter
    
    image = np.random.rand(1000, 1000)
    
    def apply_gaussian(n):
        start_time = perf_counter() - global_start
        print(f"Gaussian {n} started at {start_time:.5f}")
        gaussian_filter(image, sigma=3)
        end_time = perf_counter() - global_start
        print(f"Gaussian {n} finished at {end_time:.5f}")
        print(f"Gaussian {n} duration: {end_time - start_time:.5f}")
    
    def worker(n, start_event):
        start_event.wait()
        apply_gaussian(n)
    
    if __name__ == "__main__":
        global_start = perf_counter()
        start_event = threading.Event()
        end_event1 = threading.Event()
        end_event2 = threading.Event()
    
        thread1 = threading.Thread(target=worker, args=(1, start_event))
        thread2 = threading.Thread(target=worker, args=(2, start_event))
    
        print("sync gaussian")
        apply_gaussian(0)
        print("sync gaussian done")
    
        thread1.start()
        thread2.start()
    
        start_event.set()
    
        thread1.join()
        thread2.join()
  ```
</details>

lazyflow/operators/opResize.py
btbest: I think `_ensure_minimum_halo` would also be fine, or maybe `_enlarge_halo_to_minimum` (going by the example of `enlargeRoiForHalo`). I'm happy with any of them, do you have a favourite? 😅

tests/test_lazyflow/test_operators/test_opResize.py
btbest: Heh, I guess I didn't think about seeing if it can push beyond numpy's default strictness :)

lazyflow/operators/ioOperators/opExportSlot.py
btbest: TargetScales is the same kind of slot as ExportPath (gui input/output in export widget with validation requirements), and is implemented to work the same way. I'll refactor both of them as I want to keep their implementations in sync to make that clear.

tests/test_lazyflow/test_utility/test_io_util/test_write_ome_zarr.py
btbest: Empty `datasets`, `version` and `name` are technically ok according to spec 🙃 but yeah that makes sense

lazyflow/utility/io_util/write_ome_zarr.py
btbest: Not sure it needs an alias, but narrower typing is definitely better here (the fact that `default_value` is float or int doesn't make it clear enough)

lazyflow/utility/io_util/write_ome_zarr.py
btbest: I don't even read it as "optional" in the conventional sense of the word, so I can guarantee I will end up putting `Optional`s in return annotations out of habit in the future 🙃

lazyflow/utility/io_util/write_ome_zarr.py
btbest: no complaint here... is it because of the assignment chaining with `OrderedTranslation` or does it also complain for `TaggedShape` for you..?

lazyflow/utility/io_util/write_ome_zarr.py
btbest: Yes, this will create folders and `.zarray` files with the zarr-internal array metadata (shape, chunk shape, compression etc). Delaying each scale's array creation until the export gets there means that if something breaks (and in case exports are so expensive that it's worth trying to recover successfully written scales), you can directly see on disk how far the export managed to get. On the other hand, it's not like inspecting each of the array directories for whether they're empty or not is much extra effort... I guess I just liked seeing the folders pop up as a progress indicator.

I'm adding a docstring just to make it clear.

lazyflow/utility/io_util/write_ome_zarr.py
btbest: As previous comment, I do use `Optional` meaning basically just "nullable"...

lazyflow/utility/io_util/write_ome_zarr.py
btbest: Actually ` numpy.prod([2, 2, 0.1])` is 0.4, so this goes into the upscale branch despite being a downscaling along two axes :) It just means it doesn't go into the chain-cached route. That caching is supposed to be an optimisation for the most-likely thing to happen most-commonly afterall, so I don't think it's a problem for unusual scenarios to not optimise as well as usual ones. I figured just taking the product of the scaling factors would be a good enough proxy, but we could factor in the actual size along each axis and basically determine whether it's a net up or downscale by total number of pixels.

lazyflow/utility/io_util/write_ome_zarr.py
btbest: Both chunks and scaling are about file size optimisations, so I find it natural that different pixel types result in different file optimisations...

Edit just to add to this: It's true that the mainstay for the more popular tools is to leave chunk shape choice up to the user (ome-zarr-py, ngff-zarr, webknossos, FIJI), with defaults or examples showing shapes like "80", "128" or "128 for 3D, 256 for 2D".  [Eubi-bridge](https://github.com/Euro-BioImaging/EuBI-Bridge) has chunk shape auto-determination in its newest version and runs by a default target of 1MB. Looks like my default is around half what people seem to tend towards, so might be good to increase it.

btbest:
I actually have a couple more todos after playing around a bit more:

- add option to exclude upscales and make this the default (I think people probably generally don't want to upscale segmentations)
- rework OpResize init with `setOrConnectIfAvailable` for all inputs

and 

- handle multiple lanes (technically not this PR but the volumina counterpart)
- maybe: progress signal not one 100% bar per scale but make it global across the entire export? maybe a follow-up PR...

btbest:
With that, the remaining todos are done, except for the big one regarding interpolation.

In regards to implementation, whether I add a slot meta property `appropriate_interpolation_order` or call it something different is basically trivial - I'm happy to go with whatever else if it seems more appropriate to call the slot property "data_semantics" or something like that and have an enum for its values, which then gets mapped to interpolation order in `OpResize`. Given that there aren't any plans for other features that interpret these semantics, I'd call the duck by its name and stick with "appropriate_interpolation_order". The only thing I could think of right now is that Gui classes could pick volumina layer types automatically based on this property if we get it right, rather than hard-coding them like now.

btbest:
Ready for re-review 🥳

k-dominik:
yes that sounds about right.

With the plan of adding those to the input data already, I think it makes even more sense to not call it interpolation_order. Do you see any downsides of that?

ilastik/applets/objectClassification/opObjectClassification.py
```diff
@@ -1376,6 +1378,13 @@ class OpRelabelSegmentation(Operator):
     def setupOutputs(self):
         self.Output.meta.assignFrom(self.Image.meta)
         self.Output.meta.dtype = self.ObjectMap.meta.mapping_dtype
+        # ObjectMap determines data semantics, but the meta property isn't implemented throughout ilastik
+        if "data_semantics" in self.ObjectMap.meta:
+            self.Output.meta.data_semantics = self.ObjectMap.meta.data_semantics
+        elif "data_semantics" in self.Output.meta:
+            # Have to remove if Image had it but ObjectMap didn't
+            # (this could be deleted if we implement this property consistently)
+            self.Output.meta.pop("data_semantics")
```
k-dominik: Since this op assumes a segmentation image as input, should the `data_semantics` not always be `label`? It's hard for me to construct a different case.

tests/test_ilastik/test_workflows/conftest.py
```diff
@@ -0,0 +1,55 @@
+import shutil
+import zipfile
+from pathlib import Path
+
+import pytest
+
+
+@pytest.fixture(scope="module")
```
k-dominik: Scope here should probably be "session". Projects are extracted, data is copied, twice.

tests/test_ilastik/test_workflows/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+def sample_projects_dir(tmp_path_factory) -> Path:
+    tmp_path = tmp_path_factory.mktemp("test_projects")
+    test_data_path = Path(__file__).parent.parent / "data"
+    sample_projects_zip_path = test_data_path / "test_projects.zip"
+    sample_data_dir_path = test_data_path / "inputdata"
+
+    projects_archive = zipfile.ZipFile(sample_projects_zip_path, mode="r")
+    projects_archive.extractall(path=tmp_path)
+
+    shutil.copytree(sample_data_dir_path, tmp_path / "inputdata")
```
k-dominik: is copying all of inputdata necessary? Looks to me you're using only one of the images (that could maybe then also be a fixture). And not sure copying the data is necessary for that.

tests/test_ilastik/test_workflows/test_HeadlessPixelClassificationWorkflow.py
```diff
@@ -364,3 +352,38 @@ def test_headless_from_ome_zarr_http_uri(
...
+
+    run_headless_pixel_classification(
+        testdir,
+        project=ilp_path,
+        raw_data=raw_2d_path,
+        output_filename_format=str(output_path),
+        ignore_training_axistags=True,
+        output_format="multi-scale OME-Zarr",
+        export_source="simple segmentation",
+        export_dtype="float32",  # Segmentation is uint8 by default, force float to expose interpolated values
```
k-dominik: would be thankful for a clarification of that comment. Would float values force a specific interpolation order?

tests/test_ilastik/test_workflows/test_data_semantics.py
```diff
@@ -0,0 +1,186 @@
...
+    shell = HeadlessShell()
+    shell.openProjectFile(projectFilePath=str(project))
+    return shell.projectManager.workflow
+
+
+@pytest.mark.parametrize(
+    "export_source,expected_semantics",
+    [
+        (ObjectClassificationWorkflowPrediction.ExportNames.OBJECT_PREDICTIONS, ImageTypes.Labels),
+        (ObjectClassificationWorkflowPrediction.ExportNames.OBJECT_PROBABILITIES, ImageTypes.Intensities),
```
k-dominik: All outputs of object classification are label image types - or should be. Values are always mapped to whole objects.

tests/test_lazyflow/test_operators/test_opResize.py
```diff
@@ -0,0 +1,122 @@
+from unittest import mock
```
k-dominik, line 22: most test files add the license 🤷

tests/test_lazyflow/test_operators/test_opResize.py
```diff
@@ -0,0 +1,122 @@
...
+
+def test_interpolation_order(graph):
+    """1 is default, test 0."""
+    arr = numpy.indices((25, 25)).sum(0)
+    data = vigra.taggedView(arr, "yx")
+
+    op = OpResize(graph=graph)
+    op.RawImage.setValue(data)
+    op.TargetShape.setValue((5, 5))
+    op.InterpolationOrder.setValue(0)
```
k-dominik: would be safer to use the `InterpolationOrder` enum here, unless other parts of the code explicitly rely on the numerical values.

lazyflow/operators/opResize.py
```diff
@@ -0,0 +1,253 @@
...
+
+from lazyflow.graph import Operator, InputSlot, OutputSlot, Slot
+from lazyflow.roi import enlargeRoiForHalo, roiToSlice
+from lazyflow.rtype import SubRegion
+from lazyflow.utility.data_semantics import ImageTypes
+
+logger = logging.getLogger(__name__)
+
+
+def ceil_roi(roi: NDArray) -> NDArray:
```
k-dominik: the function is short enough, but neither the name nor the "docstring" describe accurately what it is doing. idk `to_integer_bounds` or something would maybe be better.

lazyflow/operators/opResize.py
k-dominik: neat, thank you for checking and providing code!

lazyflow/operators/opResize.py
```diff
@@ -0,0 +1,253 @@
...
+        ), f"Input image ({self.RawImage.meta.shape}) and target shape ({self.TargetShape.value}) must have same axes"
+        assert (
+            "c" not in input_axiskeys
+            or self.TargetShape.value[input_axiskeys.index("c")] == self.RawImage.meta.getTaggedShape()["c"]
+        ), "Cannot resize along channel axis"
+        if (
+            "t" in input_axiskeys
+            and self.TargetShape.value[input_axiskeys.index("t")] != self.RawImage.meta.getTaggedShape()["t"]
+        ):
+            logger.warning("Resizing along time axis. Are you sure this is what you want?")
```
k-dominik: I'm not sure it's a good idea to allow for that "because we can". I at least hope that currently, for a user, it is not possible to reach that warning, right (because other components involved in export prevent that).

lazyflow/operators/opResize.py
```diff
@@ -0,0 +1,253 @@
...
+        # E.g. 2 if the output shape is half the input shape (rather than 0.5).
+        shape_in = self.RawImage.meta.shape
+        shape_out = self.ResizedImage.meta.shape
+        self.scaling_factors = np.divide(shape_in, shape_out)
+
+        interpolation_order = self.InterpolationOrder.value
+        # If higher-order interpolation is desired, need to figure out required padding
+        assert (
+            interpolation_order in self.required_min_padding
+        ), "supports only order 0 (nearest-neighbor) or 1 (linear)"
```
k-dominik: hard coded enum constants as strings here. Potential source for future inconsistency.

lazyflow/operators/opResize.py
```diff
@@ -0,0 +1,253 @@
...
+
+        if interpolation_order > 0:
```
k-dominik: hard coded enum constants as strings here. Potential source for future inconsistency.

```suggestion
        if interpolation_order != self.Interpolation.NEAREST :
```

k-dominik:
I cannot find the conversation you're quoting from.

lazyflow/utility/io_util/write_ome_zarr.py
k-dominik: type checker stopped complaining

lazyflow/utility/io_util/write_ome_zarr.py
k-dominik: hehe, gotta fix my prod ;) but I think you got the example despite this.

btbest:
> that was for me far easier to spot with semantic image types than interpolation order

Not sure why - previously the parametrization of `test_interpolation_order_oc` directly mapped `OBJECT_PROBABILITIES` to `Interpolation.LINEAR`. Now the `data_semantics` test maps `OBJECT_PROBABILITIES` to `ImageTypes.Intensities`, so understanding that implication requires more code base digging to find out what happens to different ImageTypes elsewhere.

Fair point on interpolating the object probabilities though. It would probably be weird for a thresholding on object probabilities to give a different binary mask than the object predictions.

btbest:
> scale-separator

You mean dimension separator, i.e. slash vs. dot? If so, no, the exporter carries over the source's ome metadata, but not the zarr metadata (like chunk size, dimension separator, compression).

ilastik/applets/objectClassification/opObjectClassification.py
btbest: I realise I had this a bit upside-down: the object maps shouldn't have image semantics because they're not images... and opRelabel shouldn't be taking them over from ObjectMaps, but rather the Image already needs to have the property.

The reason this even needed to exist was because `OpBlockwiseObjectClassification` was still assembling its own metadata incorrectly (fixed now - mostly - it's still partially wrong but I erred on the side of caution with preserving old behaviour).

tests/test_ilastik/test_workflows/conftest.py
btbest: Hmm... `test_HeadlessPixelClassificationWorkflow.py` was using the fixture on function-scope up to now. I didn't want to stretch the cross-dependency further than necessary and unzipping once per module seemed good enough.

I'll see whether the tests run fine with session scope.

tests/test_ilastik/test_workflows/conftest.py
btbest: I didn't check which parts of this fixture are (still) necessary. I expect some of the other headless tests might fail because their ilps depend on some of the files here.

tests/test_ilastik/test_workflows/test_HeadlessPixelClassificationWorkflow.py
btbest: No, the test rationale is that linear interpolation would produce fractional values in the scaled segmentation. They need to be exported as float to preserve the fractions.

lazyflow/operators/opResize.py
btbest: I decided to allow it because it's not in principle wrong to scale along t, I'm just not anticipating that anyone wants to use ilastik to change the frame rate of their videos.

I'd leave that decision to if/when we decide to offer customisation of the exported scales... not that it really makes a difference for now whether this is an assert or a warning.

added note: For frame rate changes, the basic methods are also just nn or linear interpolation, so that's not an issue, but I guess since ilastik is supposed to be ML software, people might expect it to do smarter things :)

tests/test_ilastik/test_workflows/conftest.py
btbest: I've checked and looks like it's really just ~~two~~ a few files. Will update

btbest:
That's it from my end, @k-dominik let me know if there's anything else (or if I overlooked any of your comments...)

ilastik/applets/objectClassification/opObjectClassification.py
```diff
@@ -32,6 +32,7 @@
 from lazyflow.operators import OpValueCache, OpSlicedBlockedArrayCache, OpMultiArrayStacker
 from lazyflow.operatorWrapper import OperatorWrapper
 from lazyflow.request import Request, RequestPool, RequestLock
+from lazyflow.utility.data_semantics import ImageTypes
```
k-dominik: unused 🤷

btbest: 🤦

PR #3020, opened 2025-05-28
Display license via dedicated dialog

k-dominik:
previous implementation with webbrowser.open would break on MacOS in the signed app due to missing permissions :)


btbest:
That means the link in the "report issue" dialog is also broken on mac then I guess

btbest (approved):
image

window title would be nice, otherwise lgtm 👍

PR #3028, opened 2025-06-04
Try orb-based codecov upload

k-dominik:
Switched to "orb"-based upload on circleCI. Trickiest part was setting the token in the right place and configuring circleci properly. Works now.

Edit: in the end the reason why codecov stopped working was that they now require token usage on protected branches. Technically this is solved by setting the token properly. However, they deprecated their bash script for upload (that was used before in our CI) and using the orb is the way.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,8 @@
 version: 2.1
+
+orbs:
+  codecov: codecov/codecov@5.4.3
+
 jobs:
   checkout_ilastik:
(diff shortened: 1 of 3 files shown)


k-dominik:
:shrug: tests are passing locally, I suspect some depenency update - investigating

k-dominik:
looks like it's scikit-learn... have to see exactly what the problem is, but for now I'll pin.

PR #3032, opened 2025-06-20
Add Object Classification from Label images worfklow (the sequel)

k-dominik:
This replaces https://github.com/ilastik/ilastik/pull/2921.
I had to rebase on main and resolve some conflicts, so I opened a new PR to preserve the previous review with links to code.

Original Message (slightly edited/removed thinks that have been fixed via main:

-----

`original message`

Inputs such as label images from cellpose or stardist are treated a bit poorly in ilastik - connected components is run again (with possibly different connectivity) resulting in different objects in ilastik after loading such label image data. This is super inconvenient for users: Any prior analysis they did will be hard to merge (different object ids, possibly different objects).

This PR introduces a new workflow, "Object Classification [Raw Data, Label Image]" that internally doesn't do CCA anymore but merely relabels the label image to ensure consecutive labels (required by feature extraction in vigra).

see discussion in https://github.com/ilastik/ilastik/issues/2416

fixes https://github.com/ilastik/ilastik/issues/2416
closes https://github.com/ilastik/ilastik/pull/2921

Summary:
* rewrite of `OpRelabelConsecutive` into an operator that caches two results: the relabeled image and a dictionary that holds the mapping - both necessary in the workflow (relabeled image as input for feature computation, the mapping to export the original labelids to the table).
* Serialization logic for this slot that ensures valid saving of both data to the ilp file
* Added the new column (`original_oid`) to the export logic
* Added the new Object Classification from Label images workflow

There are some slightly unrelated changes (first two commits):
* Binary image layer and object id image have been treated as different sources when displaying, which is not really necessary and can increase cache usage (as two images will be accessed instead of one). This is changed by sourcing the binary layer also from the label image (which has to be computed anyhow for all three flavors of object classification).
* Removed the Binary Image input completely from the Object Classification op: it was only used for visualization.
* I renamed all usages of Binary image to Segmentation image as we used binary segmentations as well as instance segmentations as inputs there.

Changes I didn't include here because changeset is already pretty big:
* change the order of columns in export - `original_oid` column should be easier to find (and fix order to some sorting)
* Allow for export without an image: Currently ilastik forces exporting an image to export the table. If the Label Image already exists outside, and the table correctly links to it, it is really not necessary anymore to export any image.

## Todos (before merge, didn't do it yet for the sake of the review):
* [x] remove https://github.com/ilastik/ilastik/pull/2921/commits/ac9b4d5640a292ed427af271a9e1be4c278a4534, has been fixed on main
* [x] rebase/fixup commits as indicated by commit messages

`/end of original message`

-----

Thank you for the original review. Really unfortunate it's been 8 months since...
Here a summary of the review to make it a bit easier for you (only "larger-scope" comments, one liners i mostly fixed in a6f14c5fd727a8efb91af2459901b83f12115041), @btbest

The main changes from the PR are in the three commits a6f14c5fd727a8efb91af2459901b83f12115041 (review suggestions), 0ee365dac2e628afb1e904ffd36c62c71f64c6e5 (base class for LabelOps), and 0273ffb241eab6c69e9ba18dc955f49429807bae (Gui test for the new workflow).

* "Review could have been quicker if the commits documented copy-pasting (bonus points for separating copying commit from subsequent modifications)"
  * I edited the commit message of the relevant commit a bit. The 6 methods that has some overlap with `OpUnblockedArrayCache` are now mentioned. (612823d793f5f0481654cb1c5f13de9088347424), but didn't go as far as doing separate commits (no bonus points for me)
* "The semantic dependency tracking here was too hard to understand in what sense "OpObjectExtraction expects this interface"; now I get it.
  I think in this context I would keep it on OpRelabelConsecutive to document the dependency (and the fact that OpRelabelConsecutive is actually an "OpLabelVolume", even if they share no base class or naming), but remove it from OpRelabelConsecutive5D so that at least that op can focus on its job."
  * Not sure if that's what you meant but I've added a base class for the operators in 0ee365dac2e628afb1e904ffd36c62c71f64c6e5 and removed `BypassModeEnabled`. Figured in the end it was not needed.
* "I can't really see the value of nesting the cached op here, seems to me one could just put the dtype functionality in the caching op and remove the wrapper? Then the caching op also wouldn't have to double check the dtype on its own side again..." ... "Now I see OpRelabelConsecutive mirrors the internal architecture of OpLabelVolume pretty much 1:1, though it does read a lot cleaner here :) This implicit relationship has tripped me up a few times during the review... I feel it could be helpful to have a noop OpObjectsABC or sth purely to document the connection, and/or rename OpRelabelConsecutive into OpRelabelVolume. I'm immediately also jumping to "well and extract the reorder-dtype-convert-MakeLabels-reorderToInput" cascade into the base class"... Or maybe OpRelabelConsecutive5D should really be a subclass of OpLabelingABC. But I remember you've mentioned going back and forth between implementations and discarding others, so I'm happy to accept what we have here was the cleanest way to solve this."
  * I've added a base class for the operators in 0ee365dac2e628afb1e904ffd36c62c71f64c6e5
* "not a fan of the _impl pattern - i get that for Output it's copypasta and kept to maintain the interface of the base, but I'd inline the impl for RelabelDict"
  * I left it as is, makes sense as it derives from `OpUnblockedArrayCache`. I don't want to inline one of them, one more function call doesn't hurt and at least it's a familiar pattern.


k-dominik:
note: The issue with time slices that put the original PR on hold was fixed in that commit (that got fixed up in this PR) https://github.com/ilastik/ilastik/pull/2921/commits/cfc94cfff862b12bda7523e9175e08a71498b64a

btbest (approved):
Looks like the Object Probabilities test passes on CI also for Windows, which means the problem I was seeing is specific to my machine and not Windows overall. So then all good 🥳

PR #3036, opened 2025-07-01
Logging: Ignore failed session log deletions

btbest:
micro-PR - We don't want ilastik to crash during startup just because a log file is locked.

Unlikely to actually happen to a user in any case, but I somehow managed to lock a log file during debugging and now ilastik can't start 🙃


ilastik/ilastik_logging/default_config.py
```diff
@@ -76,7 +76,10 @@ def _delete_old_session_logs(*, log_dir: Path):
         [child for child in log_dir.iterdir() if re.match(SESSION_LOGFILE_NAME_PATTERN, child.name)], reverse=True
     )
     for log_file in log_files[KEEP_SESSION_LOGS:]:
-        log_file.unlink()
+        try:
+            log_file.unlink()
+        except OSError:
+            pass
```
k-dominik: maybe it would be worth logging the problem? But it looks that logging isn't configured at this stage yet... print?

PR #3040, opened 2025-07-04
Data Export: Round when normalizing float to int

btbest:
We recommend our users convert Probabilities to `uint8` and normalize `0.0 ... 1.0` to `0 ... 255`.

It looks like some people would like to [normalize to `0 ... 100`](https://forum.image.sc/t/export-probabilities-as-8-bit-images/114083) so that pixels contain integer percentage probabilities. This is currently not possible - float imprecision isn't rounded but truncated during dtype conversion and as a consequence, some probability values like `14.99998` get mismapped to `14` instead of `15`.

I went with `numpy.round(a, 5)` to maintain existing behaviour as much as possible while fixing the specific issue for Probabilities, but I am wondering whether `numpy.round(a)` (without restricting to lower decimals) wouldn't actually be the correct thing to do here.

Fixes https://github.com/ilastik/ilastik/issues/3038

(Also a minor typo fix in the other test in the file which probably made the test not actually test anything)

### DIFF
--- a/lazyflow/operators/ioOperators/opFormattedDataExport.py
+++ b/lazyflow/operators/ioOperators/opFormattedDataExport.py
@@ -26,6 +26,8 @@ from builtins import zip
 import os
 import collections
 import warnings
+from functools import partial
+
 import numpy
 from typing import Tuple
 from pathlib import Path
@@ -39,9 +41,24 @@ from lazyflow.roi import roiFromShape
 from lazyflow.operators.generic import OpSubRegion, OpPixelOperator
 from lazyflow.operators.valueProviders import OpMetadataInjector
 from lazyflow.operators.opReorderAxes import OpReorderAxes
-from lazyflow.utility.pathHelpers import PathComponents
 
 from .opExportSlot import OpExportSlot
+
+
+def normalize(minVal, maxVal, outputMinVal, outputMaxVal, target_dtype, a):
+    numerator = numpy.float64(outputMaxVal) - numpy.float64(outputMinVal)
+    denominator = numpy.float64(maxVal) - numpy.float64(minVal)
+    if denominator != 0.0:
+        frac = numpy.float32(numerator / denominator)
+    else:
+        # Denominator was zero.  The user is probably just temporarily changing the values.
+        frac = numpy.float32(0.0)
+    if numpy.issubdtype(a.dtype, numpy.floating) and numpy.issubdtype(target_dtype, numpy.integer):
+        # Float to int: Round to avoid truncation
+        result = numpy.rint(outputMinVal + (a - minVal) * frac).astype(target_dtype)
+    else:
+        result = (outputMinVal + (a - minVal) * frac).astype(target_dtype)
+    return result
 
 
 class OpFormattedDataExport(Operator):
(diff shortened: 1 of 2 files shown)


btbest:
Went with `rint` as per @Tomaz-Vieira's comment on the issue

lazyflow/operators/ioOperators/opFormattedDataExport.py
```diff
@@ -39,11 +41,26 @@
...
 from .opExportSlot import OpExportSlot
 
 
+def normalize(minVal, maxVal, outputMinVal, outputMaxVal, target_dtype, a):
+    numerator = numpy.float64(outputMaxVal) - numpy.float64(outputMinVal)
+    denominator = numpy.float64(maxVal) - numpy.float64(minVal)
+    if denominator != 0.0:
+        frac = numpy.float32(numerator / denominator)
+    else:
+        # Denominator was zero.  The user is probably just temporarily changing the values.
```
Tomaz-Vieira: I don't think this type of function should be aware of lazyflow shenanigans; Rather, it should probably just raise an Exception to be caught by the caller. This also prevents this function from failing silently on bad parameters. Whatever code that is directly dealing with slots and user input can elect not to call this function if some args are missing.

lazyflow/operators/ioOperators/opFormattedDataExport.py
```diff
@@ -39,11 +41,26 @@
 from lazyflow.operators.generic import OpSubRegion, OpPixelOperator
 from lazyflow.operators.valueProviders import OpMetadataInjector
 from lazyflow.operators.opReorderAxes import OpReorderAxes
-from lazyflow.utility.pathHelpers import PathComponents
 
 from .opExportSlot import OpExportSlot
 
 
+def normalize(minVal, maxVal, outputMinVal, outputMaxVal, target_dtype, a):
```
Tomaz-Vieira: How about some typing? Something like:

```python
IN = TypeVar("IN")
OUT = TypeVar("OUT")
SHAPE = TypeVar("SHAPE")
def normalize(
    *,
    minVal: IN,
    maxVal: IN,
    outputMinVal: OUT,
    outputMaxVal: OUT,
    output_dtype: Type[OUT],
    a: np.ndarray[SHAPE, IN]
) -> np.ndarray[SHAPE, OUT]:
```

Also, I'd force the params to be named-only; At the call site this will look like a bunch of numbers whose semantics are hard to read (e.g.: _"does `normalize(a, b, c, d)` represent `minIn`, `minOut`, `maxIn`, `maxOut` ? or maybe `minIn`, `maxIn`, `minOut`, `maxOut`?"_)

lazyflow/operators/ioOperators/opFormattedDataExport.py
```diff
@@ -39,11 +41,26 @@
...
+    numerator = numpy.float64(outputMaxVal) - numpy.float64(outputMinVal)
+    denominator = numpy.float64(maxVal) - numpy.float64(minVal)
+    if denominator != 0.0:
+        frac = numpy.float32(numerator / denominator)
```
Tomaz-Vieira: Why are `numerator` and `denominator` `float4` but `frac` is a `float32` ? In fact, if output is `float64`, I think this might throw away some precision. Maybe `frac` should be the highest precision between input and output? :thinking:

btbest: b4aaad933aef37eb84f3487b3085a0532fff461b Sounds like there were overflow issues. Out of risk aversion, I don't think I'll be touching this, given that it's not documented when exactly those problems would occur

lazyflow/operators/ioOperators/opFormattedDataExport.py
btbest: I think the behaviour is appropriate (normalising from a 0-range to anything should result in an image with 0 drange), but the comment is unnecessary (and my IDE is flagging the equals-comparison against 0.0)

btbest:
All comments should be addressed now, and I've added another test case for the conversion-only path. A re-review would be nice :)

k-dominik (approved):

PR #3042, opened 2025-07-15
OME-Zarr: Better feedback for S3 with auth

btbest:
Just a couple of improvements in error messages when trying to use S3-like web servers. This should make it easier to diagnose from the log what exactly is working and what isn't (i.e. is it attempting the right auth pathway, is auth working).

### DIFF
--- a/lazyflow/utility/io_util/OMEZarrStore.py
+++ b/lazyflow/utility/io_util/OMEZarrStore.py
@@ -261,7 +261,7 @@ def _get_zarr_cache_max_size() -> int:
 def _try_authenticated_aws_s3(uri, kwargs, mode, test_path) -> Optional[FSStore]:
     authenticated_store = FSStore(uri, mode=mode, anon=False, **kwargs)
     try:
-        logger.debug(f"Trying to access path={test_path} at uri={uri} with S3FS credentials.")
+        logger.info(f"Trying AWS S3 with S3FS credentials, path={test_path} at uri={uri}.")
         _ = authenticated_store[test_path]
     except NoCredentialsError:
         logger.warning(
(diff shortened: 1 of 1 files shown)


lazyflow/utility/io_util/OMEZarrStore.py
```diff
@@ -299,14 +299,21 @@ def _try_authenticated_s3_compatible(uri, kwargs, e, test_path) -> FSStore:
...
+            raise ConnectionError(
+                f"Authentication probably succeeded but server refused permission to read {uri}.\n"
+                f'Did you enter the right URL? On some servers, this means the requested file "{test_path}" doesn\'t exist.'
+            ) from ee
+        if isinstance(ee, NoCredentialsError):
             raise ConnectionError(
-                f"Server refused permission to read {uri}.\n"
-                f"Please ensure you have access to this bucket and your credentials are set up for S3FS."
+                f"Authentication failed when trying to read {uri}.\n"
+                "Please confirm your credentials for S3FS are set up correctly."
```
k-dominik: Not sure what's more helpful here, mentioning s3fs, or boto directly... I found the [documentation for boto](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html) more helpful than the [documentation of s3fs](https://s3fs.readthedocs.io/en/latest/#credentials).

On the other hand s3fs seems to have their own env variables for auth...

```
# export FSSPEC_S3_ENDPOINT_URL=https://...
# export FSSPEC_S3_KEY='miniokey...'
# export FSSPEC_S3_SECRET='asecretkey...'
```

In any case, maybe adding something about auth in our docs would also not be too bad: https://www.ilastik.org/documentation/basics/dataselection

btbest: I changed it to just mention both :)

But yeah I should add this to the docs...

k-dominik (approved):

Issue #3029, opened 2025-06-12
Wishlist: Features to utilise pixel sizes once available


btbest:
This issue is to collect all the great ideas of things ilastik could do once we have metadata of physical sizes of pixels. Feel free to add more.

This list is meant besides the minimum feature of reading and writing such metadata from and to tiff, h5, n5 and ome-zarr so that users don't lose the information across their workflows.

* Scale bar in volumina
* Data selection: Physical dataset dimensions
* Features: Physical filter sizes
* Thresholding: Physical units for object size threshold
* OC: Physical object size / volumes
* Tracking: Object velocity
* NN: Scaling input images to match model resolution

k-dominik:
In the same spirit..
 
* write out sizes also in hdf5 (in a way that we can convince fiji to open, too - likely to the axistags that support that)
* pixel-features: deal with anisotropy
* Export applet: spacial ranges in physical units
* Support for image pairs of different resolutions (e.g. high res raw data with lower res segmentation image)
* Gui: measurement ruler tool
* ...

Issue #2870, opened 2024-06-21
Scale bar in visualization


oanegros:
I would like a scale bar showing your current scale. 
This would require reading in pixel sizes, but that would be better anyway and you can assume some (i.e. imagej) standards.

k-dominik:
related: https://github.com/ilastik/ilastik/issues/967

oanegros:
Yes, this is related to integrating pixel size in all algorithms, but imo a lot easier to implement as you do not need to really deal with anisotropic kernels and backwards compatibility etc.

Issue #967, opened 2014-07-15
Specify feature scales relative to image resolution


stuarteberg:
Instead of asking for a raw "sigma" value specified in pixels, the feature scales could be defined in terms of actual resolution.  For ansiotropic data, the same scale would be used for all dimensions, and the actual filter widths (in pixels) would be adjusted according to the resolution in the dataset metadata.

(Of course, we would also want to give the user a convenient way to specify the resolution of each axis in their dataset...)

Update: The iiboost branch now contains a convenient way of allow the user to edit the resolution of each axis in their dataset (via the dataset properties editor window).  As soon as it is merged, that part is solved, at least.  Now we just need to use that info in the feature operator...

PR #3045, opened 2025-07-22
Pixel Sizes

Peter-Laszcz:
Add support for importing/exporting pixel size metadata via TIFF/OME-TIFF.

The addition of this feature would allow the development of features specified in #3029  as well as being a fundamental first step in solving issues #967 and #2870 .

## Checklist

- [X] Add tests.


ilastik/applets/dataSelection/dataSelectionApplet.py
```diff
@@ -269,6 +269,13 @@ def create_lane_configs(
         ignore_training_axistags: bool = False,
         stack_along: str = "z",
     ) -> List[Dict[str, DatasetInfo]]:
+        raw_tags = self.get_lane(-1).get_axistags()["Raw Data"] if self.num_lanes > 0 else None
+        for tagset in input_axes:
+            if not (tagset and raw_tags):
+                continue
+            for axis in tagset.keys():
+                if axis in raw_tags:
+                    tagset.setResolution(axis, raw_tags[axis].resolution)
```
btbest: `input_axes` are a parameter for headless mode... this shouldn't be necessary. Rather, it's not really correct to apply the resolution of the raw training data to new data being run through the project in headless mode.

ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
```diff
@@ -35,7 +36,8 @@ class DatasetColumn:
     TaggedShape = 3
     Scale = 4
     Range = 5
-    NumColumns = 6
+    PixelSizes = 6
```
btbest: I'd rename to "PixelSize" and move this to position 4, in front of the Scale

ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
```diff
@@ -162,6 +164,7 @@ def headerData(self, section: int, orientation: int, role=Qt.DisplayRole):
                 DatasetColumn.TaggedShape: "Shape",
                 DatasetColumn.Range: "Data Range",
                 DatasetColumn.Scale: "Resolution Level",
+                DatasetColumn.PixelSizes: "Pixel Sizes",
```
btbest: I prefer singular "Pixel Size" (I think it should feel more natural for users)

ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
```diff
@@ -219,6 +223,24 @@ def _getDisplayRoleData(self, index):
...
+                    if str(axis.typeFlags) == "Space" or str(axis.typeFlags) == "Time":
+                        axes.append(axis.key)
+                        resolutions.append(axis.resolution)
+                        unit = axis_units[axis.key]
+                        if unit is not None:
+                            units.append(unit)
+                        else:
+                            units.append("None")
+                return str(", ".join(f"{ax}: {res} {un}" for ax, res, un in zip(axes, resolutions, units)))
+            return "None Found"
```
btbest:
```suggestion
        if index.column() == DatasetColumn.PixelSizes:
            if not datasetInfo.axis_units:
                return UninitializedDisplayData[index.column()]
            axis_strings = [f"{tag.key}: {tag.resolution or '-'}" + f" {datasetInfo.axis_units[tag.key]}" if datasetInfo.axis_units[tag.key] else "" for tag in datasetInfo.axistags]
            return ", ".join(axis_strings)
```

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -107,13 +108,15 @@ def __init__(
         drange: Tuple[Number, Number] = None,
         working_scale: str = DEFAULT_SCALE_KEY,
         scale_locked: bool = False,
+        axis_units: str,
```
btbest:
```suggestion
        axis_units: Dict[str, str] = {},  # axis: unit string ("nm")
```

why store them as a json string while inside ilastik? (actual jsonifying is done by the Dataset slot when it gets serialized)

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -140,14 +143,16 @@ def legacy_location(self) -> str:
...
+        metadata = {"display_mode": self.display_mode, "axistags": self.axistags, "axis_units": self.axis_units}
 
         if self.drange is not None:
             metadata["drange"] = self.drange
         elif self.laneDtype == numpy.uint8:
             metadata["drange"] = (0, 255)
         if self.normalizeDisplay is not None:
             metadata["normalizeDisplay"] = self.normalizeDisplay
+        if self.axis_units is not None:
+            metadata["axis_units"] = json.loads(self.axis_units)
```
btbest:
```suggestion
```

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -166,6 +171,7 @@ def create_data_reader(self, parent: Optional[Operator] = None, graph: Optional[
 
     def to_json_data(self) -> Dict:
         return {
+            "axis_units": self.axis_units,
```
btbest:
```suggestion
            "axis_units": json.dumps(self.axis_units),
```

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -198,6 +204,8 @@ def from_h5_group(cls, data: h5py.Group, params: Dict = None):
             axisorder = data["axisorder"][()].decode("utf-8")
             params["axistags"] = vigra.defaultAxistags(axisorder)
 
+        if "axis_units" in data and data.get("axis_units"):
+            params["axis_units"] = data["axis_units"][()].decode("utf-8")
```
btbest:
```suggestion
            params["axis_units"] = json.loads(data["axis_units"][()].decode("utf-8"))
```

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -393,12 +401,19 @@ def __init__(self, *, inner_path: str, project_file: h5py.File, nickname: str =
             default_tags = vigra.AxisTags.fromJSON(self.dataset.attrs["axistags"])
         else:
             default_tags = vigra.defaultAxistags(get_default_axisordering(self.dataset.shape))
+        if "axis_units" in info_kwargs:
+            axis_units = info_kwargs.pop("axis_units")
+        elif "axis_units" in self.dataset.attrs:
+            axis_units = json.dumps(json.loads(self.dataset.attrs["axis_units"]))
+        else:
+            axis_units = json.dumps({tag: "" for tag in default_tags.keys()})
```
btbest:
```suggestion
```

This logic for the default values should be in the base class

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -451,15 +466,29 @@ def internal_paths(self) -> List[str]:
...
+        **info_kwargs,
+    ):
         self.preloaded_array = vigra.taggedView(
             preloaded_array, axistags or get_default_axisordering(preloaded_array.shape)
         )
+        if "axis_units" in info_kwargs:
+            axis_units = info_kwargs.pop("axis_units")
+        elif not axis_units:
+            axis_units = json.dumps({tag: "" for tag in axistags.keys()})
+
```
btbest: See above, shouldn't be necessary in subclasses

lazyflow/operators/ioOperators/opExport2DImage.py
```diff
@@ -83,6 +84,44 @@ def run_export(self):
...
+        extension = os.path.splitext(self.Filepath.value)[1][1:]
+
+        # now write any pixel sizes
+        if (
+            extension in ["tif", "tiff"]
+            and hasattr(self.Input.meta, "axis_units")
+            and any(self.Input.meta.axis_units.items())
+        ):
+            with tifffile.TiffFile(self.Filepath.value) as tif:
+                olddata = tif.asarray()
```
btbest: Does this read and write back the entire dataset? As in, the entire IO is doubled now?

If metadata can be appended, maybe this could work - otherwise I guess the vigra.impex should be in the `else`

lazyflow/operators/ioOperators/opStreamingH5N5Reader.py
```diff
@@ -106,6 +107,9 @@ def setupOutputs(self):
         self.OutputImage.meta.shape = dataset.shape
         self.OutputImage.meta.axistags = axistags
 
+        if "axis_units" in self._h5N5File[internalPath].attrs:
+            self.OutputImage.meta.axis_units = json.loads(self._h5N5File[internalPath].attrs["axis_units"])
```
btbest, line 111:
```suggestion
            self.OutputImage.meta.axis_units = self._h5N5File[internalPath].attrs["axis_units"]
```

lazyflow/operators/ioOperators/opTiffReader.py
```diff
@@ -100,6 +103,13 @@ def setupOutputs(self):
         self.Output.meta.axistags = vigra.defaultAxistags(axes)
         self.Output.meta.dtype = numpy.dtype(dtype_code).type
 
+        self.Output.meta.axis_units = {}
+        for key in axes:
+            self.Output.meta.axis_units[key] = ""
```
btbest:
```suggestion
        self.Output.meta.axis_units = {key: "" for key in axes}
```

lazyflow/operators/ioOperators/opTiffReader.py
```diff
@@ -138,6 +148,86 @@ def execute(self, slot, subindex, roi, result):
...
+                meta = f.pages[0].tags
+
+                # TIFF format stores resolution values as Rational tuples (numerator, denominator)
+
+                if "x" in axes:
+                    self.Output.meta.axistags.setResolution("x", self.get_tiff_res(meta.get("XResolution", None)))
+                    unit = ij_meta.get("unit", "")
+                    if isinstance(unit, str) and unit.startswith("('") and unit.endswith("',)"):
+                        unit = unit[2:-3]
+                    self.Output.meta.axis_units["x"] = self.decode_non_ascii(unit) if unit else ""
```
btbest: subfunction this and describe what's up with the code handling stringified tuples

lazyflow/operators/ioOperators/opTiffReader.py
```diff
@@ -138,6 +148,86 @@ def execute(self, slot, subindex, roi, result):
...
+    def decode_non_ascii(self, raw_string: str):
+        return raw_string.encode("utf-8").decode("unicode_escape").encode("utf-16", "surrogatepass").decode("utf-16")
+
+    def get_tiff_res(self, frac):
+        if frac and frac.value[0] != 0:
+            resolution = frac.value[1] / frac.value[0]
+            if abs(resolution - round(resolution)) < 0.001:
+                resolution = round(resolution)
+            return resolution
+        return 0
```
btbest: Docstring why rounding; especially explain rationale behind the specific threshold.

There should be tests with parameters just above and below the threshold where the difference in their expected results illustrates what the rounding does.

lazyflow/operators/ioOperators/opExport2DImage.py
```diff
@@ -83,6 +84,44 @@ def run_export(self):
...
+                data = olddata.squeeze()
+                x = None
+                y = None
+                axes = "YX"
+                if tagged_shape["c"] > 1:
+                    axes = "CYX"
+                if self.Input.meta.axis_units["x"]:
+                    x = (self.Input.meta.axis_units["x"].encode("unicode_escape").decode("ascii"),)
```
btbest: The encoding-decoding should be extracted into a function that explains why/what for.

Since there's a couple of variants of how this is done, and this happens in a few ops, it might make sense to put it in sth like `lazyflow.utility.io_util.tif_encoding`

lazyflow/operators/ioOperators/opExport2DImage.py
```diff
@@ -83,6 +84,44 @@ def run_export(self):
...
+                    "axes": axes,
+                }
+
+                tifffile.imwrite(
+                    self.Filepath.value,
+                    data,
+                    imagej=True,
+                    metadata=imagej_metadata,
+                    resolution=(
+                        1 / self.Input.meta.axistags["x"].resolution,
```
btbest: Would add a comment here like "tiff convention is to store resolution as 'pixels per space', but internally we keep 'space per pixel'"

lazyflow/operators/ioOperators/opExportMultipageTiff.py
```diff
@@ -82,13 +82,47 @@ def run_export(self) -> None:
...
+                meta_dict[unit_key] = None
+
+                if axis != "c":
+                    meta_dict[size_key] = axistags[axis].resolution
+
+                    unit_value = axis_units.get(axis, "")
+                    if unit_value != "":
+                        meta_dict[unit_key] = unit_value.encode("unicode_escape").decode("ascii")
+                    else:
+                        meta_dict[unit_key] = axis_units[axis] = ""
```
btbest: why the assignment to axis_units?

lazyflow/operators/ioOperators/opTiffReader.py
```diff
@@ -138,6 +148,86 @@ def execute(self, slot, subindex, roi, result):
...
+                        "x": "PhysicalSizeXUnit",
+                        "y": "PhysicalSizeYUnit",
+                        "z": "PhysicalSizeZUnit",
+                        "t": "TimeIncrementUnit",
+                    }
+
+                    for axis in axes:
+                        if axis.lower() == "c":
+                            continue
+                        elif axis.lower() in size_trans_0.keys():
```
btbest: why not just combine `size_trans_0` and `size_trans_1` into a single dict?

also strange naming, I'd go with `size_keys` and `unit_keys`

plus no need to do `.keys()` for `in` :)

lazyflow/operators/ioOperators/opTiffReader.py
```diff
@@ -138,6 +148,86 @@ def execute(self, slot, subindex, roi, result):
...
+                            )
+                        else:
+                            self.Output.meta.axistags.setResolution(
+                                axis.lower(), float(pixels.attrib.get(size_trans_1[axis], 0))
+                            )
+                            self.Output.meta.axis_units[axis.lower()] = self.decode_non_ascii(
+                                pixels.attrib.get(unit_trans_0[axis], "")
+                            )
+
+    def decode_non_ascii(self, raw_string: str):
```
btbest: this function should explain why it does what it does

tests/test_ilastik/test_workflows/axesOrderPreservation/testAxesOrderPreservation.py
```diff
@@ -214,37 +214,38 @@ def compare_results(
...
+            ("5t2d2c", "txyc", None, None),
+            ("5t2d2c", "xytc", None, None),
+            ("5t3d2c", "tzyxc", None, None),
+            ("5t3d2c", "ztxyc", None, None),
+            ("5t3d2c", "xyztc", None, None),
         ],
     )
     @timeLogged(logger)
-    def test_pixel_classification(self, dims, input_axes):
+    def test_pixel_classification(self, dims, input_axes, units, res):
```
btbest: about to merge the ome-zarr multiscale export PR, that comes with some example tests (and fixtures) that will be easier to work off than these headless tests

tests/test_lazyflow/test_operators/test_ioOperators/test_pixel_sizes.py
```diff
@@ -0,0 +1,112 @@
+import os
```
btbest: Still need to have a closer look at the code here, but some more tests would be good. right now this covers OpTiffReader, but you're also implementing units in OpExport2dImage, OpExportMultipageTiff and OpStreamingH5N5Reader.

Going very strict, the changes in DatasetInfo are also untested right now.

I think for this though the best approach might be a pairing session focussed just on testing.

btbest (commented):
Exported a tiff and got this. Looks like there's at least one test missing to make sure this doesn't happen :)

image

ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
```diff
@@ -219,6 +223,19 @@ def _getDisplayRoleData(self, index):
...
+                return UninitializedDisplayData[index.column()]
+            axis_strings = [
+                (
+                    f"{tag.key}: {tag.resolution or '-'}" + f" {datasetInfo.axis_units[tag.key]}"
+                    if datasetInfo.axis_units[tag.key]
+                    else ""
+                )
+                for tag in datasetInfo.axistags
+            ]
+            return ", ".join(axis_strings)
```
btbest: 1) This currently KeyErrors when reading my 5D FIJI stack because the tiff reader doesn't put `c` in the axis_units`

2) Could use a bit of refinement for pretty printing axes with no units or resolution. Maybe sth like 

```suggestion
            axis_strings = [
                (
                    f"{tag.key}: " + (f"{round(tag.resolution, 2)} {datasetInfo.axis_units[tag.key]}"
                    if datasetInfo.axis_units[tag.key] and tag.resolution
                    else str(round(tag.resolution, 2)) if tag.resolution else datasetInfo.axis_units[tag.key])
                )
                for tag in datasetInfo.axistags if tag.resolution or datasetInfo.axis_units[tag.key]
            ]
            return ", ".join(axis_strings)
```

lazyflow/operators/ioOperators/opTiffReader.py
```diff
@@ -138,6 +147,79 @@ def execute(self, slot, subindex, roi, result):
...
+                            continue
+                        elif axis.lower() in size_keys.keys():
+                            self.Output.meta.axistags.setResolution(
+                                axis.lower(), float(pixels.attrib.get(size_keys[axis], 0))
+                            )
+                            self.Output.meta.axis_units[axis.lower()] = tiff_encoding.fromASCII(
+                                pixels.attrib.get(unit_keys[axis], "")
+                            )
+                        else:
+                            continue
```
btbest: unnecessary (intellij should've highlighted this 😝 )

lazyflow/utility/io_util/tiff_encoding.py
```diff
@@ -0,0 +1,10 @@
+def fromASCII(raw_string: str):  # necessary for nonstandard unit characters, e.g mu
+    if raw_string == "":
+        return ""
+    return raw_string.encode("utf-8").decode("unicode_escape").encode("utf-16", "surrogatepass").decode("utf-16")
+
+
+def toASCII(val):  # encoding is necessary for exporting image metadata
+    if val == "":
+        return ""
+    return val.encode("unicode_escape").decode("ascii")
```
btbest: Does this correctly encode emojis? If so, would it be enough in the reverse function above to just `.encode("ascii").decode("unicode_escape")`?

tests/test_ilastik/test_workflows/test_HeadlessPixelClassificationWorkflow.py
```diff
@@ -35,6 +37,16 @@
...
+@pytest.fixture
+def pixel_classification_ilp_2d3c(sample_projects_dir: Path) -> Path:
+    return sample_projects_dir / "PixelClassification2d3c.ilp"
+
+
+@pytest.fixture
+def pixel_classification_ilp_2d_units(sample_projects_dir: Path) -> Path:
+    return sample_projects_dir / "PixelClassification2d_units.ilp"
+
+
```
btbest: `pixel_classification_ilp_2d3c` should be unnecessary here because it's already available via `conftest.py`

The `pixel_classification_ilp_2d_units` fixture should be added next to the other similar fixtures in `conftest`

tests/test_lazyflow/test_operators/test_ioOperators/testOpExportMultipageTiff.py
```diff
@@ -27,11 +27,20 @@
 import vigra
 
 from lazyflow.graph import Graph
-from lazyflow.operators.ioOperators import OpExportMultipageTiff
 from lazyflow.operators.ioOperators.opTiffReader import OpTiffReader
 from lazyflow.operators.opArrayPiper import OpArrayPiper
 from lazyflow.operators.opReorderAxes import OpReorderAxes
 from lazyflow.utility import Pipeline
+from lazyflow.operators.ioOperators import OpExportMultipageTiff, OpExport2DImage
```
btbest: Ctrl + Alt + O (`Optimize Imports`) to put these in conventional order

tests/test_ilastik/test_workflows/test_HeadlessPixelClassificationWorkflow.py
```diff
@@ -392,3 +404,31 @@ def test_headless_ome_zarr_multiscale_export(testdir, tmp_path, sample_projects_
...
+        output_filename_format=str(output_path),
+        ignore_training_axistags=True,
+        output_format="hdf5",
+    )
+
+    assert output_path.exists()
+    opReaderResult = OpInputDataReader(graph=Graph())
+    opReaderResult.FilePath.setValue(str(output_path))
+
+    assert round(opReaderResult.Output.meta.axistags["x"].resolution) == 5.0
```
btbest: numpy.isclose would be better instead of rounding one side as it allows specifying the tolerance explicitly

lazyflow/operators/ioOperators/opExportMultipageTiff.py
```diff
@@ -82,13 +82,47 @@ def run_export(self) -> None:
...
+            "y": "PhysicalSizeYUnit",
+            "z": "PhysicalSizeZUnit",
+            "t": "TimeIncrementUnit",
+        }
+
+        meta = self._opReorderAxes.Output.meta
+        axistags = meta.axistags
+        axis_units = meta.get("axis_units", {})
+
+        for axis in "".join(meta.getAxisKeys()):
```
btbest:
```suggestion
        for axis in meta.getAxisKeys():
```

lazyflow/operators/ioOperators/opExport2DImage.py
```diff
@@ -82,7 +86,41 @@ def run_export(self):
...
+                        (1 / self.Input.meta.axistags["x"].resolution)
+                        if self.Input.meta.axistags["x"].resolution != 0
+                        else 0
+                    ),
+                    (
+                        (1 / self.Input.meta.axistags["y"].resolution)
+                        if self.Input.meta.axistags["y"].resolution != 0
+                        else 0
+                    ),
+                ),  # standard ti1 / self.Input.meta.axistags["y"].resolution,ff convention is to store pixels/space but internally we prefer and use space/pixel
```
btbest:
```suggestion
                ),  # standard tiff convention is to store pixels/space but internally we prefer and use space/pixel
```

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -655,15 +670,17 @@ def __init__(
             graph=Graph(), WorkingDirectory=self.base_dir, FilePath=self.filePath, SequenceAxis=self.sequence_axis
         )
         meta = op_reader.Output.meta.copy()
+        if "axis_units" in info_kwargs:
+            meta.axis_units = info_kwargs.pop("axis_units")
```
btbest: How do you end up in a situation where `axis_units` is in `info_kwargs`?

lazyflow/operators/ioOperators/opTiffReader.py
```diff
@@ -100,6 +104,11 @@ def setupOutputs(self):
         self.Output.meta.axistags = vigra.defaultAxistags(axes)
         self.Output.meta.dtype = numpy.dtype(dtype_code).type
 
+        self.Output.meta.axis_units = {key: "" for key in axes if key != "c"}
```
btbest:
```suggestion
        self.Output.meta.axis_units = {key: "" for key in axes}
```
`axis_units` dict should always contain entries for all axes in `slot.meta.getAxisKeys()` to make our life easier

lazyflow/operators/ioOperators/opTiffReader.py
```diff
@@ -138,6 +147,77 @@ def execute(self, slot, subindex, roi, result):
...
+                        ),
+                    )
+                    # (TIFF format stores resolution values as Rational tuples (numerator, denominator))
+                    unit = self.handle_stringified_tuples(ij_meta.get(unit_keys[ax], ""))
+                    self.Output.meta.axis_units[ax] = tiff_encoding.fromASCII(unit) if unit else ""
+
+            # Look for OME-TIFF metadata (possible in FIJI hyperstacks)
+            else:
+                xml = ET.fromstring(ome_meta)
+                ns = {"ome": "http://www.openmicroscopy.org/Schemas/OME/2016-06"}
```
btbest, line 154: I assume this doesn't actually send any web requests to this URL, but just for our info: Does this mean we'd have to update the URL if the OME schema changes?

lazyflow/operators/ioOperators/opTiffReader.py
```diff
@@ -138,6 +147,77 @@ def execute(self, slot, subindex, roi, result):
...
+
+    def get_tiff_res(self, frac):
+        """
+        Rounding is employed for axes with resolutions deviating extremely slightly from integer values.
+        This is because the TIFF standard introduces floating-point precision errors
+        by storing resolutions as numerator/denominator pairs.
+        """
+        if frac and frac.value[0] != 0:
+            resolution = frac.value[1] / frac.value[0]
+            if abs(resolution - round(resolution)) < 0.00001:
```
btbest:
```suggestion
            if abs(resolution - round(resolution)) < 1e-8:
```
To stay consistent with the test.

Might be better to move this to a global constant (like `TIFF_RESOLUTION_PRECISION = 1e-8`) and then importing that constant in the test rather than hardcoding it in two places.

lazyflow/operators/ioOperators/opTiffReader.py
```diff
@@ -138,6 +147,77 @@ def execute(self, slot, subindex, roi, result):
...
+                            continue
+                        else:
+                            self.Output.meta.axistags.setResolution(
+                                axis.lower(), float(pixels.attrib.get(size_keys[axis], 0))
+                            )
+                            self.Output.meta.axis_units[axis.lower()] = pixels.attrib.get(
+                                unit_keys[axis], ""
+                            )  # OME uses unicode
+
+    def get_tiff_res(self, frac):
```
btbest:
```suggestion
    def round_resolution_to_tiff_precision(self, frac):
```

lazyflow/operators/ioOperators/opTiffReader.py
```diff
@@ -138,6 +147,77 @@ def execute(self, slot, subindex, roi, result):
...
+                        "z": "PhysicalSizeZUnit",
+                        "t": "TimeIncrementUnit",
+                    }
+
+                    for axis in axes:
+                        if axis.lower() == "c":
+                            continue
+                        else:
+                            self.Output.meta.axistags.setResolution(
+                                axis.lower(), float(pixels.attrib.get(size_keys[axis], 0))
```
btbest, line 171: How come `float()` seems to be necessary here but not in the `ij_meta` branch above?

lazyflow/utility/io_util/tiff_encoding.py
```diff
@@ -0,0 +1,57 @@
+def fromASCII(raw_string: str):  # necessary for nonstandard unit characters, e.g mu
```
btbest:
```suggestion
def from_ascii(raw_string: str):
    """
    Necessary for reading nonstandard unit characters out of TIFF tags, e.g mu
    """
```
Convention is snake case for functions :)

lazyflow/utility/io_util/tiff_encoding.py
```diff
@@ -0,0 +1,57 @@
...
+    if raw_string == "":
+        return ""
+    return raw_string.encode("utf-8").decode("unicode_escape").encode("utf-16", "surrogatepass").decode("utf-16")
+
+
+def toASCII(val):  # encoding is necessary for exporting image metadata
```
btbest:
```suggestion
def to_ascii(val):
    """
    Encoding is necessary for exporting image metadata in TIFF tags
    """
```

lazyflow/utility/io_util/tiff_encoding.py
```diff
@@ -0,0 +1,57 @@
...
+    if val == "":
+        return ""
+    return val.encode("unicode_escape").decode("ascii")
+
+
+def toOME(val):
```
btbest:
```suggestion
def to_ome(val):
```

tests/test_ilastik/data/inputdata/2d.tif
btbest, line 1: It would be good to rename this to sth like `2d_with_resolution_meta.tif` so future us can tell what's special about this file

btbest (commented):
Awesome to finally see pixel size metadata in ilastik 🥳 

A few minor things from my side, mostly nitpicks. And I'm excited to see your draft work using the new metadata once it's rebased ;)

lazyflow/operators/ioOperators/opTiffReader.py
Peter-Laszcz: Not likely; there is no active development on the schema as OME-Zarr uses NGFF. They haven't updated the schema since 2016

lazyflow/utility/io_util/tiff_encoding.py
Peter-Laszcz: ...apparently so!

lazyflow/operators/ioOperators/opTiffReader.py
Peter-Laszcz: The XML module for OME only reads the values as strings (because the XML data is a string), so they need to be cast. IJ metadata is stored more like a regular dictionary, so you an store number values.

ilastik/applets/dataSelection/opDataSelection.py
Peter-Laszcz: from_h5_group has axis_units as a parameter- when the project is loaded e.g in the pixel classification tests, the program grabs the axistags from the meta, but axis_units is still passed through as its own keyword argument when from_h5_group returns an opDataSelection class. Turns out we don't need to load the parameter because we have the meta so I've taken it out

tests/test_ilastik/test_workflows/test_HeadlessPixelClassificationWorkflow.py
```diff
@@ -392,3 +394,31 @@ def test_headless_ome_zarr_multiscale_export(testdir, tmp_path, sample_projects_
...
+        ignore_training_axistags=True,
+        output_format="hdf5",
+    )
+
+    assert output_path.exists()
+    opReaderResult = OpInputDataReader(graph=Graph())
+    opReaderResult.FilePath.setValue(str(output_path))
+
+    assert numpy.isclose(opReaderResult.Output.meta.axistags["x"].resolution, 5.0, 1e-8)
+    assert numpy.isclose(opReaderResult.Output.meta.axistags["y"].resolution, 6.0, 1e-5)
```
btbest: Why different tolerances for x and y..?

ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
```diff
@@ -219,6 +223,23 @@ def _getDisplayRoleData(self, index):
...
+        if index.column() == DatasetColumn.PixelSize:
+            if not datasetInfo.axis_units:
+                return UninitializedDisplayData[index.column()]
+            axis_strings = [
+                (
+                    f"{tag.key}: "
+                    + (
+                        f"{tag.resolution} {datasetInfo.axis_units[tag.key]}"
+                        if datasetInfo.axis_units[tag.key] and tag.resolution
+                        else str(tag.resolution) if tag.resolution else datasetInfo.axis_units[tag.key]
```
btbest:
```suggestion
                        f"{round(tag.resolution, 2)} {datasetInfo.axis_units[tag.key]}"
                        if datasetInfo.axis_units[tag.key] and tag.resolution
                        else str(round(tag.resolution, 2)) if tag.resolution else datasetInfo.axis_units[tag.key]
```

tests/test_ilastik/test_applets/dataSelection/testDataSelectionGui.py
```diff
@@ -35,9 +35,9 @@ def mock_DatasetDetailedInfoTableModel():
         m.get_scale_options = lambda _row: {"100_100_10": "100, 100, 10", "50_50_10": "50, 50, 10"} if _row > 0 else {}
         m.is_scale_locked = lambda _row: _row == 2
         data_rows = [
-            ["image", "/usr/root/image.png", "", "z: 1, y: 10, x: 10", "", ""],
-            ["image", "precomputed://http://localhost:8000", "", "z: 1, y: 10, x: 10", "100, 100, 10", ""],
-            ["image", "precomputed://http://localhost:8000", "", "z: 1, y: 10, x: 10", "50, 50, 10", ""],
+            ["image", "/usr/root/image.png", "", "z: 1, y: 10, x: 10", "", "", ""],
+            ["image", "precomputed://http://localhost:8000", "", "z: 1, y: 10, x: 10", "", "100, 100, 10", ""],
+            ["image", "precomputed://http://localhost:8000", "", "z: 1, y: 10, x: 10", "", "50, 50, 10", ""],
```
k-dominik: a bit nitpicky (sorry), and I know it is just the previous pattern evolving, and also "just" a test, but something like

```Python
        data_row_t = namedtuple("data_row_t", ("nickname", "location", "internalID", "taggedShape", "pixelSize", "scale", "range"))
# ...
            data_row_t(nickname="image", location="/usr/root/image.png", internalID="", taggedShape= "z: 1, y: 10, x: 10", pixelSize="", scale="", range=""),
            data_row_t(nickname="image", location="precomputed://http://localhost:8000", internalID="", taggedShape="z: 1, y: 10, x: 10", pixelSize="", scale="100, 100, 10", range=""),
            data_row_t(nickname="image", location="precomputed://http://localhost:8000", internalID="", taggedShape="z: 1, y: 10, x: 10", pixelSize="", scale="50, 50, 10", range=""),
# ...
```

looks more maintainable in the long term

tests/test_ilastik/data/test_projects.zip
k-dominik, line 1: The zipped test project are challenging to maintain. I checked your commits modifying it and the commit messages should address which of the project files was added/modified for what purpose to make it easier to track. I can't say we were poster children regards to this in the past, but I would really appreciate an edit to the messages there.

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -107,13 +108,15 @@ def __init__(
         drange: Tuple[Number, Number] = None,
         working_scale: str = DEFAULT_SCALE_KEY,
         scale_locked: bool = False,
+        axis_units: Dict[str, str] = {},  # { axis: physical unit string e.g. "micron" }
```
k-dominik: Code like this can result in subtle bugs, consider

```Python
class A:
    def __init__(self, axis_units={}):
        self.axis_units=axis_units

d1 = A()
d1.axis_units["x"] = 3

d2 = A()
print(d2.axis_units)
# prints {"x": 3}
```
whereas:
```Python
class B:
    def __init__(self, axis_units=None):
        self.axis_units = axis_units or {}

d1 = B()
d1.axis_units["x"] = 3

d2 = B()
print(d2.axis_units)
# prints {}
```

I'm not implying that there are any code paths that would modify `DatasetInfo.axis_units` that way - I haven't checked - but I prefer not having to worry about it.

lazyflow/operators/ioOperators/opExportMultipageTiff.py
```diff
@@ -82,13 +82,45 @@ def run_export(self) -> None:
...
+
+        for axis in meta.getAxisKeys():
+            if axis in size_trans:
+                size_key = size_trans[axis]
+                unit_key = unit_trans[axis]
+                pixels[size_key] = None
+                pixels[unit_key] = None
+
+                if axis != "c":
+                    pixels[size_key] = float(axistags[axis].resolution)  # OME uses unicode
```
k-dominik: Apologies in advance (I'm tired), more a question: How do unicode and floats mix here?

lazyflow/operators/ioOperators/opExportMultipageTiff.py
```diff
@@ -82,13 +82,45 @@ def run_export(self) -> None:
         page_buf = RoiRequestBufferIter(self._opReorderAxes.Output, self._batch_size, iterate_axes="tzc")
         page_buf.progress_signal.subscribe(self.progressSignal)
 
+        meta_dict = {
+            "axes": "".join(k.upper() for k in self._opReorderAxes.Output.meta.getAxisKeys()),
+            "SignificantBits": 8,
```
k-dominik: incredibly naive, sorry. I don't know the ome spec - to me this looks like number of bits per pixel? which should be datatype dependent?

lazyflow/operators/ioOperators/opTiffReader.py
```diff
@@ -138,6 +147,77 @@ def execute(self, slot, subindex, roi, result):
 
                 result[roi_page_ndindex] = page_data[roiToSlice(*roi_within_page)]
 
+    def setPixelSizes(self, axes, ij_meta, ome_meta):
+        with tifffile.TiffFile(self._filepath, mode="r") as f:
+            if ij_meta:
+                meta = f.pages[0].tags
+                resolution_keys = {"x": "XResolution", "y": "YResolution", "z": "spacing", "t": "finterval"}
+                unit_keys = {"x": "unit", "y": "yunit", "z": "zunit", "t": "tunit"}
```
k-dominik: gotta love it - is it really the case that the key for the x-unit has no prefix (like `y` in "yunit" for the y-unit)? Probably to enable the "one unit for all" use-case?

tests/test_lazyflow/test_operators/test_ioOperators/test_pixel_sizes.py
```diff
@@ -0,0 +1,476 @@
...
+            ["t", "z", "c", "y", "x"],
+            (15, 23, 2, 65, 58),
+            [120, 0.54321, 0.0, 6.000024000096, 13],
+            ["ms", "µm", "", "mm", "cm"],
+        ),
+    ],
+)
+def test_write_OpExportMultipageTiff(graph, tmp_path, axes, shape, resolutions, units):
+    op_data = get_data_op_with_pixel_size_meta(graph, axes, shape, resolutions, units)
+    out_path = str(tmp_path / "3d_export.tif")
```
k-dominik: I get that it has to have a name, and it's only a test, but there seem to be other dims than 3d as well in this test...

k-dominik (commented):
Hello @Peter-Laszcz - really solid work! Nicely tested, too. I had a few comments - some of them are taste :). @btbest, maybe you want to weigh in on those. Really critical for me right now is documenting the changes to the zipped projects (as this is hard to follow otherwise), and maybe the default dictionary value in DatasetInfo.

Another thought experiment: Say I export a subregion - one of the axes becomes singleton and I chose to drop it, would the metadata reflect this accurately in all file formats?

ilastik/applets/dataSelection/opDataSelection.py
btbest: Oof... is there any good way to write this that doesn't make it look like the function is expecting to be called with `None`?

lazyflow/operators/ioOperators/opExportMultipageTiff.py
Peter-Laszcz: That's my fault- accidentally pasted from OpTiffReader

tests/test_lazyflow/test_operators/test_ioOperators/test_pixel_sizes.py
Peter-Laszcz: True- How does multipage_export sound?

ilastik/applets/dataSelection/opDataSelection.py
Peter-Laszcz: > Oof... is there any good way to write this that doesn't make it look like the function is expecting to be called with `None`?

This would match what the function currently has for axistags but to make it more explicit  i could change the comment to # expects { axis: physical unit string e.g. "micron" }

k-dominik: how about making it a non optional argument and handling the rest in the subclasses? More code, but would make sure this is not called with None?

lazyflow/operators/ioOperators/opTiffReader.py
btbest: I had wondered what that's about, and I guess you're right - if there's only "unit", it's meant to apply to all axes...

ilastik/applets/dataSelection/opDataSelection.py
btbest: I think I'd rather stick with the None default then... at least it's in line with the other arguments, where we also don't know if they are actually ever None during runtime.

lazyflow/operators/ioOperators/opExportMultipageTiff.py
Peter-Laszcz: Yes- I believe this got added in by Pycharm's autofill because it wasn't there before; I've taken it back out now.

Peter-Laszcz:
> Hello @Peter-Laszcz - really solid work! Nicely tested, too. I had a few comments - some of them are taste :). @btbest, maybe you want to weigh in on those. Really critical for me right now is documenting the changes to the zipped projects (as this is hard to follow otherwise), and maybe the default dictionary value in DatasetInfo.
> 
> Another thought experiment: Say I export a subregion - one of the axes becomes singleton and I chose to drop it, would the metadata reflect this accurately in all file formats?

Just tried it and came up with an interesting result. Ilastik exports the correct pixel sizes to the correct areas in the metadata, but FIJI can't read it properly:

image

Checking with tifffile, these fields appear to have the correct numbers and format. One thing I *did* notice - not sure if it's any bearing on what's going on here- was that Ilastik seems to have added in a channel axis which has shifted the image data along one axis, i.e. the image was meant to have an XY axis set and shape of (16,16) but has been exported with an axis set of CYX and a shape of (16,16,2), also producing the unfortunate side effect of the X axis being listed as having a length of 2 in the metadata :/

lazyflow/utility/io_util/tiff_encoding.py
```diff
@@ -0,0 +1,57 @@
...
+def to_ascii(val):  # encoding is necessary for exporting image metadata
+    if val == "":
+        return ""
+    return val.encode("unicode_escape").decode("ascii")
+
+
+def to_ome(val):
+    if val.lower() in ome_units:
+        return ome_units[val.lower()]
+    return "pixel"
```
k-dominik, line 16: Not sure this is sensible without considering the axis type. "pixel" is fine for spacial axes, but for time I'd expect something like "frame"

lazyflow/utility/io_util/tiff_encoding.py
```diff
@@ -0,0 +1,57 @@
...
+
+def to_ascii(val):  # encoding is necessary for exporting image metadata
+    if val == "":
+        return ""
+    return val.encode("unicode_escape").decode("ascii")
+
+
+def to_ome(val):
+    if val.lower() in ome_units:
+        return ome_units[val.lower()]
```
k-dominik: The `ome_units` dict is not exhaustive (see https://ome-model.readthedocs.io/en/stable/developers/ome-units.html). I would've expected that there is a library out there that would handle such things somehow, but ... I couldn't find one. Now solving this would blow this PR out of proportion. In the meantime it would be good to at least log a warning if there's an unsupported unit.

tests/test_lazyflow/test_operators/test_ioOperators/test_pixel_sizes.py
```diff
@@ -0,0 +1,404 @@
...
+    np.testing.assert_array_equal(op.OutputImage.value, data)
+
+
+def test_write_read_roundtrip_h5(
+    graph,
+    tmp_path,
+    axes=["c", "y", "x"],
+    shape=(2, 58, 67),
+    resolutions=[0.0, 6.000024000096, 13],
+    units=["", "μm", "mm"],
```
k-dominik, line 378: interesting - but also kind of weird pattern (or I don't see the light yet). Using parameters that only take default values works - but it would take exactly the same space in the function body - any particular reason to do it this way?

k-dominik (commented):
Wow that is some clean git history :) Must have been a fun rebase. Thank you for going through that!

Really solid work with exemplary test coverage.

On the danger of being incredibly annoying I found a case that I would expect to work, not working:

Exporting to hdf5 (probabilities from pixel classification) (very cool with units now) and loading this data as raw data seems to fail if the input has no channel axes (then, no unit gets written, but ui seems to expect it to be there for all the axes):

<details><summary>screenshot</summary>

image

</details>
<details><summary>traceback</summary>

```pytb

ERROR 2025-09-16 16:05:57,750 excepthooks 46725 8611815616 Traceback (most recent call last):
  File "...ilastik/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py", line 145, in data
    return self._getDisplayRoleData(index)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...ilastik/ilastik/applets/dataSelection/dataLaneSummaryTableModel.py", line 75, in _getDisplayRoleData
    return model_cls._getDisplayRoleData(self, index)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...ilastik/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py", line 232, in _getDisplayRoleData
    axis_strings = [
                   ^
  File "...ilastik/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py", line 242, in <listcomp>
    if tag.resolution or datasetInfo.axis_units[tag.key]
                         ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
KeyError: 'c'

```
</details>

example data:  [blobs.zip](https://github.com/user-attachments/files/22364609/blobs.zip)

lazyflow/utility/io_util/tiff_encoding.py
Peter-Laszcz: I've actually been writing that library in my spare time 😅 there's a lot of ground for me to cover but I've already done the spatial stuff so I have exhaustive versions of the dictionaries- it'd be a straight copy/ paste job but I'd understand if that would be considered beyond the scope of the PR

tests/test_lazyflow/test_operators/test_ioOperators/test_pixel_sizes.py
Peter-Laszcz: My thought process was that the roundtrip only needed one test, but should someone want to test other values for any reason this makes it easier; thinking about it now though I don't see a situation where this would happen so happy to take it out

k-dominik: all right, fine with me :)

lazyflow/utility/io_util/tiff_encoding.py
k-dominik: Ah cool, such a library would definitely fill a niche!

Peter-Laszcz:
> Exporting to hdf5 (probabilities from pixel classification) (very cool with units now) and loading this data as raw data seems to fail if the input has no channel axes (then, no unit gets written, but ui seems to expect it to be there for all the axes):
> screenshot
> traceback
> 
> ```
> ERROR 2025-09-16 16:05:57,750 excepthooks 46725 8611815616 Traceback (most recent call last):
>   File "...ilastik/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py", line 145, in data
>     return self._getDisplayRoleData(index)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "...ilastik/ilastik/applets/dataSelection/dataLaneSummaryTableModel.py", line 75, in _getDisplayRoleData
>     return model_cls._getDisplayRoleData(self, index)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "...ilastik/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py", line 232, in _getDisplayRoleData
>     axis_strings = [
>                    ^
>   File "...ilastik/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py", line 242, in <listcomp>
>     if tag.resolution or datasetInfo.axis_units[tag.key]
>                          ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
> KeyError: 'c'
> ```
> 
> example data: [blobs.zip](https://github.com/user-attachments/files/22364609/blobs.zip)

This happened because I was allowing channel axes to display pixel sizes, because there is a niche case where  someone could potentially write a unit to the channel axis in the h5 format. However, after some thought, the commit I am pushing to fix this will stop channel pixel sizes from appearing in the data table, mainly as anything written in that field would not be used/manipulable in ilastik in any way I can fathom (however, the data is still passed through the pipeline so if a use case emerges it's somewhat futureproof)

k-dominik:
I probably wouldn't mind showing the unit value for the c axis if it's there, but let's leave it as is. For the circleCI failure you probably need to merge main into this PR again (vaguely remember running into this and fixing it).

Peter-Laszcz:
> would you say this is more or less complete now @Peter-Laszcz?
Definitely- as soon as circleCI stops segfaulting :sweat_smile:

PR #3048, opened 2025-08-06
Pixel label explorer

k-dominik:
## Checklist

- [ ] Add tests.

### DIFF
--- a/ilastik/applets/labeling/connectBlockedLabels.py
+++ b/ilastik/applets/labeling/connectBlockedLabels.py
@@ -1,0 +1,260 @@
+###############################################################################
+#   ilastik: interactive learning and segmentation toolkit
+#
+#       Copyright (C) 2011-2025, the ilastik developers
+#                                <team@ilastik.org>
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# In addition, as a special exception, the copyright holders of
+# ilastik give you permission to combine ilastik with applets,
+# workflows and plugins which are not covered under the GNU
+# General Public License.
+#
+# See the LICENSE file for details. License information is also available
+# on the ilastik web site at:
+#          http://ilastik.org/license.html
+###############################################################################
+from dataclasses import dataclass
+from enum import IntEnum, StrEnum
+from typing import Dict, Iterator, List, Optional, Sequence, Tuple, Union
+
+from lazyflow.utility.io_util.write_ome_zarr import SPATIAL_AXES
+
+
+class Neighbourhood(IntEnum):
+    NONE = 0
+    SINGLE = 1  # Axis aligned, 2D: 4, 3D: 6
+    NDIM = 2  # Full block, 2D: 8, 3D: 26
+
+
+class BlockBoundary(IntEnum):
+    NONE = 0
+    START = 1
+    STOP = 2
+
+
+class SpatialAxesKeys(StrEnum):
+    x = "x"
+    y = "y"
+    z = "z"
+
+
+_Boundaries2D_SINGLE = ((BlockBoundary.NONE, BlockBoundary.STOP), (BlockBoundary.STOP, BlockBoundary.NONE))
+_Boundaries2D_NDIM = _Boundaries2D_SINGLE + ((BlockBoundary.STOP, BlockBoundary.STOP),)
+
+
+_Boundaries3D_SINGLE = (
+    (BlockBoundary.NONE, BlockBoundary.NONE, BlockBoundary.STOP),
+    (BlockBoundary.NONE, BlockBoundary.STOP, BlockBoundary.NONE),
+    (BlockBoundary.STOP, BlockBoundary.NONE, BlockBoundary.NONE),
+)
+_Boundaries3D_NDIM = _Boundaries3D_SINGLE + (
+    (BlockBoundary.NONE, BlockBoundary.STOP, BlockBoundary.STOP),
+    (BlockBoundary.STOP, BlockBoundary.NONE, BlockBoundary.STOP),
+    (BlockBoundary.STOP, BlockBoundary.STOP, BlockBoundary.NONE),
+    (BlockBoundary.STOP, BlockBoundary.STOP, BlockBoundary.STOP),
+)
+
+
+BoundaryDescrRelative = Dict[SpatialAxesKeys, BlockBoundary]
+BoundaryDescr = Dict[SpatialAxesKeys, Union[int, None]]
+
+_INVERTED_BOUNDARIES: dict[BlockBoundary, BlockBoundary] = {
+    BlockBoundary.NONE: BlockBoundary.NONE,
+    BlockBoundary.START: BlockBoundary.STOP,
+    BlockBoundary.STOP: BlockBoundary.START,
+}
+
+_Boundaries = {
+    (Neighbourhood.NONE, 2): (),
+    (Neighbourhood.NONE, 3): (),
+    (Neighbourhood.SINGLE, 2): _Boundaries2D_SINGLE,
+    (Neighbourhood.SINGLE, 3): _Boundaries3D_SINGLE,
+    (Neighbourhood.NDIM, 2): _Boundaries2D_NDIM,
+    (Neighbourhood.NDIM, 3): _Boundaries3D_NDIM,
+}
+
+
+@dataclass(slots=True, frozen=True)
+class Region:
+    """
+    Representation of the bounding box belonging to a certain label
+    """
+
+    axistags: Sequence[str]
+    slices: Tuple[slice, ...]
+    label: int
+
+    def __post_init__(self):
+        if len(self.axistags) != len(self.slices):
+            raise ValueError(f"{self.axistags=} and {self.slices=} not matching in length")
+
+        if any(_is_unbound(sl) for sl in self.slices):
+            raise ValueError(f"Cannot construct region with unbound slicing")
+
+    @property
+    def tagged_slicing(self) -> Dict[str, slice]:
+        return dict(zip(self.axistags, self.slices))
+
+    @property
+    def tagged_center(self) -> Dict[str, float]:
+        """Returns the center of the interval defined by slicing
+
+        Stop element is assumed to be exclusive.
+        """
+        return {k: (sl.stop - 1 - sl.start) / 2 + sl.start for k, sl in self.tagged_slicing.items()}
+
+    def is_at_boundary(self, boundary: BoundaryDescr) -> bool:
+        if all(b is None for b in boundary.values()):
+            return False
+
+        is_at_boundary = True
+        for k, coord in boundary.items():
+            if coord is not None:
+                sl = self.tagged_slicing[k]
+                if sl.start == coord or sl.stop == coord:
+                    continue
+                else:
+                    return False
+
+        return is_at_boundary
+
+    def __hash__(self):
+        """
+        Hash based on volume defined by slicing.
+        In our setting regions cannot overlap in start and stop coordinates
+        -> result in a unique key.
+        """
+        return hash(tuple((sl.start, sl.stop) for sl in self.slices))
+
+    @classmethod
+    def with_slices(cls, reg: "Region", tagged_slices: Dict[str, slice]) -> "Region":
+        axistags = [k for k in tagged_slices]
+        slices = tuple([v for v in tagged_slices.values()])
+        return Region(axistags=axistags, slices=slices, label=reg.label)
+
+
+@dataclass(frozen=True)
+class Block:
+    """
+    Single block, as part of a larger blocking (knows about neighbors starting points)
+
+    Args:
+      axistags: up to 5D (tzyxc)
+      slices: Block position. Neighbours are expected to start at block stops
+      regions: Block-local regions (as in slices be within limits of block size)
+      neighborhood: Neighborhood to consider for finding neighboring blocks
+    """
+
+    axistags: Sequence[str]
+    slices: Tuple[slice, ...]
+    regions: Sequence[Region]
+    neigbourhood: Neighbourhood = Neighbourhood.NDIM
+
+    @property
+    def block_start(self) -> Tuple[int, ...]:
+        return tuple(sl.start for sl in self.slices)
+
+    @property
+    def tagged_slices(self) -> dict[str, slice]:
+        return {tag: sl for tag, sl in zip(self.axistags, self.slices)}
+
+    @property
+    def spatial_axes(self) -> List[SpatialAxesKeys]:
+        return [SpatialAxesKeys(x) for x in self.axistags if x in SPATIAL_AXES]
+
+    @property
+    def tagged_start(self) -> dict[str, slice]:
+        return {tag: sl.start for tag, sl in zip(self.axistags, self.slices)}
+
+    def boundary_regions(self, boundary: BoundaryDescrRelative, label: Optional[int] = None) -> Iterator[Region]:
+        if self.neigbourhood == Neighbourhood.NONE:
+            return
+
+        def _boundary_index_from_slice(sl: slice, boundary: BlockBoundary) -> Union[int, None]:
+            if boundary == BlockBoundary.NONE:
+                return None
+            if boundary == BlockBoundary.START:
+                return 0
+            if boundary == BlockBoundary.STOP:
+                return sl.stop - sl.start
+
+        tagged_boundary: dict[SpatialAxesKeys, Union[int, None]] = {}
+        for at, bd in boundary.items():
+            tagged_boundary[at] = _boundary_index_from_slice(self.tagged_slices[at], bd)
+
+        def _labelmatch(region_label: Union[int, None]) -> bool:
+            if label == None:
+                return True
+            else:
+                return region_label == label
+
+        for region in self.regions:
+            if region.is_at_boundary(tagged_boundary) and _labelmatch(region.label):
+                yield region
+
+    def boundaries_positive(self) -> Iterator[BoundaryDescrRelative]:
+        n_spatial = len(self.spatial_axes)
+        assert n_spatial in (2, 3)
+
+        boundary_iter = _Boundaries[(self.neigbourhood, n_spatial)]
+
+        for boundary in boundary_iter:
+            if all(x == BlockBoundary.NONE for x in boundary):
+                continue
+
+            yield dict(zip(self.spatial_axes, boundary))
+
+    def boundary_regions_positive(self) -> Iterator[Tuple[BoundaryDescrRelative, Region]]:
+        for boundary in self.boundaries_positive():
+            for boundary_region in self.boundary_regions(boundary):
+                yield boundary, boundary_region
+
+    def neighbour_start_coordinates(self, boundary: BoundaryDescrRelative) -> Tuple[int, ...]:
+        tagged_slices = self.tagged_slices
+
+        neighbour_start: dict[str, int] = {}
+        for k, sl in tagged_slices.items():
+            if k not in SPATIAL_AXES:
+                neighbour_start[k] = sl.start
+                continue
+
+            ks = SpatialAxesKeys(k)
+            b = boundary[ks]
+            if b == BlockBoundary.NONE:
+                neighbour_start[k] = sl.start
+            elif b == BlockBoundary.START:
+                neighbour_start[k] = sl.start - (sl.stop - sl.start)
+            elif b == BlockBoundary.STOP:
+                neighbour_start[k] = sl.stop
+            else:
+                # unreachable
+                raise NotImplemented()
+
+        return tuple(neighbour_start[k] for k in self.axistags)
+
+    def region_in_world(self, region: Region) -> Region:
+        tagged_region_sl = region.tagged_slicing
+        tagged_block_sl = self.tagged_slices
+        world_sl = {}
+        for k in self.axistags:
+            if k in SPATIAL_AXES:
+                sl = tagged_region_sl[k]
+                world_sl[k] = slice(sl.start + self.tagged_start[k], sl.stop + self.tagged_start[k])
+            else:
+                world_sl[k] = tagged_block_sl[k]
+
+        return region.with_slices(region, world_sl)
+
+
+def add_tagged_coords(t1: Dict[str, float], t2: Dict[str, float]) -> Dict[str, float]:
+    assert set(t1.keys()).issubset(set(t2.keys())), f"{list(t1.keys())=}, {list(t2.keys())=}"
+    return {k: t1[k] + t2[k] for k in t1.keys()}
+
+
+def _is_unbound(sl: slice) -> bool:
+    return sl.start is None or sl.stop is None
--- a/ilastik/applets/labeling/connectBlockedLabels.py
+++ b/ilastik/applets/labeling/connectBlockedLabels.py
@@ -22,6 +22,9 @@ from dataclasses import dataclass
 from enum import IntEnum, StrEnum
 from typing import Dict, Iterator, List, Optional, Sequence, Tuple, Union
 
+import vigra
+from vigra.analysis import extractRegionFeatures
+
 from lazyflow.utility.io_util.write_ome_zarr import SPATIAL_AXES
 
 
@@ -258,3 +261,107 @@ def add_tagged_coords(t1: Dict[str, float], t2: Dict[str, float]) -> Dict[str, f
 
 def _is_unbound(sl: slice) -> bool:
     return sl.start is None or sl.stop is None
+
+
+def extract_annotations(labels_data: vigra.VigraArray) -> Tuple[Region, ...]:
+    """
+
+    Args:
+      labels_data: integer valued VigraArray with axistags. Note: only spatial
+        axistags (x, y, z) are allowed. 0 values are considered background. Arrays
+        are casted to`uint32` for labeling, potentially lossy coming from `uint64`.
+
+    Returns:
+      Tuple of connected components wrapped in Region objects
+    """
+    axistags = [a.key for a in labels_data.axistags]
+    assert len(axistags) in [2, 3]
+    assert all(k in SPATIAL_AXES for k in axistags)
+
+    if len(axistags) == 3:
+        connected_components = vigra.analysis.labelVolumeWithBackground(
+            labels_data.astype("uint32"),
+            neighborhood=26,
+        )
+    else:
+        connected_components = vigra.analysis.labelImageWithBackground(
+            labels_data.astype("uint32"),
+            neighborhood=8,
+        )
+    feats = extractRegionFeatures(
+        labels_data.astype("float32"),
+        connected_components,
+        ignoreLabel=0,
+        features=["RegionCenter", "Coord<Maximum>", "Coord<Minimum>", "Minimum"],
+    )
+
+    # shape: (n_objs, ndim), we +1 the maximum bounding box value to make it
+    # compatible with slice objects, where the stop is exclusive
+    # first object is ignored -> background object.
+    max_bb = feats["Coord<Maximum>"][1:].astype("uint32") + 1
+    min_bb = feats["Coord<Minimum>"][1:].astype("uint32")
+
+    slices: list[Tuple[slice, ...]] = []
+    for min_, max_ in zip(min_bb, max_bb):
+        slices.append(tuple(slice(mi, ma) for mi, ma in zip(min_, max_)))
+
+    # we pass the label image as "image", so minimum will be the same label
+    labels = feats["Minimum"][1:].astype("uint32")
+
+    spatial_at = [at for at in axistags if at in SPATIAL_AXES]
+    regions = tuple(Region(axistags=spatial_at, slices=sl, label=label) for sl, label in zip(slices, labels))
+
+    return regions
+
+
+def connect_regions(block_dict: Dict[Tuple[int, ...], Block]) -> Dict[Region, Region]:
+    regions_dict: Dict[Region, Region] = {}  # region_world: region_world, updated with anchor as value
+
+    def get_anchor(region):
+        if region not in regions_dict:
+            return region
+
+        current = region
+        while True:
+            anchor = regions_dict[current]
+            if anchor == current:
+                return anchor
+            current = anchor
+
+    for block in block_dict.values():
+        for region in block.regions:
+            region_world = block.region_in_world(region)
+            if region_world not in regions_dict:
+                regions_dict[region_world] = region_world
+
+        for boundary, region in block.boundary_regions_positive():
+            region_world = block.region_in_world(region)
+            block_start = block.neighbour_start_coordinates(boundary)
+            if block_start not in block_dict:
+                continue
+
+            neighbour_block = block_dict[block_start]
+            boundary_in_neighbour = {k: _INVERTED_BOUNDARIES[v] for k, v in boundary.items()}
+            for reg in neighbour_block.boundary_regions(boundary_in_neighbour, label=region.label):
+                neighbour_region_world = neighbour_block.region_in_world(reg)
+                if check_overlap(region_world, neighbour_region_world):
+                    anchor_neighbour = get_anchor(neighbour_region_world)
+                    anchor_reg = get_anchor(region_world)
+
+                    regions_dict[anchor_neighbour] = anchor_reg
+
+    return regions_dict
+
+
+def check_overlap(region_a: Region, region_b: Region) -> bool:
+    assert region_a.axistags == region_b.axistags
+
+    overlap = True
+    for k, v in region_a.tagged_slicing.items():
+
+        if k not in SPATIAL_AXES:
+            continue
+        if not (v.stop >= region_b.tagged_slicing[k].start and region_b.tagged_slicing[k].stop >= v.start):
+            return False
+
+    return overlap
--- a/tests/test_ilastik/widgets/test_connectBlockedLabels.py
+++ b/tests/test_ilastik/widgets/test_connectBlockedLabels.py
@@ -1,0 +1,184 @@
+###############################################################################
+#   ilastik: interactive learning and segmentation toolkit
+#
+#       Copyright (C) 2011-2025, the ilastik developers
+#                                <team@ilastik.org>
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# In addition, as a special exception, the copyright holders of
+# ilastik give you permission to combine ilastik with applets,
+# workflows and plugins which are not covered under the GNU
+# General Public License.
+#
+# See the LICENSE file for details. License information is also available
+# on the ilastik web site at:
+#          http://ilastik.org/license.html
+###############################################################################
+import pytest
+
+from ilastik.applets.labeling.connectBlockedLabels import (
+    Block,
+    BlockBoundary,
+    BoundaryDescrRelative,
+    Neighbourhood,
+    Region,
+    SpatialAxesKeys,
+    extract_annotations,
+)
+
+
+def test_region_construction_raises_on_axistags_slices_mismatch():
+    with pytest.raises(ValueError):
+        _ = Region(axistags="z", slices=(slice(0, 1), slice(1, 2)), label=1)
+
+
+def test_region_construction_raises_on_unbound_slicing():
+    with pytest.raises(ValueError):
+        _ = Region(axistags="z", slices=(slice(None, 1),), label=1)
+
+
+def test_region_center():
+    region = Region(axistags="xy", slices=(slice(1, 2), slice(0, 42)), label=1)
+    assert region.tagged_center == dict(x=1.0, y=20.5)
+
+
+def test_region_with_slices():
+    region = Region(axistags="xy", slices=(slice(1, 2), slice(0, 42)), label=1)
+
+    new_slices = {"x": slice(5, 6), "y": slice(42, 10)}
+    region2 = Region.with_slices(region, tagged_slices=new_slices)
+    assert region2.tagged_slicing == new_slices
+
+
+@pytest.mark.parametrize(
+    "neighbourhood, expected_neighbours",
+    [
+        (Neighbourhood.NONE, []),
+        (
+            Neighbourhood.SINGLE,
+            [{"y": BlockBoundary.NONE, "x": BlockBoundary.STOP}, {"y": BlockBoundary.STOP, "x": BlockBoundary.NONE}],
+        ),
+        (
+            Neighbourhood.NDIM,
+            [
+                {"y": BlockBoundary.NONE, "x": BlockBoundary.STOP},
+                {"y": BlockBoundary.STOP, "x": BlockBoundary.NONE},
+                {"y": BlockBoundary.STOP, "x": BlockBoundary.STOP},
+            ],
+        ),
+    ],
+)
+def test_block_neighbourhood_types(neighbourhood, expected_neighbours):
+    block = Block(
+        axistags=("t", "y", "x", "c"),
+        slices=(slice(10, 11), slice(100, 200), slice(200, 250), slice(0, 1)),
+        regions=(),
+        neigbourhood=neighbourhood,
+    )
+
+    assert list(block.boundaries_positive()) == expected_neighbours
+
+
+@pytest.mark.parametrize(
+    "boundary",
+    [
+        {"x": BlockBoundary.START, "y": BlockBoundary.NONE},
+        {"x": BlockBoundary.STOP, "y": BlockBoundary.NONE},
+        {"x": BlockBoundary.NONE, "y": BlockBoundary.START},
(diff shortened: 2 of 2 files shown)


k-dominik:
> I did instead find the behaviour in the video below on brush stroke... Some table entries suddenly double up. It goes back to correct if I switch lanes back and forth. Weirdest one with the eraser, the duplicate showed a different label name. Are you able to reproduce this?

Thank you @btbest for looking into this PR. And oh wow this looks insane! I tried reproducing it but so far without success.... Need to get to the bottom of this, obviously. Do you think you could share this project with me?

ilastik/applets/counting/countingGui.py
```diff
@@ -130,13 +130,15 @@ def __init__(self, parentApplet, topLevelOperatorView):
...
-        labelSlots.labelNames = topLevelOperatorView.LabelNames
+        labelSlots = LabelingSlots(
+            labelInput=topLevelOperatorView.LabelInputs,
+            labelOutput=topLevelOperatorView.LabelImages,
+            labelEraserValue=topLevelOperatorView.opLabelPipeline.opLabelArray.eraser,
+            labelDelete=topLevelOperatorView.opLabelPipeline.opLabelArray.deleteLabel,
+            labelNames=topLevelOperatorView.LabelNames,
+        )
+
+        labelSlots.nonzeroLabelBlocks = topLevelOperatorView.NonzeroLabelBlocks
```
btbest: Why not add this to the instantiation above..?

tests/test_lazyflow/test_utility/test_pipeline.py
```diff
@@ -0,0 +1,99 @@
...
+class OpNoInput(Operator):
+    Output = OutputSlot()
+
+    def setupOutputs(self):
+        self.Output.meta.shape = (1, 2, 3)
+        self.Output.meta.dtype = "uint8"
+
+
+class OpOptionalInput(OpNoInput):
+    SomeOptionalSlot = InputSlot(value="notSet")
```
btbest, line 36: _Technically_, these aren't `optional`, they're `default` :)

tests/test_lazyflow/test_utility/test_pipeline.py
```diff
@@ -0,0 +1,99 @@
...
+        start_op = pl.add(OpNoInput)
+        op = pl.add(Op)
+
+        assert op.Input.upstream_slot == start_op.Output
+        assert op.Output.meta.shape == start_op.Output.meta.shape
+
+
+def test_slot_kwargs(graph):
+    with Pipeline(graph=graph) as pl:
+        start_op = pl.add(OpOptionalInput, SomeOptionalSlot="123")
```
btbest, line 77: Is this behaviour specific to slots with default values or could this test run with `Op`?

ilastik/shell/gui/shellWidgets.py
```diff
@@ -138,21 +214,44 @@ def __init__(self, parent):
 
         self._centralStack = CentralWidgetStack()
 
-        self.insertWidget(0, self._mainControls)
-        self.insertWidget(1, self._centralStack)
+        self._secondaryStack = SecondaryControlsStack()
+        self._secondary_controls_hadle = None
```
btbest:
```suggestion
        self._secondary_controls_handle = None
```

ilastik/shell/gui/shellWidgets.py
```diff
@@ -130,6 +205,7 @@ def __init__(self, parent):
             parent=parent,
         )
         self.setChildrenCollapsible(False)
+        self._secondary_handle = None
```
btbest: What's this for?

ilastik/shell/gui/shellWidgets.py
```diff
@@ -171,8 +270,67 @@ def setActiveCentralWidget(self, centralWidget: Union[QWidget, None]):
 
         self._setActiveStackWidget(self._centralStack, centralWidget)
 
+    def setActiveSecondaryControls(self, secondaryControlsWidget: Union[QWidget, None]):
+        if secondaryControlsWidget is not None:
```
btbest:
```suggestion
        if secondaryControlsWidget is None:
            self._secondaryStack.setVisible(False)
            return
```

Super minor matter of taste, but I prefer inverting the condition + early return so that the rest of the function can be on base indent level.

ilastik/shell/gui/shellWidgets.py
```diff
@@ -20,23 +20,76 @@
...
+
+class LabeledHandle(QSplitterHandle):
+
+    def __init__(self, orientation, parent: "HorizontalMainSplitter"):
+        super().__init__(orientation, parent)
+        self.is_collapsed: bool
+        self.toggle_button = VerticalButton("Secondary controls", self)
+        self.toggle_button.setAttribute(Qt.WA_TransparentForMouseEvents)
+        self.toggle_button.setFocusPolicy(Qt.NoFocus)
+        self.setToolTip("Secondary controls. Double click to show/hide, click and drag to resize.")
```
btbest:
```suggestion
        self.setToolTip("Double click to show/hide, click and drag to resize.")
```

Not sure this adds anything for the user.

Or... call this in `setButtonText` with `f"{text}. Double click to..."` :D

ilastik/shell/gui/shellWidgets.py
```diff
@@ -116,12 +169,34 @@ def __init__(self, parent: Optional[QWidget] = None) -> None:
...
+        self.setLineWidth(0)
+
+
 class HorizontalMainSplitter(QSplitter):
     """Widget intended for the main GUI elements
 
     Attrs:
-      mainControls: Contains AppletDrawer and ViewerControlstack
-      centralStack: Contains CentralWidget
+      imageSelectionCombo: combobox for image index switching
```
btbest, line 199: It also has the `appletBar`..?

ilastik/shell/gui/shellWidgets.py
```diff
@@ -138,21 +214,44 @@ def __init__(self, parent):
...
+        self.setSizes([sz[0], sz[1] + sz[2], 0])
         for stacked_widget in [self._viewerControlStack, self._centralStack]:
             for i in reversed(list(range(stacked_widget.count()))):
                 lastWidget = stacked_widget.widget(i)
                 stacked_widget.removeWidget(lastWidget)
 
+        self._clear_secondary_control_stack()
+
+    def _clear_secondary_control_stack(self):
+
```
btbest:
```suggestion
```

ilastik/shell/gui/shellWidgets.py
```diff
@@ -171,8 +270,67 @@ def setActiveCentralWidget(self, centralWidget: Union[QWidget, None]):
...
     def setImageSelectionGroupVisible(self, visible: bool):
         self._imageSelectionGroup.setVisible(visible)
+
+    def createHandle(self):
+        n_widgets = self.count()
+
+        if n_widgets == 2:
+            if self._secondary_controls_hadle is None:
+                self._secondary_controls_hadle = LabeledHandle(self.orientation(), self)
+            return self._secondary_controls_hadle
```
btbest: How this creates a LabeledHandle only for the second splitter and not the first one (between appletBar and centralStack) is a mystery to me...

ilastik/applets/labeling/connectBlockedLabels.py
```diff
@@ -0,0 +1,378 @@
...
+    (Neighbourhood.NONE, 2): (),
+    (Neighbourhood.NONE, 3): (),
+    (Neighbourhood.SINGLE, 2): _Boundaries2D_SINGLE,
+    (Neighbourhood.SINGLE, 3): _Boundaries3D_SINGLE,
+    (Neighbourhood.NDIM, 2): _Boundaries2D_NDIM,
+    (Neighbourhood.NDIM, 3): _Boundaries3D_NDIM,
+}
+
+
+@dataclass(slots=True, frozen=True)
```
btbest: `slots=True` available only in py3.10 and up

ilastik/applets/labeling/connectBlockedLabels.py
```diff
@@ -0,0 +1,378 @@
...
+    Single block, as part of a larger blocking (knows about neighbors starting points)
+
+    Args:
+      axistags: up to 5D (tzyxc)
+      slices: Block position. Neighbours are expected to start at block stops
+      regions: Block-local regions (as in slices be within limits of block size)
+      neighborhood: Neighborhood to consider for finding neighboring blocks
+    """
+
+    axistags: Sequence[str]
```
btbest, line 210:
```suggestion
    axiskeys: Sequence[str]
```

😬

Even if the "tagged shape" naming breaks the rule, the code base is mostly consistent in things that are named "tags" actually being vigra.AxisTags. Would be nice to stay that way.

Edit: Nevermind, apparently not even vigra is internally consistent on this 🤦 

image

ilastik/applets/labeling/connectBlockedLabels.py
```diff
@@ -0,0 +1,378 @@
...
+    def block_start(self) -> Tuple[int, ...]:
+        return tuple(sl.start for sl in self.slices)
+
+    @property
+    def tagged_slices(self) -> dict[str, slice]:
+        return {tag: sl for tag, sl in zip(self.axistags, self.slices)}
+
+    @property
+    def spatial_axes(self) -> List[SpatialAxesKeys]:
+        return [SpatialAxesKeys(x) for x in self.axistags if x in SPATIAL_AXES]
```
btbest, line 225: Do we want to maybe move `SPATIAL_AXES` out of the ome-zarr context into something more central like `utility.shapes`?

`from lazyflow.utility.io_util.write_ome_zarr import SPATIAL_AXES` just reads a bit weirdly in this file.

The StrEnum implementation seems to play more nicely with type checking on my end... although it's a bit inconvenient that it doesn't support `in`.

ilastik/applets/labeling/connectBlockedLabels.py
```diff
@@ -0,0 +1,378 @@
...
+class Block:
+    """
+    Single block, as part of a larger blocking (knows about neighbors starting points)
+
+    Args:
+      axistags: up to 5D (tzyxc)
+      slices: Block position. Neighbours are expected to start at block stops
+      regions: Block-local regions (as in slices be within limits of block size)
+      neighborhood: Neighborhood to consider for finding neighboring blocks
+    """
```
btbest, line 208:
```suggestion
    """
    Single block, as part of a larger blocking (knows about neighbours starting points)

    Args:
      axistags: up to 5D (tzyxc)
      slices: Block position. Neighbours are expected to start at block stops
      regions: Block-local regions (as in slices be within limits of block size)
      neighbourhood: Neighbourhood to consider for finding neighbouring blocks
    """
```

On the one hand, code convention is American spelling without the extra "u" (cf signatures of the vigra functions used further down). On the other hand, we already have "Neighbours" in OpLazyConnectedComponents, so it's not entirely out of pattern in ilastik 😆 The docstring should at least be consistent with the spelling in the code though - in particular for the names of the args it lists :P

ilastik/applets/labeling/connectBlockedLabels.py
```diff
@@ -0,0 +1,378 @@
...
+            b = boundary[ks]
+            if b == BlockBoundary.NONE:
+                neighbour_start[k] = sl.start
+            elif b == BlockBoundary.START:
+                neighbour_start[k] = sl.start - (sl.stop - sl.start)
+            elif b == BlockBoundary.STOP:
+                neighbour_start[k] = sl.stop
+            else:
+                # unreachable
+                raise NotImplemented()
```
btbest:
```suggestion
                raise NotImplementedError()
```

ilastik/applets/labeling/connectBlockedLabels.py
```diff
@@ -0,0 +1,378 @@
...
+    axistags: Sequence[str]
+    slices: Tuple[slice, ...]
+    label: int
+
+    def __post_init__(self):
+        if len(self.axistags) != len(self.slices):
+            raise ValueError(f"{self.axistags=} and {self.slices=} not matching in length")
+
+        if any(_is_unbound(sl) for sl in self.slices):
+            raise ValueError(f"Cannot construct region with unbound slicing")
```
btbest:
```suggestion
            raise ValueError("Cannot construct region with unbound slicing")
```

tests/test_ilastik/widgets/test_LabelExplorer.py
```diff
@@ -0,0 +1,260 @@
...
+) -> Tuple[InputSlot, LabelExplorerWidget]:
+    ex = LabelExplorerWidget(label_pipeline_non_blocked.NonzeroBlocks, label_pipeline_non_blocked.Output)
+    qtbot.addWidget(ex)
+    return label_pipeline_non_blocked.LabelSinkInput, ex
+
+
+@pytest.mark.parametrize(
+    "gui_variant, expected_neighbourhood",
+    [
+        pytest.param("label_explorer_blocked", Neighbourhood.SINGLE, id="blocked"),
```
btbest: Once again pytest being great with typing: pytest.param breaks it and all parameters in the test are now simply typed as `ParameterSet` :D

In general I do prefer to have the IDs, and this pytest.param syntax is better than just passing `ids` to `parametrize`, because it keeps the ID with the actual case parameters. Sad that one has to choose to have either actually convenient IDs, or typing.

tests/test_ilastik/widgets/test_LabelExplorer.py
```diff
@@ -0,0 +1,260 @@
...
+        pytest.param("label_explorer_non_blocked", 1, id="non-blocked"),
+    ],
+)
+def test_update_on_update(
+    qtbot,
+    gui_variant,
+    request,
+    expected_n_blocks: int,
+):
+    islot = InputSlot
```
btbest:
```suggestion
    islot: InputSlot
```

ilastik/applets/labeling/connectBlockedLabels.py
```diff
@@ -0,0 +1,378 @@
...
+import vigra
+from vigra.analysis import extractRegionFeatures
+
+from lazyflow.utility.io_util.write_ome_zarr import SPATIAL_AXES
+
+
+class Neighbourhood(IntEnum):
+    NONE = 0
+    SINGLE = 1  # Axis aligned, 2D: 4, 3D: 6
+    NDIM = 2  # Full block, 2D: 8, 3D: 26
```
btbest:
```suggestion
class Neighbourhood(IntEnum):
    """
    Connectivity paradigm denoting which blocks are considered within a particular block's neighbourhood.
    The current simplified algorithm only takes into account direct neighbours with a shared block edge/side, i.e. neighbours along the image axes (SINGLE).
    A precise algorithm would need to take into account the complete neighbourhood (NDIM).
    The value corresponds to the number of hops, as in the connectivity parameter for skimage.measure.label (see their docs for more context).
    """
    NONE = 0
    SINGLE = 1  # Axis aligned, 2D: 4, 3D: 6
    NDIM = 2  # Axis aligned and diagonals, 2D: 8, 3D: 26
```

As mentioned in person, I find this nomenclature very unintuitive, to the point that I couldn't figure out what any of these terms (or the numbers in the comments) mean.

ilastik/applets/labeling/connectBlockedLabels.py
```diff
@@ -0,0 +1,378 @@
...
+
+
+def connect_regions(block_dict: Dict[Tuple[int, ...], Block]) -> Dict[Region, Region]:
+    """
+    Connect regions between a sparse blocking
+    """
+    regions_dict: Dict[Region, Region] = {}  # region_world: region_world, updated with anchor as value
+
+    def get_anchor(region):
+        if region not in regions_dict:
```
btbest, line 374:
```suggestion
    def get_anchor(region: Region) -> Region:
        """
        An anchor is the first already seen region connected to `region` through any number of other regions.
        """
        if region not in regions_dict:
```

k-dominik:
label deletion should be sorted - but label merging can also result in inconsistent label names.. 🫨

ilastik/applets/labeling/connectBlockedLabels.py
```diff
@@ -0,0 +1,378 @@
...
+        In our setting regions cannot overlap in start and stop coordinates
+        -> result in a unique key.
+        """
+        return hash(tuple((sl.start, sl.stop) for sl in self.slices))
+
+    @classmethod
+    def with_slices(cls, reg: "Region", tagged_slices: Dict[str, slice]) -> "Region":
+        axistags = [k for k in tagged_slices]
+        slices = tuple([v for v in tagged_slices.values()])
+        return Region(axistags=axistags, slices=slices, label=reg.label)
```
btbest: Kinda weird for a class method to take an instance of its own class as an argument. It's more convenient like this:

```suggestion
    def with_slices(self, tagged_slices: Dict[str, slice]) -> "Region":
        axistags = [k for k in tagged_slices]
        slices = tuple([v for v in tagged_slices.values()])
        return cls(axistags=axistags, slices=slices, label=self.label)
```

Alternatively I'd make it a real alternative constructor `from_slices` that takes slices and `label: int`, not an existing instance.

k-dominik:
Allright- I could not break label deletion, or label merging anymore, but then again...

tests/test_ilastik/widgets/test_connectBlockedLabels.py
```diff
@@ -0,0 +1,309 @@
...
+        ({"x": BlockBoundary.NONE, "y": BlockBoundary.STOP}, 3),
+        ({"x": BlockBoundary.START, "y": BlockBoundary.START}, 1),
+        ({"x": BlockBoundary.STOP, "y": BlockBoundary.START}, 1),
+        ({"x": BlockBoundary.START, "y": BlockBoundary.STOP}, 1),
+        ({"x": BlockBoundary.STOP, "y": BlockBoundary.STOP}, 1),
+    ],
+    ids=["left", "right", "top", "bottom", "top-left", "top-right", "bottom-left", "bottom_right"],
+)
+def test_block_boundary_regions_2d(boundary: BoundaryDescrRelative, expected_regions):
+    # TODO: test that the correct regions are returned!
```
btbest: image

ilastik/applets/labeling/connectBlockedLabels.py
```diff
@@ -0,0 +1,378 @@
...
+
+    def boundaries_positive(self) -> Iterator[BoundaryDescrRelative]:
+        n_spatial = len(self.spatial_axes)
+        assert n_spatial in (2, 3)
+
+        boundary_iter = _Boundaries[(self.neigbourhood, n_spatial)]
+
+        for boundary in boundary_iter:
+            if all(x == BlockBoundary.NONE for x in boundary):
+                continue
```
btbest:
```suggestion
            assert any(x != BlockBoundary.NONE for x in boundary), "shouldn't define nonsense boundary descriptions"
```

Not sure what this check contributes otherwise - `boundary_iter` only contains globals constants, so dev error if there's something wrong in there

ilastik/applets/labeling/connectBlockedLabels.py
```diff
@@ -0,0 +1,378 @@
...
+                return 0
+            if boundary == BlockBoundary.STOP:
+                return sl.stop - sl.start
+
+        tagged_boundary: dict[SpatialAxesKeys, Union[int, None]] = {}
+        for at, bd in boundary.items():
+            tagged_boundary[at] = _boundary_index_from_slice(self.tagged_slices[at], bd)
+
+        def _labelmatch(region_label: Union[int, None]) -> bool:
+            if label == None:
```
btbest:
```suggestion
            if label is None:
```

ilastik/applets/labeling/connectBlockedLabels.py
```diff
@@ -0,0 +1,378 @@
...
+}
+
+_Boundaries = {
+    (Neighbourhood.NONE, 2): (),
+    (Neighbourhood.NONE, 3): (),
+    (Neighbourhood.SINGLE, 2): _Boundaries2D_SINGLE,
+    (Neighbourhood.SINGLE, 3): _Boundaries3D_SINGLE,
+    (Neighbourhood.NDIM, 2): _Boundaries2D_NDIM,
+    (Neighbourhood.NDIM, 3): _Boundaries3D_NDIM,
+}
```
btbest: These globals are a bit mixed up in their order, which obscures what's related to what. They also added confusion because they look like hard-coded enumerations, but then clearly lack all combinations with `START`. I feel like it's taken me too long to figure out even the basic concepts used in this algorithm, and I still haven't figured out the algorithm itself. The small refactor below provides a lot more clarity to me as to which globals belong with which, and what's their purpose/context. (It also provides a more natural place for the docstring I'm suggesting in the comment on `boundaries_positive` otherwise 😄)

```suggestion
_INVERTED_BOUNDARIES: dict[BlockBoundary, BlockBoundary] = {
    BlockBoundary.NONE: BlockBoundary.NONE,
    BlockBoundary.START: BlockBoundary.STOP,
    BlockBoundary.STOP: BlockBoundary.START,
}


BoundaryDescrRelative = Dict[SpatialAxesKeys, BlockBoundary]
BoundaryDescr = Dict[SpatialAxesKeys, Union[int, None]]


class AllPositiveBoundariesRelative(dict):
        """
        Boundary descriptions in the positive directions.
        E.g. if the origin is thought to be the top-left corner of a 2d image, positive boundaries are in the bottom and right directions.
        Independent of axes, this simply translates into all boundaries in the STOP-direction of slices.
        """
    _Boundaries2D_SINGLE = ((BlockBoundary.NONE, BlockBoundary.STOP), (BlockBoundary.STOP, BlockBoundary.NONE))
    _Boundaries2D_NDIM = _Boundaries2D_SINGLE + ((BlockBoundary.STOP, BlockBoundary.STOP),)

    _Boundaries3D_SINGLE = (
        (BlockBoundary.NONE, BlockBoundary.NONE, BlockBoundary.STOP),
        (BlockBoundary.NONE, BlockBoundary.STOP, BlockBoundary.NONE),
        (BlockBoundary.STOP, BlockBoundary.NONE, BlockBoundary.NONE),
    )
    _Boundaries3D_NDIM = _Boundaries3D_SINGLE + (
        (BlockBoundary.NONE, BlockBoundary.STOP, BlockBoundary.STOP),
        (BlockBoundary.STOP, BlockBoundary.NONE, BlockBoundary.STOP),
        (BlockBoundary.STOP, BlockBoundary.STOP, BlockBoundary.NONE),
        (BlockBoundary.STOP, BlockBoundary.STOP, BlockBoundary.STOP),
    )

    _Boundaries = {
        (Neighbourhood.NONE, 2): (),
        (Neighbourhood.NONE, 3): (),
        (Neighbourhood.SINGLE, 2): _Boundaries2D_SINGLE,
        (Neighbourhood.SINGLE, 3): _Boundaries3D_SINGLE,
        (Neighbourhood.NDIM, 2): _Boundaries2D_NDIM,
        (Neighbourhood.NDIM, 3): _Boundaries3D_NDIM,
    }

    @classmethod
    def with_axes(
        cls, spatial_axes: Sequence[SpatialAxesKeys], neighbourhood: Neighbourhood
    ) -> Iterator[BoundaryDescrRelative]:
        # This method is Block.boundaries_positive
        assert len(spatial_axes) in (2, 3), "cannot handle higher than 3-dimensional neighbourhoods"
        all_positive_boundary_descriptions = cls._Boundaries[(neighbourhood, len(spatial_axes))]

        for boundary in all_positive_boundary_descriptions:
            yield cls(zip(spatial_axes, boundary))
```

We would then remove `boundaries_positive()` from `Block` and change `boundary_regions_positive` like below, which now more obviously denotes what's related to the actual `Block` instance, and what's simply global constants in this Block's context:

```
    def boundary_regions_positive(self) -> Iterator[Tuple[BoundaryDescrRelative, Region]]:
        for boundary in AllPositiveBoundariesRelative.with_axes(self.spatial_axes, self.neigbourhood):
            for boundary_region in self.boundary_regions(boundary):
                yield boundary, boundary_region
```

And of course `test_block_neighbourhood_types` needs to change the assert to `assert list(PositiveBoundaryDescriptions.with_axes(block.spatial_axes, block.neigbourhood)) == expected_neighbours
`

ilastik/applets/labeling/connectBlockedLabels.py
```diff
@@ -0,0 +1,378 @@
...
+                return True
+            else:
+                return region_label == label
+
+        for region in self.regions:
+            if region.is_at_boundary(tagged_boundary) and _labelmatch(region.label):
+                yield region
+
+    def boundaries_positive(self) -> Iterator[BoundaryDescrRelative]:
+        n_spatial = len(self.spatial_axes)
```
btbest:
```suggestion
    def boundaries_positive(self) -> Iterator[BoundaryDescrRelative]:
        """
        Returns the descriptions of boundaries in the positive direction along each axis.
        E.g. if the origin is thought to be the top-left corner of a 2d image, positive boundaries are in the bottom and right directions.
        """
        n_spatial = len(self.spatial_axes)
```
        
If this method doesn't get removed in the refactor, this docstring at least helps a bit

tests/test_ilastik/widgets/test_connectBlockedLabels.py
```diff
@@ -0,0 +1,309 @@
...
+    regions_dict = connect_regions(block_dict)
+
+    assert len(regions_dict) == 5
+    assert len([k for k, v in regions_dict.items() if k == v]) == 4
+
+    merged_region = next(k for k, v in regions_dict.items() if k != v)
+
+    parent_region = regions_dict[merged_region]
+
+    assert merged_region.label == parent_region.label == 1
```
btbest, line 399: I wanted to suggest that this should assert the merged and parent region are actually the correct regions from within the blocks, but I forgot that it's not possible to link the world regions back to the block regions they came from 🙃

ilastik/applets/labeling/labelExplorer.py
```diff
@@ -0,0 +1,267 @@
...
+        if not self._table_initialized:
+            QTimer.singleShot(0, self.sync_state)
+        super().paintEvent(a0)
+
+    def _setupUi(self):
+        layout = QVBoxLayout()
+        self.tableWidget = QTableWidget()
+        self.tableWidget.setColumnCount(len(self._display_axistags) + 1)
+        self.tableWidget.setHorizontalHeaderLabels(self._display_axistags + ["label"])
+        self.tableWidget.setSelectionBehavior(QAbstractItemView.SelectRows)
```
btbest, line 132:
```suggestion
        self.tableWidget.setSelectionBehavior(QAbstractItemView.SelectRows)
        self.tableWidget.setSelectionMode(QAbstractItemView.SingleSelection)
```

I could see cool things happening with multi-selection (was almost hoping for a context menu to pop up when I select multiple and right-click), but until then single-selection is nicer.

tests/test_ilastik/widgets/test_LabelExplorer.py
```diff
@@ -0,0 +1,260 @@
...
+        # delete all labels
+        label_pipeline_non_blocked.DeleteLabel.setValue(1)
+        label_pipeline_non_blocked.DeleteLabel.setValue(-1)
+
+        assert wrapped_populate_table.call_count == 2
+        assert len(label_explorer._block_cache) == 0
+        assert label_explorer.tableWidget.rowCount() == 0
+
+
+def test_delete_labels_empty_table_blocked(
```
btbest, line 197: Now that I saw the fix for the deadlock, I'm weirded out that this test didn't catch it...

btbest (commented):
What to say, it's a much-needed addition that we should've probably worked on years ago 🙃 I love how this works in the GUI. Once this is released, we might consider updating the table shared at https://forum.image.sc/t/comparaison-of-some-tools-for-3d-dense-ground-truth-annotations/38918. It lists ilastik as not having a label explorer, which now it finally does ;)

image

In terms of styling, it looks a bit more meh on Windows compared to your screen cap. I was trying to get a 1px padding on the vertical button but didn't manage (not sure it's worth the effort). If the top margin of the table can be eliminated to align with the tops of the remaining interface, that would be nice. The bottom margin makes it not quite line up with the bottom edge of the other widgets too (the "Crosshairs" text would be the reference), but that's much less obvious.

Overall from the code side I found it hard to wrap my head around the connected-labels code and I've made some suggestions that I think would've helped me understand it faster (I still don't fully understand the algorithm in `connect_labels` 😅). Ironically I feel there are docstrings in functions whose purpose is pretty obvious from the name, and no explanations for the complex functionality like the methods in `Block`, and `connect_labels`.

Most of my comments are real nitpicks though, the only substantial one is that `slots=True` is only available from [py3.10 onwards](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass).

ilastik/applets/counting/countingGui.py
k-dominik: ah rebase fallout thanks for the catch!

ilastik/applets/labeling/connectBlockedLabels.py
k-dominik: yeah. I'm all for re-usability but wanted to talk to you before refactoring and ... forgot. To me something like `lazyflow.base` sounds sensible/findable.

ilastik/applets/labeling/connectBlockedLabels.py
k-dominik: guess I got neigh<i>bored</i> writing documentation ;)

tests/test_ilastik/widgets/test_LabelExplorer.py
k-dominik: that was because I picked up the wrong function from the op. While this function works, and does as expected, for "reasons" the deletion via the ui is handled differently. If the test was set up correctly - it would've hung. (I was also really surprised you could make it hang like that 😢 )

tests/test_ilastik/widgets/test_connectBlockedLabels.py
k-dominik: looks like the test does this :) enough todos from the review already, don't need my past self chiming in.

ilastik/shell/gui/shellWidgets.py
k-dominik: True... And for windows I'll look into making it look a bit more like macos - effectively hiding all the buttonness...

tests/test_ilastik/widgets/test_connectBlockedLabels.py
btbest: (for posterity: what the todo means is that the test currently only asserts on the length of returned regions, not that the regions are the correct ones)

ilastik/applets/labeling/connectBlockedLabels.py
k-dominik: hm I'll leave out the lines about the "current simplified algorithm"... because it's not true. The algorithm will handle all of the enums faithfully. Also, Both `None` and `Single` are currently in use. I just found that the overhead of using NDIM is currently not worth it.

And just for the possible NDIM is not used sentiment - as soon as we want to edit dense labels I think we will have to use that enum (and again, no change to the algorithm needed).

I'll leave a comment though in `labelExplorer.LabelExplorerWidget` where the values are chosen for the given cache.

ilastik/applets/labeling/connectBlockedLabels.py
k-dominik: while 3.9 will go EOL end of October, I always complain about people making their libraries incompatible with older versions just for the sake of it, so will specify slots explicitly. Thank you for catching!

ilastik/applets/labeling/connectBlockedLabels.py
k-dominik: Can get behind the classmethod to instance method.

ilastik/applets/labeling/connectBlockedLabels.py
k-dominik: Guess what, there was another type in the variable name :D

ilastik/applets/labeling/connectBlockedLabels.py
k-dominik: yeah that's a bit rough, but won't tackle a sensible unification here.

ilastik/applets/labeling/connectBlockedLabels.py
k-dominik: aye, that was a guard when I was still dynamically generating the boundaries and not using constants. Good catch. I guess just to be safe (for the future) and communicate intent I can make it an assert...

ilastik/applets/labeling/connectBlockedLabels.py
k-dominik: ha :D this is something that I definitely want to change to a more sensible value than the first occurrence....

tests/test_lazyflow/test_utility/test_pipeline.py
k-dominik: Well in the lazyflow sense this slot is an optional connection

tests/test_lazyflow/test_utility/test_pipeline.py
k-dominik: Op has an "Input" so it can not be used as the first operator in a pipeline. Otherwise yes

tests/test_ilastik/widgets/test_LabelExplorer.py
k-dominik: ah I guess this sucks :/ yeah. Otherwise one would get tuple(with, the, actual, types). If it is any consolidation I added type annotations to the function itself now.

I kind of prefer to have the ids close to the parameter set.

tests/test_ilastik/widgets/test_connectBlockedLabels.py
k-dominik: Hm yes that was a conscious decision as I didn't see any use. Now there could be some use in testing, but I don't want to implement a feature that is used solely in tests and has no utility in the application. I don't think it's critical at this point and would avoid adding to this already large PR.

tests/test_ilastik/widgets/test_connectBlockedLabels.py
k-dominik: bummer :), terrible variable naming. Gonna rename to `expected_n_regions` at least

k-dominik: okay that wasn't so much work - for posterity: Todo done.

ilastik/shell/gui/shellWidgets.py
k-dominik: it can be made private actually, which I really like :)

ilastik/shell/gui/shellWidgets.py
k-dominik: awwwww

ilastik/shell/gui/shellWidgets.py
k-dominik: nothing? :)

ilastik/shell/gui/shellWidgets.py
k-dominik: yeah no, less indentation is always good :)

ilastik/shell/gui/shellWidgets.py
k-dominik: well, was too happy too early - still needed to be exposed.

ilastik/shell/gui/shellWidgets.py
k-dominik: it counts the number of widgets it has. Good reminder to actually override `addWidget`  in this class, as it's not supposed to be public.

k-dominik: ah oh, I meant insertWidget and apparently already made it private.

ilastik/applets/labeling/labelExplorer.py
```diff
@@ -0,0 +1,271 @@
...
+        # Hack: Initialize new widget after lane removal
+        # need to trigger initialization manually - while the widget is on the
+        # not on the screen
+        if not self._table_initialized:
+            QTimer.singleShot(0, self.sync_state)
+        super().paintEvent(a0)
+
+    def _setupUi(self):
+        layout = QVBoxLayout()
+        self.tableWidget = QTableWidget()
```
btbest, line 129:
```suggestion
    def _setupUi(self):
        layout = QVBoxLayout()
        layout.setContentsMargins(0, 0, 0, 0)
        self.tableWidget = QTableWidget()
```
Found the source of the irritating top-end misalignment (and agree, I love the handle like this)

image

k-dominik: the things you can't unsee once you noticed. Thank you for finding the fix - it's satisfying to look at now :)

tests/test_ilastik/widgets/test_LabelExplorer.py
k-dominik: added a test for deletion via the `deleteLabel` slot, too (this is the one that hung before)

ilastik/applets/labeling/connectBlockedLabels.py
k-dominik: *typo 🙄

ilastik/applets/labeling/connectBlockedLabels.py
k-dominik: Okay I created `lazyflow.base` now and only moved a minimum there. In future I think it's good to keep in mind and liberally move things there...

ilastik/applets/labeling/connectBlockedLabels.py
k-dominik: This is a good suggestion! I loosely followed it. Origin in image analysis is almost always top left btw, nothing one needs to state imo.

tests/test_lazyflow/test_utility/test_pipeline.py
btbest: What I meant is exactly that `InputSlot(value=1.0)` isn't "optional" in the lazyflow sense like `InputSlot(optional=True)`. The true optional slot doesn't prevent setupOutputs if you set it to None or make it otherwise unready, while the input with default value does. I was just being pedantic though, it's fine for the test.

ilastik/shell/gui/shellWidgets.py
btbest: Note to self in case I need to understand this again for some reason: `n_widgets == 2` is the right condition because the function is called multiple times after each widget in the horizontal stack is added. So when there are currently two widgets, applet bar and main stack, the splitter handle to the next (third) widget is the label explorer handle.

Also this goes until `n_widgets == 4`, because `_viewerControlPlaceholder` and `_centralWidgetPlaceholder` instantiate QWidgets with `self` as parent.

ilastik/applets/labeling/connectBlockedLabels.py
btbest: I love it, there are still both US and UK spelling in this docstring 😅 spelling should just be illegal in general

ilastik/shell/gui/shellWidgets.py
k-dominik: it goes until `n_widgets == 5` (transient during init), as both placeholder get reparented few lines down (then count is 3). I'll make it more obvious and instantiate them with the intended parents right away. (count will go up to three then)

ilastik/applets/labeling/connectBlockedLabels.py
k-dominik: 🫠

k-dominik: I've changed it to American English everywhere...

ilastik/applets/labeling/connectBlockedLabels.py
btbest: Really nice docstring now 💯

tests/test_ilastik/widgets/test_connectBlockedLabels.py
```diff
@@ -0,0 +1,399 @@
...
+        {"x": BlockBoundary.STOP, "y": BlockBoundary.NONE},
+        {"x": BlockBoundary.NONE, "y": BlockBoundary.START},
+        {"x": BlockBoundary.NONE, "y": BlockBoundary.STOP},
+        {"x": BlockBoundary.START, "y": BlockBoundary.START},
+        {"x": BlockBoundary.STOP, "y": BlockBoundary.START},
+        {"x": BlockBoundary.START, "y": BlockBoundary.STOP},
+        {"x": BlockBoundary.STOP, "y": BlockBoundary.STOP},
+        {"x": BlockBoundary.NONE, "y": BlockBoundary.NONE},
+    ],
+    ids=["left", "right", "top", "bottom", "top-left", "top-right", "bottom-left", "bottom_right", "none"],
```
btbest:
```suggestion
    ids=["left", "right", "top", "bottom", "top-left", "top-right", "bottom-left", "bottom-right", "none"],
```

This also occurs two more times further down 😬 Didn't notice it in the first review round

ilastik/applets/labeling/connectBlockedLabels.py
btbest: I would've also moved the definition of `_INVERTED_BOUNDARIES` either closer to `BlockBoundary`, or alternatively closer to `connect_regions` :)

tests/test_ilastik/widgets/test_connectBlockedLabels.py
btbest: Agree

btbest (approved):
Love it, let's get it in 🚀

PR #3052, opened 2025-08-12
OME-Zarr export: Fix corner cases when matching input scales

btbest:
😭 

* input has no channels -> KeyError "c"
* input has a bunch of downscales smaller than the working scale, but export gets cropped tiny -> matched downscale shapes can have 1 and 0 along xyz


k-dominik (approved):
Looks good!

PR #3055, opened 2025-08-14
Log data selection errors

btbest:
Tiny change for better debuggability. Noticed this because it made errors while working with multiscale datasets across roles hard to debug, but it's independent of that feature/fix.


btbest:
I agree :)

PR #3058, opened 2025-08-22
Lazyflow.Slot: Fix multi-input subslots not signalling unreadiness

btbest:
Closely related to #3057 (sorry it ended up two PRs, I only discovered this one now 🤦 ), but slightly different.

The new test illustrates the issue: The previous PR was about a higher-level slot not signalling its unreadiness when disconnecting.

This PR is about the higher-level slot not knowing when one of its subslots becomes unready.

There is still one further bug relating to higher-level slots that even this PR doesn't solve, illustrated by the below test. I'm leaving that one untouched because it's not necessary to make multiscale work - but happy to suggest a fix if we want it (this requires going a bit deeper into multi-slot value logic though).
```
def test_set_None_multi_input_sets_output_unready(graph):
    op = OpMultiInput(graph=graph)
    op.MultiInput.resize(1)
    op.MultiInput[0].setValue(True)
    assert op.Output.ready()
    op.MultiInput.setValue(None)
    assert op.MultiInput[0].value is None  # Fails. setValue did nothing instead of propagating to the subslot like it should
    assert not op.Output.ready()  # Doesn't get here of course, but if you remove the above assert, this also fails
```


btbest:
> As for the question about .setValue... This should probably not be possible on a multi-level slot and raise. There is .setValues, but it is required to be of the length of subslots.

I'd be on board with removing the logic for multislots inside `setValue` (propagating to subslots) and raising on `level > 0` instead.

> Currently resizing the input will not signal unreadyness, so one can end up in a state where Outerslot.ready() == False, but _sig_unready has not been fired...

I suppose I should have written "There is still _at least_ one further bug _that I know of_"... I'd be tempted to try to fix the two things we've found here now and see if anything breaks. In principle I see a chance this might fix some of the random SlotNotReadyErrors volumina can trigger when scrolling fast or switching applets.

PR #3061, opened 2025-08-28
  switch to qtpy - qt-api "agnostic" qt-python-bindings

k-dominik:
In order to prepare enabling/switching to Pyside6 (qt6!) I propose to go via qtpy. This enables us to switch qt apis more flexibly. I already tested volumina with Pyside6. With this PR, this does not require any code changes (tested python 3.13 with pyside6). Note that the qt bindings with this pr are still PyQt5.

builds on top of #3060 

Summary:

    * change all `PyQt5` import to `qtpy`
    * change use of `qApp` to `QApplication.instance()` (doesn't exist in pyside)
    * `PyQt5.QtCore.pyqtSignal` -> `qtpy.QtCore.Signal`
    * `PyQt5.QtCore.pyqtSlot` -> `qtpy.QtCore.Slot`
    * `sip.isdeleted` -> `not qtpy.compat.isalive`

Note: Should definitely be squash and merge for revertability.


btbest (approved):
I ran this + the volumina PR in a newly created env.

I tried loading and clicking through my demo project files in each workflow, interacted with a bunch of widgets and ran a couple of exports mostly randomly. I didn't find anything unusual.

We should do quite a thorough manual verification on the next release :)

PR #3063, opened 2025-08-28
Magic multiscales across roles

btbest:
Pretty-handling for multiscales in different roles:

- When there are several multiscales in different roles: Make sure switching scales in all of them at the same time works
- When there are several multiscales in different roles: Disable scale options that don't have a match in the other roles in the drop-down box
- When adding a new multiscale, but there is already a dataset in another role: Find a scale whose shape matches the other dataset's and preset it
- When adding a new dataset, but there is already a multiscale in another role: Check if that one has a scale that matches the new dataset, and switch it if it does

Feels a bit quick-and-dirty, but it's at least in a state that's worth having a look at...


btbest:
Hmm, looks like the new scale-disabling test might be flaky on mac :/ will have a look at that or maybe skip it

ilastik/applets/dataSelection/dataSelectionGui.py
```diff
@@ -669,6 +701,29 @@ def addMultiscaleDataset(self, roleIndex, laneIndex):
...
+            # Adding a multiscale to another role: switch to the scale that matches the dataset(s) in other role(s)
+            dataset_group = self.topLevelOperator.get_lane(laneIndex).DatasetGroupOut
+            for other_role_index, role_dataset_slot in enumerate(dataset_group):
+                if not role_dataset_slot.ready() or other_role_index == roleIndex:
+                    continue
+                shape_in_other_lane = dict(
+                    zip(role_dataset_slot.value.axistags.keys(), role_dataset_slot.value.laneShape)
+                )
+                try:
+                    info.switch_to_scale_with_shape(shape_in_other_lane)
```
k-dominik:
```suggestion
                shape_in_other_role = dict(
                    zip(role_dataset_slot.value.axistags.keys(), role_dataset_slot.value.laneShape)
                )
                try:
                    info.switch_to_scale_with_shape(shape_in_other_role)
```

tests/test_ilastik/test_applets/dataSelection/testDataSelectionGui.py
```diff
@@ -91,3 +93,80 @@ def test_locked_scale_select_does_not_trigger_gui_and_informs_user(
...
+    model, enabled, disabled = mismatching_model
+    table_view = DatasetDetailedInfoTableView(None)
+    table_view.setModel(model)
+    # Make sure qtbot window is big enough to `paint` also the Scale column
+    table_view.resize(1200, 300)
+    return prepare_widget(qtbot, table_view), enabled, disabled
+
+
+def test_scale_select_disables_scale_options_not_available_in_other_roles(
+    dataset_table_with_mismatching_roles, mock_gui, intercept_info_popup
```
k-dominik: Love them or hate them - fixtures. In any case, my LSP will never figure out the types of these things unless explicitly annotated (does yours?). Can't live without it anymore ;)

tests/test_ilastik/test_applets/dataSelection/testOpDataSelection.py
```diff
@@ -897,6 +903,47 @@ def test_scale_updates_dataset_info(self, op, datasetInfo):
...
+    like OpObjectExtraction._checkConstraints or the tagged-shape check in OpSimpleStacker (multicut)."""
+
+    Role1 = InputSlot()
+    Role2 = InputSlot()
+    Role3 = InputSlot()
+    Output = OutputSlot()
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.setup_calls = 0
```
k-dominik: operators already track `._setup_count`, or does it fire too often for some reason?

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -1016,6 +1058,12 @@ def propagateDirty(self, slot, subindex, roi):
         # Output slots are directly connected to internal operators
         pass
 
+    def _ensure_transaction_in_progress(self, *_):
+        # Setting scale while the op is fully ready is forbidden.
+        # Doing this will set scale on the internal OpDataSelection for each data role one by one,
+        # causing setupOutputs throughout the workflow with different data roles providing different data shape.
+        assert not self.ScaleChangeFinished.ready(), "must disconnect ScaleChangeFinished first before changing scale"
```
k-dominik: I like the pattern, but if you want to rely on it (and also I've seen you test this behavior) it should probably not be an assert, but raise a dedicated exception?

k-dominik (commented):
Not sure if I share your concern about this feeling "quick-and-dirty". Looks quite good, and solid tests, too. And I poked it in various ways but didn't manage to make it behave in an unexpected way.

You mentioned in one of your commit messages, that OC from Predictions would not raise if there is a mismatch in the Atlas role. It would be cool if you could add dataconstraint errors to remaining workflows to make this behavior uniform (Didn't check, but multicut has many secondary inputs).

tests/test_ilastik/test_applets/dataSelection/testDataSelectionGui.py
btbest: The typing doesn't work for me either :( I tried replacing the fixtures with plain functions, but it looks like qtbot doesn't render when it's used from inside the test instead of inside fixtures. Any idea how to fix that..?

tests/test_ilastik/test_applets/dataSelection/testOpDataSelection.py
btbest: Looks like it's fine 🙃 I vaguely remembered this might exist, but putting `setup_calls` there was quicker than looking for what it's called and checking whether it does what I need. Thanks for pointing it out 🎉

ilastik/applets/dataSelection/opDataSelection.py
btbest: AssertionError seemed appropriate, since this is supposed to be a "development-time" check. `DeveloperError` 😅 ?

tests/test_ilastik/test_applets/dataSelection/testDataSelectionGui.py
k-dominik: It sort of works when annotating the arguments. so here it would be something like 
```suggestion
    dataset_table_with_mismatching_roles: Tuple[DatasetDetailedInfoTableView, List[int], List[int]], mock_gui: mock.Mock, intercept_info_popup: mock.Mock
```
gives me the correct types inside the tests at least.

ilastik/applets/dataSelection/opDataSelection.py
k-dominik: Hmmm `NonTransactionSlotChange`? :D

PR #3068, opened 2025-09-09
Fix imports from PyQt5 -> qtpy

k-dominik:
Sad sad - this one slipped through the cracks of #3061 

should have been in:
1c39916ca3cce3d111b31f8bdea37ccbe15f3a16

No clue why `conda run` stopped working for running the tests, but it was consistent (of course not reproducible locally). So I changed it to invoking pytest directly, which seems to do the trick.


btbest:
Green CI so should be fine now?

PR #3070, opened 2025-09-15
Export dialog on selected lane

### LINKED ISSUE #3064: Multiscale Export Settings / scales preview with multiple lanes
This does not require #3063. But I discovered it reviewing it, as it makes it soooo easy to add multiple images with matching scales. I went for our 2dcells.zarr example and loaded all 3 res layers in separate lanes (Object Classification [raw + pred], where raw == pred ;)).

I tried to summarize the problem in the following screenshot:

Image

##### Problem - preview scales are based on the last lane

In my case the export dialog shows me a preview of the scales - so I would expect all my export data to have such scales (`"s2"`), or at least the one I've selected in the table (first one). As it turns out, also evidenced by the "file explorer", only one of the three datasets even gets this downscale.
I also still find it weird that we would arrive at different downscales than present in the input data (as in the `"s0"` and `"s1"` case, both missing `"s2"`). Data type is the same between input and output (which I still consider not a good excuse - output, in my opinion, should always have the same <i>down</i>scales as input).

btbest:
Export dialog fixes:

* Base the export off the currently selected lane rather than the last
* Remove the pre-cropping of the dialog's "model operator" so that the export shape preview can be correct when there are smaller lanes than the one the dialog is based on. The bounds for the subregion widgets are now passed explicitly to the dialog (which was previously using the model op's cropped shape to obtain the bounds).
* Also hopefully finally got the scale matching right >.<
image

Fixes #3064

### DIFF
--- a/lazyflow/utility/io_util/write_ome_zarr.py
+++ b/lazyflow/utility/io_util/write_ome_zarr.py
@@ -58,13 +58,21 @@ SPATIAL_AXES: List[Axiskey] = ["z", "y", "x"]
 SINGE_SCALE_DEFAULT_KEY = "s0"
 
 
+def _rescale_size(size: int, factor: float) -> int:
+    """
+    Rescale a single dimension of a shape.
+    Floor-round to match behavior of OpResize, and ensure minimum size is 1.
+    """
+    return max(int(size / factor), 1)
+
+
 def match_target_scales_to_input_excluding_upscales(
     export_shape: TaggedShape, input_scales: Multiscales, input_key: str
 ) -> Multiscales:
     """We assume people don't generally want to upscale lower-resolution segmentations to raw scale."""
     # Since Multiscales is ordered largest-to-smallest, simply drop matching scales before input_key.
     all_matching_scales = _match_target_scales_to_input(export_shape, input_scales, input_key)
-    assert input_key in all_matching_scales
+    assert input_key in all_matching_scales, "generated scales don't include source scale"
     start = list(all_matching_scales.keys()).index(input_key)
     keep_scales = list(all_matching_scales.keys())[start:]
     return ODict((k, all_matching_scales[k]) for k in keep_scales)
@@ -72,7 +80,9 @@ def match_target_scales_to_input_excluding_upscales(
 
 def _match_target_scales_to_input(export_shape: TaggedShape, input_scales: Multiscales, input_key: str) -> Multiscales:
     def _eq_shape_permissive(test: TaggedShape, ref: TaggedShape) -> bool:
-        """Check if two shapes are equal, but allow reference shape to be missing axes, and ignore channel."""
+        """
+        Check if two shapes are equal. Ignore channel and allow `test` to have additional axes (but no dropped axes).
+        """
         return all(a not in ref or a == "c" or test[a] == ref[a] for a in test.keys())
 
     source_scale_shape = input_scales[input_key]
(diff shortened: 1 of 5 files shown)


lazyflow/utility/io_util/write_ome_zarr.py
```diff
@@ -84,14 +94,14 @@ def _eq_shape_permissive(test: TaggedShape, ref: TaggedShape) -> bool:
...
         for scale_key, scale_factors in input_scalings.items():
-            scaled_shape = ODict([(a, int(size / scale_factors[a])) for a, size in export_shape.items()])
+            scaled_shape = ODict([(a, _rescale_size(size, scale_factors[a])) for a, size in export_shape.items()])
             spatials = [a for a in SPATIAL_AXES if a in scaled_shape]
-            if all(scaled_shape[a] <= 1 for a in spatials):
-                break
-            target_shapes.append(scaled_shape)
-        target_scales = ODict(zip(input_scales.keys(), target_shapes))
+            if all(scaled_shape[a] <= 1 for a in spatials) and scale_key != input_key:
+                break  # Exclude scales that would be just 1px, but make sure the source scale itself is included
```
k-dominik: Why are scales that result in size 1 excluded? 0, I would understand... You allow scales of size one, but only if it's the original one. One more special case to consider/remember?

btbest: Scale shapes should be unique (debatably even "must"), so at most there should be one scale with size 1. At the same time, scales with size 1 have no value in a multiscale context. Either you also have a larger scale available, or it's pointless to work multiscale with the image. Making sure the source scale stays was meant to provide at least some continuity from source to derived dataset even in this degenerate case... on the technical side, it's a workaround to make sure the scales list doesn't end up empty. So there was also an aspect of trying to be permissive (or avoiding having to add validation logic for the dialog's OK button 🙂 )
 
Our settings dialog currently doesn't forbid users from cropping images to 1d or "0d" / size 1, even though volumina explicitly doesn't like such images (though it seems it handles them fine). This is the error you found - you also get this if you just try to load a 1d or 0d image in Input Data. Forbidding such cropping would be my preferred option, then we can be sure there is never a size-1 scale (*). Alternatively, we could say volumina should accept such datasets, record the error as a bug to be fixed, and I change this PR to always allow one size-1 scale to trickle through.

(*) In that case, should change this to exclude not just 0d but also 1d scales

btbest: I've gone round and round on this a few times, keep coming back to the same result, and at this point feel like I'm wasting way too much time on this corner case.

I've made it more strict now so that it excludes anything that isn't at least 2d, but the source scale always has to stay so that we don't end up with nothing to export. This way, the behaviour when choosing "multi-scale OME-Zarr" is consistent with behaviour of the other n-dimensional formats like h5. I.e. to allow arbitrary sizes as far as the format from lazyflow's side is concerned, and let the user run into whatever other errors volumina or ilastik produce when you choose nonsense.

I considered whether it's better to match all scales even if multiple end up having just 1px, but discarded that again - if we expect downstream to handle missing upscales, we can expect it to handle missing downscales. And this certainly avoids the nonsense of non-unique scale shapes (and scaling level metadata).

The right approach for said errors with choosing tiny/nonsense subregions is to add validation to the subregion settings (in a separate PR, or happy to add it here if this would help)...

k-dominik: Thank you for considering my comment. I agree with your reasoning. Let's do the validation of the subregion separately.

k-dominik (approved):

PR #3073, opened 2025-09-22
Export: Use shutil.rmtree for Delete All

btbest:
Another quite small fix: The "Delete All" button in the export applet didn't work for ome-zarr exports because they are folders, so `os.remove` gives `PermissionError` on Windows. `shutil.rmtree` seemed like the obvious choice because whatever we do here needs to be able to delete nested subfolders and files.

### DIFF
--- a/ilastik/applets/dataExport/dataExportGui.py
+++ b/ilastik/applets/dataExport/dataExportGui.py
@@ -1,10 +1,7 @@
-from __future__ import absolute_import
-from __future__ import division
-
 ###############################################################################
 #   ilastik: interactive learning and segmentation toolkit
 #
-#       Copyright (C) 2011-2014, the ilastik developers
+#       Copyright (C) 2011-2025, the ilastik developers
 #                                <team@ilastik.org>
 #
 # This program is free software; you can redistribute it and/or
@@ -21,10 +18,9 @@ from __future__ import division
 # on the ilastik web site at:
 # 		   http://ilastik.org/license.html
 ###############################################################################
-from builtins import range
-from past.utils import old_div
 import os
 import threading
+import shutil
 from functools import partial
 
 import qtpy.compat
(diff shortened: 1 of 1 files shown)


ilastik/applets/dataExport/dataExportGui.py
```diff
@@ -500,8 +495,8 @@ def deleteAllResults(self):
             operatorView = innerOp
             pathComp = PathComponents(operatorView.ExportPath.value, operatorView.WorkingDirectory.value)
             if os.path.exists(pathComp.externalPath):
-                os.remove(pathComp.externalPath)
-            # we need to toggle the dirts state in order to enforce a frech dirty signal
+                shutil.rmtree(pathComp.externalPath)
```
k-dominik: Hmmmmmmm I'm soooo paranoid :D. Should we do something special here? Only `shutil.rmtree` if `pathComp.externalPath` is a folder, and pathComp.externalPath.endsWith(".zarr")` or something?

btbest: Yeah we need to actually, just to handle regular files... `rmtree` can _only_ delete folders...

ilastik/applets/dataExport/dataExportGui.py
```diff
@@ -500,8 +495,11 @@ def deleteAllResults(self):
             operatorView = innerOp
             pathComp = PathComponents(operatorView.ExportPath.value, operatorView.WorkingDirectory.value)
             if os.path.exists(pathComp.externalPath):
-                os.remove(pathComp.externalPath)
-            # we need to toggle the dirts state in order to enforce a frech dirty signal
+                if os.path.isdir(pathComp.externalPath) and pathComp.externalPath.endswith(".zarr"):
```
k-dominik: My bad for not thinking about this yesterday, but n5 also results in folders.

k-dominik (approved):

PR #3077, opened 2025-10-02
Fix conda CI

btbest:
Atm it seems both miniforge and setup-miniconda ship a slightly-outdated `conda` 25.3.1 and a slightly-outdated `conda-libmamba-solver` 25.3.0 (I guess setup-miniconda has the same because it just installs miniforge...).

When the workflows update `conda` to 25.9.0, this removes a "Spinner" class that `conda-libmamba-solver` 25.3.0 depends on (cf. https://github.com/conda/conda/releases/tag/25.9.0: `Remove class conda.common.io.Spinner. Use conda.reporters.get_spinner instead. (https://github.com/conda/conda/pull/15221)`). This leads to:
```
>>> conda install -n base -c conda-forge setuptools_scm
Error while loading conda entry point: conda-libmamba-solver (cannot import name 'Spinner' from 'conda.common.io' (C:\Miniforge\Lib\site-packages\conda\common\io.py))
CondaValueError: You have chosen a non-default solver backend (libmamba) but it was not recognized. Choose one of: classi
```
(on appveyor https://ci.appveyor.com/project/ilastik/ilastik/builds/52841385 and github actions https://github.com/ilastik/ilastik/actions/runs/18186833710/job/51772778076)
reproducible locally by downloading & installing miniforge3, followed by `conda update -n base -c conda-forge conda` and then erroring on any further conda commands.

Updating `conda-libmamba-solver` to 25.4.0 along with conda fixes this.

We should be able to revert this when the versions shipped with miniforge / setup-miniconda are newer.

### DIFF
--- a/.github/workflows/core-conda-bld.yml
+++ b/.github/workflows/core-conda-bld.yml
@@ -17,13 +17,14 @@ jobs:
           fetch-depth: 0
       - uses: conda-incubator/setup-miniconda@v3
         with:
-          auto-update-conda: true
           auto-activate-base: true
           activate-environment: ""
           channel-priority: strict
           miniforge-version: latest
           conda-solver: libmamba
           conda-remove-defaults: "true"
+      - name: update conda
+        run: conda update -n base -c conda-forge conda conda-libmamba-solver
       - name: install xvfb/deps
         run: |
           sudo apt-get update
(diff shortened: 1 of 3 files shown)


k-dominik (approved):
awesome! Could make sense to open an issue with conda, as they don't seem to have a proper pin for libmamba solver then....

PR #3081, opened 2025-10-11
Add layer priority

k-dominik:
This can be used, to e.g. favor the raw data layer.

While this does not change the performance of the full render of the full screen, for the user there might still be a benefit in seeing the raw data context first.

Before (current main branch):

https://github.com/user-attachments/assets/3fc7e4e2-04af-41e9-a57c-12087d59ec68

This PR:

https://github.com/user-attachments/assets/0edcd4f7-047c-4716-a445-13344216f7a2

Todos:
* [x] check which other workflows can benefit. Apparently modifying `LayerViewerGui` and `LabelingGui` goes a long way.
* [x] Needs https://github.com/ilastik/volumina/pull/338, tests will fail until then --> `volumina>=1.3.20`

### DIFF
--- a/ilastik/applets/layerViewer/layerViewerGui.py
+++ b/ilastik/applets/layerViewer/layerViewerGui.py
@@ -1,7 +1,7 @@
 ###############################################################################
 #   ilastik: interactive learning and segmentation toolkit
 #
-#       Copyright (C) 2011-2014, the ilastik developers
+#       Copyright (C) 2011-2025, the ilastik developers
 #                                <team@ilastik.org>
 #
 # This program is free software; you can redistribute it and/or
@@ -19,7 +19,7 @@
 # 		   http://ilastik.org/license.html
 ###############################################################################
 # Python
-from builtins import range
+from enum import IntEnum
 import os
 from functools import partial
 import logging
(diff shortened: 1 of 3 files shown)


btbest:
```diff
    def createLabelLayer(self, direct=False):
      # ...
-            labellayer = ColortableLayer(labelsrc, colorTable=self._colorTable16, direct=direct)
+            labellayer = ColortableLayer(labelsrc, colorTable=self._colorTable16, direct=direct, request_priority=11)
      # ...
```
My attempt to get labels drawn before predictions, but it was a bit counterproductive. Without the prio, at least the label requests mostly remain ordered, so that labels drawn onto multiple tiles end up rendered mostly at the same time. With the prio, label bits on one tile often get rendered with long time gaps to the remaining label bits on other tiles. Overall a bit more sluggish without label prio, but the disjunct rendering with prio just feels weird.

(Also would use an IntEnum if we do add more than one fixed priority)

ilastik/applets/layerViewer/layerViewerGui.py
```diff
@@ -68,6 +68,19 @@
 # ===----------------------------------------------------------------------------------------------------------------===
 
 
+class LayerPriority(IntEnum):
+    """Predefined (by experimentation) priority for different data layer types"""
```
btbest: What do you mean by "experimentation"? Do the absolute values have any implication on performance or such?

k-dominik: :)

no - the order does. I guess I wanted to make clear exactly what you understood is not the case. But I guess...

k-dominik:
okay, _now_ it's ready for a review :)

PR #3090, opened 2025-10-28
wsdt: clamp Alpha GUI input to [0,1] to prevent setting out-of-range …

### LINKED ISSUE #2975: Multicut: Disallow values > 1 for alpha
The parameter for blending probability maps with inverted distance transform should be in the range from 0..1. Now, I've already seen people using other values, so we still have to support loading those, but not setting them anymore outside the range

ArnavBallinCode:
# PR: wsdt: clamp Alpha GUI input to [0,1] to prevent setting out-of-range blending values

Clamp the WSDT Alpha GUI input to the valid [0,1] range when writing values from the GUI back into the operator. This prevents the GUI from setting out-of-range blending parameters while preserving the ability to load legacy project files that may contain values >1.

## What
- When the GUI writes Alpha into the operator, clamp the value to `[0.0, 1.0]`.
- Keep the spinbox display range unchanged so legacy project files with `Alpha > 1` still show the original value.

## Why
- Alpha blends probability maps with the inverted distance transform and is semantically defined in `[0,1]`.
- Backwards-compatible: existing project files with `Alpha > 1` still load and display unchanged, but new GUI changes cannot set values outside `[0,1]`.

## Changes
- Modified: `ilastik/applets/wsdt/wsdtGui.py`
  - In `configure_operator_from_gui()`, read the spinbox value, clamp to `[0.0, 1.0]`, and call `op.Alpha.setValue(alpha_clamped)`.

## Testing
- Manual: set `alpha_box` to `>1` in GUI, call `configure_operator_from_gui()`, confirm operator receives `1.0`.
- Recommended follow-up: add a unit test that constructs an offscreen `QApplication`, sets `alpha_box`, calls `configure_operator_from_gui()`, and asserts that the operator slot receives a clamped value.

## Compatibility
- Loading old projects with `Alpha > 1` still shows the original value in the GUI.
- New GUI changes will always be clamped to `[0,1]`.

## Follow-ups (recommended)
- Add a defensive clamp in the operator runtime (e.g., `ilastik/applets/wsdt/opWsdt.py`) so execution always uses a safe value regardless of source.
- Add the unit test for GUI→operator clamping and a brief changelog entry.

## Fixes
- Fixes #2975

## Checklist
- [ ] Add tests. (recommended)
- [ ] Update user documentation. (not required for GUI-only change)
- [x] Reference relevant issue(s): #2975
- [x] Branch contains single commit (rebased / clean)


tests/test_ilastik/test_wsdt_alpha.py
```diff
@@ -0,0 +1,155 @@
...
+    last_alpha = 0.5
+    
+    # Check if value is outside valid range [0, 1]
+    needs_dialog = (last_alpha is None or alpha_val != last_alpha) and (alpha_val < 0.0 or alpha_val > 1.0)
+    
+    # Should not need dialog for valid value
+    assert not needs_dialog
+    
+    # Value should pass through unchanged
+    assert alpha_val == 0.8
```
k-dominik: Sorry, but this test does not test anything. It's completely standalone. What would be the point?

tests/test_ilastik/test_wsdt_alpha.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def test_alpha_validation_logic_above_range():
```
k-dominik: not a real test - does not test any code

tests/test_ilastik/test_wsdt_alpha.py
```diff
@@ -0,0 +1,155 @@
...
+    
+    # Should trigger dialog
+    assert needs_dialog
+    
+    # Simulate clamping
+    alpha_clamped = min(max(alpha_val, 0.0), 1.0)
+    assert alpha_clamped == 1.0
+
+
+def test_alpha_validation_logic_below_range():
```
k-dominik: again this does not test anything

tests/test_ilastik/test_wsdt_alpha.py
```diff
@@ -0,0 +1,155 @@
...
+    
+    # Should trigger dialog
+    assert needs_dialog
+    
+    # Simulate clamping
+    alpha_clamped = min(max(alpha_val, 0.0), 1.0)
+    assert alpha_clamped == 0.0
+
+
+def test_alpha_validation_logic_unchanged_out_of_range():
```
k-dominik: also not testing any code.

tests/test_ilastik/test_wsdt_alpha.py
```diff
@@ -0,0 +1,155 @@
...
+    last_alpha = 2.5  # Same as current value
+    
+    # Check if value changed
+    needs_dialog = (last_alpha is None or alpha_val != last_alpha) and (alpha_val < 0.0 or alpha_val > 1.0)
+    
+    # Should NOT trigger dialog since value hasn't changed
+    assert not needs_dialog
+
+
+def test_alpha_tooltip_content(qapp):
```
k-dominik: This test may look a little bit like it would be testing something, but again, it does not test any code outside the test itself.

tests/test_ilastik/test_wsdt_alpha.py
```diff
@@ -0,0 +1,155 @@
...
+        )
+        
+        tooltip = gui.alpha_box.toolTip()
+        
+        # Tooltip should mention the valid range and legacy compatibility
+        assert "0.0" in tooltip and "1.0" in tooltip
+        assert "legacy" in tooltip.lower() or "compatibility" in tooltip.lower()
+
+
+def test_alpha_clamping_min_max():
```
k-dominik: I assume the `min` and `max` are already well tested and don't need further testing here.

tests/test_ilastik/test_wsdt_alpha.py
```diff
@@ -0,0 +1,155 @@
...
+    """Test the min/max clamping function."""
+    # Test clamping values outside [0, 1]
+    assert min(max(-0.5, 0.0), 1.0) == 0.0
+    assert min(max(2.5, 0.0), 1.0) == 1.0
+    assert min(max(0.5, 0.0), 1.0) == 0.5
+    assert min(max(0.0, 0.0), 1.0) == 0.0
+    assert min(max(1.0, 0.0), 1.0) == 1.0
+
+
+def test_alpha_dialog_flow_with_mock(qapp):
```
k-dominik: this test also does not test any code - everything that is asserted here, is created inside the test. Or would you think otherwise?

tests/test_ilastik/test_wsdt_alpha.py
ArnavBallinCode: I was seeing if it works through another test file, forgot to remove it...

ilastik/applets/wsdt/wsdtGui.py
```diff
@@ -252,7 +290,33 @@ def configure_operator_from_gui(self):
...
             op.Sigma.setValue(self.sigma_box.value())
             op.MinSize.setValue(self.min_size_box.value())
-            op.Alpha.setValue(self.alpha_box.value())
+            
+            # Handle Alpha value with validation and user confirmation
+            alpha_val = float(self.alpha_box.value())
+
+            def _confirm_fn(v):
+                # Show confirmation dialog and return True to clamp, False to keep
+                msg = QMessageBox(self)
```
emilmelnikov: This could be somewhat simplified with https://doc.qt.io/qt-6/qmessagebox.html#warning (not as informative but probably OK for such a rare case).

ilastik/applets/wsdt/wsdtGui.py
```diff
@@ -252,7 +290,33 @@ def configure_operator_from_gui(self):
...
+                msg.setInformativeText("Would you like to clamp the value to the valid range?")
+
+                clamp_button = msg.addButton("Clamp to Valid Range", QMessageBox.AcceptRole)
+                keep_button = msg.addButton("Keep Current Value", QMessageBox.RejectRole)
+                msg.setDefaultButton(clamp_button)
+
+                msg.exec_()
+                return msg.clickedButton() == clamp_button
+
+            alpha_val, self._last_alpha_value = _process_alpha_change(
```
emilmelnikov: Why `last_alpha_value` is even necessary?..

```python
if not (0.0 <= alpha_val <= 1.0):
    clamp_button = QMessageBox.warning(
        self,
        "Alpha value is out of range",
        # TODO: Replace with informative text
        (
            "<p>Alpha is out of range</p>"
            "<p>Clamp alpha to [0;1]?</p>"
        ),
        QMessageBox::Ok | QMessageBox::Cancel,
        QMessageBox::Cancel,
    )
    if clamp_button == QMessageBox.Ok:
        alpha_val = max(0.0, min(alpha_val, 1.0))
```

ilastik/applets/wsdt/wsdtGui.py
ArnavBallinCode: Using `QMessageBox.warning()` will definitely simplify this. Let me refactor to use that static method instead of creating a custom dialog. Will update shortly.

ilastik/applets/wsdt/wsdtGui.py
ArnavBallinCode: The `_last_alpha_value` tracking was added to avoid showing the dialog repeatedly when other parameters change (as Dominik mentioned in the first review). However, thinking about it more - if we just check `if not (0.0 <= alpha_val <= 1.0)` on every configure call, it will show the dialog every time ANY parameter changes when alpha happens to be out of range.

Would it be acceptable to show the dialog on every parameter change if alpha is out of range? Or should we keep some tracking to only show it when alpha itself changes? 

I could simplify by removing `_last_alpha_value` if showing the dialog multiple times is okay for this rare edge case.

emilmelnikov: Right, I somehow missed the discussion above, sorry.

If you decide to show the dialog only once, I think it's better to save the result of the dialog as a boolean flag instead of the last alpha value, something like `self._enforce_alpha_clamp = clamp_button == QMessageBox.Ok` (and of course mention this in the dialog text).

tests/test_ilastik/test_wsdt_alpha.py
ArnavBallinCode: removed

tests/test_ilastik/test_wsdt_alpha.py
ArnavBallinCode: removed

tests/test_ilastik/test_wsdt_alpha.py
ArnavBallinCode: removed

ArnavBallinCode:
Closing this PR for now as I haven’t been able to find the bandwidth to follow through properly ,  thanks for the reviews and feedback.

Issue #3091, opened 2025-10-28
Tiff reader should assume same pixel size unit for all axes if only x available


btbest:
This is a semi-bug.

### Context: Tiffs commonly only contain a unit name for the x-axis

For clarity: Pixel size metadata like "5nm" are composed of a number ("resolution") and a string ("unit").

I have collected a few public example tiffs:

1) The `Helmstaedter_et_al_SUPPLinformation3c.tif` file from our [example dataset](https://data.ilastik.org/helmstaedter.zip)
2) [File S1](https://journals.plos.org/plosbiology/article/file?type=supplementary&id=10.1371/journal.pbio.3000972.s001) from [a random paper](https://doi.org/10.1371/journal.pbio.3000972) I found while looking for papers with supplemental tiffs
3) [Image 642531 of Figure S3B](https://www.ebi.ac.uk/biostudies/JCB/studies/S-JCBD-201703061) of some [paper on EBI BioStudies](https://dx.doi.org/10.1083%2Fjcb.201703061). (Need to generate the .zip download and then dig deep to unnest the tif file)
4) [Fig 1C / Image 5](https://odr.stowers.org/publications/LIBPB-2417/Figure%201/C/Image005.tif) of a [paper on Stowers ODR](https://www.stowers.org/research/publications/libpb-2417). (Warning: Server seemed to be very afraid of contact as I was trying to browse)

All of these have the issue that they only keep a unit for x, so ilastik loads them as having "pixel" for the y and z unit. It looks like we should adopt the assumption that if there is only a unit string for x, it also applies to y.

I haven't found any public tiff example with z besides the Helmstaedter dataset, but that one only keeps one "unit", which is "nm". I am guessing that this tiff was originally written by FIJI. If you open any image in FIJI, add pixel size information in the Properties (with the same units for x, y, and/or z), FIJI drops identical strings for y and z. Hence, the reader should assume that the unit applies also to z, at least when it's reading it from the `imagej_metadata`.

### Fun notes

I don't think we need to feel too terrible about not getting this to "just work", even with OME-TIFF. I tried all of the data repositories OME lists [themselves](https://docs.openmicroscopy.org/ome-model/5.6.3/ome-tiff/#public-image-repositories-allowing-image-downloads-as-ome-tiff), and:

* none of OME's own example datasets have physical size metadata
[...]

PR #3093, opened 2025-10-28
Tiff export: Fix OME metadata

### LINKED ISSUE #3092: FIJI can't read metadata from ilastik's multipage tiffs
FIJI seems to pack the entire OME-XML from tifffile into the Description tag. This is the output of the Bio-Formats importer (with "show ome xml" ticked"). Note line 3 in the output containing a wrapped full ome-xml structure with the actual metadata:

```
<OME xmlns="http://www.openmicroscopy.org/Schemas/OME/2016-06" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.openmicroscopy.org/Schemas/OME/2016-06 http://www.openmicroscopy.org/Schemas/OME/2016-06/ome.xsd">
<Image ID="Image:0" Name="helmstaedter-with-nonsense-meta_Probabilities.tiff">
<Description><?xml version="1.0" encoding="UTF-8"?><OME xmlns="http://www.openmicroscopy.org/Schemas/OME/2016-06" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.openmicroscopy.org/Schemas/OME/2016-06 http://www.openmicroscopy.org/Schemas/OME/2016-06/ome.xsd" UUID="urn:uuid:3b641b27-b42a-11f0-8b8b-a677adb69c3f" Creator="tifffile.py 2025.10.16"><Image ID="Image:0" Name="Image0"><Pixels ID="Pixels:0" DimensionOrder="XYCZT" Type="float" SizeX="256" SizeY="256" SizeC="2" SizeZ="256" SizeT="1" PhysicalSizeX="16.5000165000165" PhysicalSizeXUnit="nm" PhysicalSizeY="16.5000165000165" PhysicalSizeYUnit="cm" PhysicalSizeZ="25.0" PhysicalSizeZUnit="µm" TimeIncrement="0.0" TimeIncrementUnit="frame"><Channel ID="Channel:0:0" SamplesPerPixel="1"><LightPath/></Channel><Channel ID="Channel:0:1" SamplesPerPixel="1"><LightPath/></Channel><TiffData IFD="0" PlaneCount="512"/></Pixels></Image></OME></Description>
<Pixels BigEndian="false" DimensionOrder="XYCZT" ID="Pixels:0" Interleaved="false" PhysicalSizeX="1.0" PhysicalSizeXUnit="µm" PhysicalSizeY="1.0" PhysicalSizeYUnit="µm" SignificantBits="32" SizeC="1" SizeT="512" SizeX="256" SizeY="256" SizeZ="1" Type="float">
<Channel ID="Channel:0:0" SamplesPerPixel="1">
<LightPath/>
<MetadataOnly/>
```

This issue is for us to record what's the problem, because I remember this came up at some point in the past and I don't remember if the problem was tifffile, FIJI, or us :) If it's not us, we should probably report this.

## Reproduce
1. Data conversion workflow
2. Any image with pixel size metadata
3. Export "multipage tiff"
4. Try to load the exported file in FIJI using Bio-Formats -> no pixel size metadata
5. Compare by loading the same file in tifffile, or re-load it in another workflow in ilastik -> metadata totally fine (as it should be, this is test-covered)

btbest:
"frame" is not a valid OME-TIFF time unit. Putting this into the metadata makes other readers refuse to read our files.

This fixes #3092 at least for FIJI, which is more permissive. bioio still refuses because `""` is also not a valid ome time unit 🙃

### DIFF
--- a/lazyflow/utility/io_util/tiff_encoding.py
+++ b/lazyflow/utility/io_util/tiff_encoding.py
@@ -8,7 +8,7 @@ def to_ome(val, axis):
             return val
         if val.lower() in ome_time_units:
             return ome_time_units[val.lower()]
-        return "frame"
+        return ""
     if val in case_sensitive_lengths:
         return val
     if val.lower() in ome_length_units:
--- a/lazyflow/operators/ioOperators/opExportMultipageTiff.py
+++ b/lazyflow/operators/ioOperators/opExportMultipageTiff.py
@@ -98,20 +98,19 @@ class OpExportMultipageTiff(Operator):
         meta = self._opReorderAxes.Output.meta
         axistags = meta.axistags
         axis_units = meta.get("axis_units", {})
-        pixels = {}
+        ome_pixels = {}
 
         for axis in meta.getAxisKeys():
             if axis in size_trans:
                 size_key = size_trans[axis]
                 unit_key = unit_trans[axis]
-                pixels[size_key] = None
-                pixels[unit_key] = None
 
-                if axis != "c":
-                    pixels[size_key] = float(axistags[axis].resolution)
-                    unit_value = axis_units.get(axis, "")
-                    pixels[unit_key] = str(tiff_encoding.to_ome(unit_value, axis))
-        meta_dict["Pixels"] = pixels
+                if axistags[axis].resolution:
+                    ome_pixels[size_key] = float(axistags[axis].resolution)
+                ome_unit = tiff_encoding.to_ome(axis_units.get(axis, ""), axis)
+                if ome_unit:
+                    ome_pixels[unit_key] = ome_unit
+        meta_dict["Pixels"] = ome_pixels
 
         with tifffile.TiffWriter(self.Filepath.value, byteorder="<", ome=True) as writer:
             writer.write(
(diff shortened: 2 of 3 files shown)


lazyflow/utility/io_util/tiff_encoding.py
```diff
@@ -8,7 +8,7 @@ def to_ome(val, axis):
             return val
         if val.lower() in ome_time_units:
             return ome_time_units[val.lower()]
-        return "frame"
```
k-dominik, line 11: that one is on me :(((

btbest: 😅

btbest:
Now we omit meta entries for axes with null-metadata. Still quirks on FIJI and bioio's side: FIJI doesn't like it when `PhysicalSizeZUnit` contains "pixel" and drops both this and `PhysicalSizeZ` in that case. The most common way this would occur for users is when the unit strings were omitted for y and z by the writer, so would be mitigated if we fix #3091. Bioio is fine with having "pixel" for `PhysicalSizeZUnit`, but it dropped the property when I put "μm" in it. I guess it can't handle the `μ` character.

btbest:
Fixes #3092 as close to "properly" as reasonable now

PR #3109, opened 2025-11-24
Fix OpObjectExtraction setup

btbest:
Since it looks like I won't be implementing cross-role metadata porting in `OpObjectExtraction` anymore, I figured we might still want this fix independently.

This fixes a bug where `OpObjectExtraction` never called `setupOutputs` because `LabelImageCacheInput` is never ready. The code looks like it's supposed to be used when saving/loading Tracking projects, but even there it doesn't seem this input is serialized. If it was, it would be broken, given that `setupOutputs` contains two errors :)

Now the setup is actually called, this might end up modifying the block shape in the cache, but it looks like a sensible modification to make...

### DIFF
--- a/ilastik/applets/objectExtraction/opObjectExtraction.py
+++ b/ilastik/applets/objectExtraction/opObjectExtraction.py
@@ -338,7 +338,7 @@ class OpObjectExtractionBase(Operator, ABC):
     # For other workflows, output has rtype=ArrayLike, indexed by (t)
 
     CleanLabelBlocks = OutputSlot()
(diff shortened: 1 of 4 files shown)


k-dominik (approved):
The cache will be setup to an inefficient value (whole timeseries per default), which would for sure result in slowdowns if this output was shown.
This can happen in object feature selection, where "Object centers" layer is available to view.
Activating this layer will in the old setting result in the whole time series being processed and, hence, in a potentially long delay! (I felt it was faster to show the centers of the first frame).
Awesome that you fixed this and also actually fixed the implementation.

PR #3110, opened 2025-12-01
Unify pixel size across roles in data selection

### LINKED ISSUE #3101: OC-from-segmentation/labels loses pixel size when only raw data has it
## Describe the bug
In "Object classification from segmentation" and "Object classification from labels", when adding raw data that has pixel size metadata plus a segmentation that doesn't, then the export also doesn't have the metadata.

Users shouldn't run into this as long as they stick with ilastik: They'd most likely be exporting their segmentation as h5 or tiff, in which case both raw data and segmentation will have the pixel size. But oc-from-segmentation and -from-labels is mainly for users providing the segmentation using an external tool, which may or may not be preserving the metadata (probably not).

## Expected behavior
The export should have the metadata as long as any of the inputs have it.

## To Reproduce
1. From raw data that has pixel size meta, make a segmentation that doesn't have it. E.g. object classification (export object identities) on helmstaedter using ilastik 1.4.1post1. If using 1.4.2b4, export to a format that doesn't support pixel size yet, like npy, n5, tiff sequence, or ome-zarr.
2. Start an object classification from segmentation project
3. Add the raw data with pixel size meta
4. Add the object identities image as segmentation
5. Go through workflow
6. Export to multipage tiff or h5 -- no pixel size metadata even though it should be there
7. Check e.g. by importing in fiji (ilastik4ij for h5 or bio-formats for tiff), or for h5 should be under `h5file.attrs["axis_units"]` and `h5file.attrs["axistags"][0]["resolution"]`.

 - ilastik 1.4.2b4

btbest:
This unifies pixel size metadata across roles in data selection.

* One role has metadata, others don't -> copy
* One role has metadata, other has different meta -> no copying (also not to roles that had no meta at all) + warning

image

I provoked this in oc-from-segmentation using the Helmstaedter dataset (which has pixel size), and a segmentation whose pixel size I borked using FIJI :)

The copying happens in the backend, so copied metadata won't show up in the input data table. There the user will continue to see the metadata the file may or may not have contained. 

All layer exports in all workflows will now contain uniform pixel size - unless the user received the warning and ignored it.

I kept all the tests in `test_pixel_sizes.py` since this way the test file keeps an overview where we have pixel size features in the code base. Could of course move them to `testDataSelectionGui` or `testDataSelectionApplet`.

Fixes #3101 

## Checklist
- [X] Add tests.
- [ ] Update [user documentation](https://github.com/ilastik/ilastik.github.io)  - don't think this needs docs

### DIFF
--- a/ilastik/applets/dataSelection/opDataSelection.py
+++ b/ilastik/applets/dataSelection/opDataSelection.py
@@ -1084,7 +1084,7 @@ class OpDataSelectionGroup(Operator):
         if not roles_with_pixel_size:
             return
         eq_all_with_pixel_size = all(
-            self.eq_pixel_size_spatial(slot1, slot2)
+            self.eq_resolution_and_units_xyzt(slot1, slot2)
             for slot1, slot2 in itertools.combinations(roles_with_pixel_size, 2)
         )
         if eq_all_with_pixel_size:
@@ -1124,12 +1124,12 @@ class OpDataSelectionGroup(Operator):
                 ax = target_tag.key
                 if ax in source.meta.axistags:
                     target_tag.resolution = source.meta.axistags[ax].resolution
-            if "axis_units" in source.meta:
+            if "axis_units" in source.meta and source.meta.axis_units:
                 target.meta.axis_units = source.meta.axis_units
             target.meta[OpDataSelectionGroup.META_COPY_KEY] = source_role
 
     @staticmethod
-    def eq_pixel_size_spatial(slot1: Slot, slot2: Slot):
+    def eq_resolution_and_units_xyzt(slot1: Slot, slot2: Slot):
         eq_res = OpDataSelectionGroup._eq_resolutions(slot1, slot2)
         eq_units = OpDataSelectionGroup._eq_units(slot1, slot2)
         return eq_res and eq_units
--- a/tests/test_lazyflow/test_operators/test_ioOperators/test_pixel_sizes.py
+++ b/tests/test_lazyflow/test_operators/test_ioOperators/test_pixel_sizes.py
@@ -477,20 +477,6 @@ def dataset_other_pixel_size(inputdata_dir) -> FilesystemDatasetInfo:
 
 
 def test_DataSelection_roles_copies_from_raw_data(graph, dataset_with_pixel_size, dataset_no_pixel_size):
-    dataset_infos = [dataset_with_pixel_size, dataset_no_pixel_size, dataset_no_pixel_size]
-    op_data = setup_op_data_lane(dataset_infos, graph)
-
-    for i in range(3):
-        assert_eq_spatial_pixel_size(
-            op_data.ImageGroup[i], dataset_with_pixel_size.axistags, dataset_with_pixel_size.axis_units
-        )
-    assert OpDataSelectionGroup.META_COPY_KEY in op_data.ImageGroup[1].meta
-    assert op_data.ImageGroup[1].meta[OpDataSelectionGroup.META_COPY_KEY] == "Raw Data"
-    assert OpDataSelectionGroup.META_COPY_KEY in op_data.ImageGroup[2].meta
-    assert op_data.ImageGroup[2].meta[OpDataSelectionGroup.META_COPY_KEY] == "Raw Data"
-
-
-def test_DataSelection_roles_copies_from_raw_data_2roles(graph, dataset_with_pixel_size, dataset_no_pixel_size):
     dataset_infos = [dataset_with_pixel_size, dataset_no_pixel_size, None]
     op_data = setup_op_data_lane(dataset_infos, graph)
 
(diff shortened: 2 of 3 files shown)


tests/test_lazyflow/test_operators/test_ioOperators/test_pixel_sizes.py
```diff
@@ -423,3 +432,246 @@ def test_write_read_roundtrip_h5(
...
+        assert_eq_spatial_pixel_size(
+            op_data.ImageGroup[i], dataset_with_pixel_size.axistags, dataset_with_pixel_size.axis_units
+        )
+    assert OpDataSelectionGroup.META_COPY_KEY in op_data.ImageGroup[1].meta
+    assert op_data.ImageGroup[1].meta[OpDataSelectionGroup.META_COPY_KEY] == "Raw Data"
+    assert OpDataSelectionGroup.META_COPY_KEY in op_data.ImageGroup[2].meta
+    assert op_data.ImageGroup[2].meta[OpDataSelectionGroup.META_COPY_KEY] == "Raw Data"
+
+
+def test_DataSelection_roles_copies_from_raw_data_2roles(graph, dataset_with_pixel_size, dataset_no_pixel_size):
```
k-dominik:
```suggestion
def test_DataSelection_roles_copies_from_raw_data_2(graph, dataset_with_pixel_size, dataset_no_pixel_size):
```
probably accidental ctrl+v?

tests/test_lazyflow/test_operators/test_ioOperators/test_pixel_sizes.py
```diff
@@ -423,3 +432,246 @@ def test_write_read_roundtrip_h5(
...
+
+def test_DataSelection_roles_copies_from_raw_data(graph, dataset_with_pixel_size, dataset_no_pixel_size):
+    dataset_infos = [dataset_with_pixel_size, dataset_no_pixel_size, dataset_no_pixel_size]
+    op_data = setup_op_data_lane(dataset_infos, graph)
+
+    for i in range(3):
+        assert_eq_spatial_pixel_size(
+            op_data.ImageGroup[i], dataset_with_pixel_size.axistags, dataset_with_pixel_size.axis_units
+        )
+    assert OpDataSelectionGroup.META_COPY_KEY in op_data.ImageGroup[1].meta
```
k-dominik, line 488: at least one test could assert that the meta source role has META_COPY_KEY not set.

tests/test_lazyflow/test_operators/test_ioOperators/test_pixel_sizes.py
```diff
@@ -423,3 +432,246 @@ def test_write_read_roundtrip_h5(
...
+    axes1 = default_axes[: len(res1)]
+    shape1 = default_shape[: len(res1)]
+    axes2 = default_axes[: len(res2)]
+    shape2 = default_shape[: len(res2)]
+    op_data1 = get_data_op_with_pixel_size_meta(graph, axes1, shape1, res1, units1 or [])
+    op_data2 = get_data_op_with_pixel_size_meta(graph, axes2, shape2, res2, units2 or [])
+    if not units1:
+        del op_data1.Output.meta["axis_units"]
+    if not units2:
+        del op_data2.Output.meta["axis_units"]
```
k-dominik: I'm sure this has a reason, but I cannot see it easily. Is it because get_data_op_with_pixel_size_meta sets up the op in a way that would be different than `OpDataSelectionGroup.eq_pixel_size_spatial` expects? And if so wouldn't it be better to fix the function?

ilastik/applets/dataSelection/opDataSelection.py
```diff
@@ -1073,6 +1078,109 @@ def _ensure_transaction_in_progress(self, *_):
...
+            for target_tag in target.meta.axistags:
+                ax = target_tag.key
+                if ax in source.meta.axistags:
+                    target_tag.resolution = source.meta.axistags[ax].resolution
+            if "axis_units" in source.meta:
+                target.meta.axis_units = source.meta.axis_units
+            target.meta[OpDataSelectionGroup.META_COPY_KEY] = source_role
+
+    @staticmethod
+    def eq_pixel_size_spatial(slot1: Slot, slot2: Slot):
```
k-dominik: Maybe rename to `eq_scale_and_units_spatial` as there would be various ways of arriving at the same pixels size without sharing those.

tests/test_lazyflow/test_operators/test_ioOperators/test_pixel_sizes.py
btbest: No it's intentional, this tests a setup with only two roles having datasets 😅 maybe `test_DataSelection_roles_copies_from_raw_data_only_two_roles` is unambiguous. or `test_DataSelection_roles_copies_from_raw_data_empty_tertiary`

tests/test_lazyflow/test_operators/test_ioOperators/test_pixel_sizes.py
btbest: oh yeah definitely! they all should

tests/test_lazyflow/test_operators/test_ioOperators/test_pixel_sizes.py
k-dominik: ah right - yeah I think any of your suggestions will make it easier to understand!

tests/test_lazyflow/test_operators/test_ioOperators/test_pixel_sizes.py
btbest: I don't remember, and the test runs fine without it... It was probably related to checks like `if "axis_units" in slot.meta` at some point. I've cleaned up the last instance of that to check also the value at the key instead of just its presence.

ilastik/applets/dataSelection/opDataSelection.py
btbest: I also added the axes explicitly to the name now because "spatial" confused me there for a sec. Apparently I sometimes consider time a fourth space dimension, and sometimes I don't.

PR #3116, opened 2025-12-16
Fix crash in Pixel Classification mask display

### LINKED ISSUE #3104: Prediction Mask allows float image (but then throws errors)
## Describe the bug
In Pixel Classification, when a float image is used as the optional Prediction Mask input, volumina throws a bunch of NotImplementedErrors. The mask does actually work though, so if the Mask layer is hidden, everything is fine.

Traceback looks like it's just a wrong layer type.

## Expected behavior
The mask display should also work, or if that's really not possible, default-hide it.

## To Reproduce
Prep: Use object classification with a single object class to make a binary mask, but in the export settings convert it to float (normalize `0 ... 1` to `0.0 ... 1.0`)
Within the pixel classification workflow
1. Add the raw data from which the mask was made
2. Switch to the Prediction Mask tab and add the binary mask stored as float
3. Continue to feature selection and select any
4. Continue to Training -> errors start popping up
(5. Turn off visibility for Mask to get rid of errors)

## Traceback

```pytb
ERROR 2025-11-06 16:10:19,062 excepthooks 22684 10356 Traceback (most recent call last):
  File "C:\Program Files\ilastik-1.4.2b5\Lib\site-packages\lazyflow\request\request.py", line 383, in _execute
    self._result = self.fn()
                   ^^^^^^^^^
  File "C:\Program Files\ilastik-1.4.2b5\Lib\site-packages\volumina\tiling\tileprovider.py", line 437, in _fetch_layer_tile
    img = ims_req.wait()
          ^^^^^^^^^^^^^^
  File "C:\Program Files\ilastik-1.4.2b5\Lib\site-packages\volumina\pixelpipeline\imagesources\colortable.py", line 83, in wait
    return self.toImage()
           ^^^^^^^^^^^^^^
  File "C:\Program Files\ilastik-1.4.2b5\Lib\site-packages\volumina\pixelpipeline\imagesources\colortable.py", line 117, in toImage
    raise NotImplementedError()
NotImplementedError

```

Confirmed in 1.4.2b5 on Windows and Linux (the stack trace is pretty clearly unrelated to os anyway)

abhishekblue:
Fixes #3104 

Fixed the crash by converting the mask data from floats to integers (`uint32`) before the viewer displays it. This ensures the viewer gets the correct data type it expects.

## Checklist

- [ ] Add tests.

### DIFF
--- a/ilastik/applets/pixelClassification/pixelClassificationGui.py
+++ b/ilastik/applets/pixelClassification/pixelClassificationGui.py
@@ -654,7 +654,6 @@ class PixelClassificationGui(LabelingGui):
         # Show the mask over everything except labels
         maskSlot = self.topLevelOperatorView.PredictionMasks
         if maskSlot.ready():
-            print("!!! DEBUG: I AM RUNNING THE NEW CODE !!!")
             op_cast_mask = OpPixelOperator(parent=maskSlot.operator)
             op_cast_mask.Input.connect(maskSlot)
             op_cast_mask.Function.setValue(lambda x: x.astype(numpy.uint32))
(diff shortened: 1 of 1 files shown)


btbest:
Hi @abhishekblue , I think the approach looks promising! I'm concerned about whether plugging a dtype converter is really always a good thing to do in all situations. Dtype conversion is also something that happens in various places across the code base, it would be good if you could check how it's done elsewhere and stay in pattern.

btbest:
Closing for inactivity

PR #3117, opened 2025-12-16
fix format_exception TypeError on Python 3.10+

### LINKED ISSUE #3103: NN masks tiktorch errors
## Describe the bug
During 1.4.2b4 manual verification, Mac wasn't able to run `affable-shark` and `joyful-deer`, but `shivering-raccoon` worked. Decided to release anyway.

The models would download fine but there would be an error during initialization: `TypeError: format_exception() got an unexpected kwarg 'etype'` in `applets/neuralNetwork/modelStateControl.py` line 411 with no further stack trace.

Whatever the underlying issue was, we need to fix the function call in `modelStateControl.py` - this is broken independently of OS

See also https://docs.python.org/3.11/library/traceback.html#traceback.print_exception:

> Changed in version 3.10: The etype parameter has been renamed to exc and is now positional-only.

## Expected behavior
The underlying error should be reported, not raise an exception while trying to report it.

## Reproduce
```
from ilastik.applets.neuralNetwork.modelStateControl import ModelStateControl
from qtpy.QtWidgets import QApplication

app = QApplication(["name"])
msc = ModelStateControl()
msc._showErrorMessage(Exception("bla"))
```

## Traceback
```pytb
  File "\ilastik\utility\gui\threadRouter.py", line 88, in routed
    val = func(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^
  File "\ilastik\applets\neuralNetwork\modelStateControl.py", line 411, in _showErrorMessage
    logger.error("".join(traceback.format_exception(etype=type(exc), value=exc, tb=exc.__traceback__)))
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: format_exception() got an unexpected keyword argument 'etype'
```

## Desktop
(Any)

abhishekblue:
Fixes #3103 

Updated the error logging function to be compatible with Python 3.10 rules.


ilastik/applets/neuralNetwork/modelStateControl.py
```diff
@@ -408,7 +408,7 @@ def _check_model_compatible(model_info):
 
     @threadRouted
     def _showErrorMessage(self, exc):
-        logger.error("".join(traceback.format_exception(etype=type(exc), value=exc, tb=exc.__traceback__)))
+        logger.error("".join(traceback.format_exception(type(exc), value=exc, tb=exc.__traceback__)))
```
k-dominik: from the Python docs:

> Since Python 3.10, instead of passing value and tb, an exception object can be passed as the first argument. If value and tb are provided, the first argument is ignored in order to provide backwards compatibility.

So something like 
```suggestion
        logger.error("".join(traceback.format_exception(exc)))
```

should do the trick

k-dominik:
Hey @abhishekblue, we would really like to merge this change soon. If there are no updates until the end of this week, we'll pick this PR up ourselves :)

PR #3119, opened 2025-12-18
Fix cache deserialization for ops using OpBlockedArrayCache

k-dominik:
`.setInSlot` should not be called from outside of slot. Deserialization should always go via `Slot.__setitem__` which handles `setItem` of the operators.

I made `.setInSlot` private now so that invoking it directly is clearly discouraged.

Workflows that were affected by this (check mark indicates confirmed fix):

* [x] Object classification from Predictions
* [x] Object classification from Segmentation
* [x] All tracking workflows
  * [x] Tracking from preds
  * [x] Animal tracking from preds
  * [x] Tracking from seg
  * [x] Animal tracking from seg
  * [x] Tracking w learning from preds
  * [x] Tracking w learning from segs 
* [x] ~~Carving~~ no `OpBlockedArrayCache` serialization here.
* [x] ~~Multicut~~ this one was actually working

Some timings to show benefit ;)

| project | timinig[^1] main [s] | timinig[^1] here [s] |
| --- | --- | --- |
| large 2d oc from pred | 16 | 6 | 
| large 3d oc from seg | 31 | 7 |

[^1]: timing measured from start loading to 0 requests - make sure to save in Object Classification or object feature selection applet (binary image needs to be shown) to measure this.

## Checklist

- [ ] Add tests.

### DIFF
--- a/ilastik/applets/objectExtraction/opObjectExtraction.py
+++ b/ilastik/applets/objectExtraction/opObjectExtraction.py
@@ -520,17 +520,18 @@ class OpRegionFeatures(Operator):
 
     def setupOutputs(self):
         if self.LabelVolume.meta.axistags != self.RawVolume.meta.axistags:
-            raise Exception(
-                f"raw and label axis tags do not match ({self.RawVolume.meta.axistags} != {self.LabelVolume.meta.axistags})."
+            raise DatasetConstraintError(
+                "Object Feature Selection",
+                f"Raw and label axis tags do not match ({self.RawVolume.meta.axistags} != {self.LabelVolume.meta.axistags}).",
             )
 
         taggedOutputShape = self.LabelVolume.meta.getTaggedShape()
         taggedRawShape = self.RawVolume.meta.getTaggedShape()
 
         if not numpy.all(list(taggedOutputShape.get(k, 0) == taggedRawShape.get(k, 0) for k in "txyz")):
-            raise Exception(
-                "shapes do not match. label volume shape: {}."
-                " raw data shape: {}".format(self.LabelVolume.meta.shape, self.RawVolume.meta.shape)
+            raise DatasetConstraintError(
+                "Object Feature Selection",
+                f"Shapes do not match. Label volume shape: {self.LabelVolume.meta.shape} Raw data shape: {self.RawVolume.meta.shape}.",
             )
 
         self.Output.meta.shape = (taggedOutputShape["t"],)


lazyflow/operators/opLabelVolume.py
```diff
@@ -264,12 +263,13 @@ def propagateDirty(self, slot, subindex, roi):
             self.Output.setDirty(outroi)
             self.CachedOutput.setDirty(outroi)
 
-    def setInSlot(self, slot, subindex, roi, value):
-        #    "Invalid slot for setInSlot(): {}".format( slot.name )
+    def _setInSlot(self, slot, subindex, roi, value):
+        #    "Invalid slot for _setInSlot(): {}".format( slot.name )
         # Nothing to do here.
         # Our Input slots are directly fed into the cache,
         #  so all calls to __setitem__ are forwarded automatically
```
btbest, line 270:
```suggestion
```

seems obsolete now :)

btbest (commented):
I tried to experience the speedup but didn't manage... just for documentation it would be nice if you can describe in what situation there is a speedup 😅  but lgtm

I did find a small problem (easy fix):

oc-from-segmentation: While trying to reproduce, I was putting mitocheck (txyc) into raw data, and an apparently misordered segmentation (zxyc) as secondary data.

Problem: Instead of the "fix this dataset" dialog to adjust axis order, on this PR I just get a "here's an error" dialog, with no option. It's because OpRegionFeatures.setupOutputs raises Exception, not DatasetConstraintError

```
opObjectExtraction.py", line 531, in setupOutputs
    raise Exception(
Exception: shapes do not match. label volume shape: (1, 1344, 1024, 53, 1). raw data shape: (53, 1344, 1024, 1, 1)
```

I guess OpRegionFeatures is now higher up in the chain, or previously wasn't `ready` this early..? If this doesn't immediately make sense to you, maybe worth checking - otherwise it's a quick fix of just replacing the Exception.

lazyflow/operators/opLabelVolume.py
k-dominik: Any operator with a cache deep inside needs to implement `_setInSlot(...): pass`...

Currently, any op where `SomeSlot[slicing] = value` is called needs to implement `_setInSlot(...): pass`) in order for the forwarding to a (possibly) nested cache op can be done - which - idk, I find a bit too much noise (as well as these comments throughout the code base that you highlight). I was still debating though how to safely do so that one could not forget to implement it in the ops that actually need it (like caches)... So for this PR I'm gonna leave it in, but will open an issue.

PR #3120, opened 2025-12-22
OC from PC: Fix loading of object labels - connect to blocked cache for probabilities

### LINKED ISSUE #2469: Pixel Classification + Object Classification: Object annotations lost when saving/re-opening
## Describe the bug

Object annotations are not saved (or not loaded) in the *Pixel Classification + Object Classification* workflow.
This is specific to that workflow: there is no such problem in plain *Object Classification*.

## Expected behavior

Object labels should be saved and restored when a previously saved project is opened.

## To Reproduce

Within the Pixel Classification + Object Classification workflow:
1. Add some data in the data selection applet
2. Select some pixel features
3. Add some pixel labels, do live update
4. Do thresholding
5. Select some object features
6. Add some object labels, do live update
7. Save the project and quit
8. Start ilastik again, and load the previous project
9. Previously added object labels are gone

## Desktop
 - ilastik ilastik/ilastik@913a33d3d656ca5001c7502ff1324a182fefb3c6
 - volumina ilastik/volumina@0cc582a7778df1c8551be94d8de5bcfc1f3b330a
 - OS: macOS 11.4

k-dominik:
Previous connection to the sliced cached (for gui) probabilities introduced
a) performance issues, and b) unnecessary dirty probs (on fixClassifier) change.
Introduced a blocked cached output from pixel classification for connection to the OC part.

Furthermore, this PR introduces a `.hasCacheValue()` function on `OpValueCache` which allows to check if the cache value has been set without triggering upstream calculations. This has been used in various places - I changed those to this side-effect-free way. In PC+OC the side effect of requesting the classifier cache introduced considerable amount of delays between brushstrokes in PC (only if OC part was already "hot").

Fixes #2469 #2244

## Checklist

- [too small of a change] Add docstrings and comments.
- [won't spend the time - workflow is already marked as deprecated and will be removed] Add tests.
- [no update necessary] Update [user documentation](https://github.com/ilastik/ilastik.github.io).
- [x] Rebase commits into a logical sequence.  --> should be squash and merge

### DIFF
--- a/ilastik/applets/pixelClassification/pixelClassificationApplet.py
+++ b/ilastik/applets/pixelClassification/pixelClassificationApplet.py
@@ -37,7 +37,7 @@ class PixelClassificationApplet(StandardApplet):
             if (
                 self._topLevelOperator.classifier_cache.Output.ready()
                 and self._topLevelOperator.classifier_cache.fixAtCurrent.value is True
-                and self._topLevelOperator.classifier_cache.Output.value is None
+                and not self._topLevelOperator.classifier_cache.hasCacheValue()
             ):
                 # When the classifier is deleted (e.g. because the number of features has changed,
                 #  then notify the workflow. (Export applet should be disabled.)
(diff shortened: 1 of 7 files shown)


btbest:
I tried pc+oc with mitocheck. The pc part is fine, but when I move on to Thresholding, volumina fails to load half the tiles, gets stuck with infinite progress bar and I can't recover from this. It seems to be fine on main...

btbest (approved):
I'm not sure how it was possible before to only compute the probablities for the current viewport (at least with mitocheck), and now it has to compute probabilities for the entire dataset. Honestly, that kinda sucks.

But it sucks less than losing labeling, which... basically makes pc+oc project files mostly useless on current main.

k-dominik:
Okay this should be ready now. As nothing of the previous attempt survived I dared to force push. Will update PR description...

btbest (approved):
❤️

Issue #3129, opened 2026-01-09
Axis units should use normalised terms


btbest:
## Description

We currently allow arbitrary strings for axis units. This means that when we compare units, for example `{"x": "micron"}` will be unequal `{"x": "micrometer"}`, even though these are the same unit.

Request: All operators that read units should normalise them to a "consensus term" when the string read from the dataset conforms to a known unit.

## Scenario / Problem

Currently it's possible to load a FIJI tiff that says "micron", and then export e.g. pixel classification Probabilities in "multipage tiff" format. This will normalise the term to OME-TIFF standard, which means the Probabilities file will say "µm". When taking these into object classification, loading the original raw data file and the Probabilities file, ilastik will detect the different strings as a unit mismatch and warn about it.

## Possible implementation

The OME-TIFF normalisation in `tiff_encoding.py` is probably a good place to start.

btbest:
For GUI purposes, it would be nice to have the shortest string possible, so e.g. "µm" instead of "micrometer". I wonder if we directly want to make the ilastik-internal standard term being the shortest one so the GUI can just use it, or if it would make sense to have GUI-preferred terms being separate.

Looking at `tiff_encoding.py`, it seems the OME-TIFF standard terms are basically the shortest ones, so I'd be in favour of just normalising all known variants to the OME-TIFF terms in `OpInputDataReader`.

PR #3130, opened 2026-01-09
Integrate Precomputed and OME-Zarr into pixel size system

btbest:
Currently, the [OME-Zarr export](https://github.com/ilastik/ilastik/pull/2907) is sort-of aware of pixel size metadata, because when both the input data and the export are OME-Zarr, it ports all metadata, including resolution and units.

In the meantime, ilastik is now [able to read pixel size metadata](https://github.com/ilastik/ilastik/pull/3045) from tiff and h5, it is displayed in data selection, and also written when exporting to tiff and h5. Awkwardly though, this is entirely unaware of existing logic for OME-Zarr metadata and vice versa.

This PR integrates both multiscale formats (Precomputed and OME-Zarr) with the new pixel size system: Reading from either format sets the metadata on the operators now, and exporting to OME-Zarr now takes the metadata from the operator instead of porting it over directly from the input.

This makes it possible to seamless take pixel size metadata along while moving between tiff, h5, and OME-Zarr.

Obvious room for improvement would be axis units: On the export side, this currently just writes whatever it says on the operator. It should be fine, since the standard unit strings for OME-Zarr are [recommendations, not mandatory](https://ngff.openmicroscopy.org/0.4/#axes-md). Of course it would probably be nicer to normalise to the recommended terms there. This could exacerbate #3129 though, because the OME-Zarr terms are different from the OME-TIFF terms :)

## Checklist

- [X] Add tests.
- [X] Update [user documentation](https://github.com/ilastik/ilastik.github.io) - Nothing to update. We still don't really explain anything about pixel size metadata in the docs.

### DIFF
--- a/lazyflow/utility/io_util/OMEZarrStore.py
+++ b/lazyflow/utility/io_util/OMEZarrStore.py
@@ -192,8 +192,8 @@ def _validate_transforms(
 @dataclass(frozen=True)
 class OMEZarrTranslations:
     """
-    Specifically for metadata that ilastik does _not_ use internally.
-    It is used for porting metadata from an OME-Zarr input to export.
+    Used for porting translation metadata from an OME-Zarr input to export.
+    Consider renaming/generalising this if using slot.meta.ome_zarr_translations for other purposes.
     """
 
     multiscale_translation: Optional[OrderedDict[Literal["t", "c", "z", "y", "x"], float]]  # {axis: translation}
@@ -205,22 +205,19 @@ class OMEZarrTranslations:
     def from_multiscale_spec(cls, multiscale_spec: OME_ZARR_MULTISCALE) -> "OMEZarrTranslations":
         def get_translation(
             transforms: Union[None, ValidTransformations, InvalidTransformationError],
-            axes: Sequence[Literal["t", "c", "z", "y", "x"]],
         ) -> Union[OrderedDict[Literal["t", "c", "z", "y", "x"], float], None]:
+            axes = [tag.key for tag in _axistags_from_multiscale(multiscale_spec)]
             translation = None
             if transforms and isinstance(transforms, tuple) and transforms[1]:
                 translation = OrderedDict(zip(axes, transforms[1].values))
             return translation
 
-        axes = [tag.key for tag in _axistags_from_multiscale(multiscale_spec)]
-        multiscale_translation = get_translation(
-            _validate_transforms(multiscale_spec.get("coordinateTransformations")), axes
-        )
+        multiscale_translation = get_translation(_validate_transforms(multiscale_spec.get("coordinateTransformations")))
         dataset_translations = OrderedDict(
             [
                 (
                     scale["path"],
-                    get_translation(_validate_transforms(scale.get("coordinateTransformations", [])), axes),
+                    get_translation(_validate_transforms(scale.get("coordinateTransformations", []))),
                 )
                 for scale in multiscale_spec["datasets"]
             ]
@@ -625,7 +622,7 @@ class OMEZarrStore(MultiscaleStore):
                     "shape": zarray.shape,
                 }
                 logger.info(f"Initializing scale {scale_key} took {timer.seconds()*1000} ms.")
-        self.ome_meta_for_export = OMEZarrTranslations.from_multiscale_spec(self._multiscale_spec)
+        self.ome_zarr_translations = OMEZarrTranslations.from_multiscale_spec(self._multiscale_spec)
         super().__init__(
             uri=uri,
             dtype=dtype,
--- a/lazyflow/utility/io_util/multiscaleStore.py
+++ b/lazyflow/utility/io_util/multiscaleStore.py
@@ -32,13 +32,14 @@ from lazyflow.slot import OutputSlot
 # See MultiscaleStore docstring for details
 Multiscale = OrderedDict[str, "Scale"]
 DEFAULT_SCALE_KEY = ""
-DEFAULT_DISPLAY_RESOLUTION = 1
-DEFAULT_VIGRA_RESOLUTION = 0
+DEFAULT_DISPLAY_RESOLUTION = 1.0
+DEFAULT_VIGRA_RESOLUTION = 0.0
 
 
 def set_multiscale_meta(slot: OutputSlot, multiscale: Multiscale, active_scale_key: str):
     """Updates slot.meta with multiscale, and pixel size for active scale."""
     assert active_scale_key in multiscale, f"Tried to set slot meta for non-existent scale {active_scale_key}"
+    assert slot.meta.axistags is not None, "multiscale can not be used to update slot metadata missing axistags."
     active_scale = multiscale[active_scale_key]
     if active_scale.has_pixel_size():
         for axis, res in active_scale.resolution.items():
@@ -56,9 +57,7 @@ class Scale:
 
     def __post_init__(self):
         if self.resolution is None:
-            object.__setattr__(
-                self, "resolution", OrderedDict([(k, float(DEFAULT_VIGRA_RESOLUTION)) for k in self.shape])
-            )
+            object.__setattr__(self, "resolution", OrderedDict([(k, DEFAULT_VIGRA_RESOLUTION) for k in self.shape]))
         if self.units is None:
             object.__setattr__(self, "units", OrderedDict([(k, "") for k in self.shape]))
         if self.shape.keys() != self.resolution.keys() or self.shape.keys() != self.units.keys():
--- a/lazyflow/utility/io_util/write_ome_zarr.py
+++ b/lazyflow/utility/io_util/write_ome_zarr.py
@@ -23,7 +23,7 @@ import logging
 from collections import OrderedDict as ODict
 from functools import partial
 from pathlib import Path
-from typing import List, Tuple, Dict, OrderedDict, Optional, Iterable, Any, Union
+from typing import List, Tuple, Dict, OrderedDict, Optional, Iterable, Union
 
 import numpy
 import vigra
@@ -261,7 +261,7 @@ def _get_multiscale_transformations(
     export_offset: Optional[TaggedShape],
     export_axiskeys: List[Axiskey],
     ilastik_meta: Dict,
-    input_ome_meta: Optional[OMEZarrTranslations],
+    input_translations: Optional[OMEZarrTranslations],
     input_scale_key: Optional[str],
 ) -> Optional[List[Dict]]:
     """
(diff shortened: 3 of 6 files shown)


lazyflow/utility/io_util/multiscaleStore.py
```diff
@@ -20,14 +20,76 @@
...
-Multiscales = OrderedDict[str, OrderedDict[Literal["t", "c", "z", "y", "x"], int]]
+Multiscale = OrderedDict[str, "Scale"]
 DEFAULT_SCALE_KEY = ""
+DEFAULT_DISPLAY_RESOLUTION = 1
+DEFAULT_VIGRA_RESOLUTION = 0
+
+
+def set_multiscale_meta(slot: OutputSlot, multiscale: Multiscale, active_scale_key: str):
+    """Updates slot.meta with multiscale, and pixel size for active scale."""
+    assert active_scale_key in multiscale, f"Tried to set slot meta for non-existent scale {active_scale_key}"
```
k-dominik, line 41:
```suggestion
    assert active_scale_key in multiscale, f"Tried to set slot meta for non-existent scale {active_scale_key}"
    assert slot.meta.axistags is not None, "multiscale can not be used to update slot metadata missing axistags."
```

being paranoid, but from the signature it looks like one can put in any OutputSlot, but in fact it has to have some metadata already configured.

lazyflow/utility/io_util/multiscaleStore.py
```diff
@@ -20,14 +20,76 @@
...
 
+from lazyflow.base import Axiskey, TaggedShape
+from lazyflow.slot import OutputSlot
+
 # See MultiscaleStore docstring for details
-Multiscales = OrderedDict[str, OrderedDict[Literal["t", "c", "z", "y", "x"], int]]
+Multiscale = OrderedDict[str, "Scale"]
 DEFAULT_SCALE_KEY = ""
+DEFAULT_DISPLAY_RESOLUTION = 1
+DEFAULT_VIGRA_RESOLUTION = 0
```
k-dominik: out of curiosity - why are both ints? According to 0.4 spec the coordinate transformation data is float.

lazyflow/utility/io_util/OMEZarrStore.py
```diff
@@ -188,38 +190,44 @@ def _validate_transforms(
 
 
 @dataclass(frozen=True)
-class OMEZarrMultiscaleMeta:
+class OMEZarrTranslations:
     """
     Specifically for metadata that ilastik does _not_ use internally.
     It is used for porting metadata from an OME-Zarr input to export.
     """
```
k-dominik, line 197: The docstring is probably left over from refactoring. Translation is used in export, even combined with export offsets in case of subregions (`combine_offsets`). Or to state it differently - are units "used internally" because they are a member of the Scale class?

lazyflow/utility/io_util/write_ome_zarr.py
```diff
@@ -407,19 +379,29 @@ def _write_ome_zarr_and_ilastik_metadata(
...
     export_axiskeys = list(next(iter(export_scalings.values())).keys())
+    export_resolution_along_scaled, export_resolution_along_unscaled = _get_resolution_as_split_scaling(
+        export_axiskeys, ilastik_meta, export_scalings, input_scales
+    )
 
-    axes = _get_axes_meta(export_axiskeys, input_ome_meta)
-    datasets = _get_datasets_meta(export_scalings, export_offset, input_scales, input_scale_key, input_ome_meta)
+    axes = _get_axes_meta(export_axiskeys, ilastik_meta)
+    datasets = _get_datasets_meta(export_scalings, export_resolution_along_scaled)
     ome_zarr_multiscale_meta = {"axes": axes, "datasets": datasets, "version": "0.4"}
```
k-dominik, line 383: following the logic of the refactor from Multiscales -> Multiscale in these should probably be `ome_zarr_multiscales_meta`? And it would be a bit redundant having the `ome_zarr_` part, but I'm not proposing to remove it.

lazyflow/utility/io_util/write_ome_zarr.py
```diff
@@ -206,187 +207,158 @@ def _write_block(zarray: zarr.Array, roi, data):
...
+      of the multiscale export, so multiscale transforms seems appropriate.
+
+    :param export_resolution_along_unscaled: For the multiscale "scale" metadata (pixel size along unscaled axes like t)
+    :param export_offset: From export subregion settings, in export source slot's axis order (not export target axes)
+    :param export_axiskeys: Export target axes
+    :param ilastik_meta: To get resolution from axistags (to convert offset to translation)
+    :param input_ome_meta: To extract existing dataset and multiscale translations
+    :param input_scale_key: To extract existing dataset translations from input_ome_meta
+
+    Returns None or the transformations adjusted to export axes as OME-Zarr conforming dicts.
```
k-dominik, line 285: I guess same comment as before somewhere - shouldn't all mentions of "multiscale" here be "multiscales" as in referring to ome-zarr multiscales metadata?

lazyflow/utility/io_util/write_ome_zarr.py
```diff
@@ -206,187 +207,158 @@ def _write_block(zarray: zarr.Array, roi, data):
...
+    """
+
+    def combine_offsets(
```
k-dominik: Not sure why some functions (e.g. `_get_resolution_as_split_scaling`) that are used only once are at top level but private, and this one is moved into the function. It does use the `input_scale_key` parameter directly from the parent scope, while other ones (`export_axiskeys`, `export_offset`) are passed as is as parameters 🤷 .

k-dominik (commented):
Very cool that scale handling is now working end to end for tiff, precomputed and ome-zarr and h5 with all combinations (well - except precomputed, which we don't export to). 

The basic interface for scale and units is now via

* `OutputSlot.meta.axistags`, via `.resolution` access, and
* `OutputSlot.meta.axis_units`.

This should maybe go into the message of the merge commit.

Now having, in principle, new standard keys in metadict, I would prefer to have these (and also the old ones) documented (#3133), but that's outside the scope of this PR.

Other comments:

* You reference a commit in the latest commit message. At least I cannot find the commit 66e5361a.

* I had multiple comments relating to the commit message of your first commit here:
   > Usage is now strictly: "one multiscale" describes one image with multiple scales.      Plural "multiscales" now only appears due to usage in the OME-Zarr spec.

  I saw the term "multiscale" refer to global (as in applies to all datasets) properties (such as transformations) and dataset to refer to the scales. With the refactor the term "multiscale" now got a very specific meaning. And I get that "multiscale transformation" makes sense somehow to assume it's a transformation that applies to all datasets, but I would prefer a clearer distinction, like "global transformation", or by all means "mutliscales transformation" as it is the transformation of the multiscales dictionary in the spec...

lazyflow/utility/io_util/write_ome_zarr.py
btbest: Err... yeah honestly I could probably just inline `combine_offsets`. Leftover because before this PR the same logic was used in two places.

lazyflow/utility/io_util/multiscaleStore.py
btbest: 😇 because then my IDE doesn't complain about float eq comparisons. I'm not sure what's supposed to be the problem with eq on floats when it's known that the value is exactly 1.0 or 0.0 respectively; those have exact float representations as far as I'm aware. (Edit: Seems they smartened it up at some point, it's not actually complaining)

Happy to make them floats and ignore my IDE :)

(Also note these two constants are independent of OME-Zarr)

I did take care that the actual values of these constants aren't written into properties (except with explicit float conversion), but I agree that it's probably better just to have them actually be floats like they're supposed to...

lazyflow/utility/io_util/OMEZarrStore.py
btbest: The docstring is still accurate to its original meaning, it's just evidently a bad phrasing from the get-go 😅

> are units "used internally" because they are a member of the Scale class?

Exactly. "internally" for me means any part of ilastik that isn't IO 😄  The metadata inside OMEZarrTranslations  (or genreally `slot.meta.ome_zarr_meta`) are exclusively meant for use by the OME-Zarr export, nowhere else inside ilastik. The expectation was that we would at some point get rid of this property / class, because we use all metadata. But for the translations I'm not sure that's on the roadmap. Maybe it would make sense to rename this to "ome_zarr_translations" now, and if we do for some reason want to add ome-zarr specific features somewhere in ilastik that use the translations, that's fine.

lazyflow/utility/io_util/write_ome_zarr.py
btbest: No, because it's the meta for a single entry of the "multiscales" array: `root.attrs["multiscales"] = [ome_zarr_multiscale_meta]`

The "ome_zarr_" prefix here is supposed to drive home that this is actually the final dict according to the OME-Zarr spec 🤷

lazyflow/utility/io_util/write_ome_zarr.py
btbest: Same as above; all of these describe metadata of a single entry of the "multiscales" array.

lazyflow/utility/io_util/OMEZarrStore.py
k-dominik: Okay I guess you found a way to draw the line there. I would probably prefer to add them as first class metadata... I mean what difference does it make to say it's only used for export. We could annotate the units then with a special doc to say they are only used for display in data selection and then in export...

lazyflow/utility/io_util/write_ome_zarr.py
k-dominik: ah right, missed the wrapping in the list :)

k-dominik: although the spec:

> Each "multiscales" dictionary MUST contain the field ...

but keeping in mind that it is a list of those dicts makes sense to call it a multiscale dict. I now see the light. Also makes sense then wrt naming the outer transformation -> "multiscale trafo"

Edit: really wish the spec wouldn't call it "multiscales" dictionary :)

btbest: Not the only part where the spec's phrasing leaves room for improvement from my point of view :)

lazyflow/utility/io_util/OMEZarrStore.py
btbest: The difference is that we have plans for more features with the units, but we don't have plans for features involving translations. I changed the docstring and slot.meta property name (now `ome_zarr_translations`), so hopefully that should be clearer now

btbest: Ah right

> first class metadata

The only way I see to make this "first class" is by adding translation to the `Scale` class, and then for OME-Zarr to sum up the translations for each dataset and keep them in each Scale. The Scale objects don't currently make it into the slot metadata, so for the export to actually receive the metadata, we would then also have to change this: `slot.meta.scales = OrderedDict([(key, scale.shape) for key, scale in multiscale.items()])`. And the export has to split the individual Scales' translations into the part that they all share (multiscale translation) and their individual part (dataset translation).

Surely doable, just adds a bunch of logic and metadata carried around on slots that we then don't use... "internally" 😬

k-dominik: to me it looks like the minimum to support in transformations would be the dataset transformation to correct for sub-pixel scaling offsets. This is not really debatable?! And while ilastik might not use this info, volumina _should_.

btbest: I mean, sure, it's important metadata. I just don't see anything that needs to be done with it in the scope of this PR.

btbest: As in, I would rather first have volumina, or whatever ilastik use case we come up with, define what format it wants for this information, and then we can go back to this and feed the ome-zarr metadata through slots in the appropriate way. Same as we did with pixel size.

k-dominik: okay, since translation does not impact pixel size (scope of this PR) should be deferred.

btbest: Just to confirm though, since now I've renamed a slot meta property: As far as I'm aware, we don't actually serialize slot metadata, or do we? Could renaming it cause compatibility issues with old project files?

k-dominik: the only metadata that is written (from the meta dict) is related to blocks in the blocked caches and masks. So I think what could constitute a design flaw (as we need to infer it all the time) is at least not causing problems for the case of renaming metadata fields.

PR #3131, opened 2026-01-10
Fix: use GrayscaleLayer for float prediction masks to avoid volumina …

soumyalimje:
When a Prediction Mask was supplied as a floating-point image (values in [0.0, 1.0]), volumina's color-table visualization code raised a NotImplementedError because the ColorTableSource expects integer indices into the color table. The mask data itself was valid and used by computation, but the visualization path crashed whenever the mask layer was visible.

This PR:

Detects floating-point dtype when constructing a binary mask visualization and uses a GrayscaleLayer for float-typed masks instead of ColortableLayer.
Keeps existing ColortableLayer behavior for integer-typed masks (preserves original binary visualization for uint8 masks).
Adds a unit test that monkeypatches volumina-related symbols and verifies:
float dtype -> GrayscaleLayer created and normalization set (default (0.0,1.0))
integer dtype -> ColortableLayer created with a 256-entry colortable

### DIFF
--- a/ilastik/applets/dataSelection/opDataSelection.py
+++ b/ilastik/applets/dataSelection/opDataSelection.py
@@ -551,6 +551,8 @@ class MultiscaleUrlDatasetInfo(DatasetInfo):
         self.url = url
         op_reader = OpInputDataReader(graph=Graph(), FilePath=self.url)
         meta = op_reader.Output.meta.copy()
+        # remember whether nickname was auto-generated so we don't overwrite user edits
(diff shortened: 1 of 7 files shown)


btbest:
Hi @soumyalimje, thanks for looking into this. There is already a PR open for this issue (https://github.com/ilastik/ilastik/pull/3116). It would be good if you could look into the proposed solution there and work together to figure out which is the best way to go.

btbest:
Closing for inactivity

PR #3132, opened 2026-01-11
avoid overwriting input .h5 by defaulting feature table filename to {nickname}_table and preserving suffix on format change

soumyalimje:
Summary When the dataset nickname equals an input HDF5 filename (e.g. test_data.h5), the old default export path for feature tables was {dataset_dir}/{nickname}.csv. If the user confirmed that and later switched export format to HDF5 (or accepted defaults), the default became {dataset_dir}/{nickname}.h5 and could overwrite the original input file. This change prevents that data-loss scenario.

What I changed

Default export path now uses {dataset_dir}/{nickname}_table.csv instead of {dataset_dir}/{nickname}.csv.
[exportObjectInfoDialog.py](vscode-file://vscode-app/private/var/folders/57/xkrbhh5x4sd2nd524g0l77tc0000gn/T/AppTranslocation/7528DDE9-257A-4FA4-9360-A9DAD0F50FA1/d/Visual%20Studio%20Code-11.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
When the export format is changed in the dialog, if the current path contains the {nickname} placeholder and does not contain [_table](vscode-file://vscode-app/private/var/folders/57/xkrbhh5x4sd2nd524g0l77tc0000gn/T/AppTranslocation/7528DDE9-257A-4FA4-9360-A9DAD0F50FA1/d/Visual%20Studio%20Code-11.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), the code replaces {nickname} with {nickname}_table before swapping the extension. This ensures format switches yield {nickname}_table.h5 instead of {nickname}.h5.
Why this fixes the bug

The export filename no longer defaults to a plain {nickname} + .h5, so the export dialog will not suggest an output filename that is identical to the input .h5 file.
The change is conservative: it only adjusts placeholder-based defaults and user-chosen filenames remain untouched (so we don't silently rewrite intentionally chosen names).
How to test

Manual UI:
Open Object Classification → Export → Configure Feature Table Export.
Confirm default path is .../{nickname}_table.csv.
Switch format to HDF5; confirm path becomes .../{nickname}_table.h5.
Optionally run an export and confirm input file is not overwritten.
CI: Run unit tests / full test-suite (no behavior changes expected elsewhere).

### DIFF
--- a/ilastik/applets/layerViewer/layerViewerGui.py
+++ b/ilastik/applets/layerViewer/layerViewerGui.py
@@ -392,6 +392,24 @@ class LayerViewerGui(with_metaclass(LayerViewerGuiMetaclass, QWidget)):
         # This works perfectly for uint8.
         # For uint32, etc., values of 256,512, etc. will be appear 'off'.
         # But why would you use uint32 for a binary mask anyway?
+        # If the mask is floating point, volumina's ColortableSource cannot
+        # handle floats (it expects integer indices into the color table).
+        # In that case, use a GrayscaleLayer for visualization instead.
+        try:
+            import numpy as _np
+
+            if _np.issubdtype(slot.meta.dtype, _np.floating):
+                src = createDataSource(slot)
+                layer = GrayscaleLayer(src, window_leveling=False, priority=priority)
+                # For float masks we usually expect values in [0.0, 1.0].
+                # Use the slot's display range if available, otherwise default to (0.0, 1.0).
+                layer.set_normalize(0, (slot.meta.normalizeDisplay and slot.meta.drange) or (0.0, 1.0))
+                return layer
+        except Exception:
+            # If anything goes wrong with dtype detection, fall back to the colortable
+            # behavior below so we at least show something instead of crashing.
+            pass
+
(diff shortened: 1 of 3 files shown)


soumyalimje:
Thanks for the quick review and the useful suggestion — I implemented the requested changes and rebased the PR onto the current main. I chose GrayscaleLayer for float masks to avoid volumina’s colortable crash without altering data (no quantization). For feature tables I default to {nickname}_table and only insert the placeholder suffix to avoid overwriting input .h5.

btbest:
... it doesn't look rebased to me :)

It would be good if we can keep the matters separate, so let's stay focussed on the issue this PR addresses in this PR and leave the other PR's matters over there.

By asking for your reasoning, I wanted to know why you think going with the "_table" suffix is the best way to do this. Does adding a fixed suffix to a file name that the user chose make for an intuitive UX? Is "_table" a good suffix to use in this case?

btbest:
Closing for inactivity

PR #3135, opened 2026-01-13
Fix Object Classification Classifier dirtyness on empty lane removal

### LINKED ISSUE #3134: Object Classification (from label image): retrains after headless
## Describe the bug
Taking a properly trained object classification (from label image) project and doing headless processing will result in:
a) a valid output
b) ilastik trying to retrain the classifier after the processing

also confirmed for Object Classification (segmenation)

This can lead to some issues:
1. Takes more time for something that is not needed
2. might even fail if raw data cannot be accessed

## Expected behavior
ilastik should not retrain, there is no reason to.

## To Reproduce

1. Train an object classification project, save
2. Call it in headless

k-dominik:
Unreadyness does not imply change (in my view).

An operator with level=1 input may, or may not set its output dirty when connecting/disconnecting slots (e.g. classifier where images are added, removed but training data didn't change).
While the output may become unready in transition, the output might not change - operator should decide what should happen).

fixes #3134

## Checklist

- [x] Add tests.

### DIFF
--- a/ilastik/applets/objectClassification/opObjectClassification.py
+++ b/ilastik/applets/objectClassification/opObjectClassification.py
@@ -1,7 +1,7 @@
 ###############################################################################
 #   ilastik: interactive learning and segmentation toolkit
 #
-#       Copyright (C) 2011-2025, the ilastik developers
+#       Copyright (C) 2011-2026, the ilastik developers
 #                                <team@ilastik.org>
 #
 # This program is free software; you can redistribute it and/or
@@ -18,7 +18,7 @@
 # on the ilastik web site at:
 # 		   http://ilastik.org/license.html
 ###############################################################################
-from typing import Tuple
+from typing import Set, Tuple
 import numpy
 import numpy.typing as npt
 import time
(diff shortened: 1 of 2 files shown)


btbest (commented):
The change looks semantically correct to me - I agree that unreadiness is the wrong signal to listen to here.

I'm not sure if disconnect isn't a step too far. Wouldn't there now be a risk of inputs becoming dirty and still not clearing the cache? But yeah - the generic cache operators can't decide whether input dirtiness or undreadiness requires cache emptying, the specific subclass needs to decide this.

I tried a bunch of stuff and I feel the cache emptying still isn't quite on point (these are unchaged by this PR though): Adding a label class triggers retrain even before actually adding labels. Clicking with the eraser triggers a retrain even when nothing is erased in the process. Adding a new lane triggers a retrain (even though of course without labels this doesn't change the ML problem).

But this one does change with this PR: Currently, *removing* a lane also triggers a retrain. Arguably, it only should when the lane actually had labels, but the retrain happens regardless of that. That's fine. With this PR, removing a lane does *not* trigger a retrain, even if the lane had labels. So I can now add a lane, add labels in it, train the classifier, then remove that lane along with its labels, and then go straight to export and it still applies the old classifier trained with those now deleted labels.

I guess in a way we want that, because batch processing (and I think headless?) work by adding a temporary lane for the dataset they process. So I guess we don't get around checking whether a to-be-removed lane had labels?

PR #3140, opened 2026-01-15
generate container-internalpath nicknames for multiscale URLs and pre…

soumyalimje:
Fixes an issue where multiscale dataset URLs produced ambiguous nicknames like "s1" (the scale key) which is unhelpful in the UI and breaks when the user changes scales.
Now automatically-generated nicknames for multiscale URLs include the container base name and the internal path/scale key (e.g. 2d_cells_apoptotic_1channel-s1).
Preserve explicit user-edits: if the user sets a custom nickname, it will not be overwritten when the active scale changes.
What I changed (behavior & examples)

Before:
URL: [https://data.ilastik.org/2d_cells_apoptotic_1channel.zarr/s1](vscode-file://vscode-app/private/var/folders/57/xkrbhh5x4sd2nd524g0l77tc0000gn/T/AppTranslocation/7528DDE9-257A-4FA4-9360-A9DAD0F50FA1/d/Visual%20Studio%20Code-11.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Auto nickname: [s1](vscode-file://vscode-app/private/var/folders/57/xkrbhh5x4sd2nd524g0l77tc0000gn/T/AppTranslocation/7528DDE9-257A-4FA4-9360-A9DAD0F50FA1/d/Visual%20Studio%20Code-11.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
After:
URL: [https://data.ilastik.org/2d_cells_apoptotic_1channel.zarr/s1](vscode-file://vscode-app/private/var/folders/57/xkrbhh5x4sd2nd524g0l77tc0000gn/T/AppTranslocation/7528DDE9-257A-4FA4-9360-A9DAD0F50FA1/d/Visual%20Studio%20Code-11.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Auto nickname:2d_cells_apoptotic_1channel-s1
If the user manually edits the nickname (via dataset info), their value is preserved and will not be auto-updated on scale changes.


btbest:
Hi @soumyalimje, you might be on to something - updating the nickname with scale switch is something we want afterall. Note that as I mentioned when I first wrote that parametrized test, if you do get this to work (which would be great!), then the expected nickname is the same for all test cases. Besides this

- you'll need to rebase on main
- you'll need to investigate the test failure(s)
- the nickname generating logic shouldn't be complex enough to warrant its own module
- there is already a sensible place for the test to go (see feedback on the existing PR for the same issue)

btbest:
Closing for inactivity

Issue #3067, opened 2025-09-09
Nickname for multiscales when path is given explicitly


k-dominik:
Nickname will be the scale key if scale key is given in address - however, user might change the scale key and in general I'm not sure this is the information that should be in the nickname then.

Examples:

| URI | nickname |
| --- | --- |
| `https://data.ilastik.org/2d_cells_apoptotic_1channel.zarr` | `2d_cells_apoptotic_1channel` |
| `https://data.ilastik.org/2d_cells_apoptotic_1channel.zarr/s1` | `s1` |

As the UI allows switching dimensions, I don't think this is so helpful.

btbest:
The target design for this (todo at some point) was to do nicknames like h5/n5, where we do `{container}-{internalpath}`. In this case, "2d_cells_apoptotic_1channel-s1". This would need to update when the scale is changed, which could be a bit fiddly.

PR #3142, opened 2026-01-19
fix: Nickname From URI

TaniaRocks:
## Checklist

This PR implements a simple nickname generation for multiscale datasets. The nickname is derived from the dataset URI and includes the scale key if specified (e.g., 2d_cells_apoptotic_1channel-s1). It helps users quickly identify datasets when adding them to a project.

- [ ] Add tests.


btbest:
Ref #3067

btbest:
Unfortunately I think adding a nickname to the preview when selecting multiscales that then disappears when the multiscale is added to the project actively worsens UX. I also wouldn't call this a "fix" for the referenced issue as it doesn't address the problem described there.

PR #3144, opened 2026-01-23
Add option to disable pixel size mismatch warning

### LINKED ISSUE #3138: Config option to turn off pixel size warning (formerly: warning pops up for unrelated lanes when adding data to other lanes)
## Describe the bug
It's possible to add data in multiple roles of the same shape, but different resolution metadata. The user will be presented with a nag screen. However Adding a new line, even only with a single role will bring up that same nag screen (for the old lane, that one has acknowledged already), which becomes really annoying.

## Expected behavior
Nag ones, and leave users alone.

## To Reproduce
Within the any workflow that has multiple roles, I tried object classification for example
1. Add some data with resolution metadata to the first lane, first role
2. Add data with different resolution metadata (same shape) to the next role, nag screen appears - ok.
3. Add new data to a new lane, any role -> nag screen appears :(((((

 - ilastik version: 1.4.2b6
 - OS: all

Ashu-Rajput08:
Fixes #3138

This PR adds a small opt-out toggle for the pixel size mismatch warning in the
DataSelection applet.

The default behavior is unchanged. When enabled (default), ilastik continues
to warn about pixel size mismatches as before. When disabled, the warning
is suppressed for the current session, avoiding repeated nag dialogs after
the user has already acknowledged the mismatch.

This provides a lightweight and non-intrusive way to opt out of the warning,
and can be extended in the future to support project- or UI-level settings.

### DIFF
--- a/ilastik/applets/dataSelection/opDataSelection.py
+++ b/ilastik/applets/dataSelection/opDataSelection.py
@@ -55,6 +55,12 @@ from lazyflow.utility.helpers import get_default_axisordering, eq_shapes
 from lazyflow.utility.io_util.multiscaleStore import DEFAULT_SCALE_KEY, Multiscale
 from lazyflow.utility.pathHelpers import splitPath, globH5N5, globNpz, PathComponents, uri_to_Path
 
+# -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


btbest:
Closing for inactivity

PR #3152, opened 2026-02-19
Add Label Explorer for Object Labels (Object Classification, Tracking)

k-dominik:
Added Label Explorer for Object labels

Pixel Label Explorer seems to be very useful but only having it for some workflows (so far only the pixel classification derived ones) is surprising (at least some users have told me this) and arguably, object classification workflows need this too. Try to find the one Annotation in tracking that marked a non-dividing cell as "dividing"... Being able to cycle through all annotations makes this at least conceivable.

Implementation wise the first commit here is a big refactor starting off from the `PixelLabelExplorer` into a generic `LabelExplorer` and the implementation specific to `PixelLabelExplorer`. Some smaller changes (abstract methods, `AnnotationAnchor`) are supposed to make it a little easier to implement subclasses. In the end there's two methods to implement and `__init__`.

Second commit adds the ObjectLabel Explorer with tests. From a user perspective there should not be any surprises using it (meaning it's exactly the same as in Pixel Classification).

Demo (Tracking):

https://github.com/user-attachments/assets/c9a7a81c-ed6d-41fd-b7cd-9ed8e726a5e2

Todos:

- [x] Add tests.
- [x] Update [user documentation](https://github.com/ilastik/ilastik.github.io). -> https://github.com/ilastik/ilastik.github.io/pull/331

related: #3048 https://github.com/ilastik/ilastik.github.io/pull/331

### DIFF
--- a/ilastik/applets/objectClassification/objectClassificationGui.py
+++ b/ilastik/applets/objectClassification/objectClassificationGui.py
@@ -154,7 +154,7 @@ class ObjectClassificationGui(LabelingGui):
     def centralWidget(self):
         return self
 
-    def secondaryControlsWidget(self) -> Union[ObjectLabelExplorerWidget, None]:
+    def secondaryControlsWidget(self) -> ObjectLabelExplorerWidget:
         return self._show_label_explorer()
 
     def stopAndCleanUp(self):
--- a/ilastik/applets/objectClassification/objectLabelExplorer.py
+++ b/ilastik/applets/objectClassification/objectLabelExplorer.py
@@ -23,6 +23,7 @@ import logging
 import numpy
 import numpy.typing as npt
 
+from ilastik.applets.objectExtraction.opObjectExtraction import default_features_key
 from ilastik.widgets.labelExplorer import AnnotationAnchor, LabelExplorerBase
 from lazyflow.base import Axiskey
 from lazyflow.slot import Slot
@@ -52,7 +53,7 @@ class ObjectLabelExplorerWidget(LabelExplorerBase):
         self._label_slot = label_slot
         self._setupUi()
 
-        self.unsubscribe_fns.append(label_slot.notifyDirty(self.populate_table))
+        self.add_unsubscribe_fn(label_slot.notifyDirty(self.populate_table))
 
     def initialize_table(self):
         """
@@ -78,14 +79,14 @@ class ObjectLabelExplorerWidget(LabelExplorerBase):
                 labels_filtered[timestep] = nz[0]
 
         if not labels_filtered:
-            []
+            return []
 
         feats: dict[Timepoint, PluginDict] = self._features_slot(list(labels_filtered.keys())).wait()
 
         labs: list[AnnotationAnchor] = []
         for timepoint, non_zero_indices in labels_filtered.items():
             for n in non_zero_indices:
-                region_center = feats[timepoint]["Default features"]["RegionCenter"][n]
+                region_center = feats[timepoint][default_features_key]["RegionCenter"][n]
                 if region_center.shape == (2,):
                     z = 0.0
                     x, y = region_center
--- a/tests/test_ilastik/widgets/test_ObjectLabelEplorer.py
+++ b/tests/test_ilastik/widgets/test_ObjectLabelEplorer.py
@@ -26,7 +26,7 @@ import pytest
 import vigra
 
 from ilastik.applets.objectClassification.objectLabelExplorer import ObjectLabelExplorerWidget
-from ilastik.applets.objectExtraction.opObjectExtraction import OpAdaptTimeListRoi
+from ilastik.applets.objectExtraction.opObjectExtraction import default_features_key, OpAdaptTimeListRoi
 from lazyflow.operators.opArrayPiper import OpArrayPiper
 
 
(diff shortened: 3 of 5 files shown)


ilastik/applets/objectClassification/objectClassificationGui.py
```diff
@@ -152,6 +154,9 @@ class ObjectClassificationGui(LabelingGui):
     def centralWidget(self):
         return self
 
+    def secondaryControlsWidget(self) -> Union[ObjectLabelExplorerWidget, None]:
```
btbest: Under what circumstances does this return None? _show_label_explorer doesn't look like it has a possible None return

ilastik/applets/objectClassification/objectLabelExplorer.py
```diff
@@ -0,0 +1,103 @@
...
+
+        if not labels_filtered:
+            []
+
+        feats: dict[Timepoint, PluginDict] = self._features_slot(list(labels_filtered.keys())).wait()
+
+        labs: list[AnnotationAnchor] = []
+        for timepoint, non_zero_indices in labels_filtered.items():
+            for n in non_zero_indices:
+                region_center = feats[timepoint]["Default features"]["RegionCenter"][n]
```
btbest: This feature always exists regardless of user feature selection? (Seems that way, but that in itself surprises me)

ilastik/applets/objectClassification/objectLabelExplorer.py
```diff
@@ -0,0 +1,103 @@
...
+        label_slot: Slot,
+        axiskeys: list[Axiskey],
+        parent=None,
+    ):
+        super().__init__(axiskeys=axiskeys, parent=parent)
+        self._features_slot = features_slot
+        self._label_slot = label_slot
+        self._setupUi()
+
+        self.unsubscribe_fns.append(label_slot.notifyDirty(self.populate_table))
```
btbest: Easy-to-forget inherited property... I wonder if it's worth forcing this as an abstractproperty on the base. Would look a bit weird on the subclasses:
```
self._unsubscribe_fns = []

@property
def unsubscribe_fns(self):
    return self._unsubscribe_fns
```
Probably not worth it, how many more label explorers are we going to add beyond this one (and maybe multicut)...

tests/test_ilastik/widgets/test_ObjectLabelEplorer.py
```diff
@@ -0,0 +1,243 @@
...
+def label_op(graph) -> OpArrayPiper:
+    labels = {0: numpy.array([0.0, 0.0, 0.0])}
+    op_annotations = OpArrayPiper(graph=graph)
+    op_annotations.Input.setValue(labels)
+    return op_annotations
+
+
+@pytest.fixture
+def feature_op_2d(graph) -> OpAdaptTimeListRoi:
+    features = numpy.array([{"Default features": {"RegionCenter": numpy.array([[0.0, 0.0], [1.0, 2.0], [3.0, 4.0]])}}])
```
btbest: This makes me wonder "Is "RegionCenter" guaranteed to exist at runtime?" again; can I take the fact that it's provided as a given for the test to mean that it is guaranteed?

tests/test_ilastik/widgets/test_ObjectLabelEplorer.py
```diff
@@ -0,0 +1,243 @@
...
+        pytest.param("label_explorer_3d", id="3d"),
+    ],
+)
+class TestObjectLabelExplorer:
+    def test_construct(
+        self,
+        qtbot,
+        gui_variant,
+        request,
+    ):
```
btbest:
```suggestion
    def test_construct(self, qtbot, gui_variant, request):
```

didn't look like black to me 🙃

tests/test_ilastik/widgets/test_ObjectLabelEplorer.py
```diff
@@ -0,0 +1,243 @@
...
+            assert wrapped_populate_table.call_count == 4
+
+            # remove the last annotations
+            label_op.Input.setValue({0: numpy.array([0.0, 0.0])})
+            assert label_explorer.tableWidget.rowCount() == 0
+            assert wrapped_populate_table.call_count == 5
+
+
+@pytest.mark.parametrize(
+    "gui_variant, expected_positions",
```
btbest, line 173: Hehe, had to try if parametrize on the class can be nested with a second parametrize on a method and yes it can, and no it's not the right thing here :)

ilastik/applets/objectClassification/objectClassificationGui.py
k-dominik: 👍 copy paste from base class ;)

ilastik/applets/objectClassification/objectLabelExplorer.py
k-dominik: yes those default features are defined in https://github.com/ilastik/ilastik/blob/a3821dc71d40eff479b5ec75dc2e624e19bc329f/ilastik/applets/objectExtraction/opObjectExtraction.py#L54

I wanted to use `default_features_key` but forgot - good reminder ;)

tests/test_ilastik/widgets/test_ObjectLabelEplorer.py
k-dominik: see https://github.com/ilastik/ilastik/pull/3152#discussion_r2858905127

tests/test_ilastik/widgets/test_ObjectLabelEplorer.py
k-dominik: technically this is black! (note the trailing comma, it communicates intend!) but I agree that this is as fine in one line as in multiple :)

ilastik/applets/objectClassification/objectLabelExplorer.py
k-dominik: Good point. I won't overengineer, but ended up at something like

```
    def __init__(....):
        ...
        self._unsubscribe_fns: list[Callable[..., Any]] = []

    def add_unsubscribe_fn(self, fn: Callable[..., Any]):
        self._unsubscribe_fns.append(fn)

k-dominik: ah yeah, and multicut is for now the only missing one....

PR #3160, opened 2026-03-17
PixelLabelExplorer: Ensure anchor is an actual annotated pixel

k-dominik:
the previous implementation took region centers as anchor points, which often led to anchor points not being on an actual annotation. Usually good enough to identify the annotation, but it could still lead to ambiguity.
This commit ensures that the anchor point is an actual pixel of the connected annotated block.
The point is kind of random, but it's guaranteed to point to an annotated pixel.

Adds a `tagged_pixel_anchor` property to Regions that maps to an annotated pixel.

Before:

Screenshot 2026-03-17 at 16 06 37

With this PR:

Screenshot 2026-03-17 at 16 07 12

anchor in the center of the "crosshairs" (checkbox must be checked to see the crosshairs).

## Checklist

- [x] Add tests.
- [na] Reference relevant issues and other pull requests.

### DIFF
--- a/ilastik/applets/labeling/connectBlockedLabels.py
+++ b/ilastik/applets/labeling/connectBlockedLabels.py
@@ -22,6 +22,7 @@ from dataclasses import dataclass
 from enum import IntEnum, StrEnum
 from typing import Dict, Iterator, List, Optional, Sequence, Tuple, Union
 
+import numpy
 import vigra
 from vigra.analysis import extractRegionFeatures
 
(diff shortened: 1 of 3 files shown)


ilastik/applets/labeling/connectBlockedLabels.py
```diff
@@ -340,21 +350,27 @@ def extract_annotations(labels_data: vigra.VigraArray) -> Tuple[Region, ...]:
...
     slices: list[Tuple[slice, ...]] = []
-    for min_, max_ in zip(min_bb, max_bb):
-        slices.append(tuple(slice(mi, ma) for mi, ma in zip(min_, max_)))
-
-    # we pass the label image as "image", so minimum will be the same label
-    labels = feats["Minimum"][1:].astype("uint32")
-
-    spatial_at = [at for at in axistags if at in SPATIAL_AXES]
-    regions = tuple(Region(axistags=spatial_at, slices=sl, label=label) for sl, label in zip(slices, labels))
+    anchors: list[Tuple(float, ...)] = []
```
btbest:
```suggestion
    anchors: list[Tuple[float, ...]] = []
```

PR #3162, opened 2026-03-18
Try ome zarr for all paths

btbest:
This feels really dirty, but I feel that accommodating Precomputed or theoretical other multiscale formats is less important than getting webknossos zarr links to work.

Context: Webknossos doesn't just not have ".zarr" in their links by default; it's actively impossible to get a webknossos link with ".zarr" in it. When trying to name a dataset "mydata.zarr", it actually removes the "." and appends some internal dataset ID to the name 🙃 

Try this: https://data-humerus.webknossos.org/data/annotations/zarr/XGYPNVQ9u_QkB-Pz/em.tif (yes this ome-zarr is called .tif 👍 which makes it a joy for FIJI/MoBIE. But that's my fault for not renaming the dataset on upload)

## Checklist

- :) Add tests.
- na Reference relevant issues and other pull requests.
- [ ] Update [user documentation](https://github.com/ilastik/ilastik.github.io).  -- maybe? Need to check
- na Add screenshots / screen recordings.

### DIFF
--- a/ilastik/applets/dataSelection/multiscaleDatasetBrowser.py
+++ b/ilastik/applets/dataSelection/multiscaleDatasetBrowser.py
@@ -36,7 +36,7 @@ from qtpy.QtWidgets import (
     QTextBrowser,
     QVBoxLayout,
 )
-from requests.exceptions import SSLError, ConnectionError
+from requests.exceptions import SSLError, ConnectionError as RequestsConnectionError
 
 from lazyflow.utility import isUrl
 from lazyflow.utility.io_util.OMEZarrStore import OMEZarrStore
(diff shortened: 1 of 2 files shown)


ilastik/applets/dataSelection/multiscaleDatasetBrowser.py
```diff
@@ -81,16 +81,18 @@ def run(self):
         try:
             store = self.store_init()
             self.success.emit(store)
-        except Exception as e:
+        except (ConnectionError, RequestsConnectionError) as e:
```
k-dominik, line 84: til there's a `ConnectionError` in the stdlib

btbest: tbh I don't remember if I encountered it while implementing this back in January, or if it's just a precaution. But yeah I had also already forgotten again that's a thing 🙃

PR #3165, opened 2026-03-20
Add result export examples to pixel classification notebook

### LINKED ISSUE #2608: write out result image in notebook example
https://github.com/ilastik/ilastik/blob/f42c1bb0df75ff4ab590e9874f9f4ab29129a20f/notebooks/pixel_classification_api/pixel-classification-api.ipynb?short_path=6582d79#L53

the notebook example for pixel classification API currently ends in displaying the image. Many users will probably want to write out the result. We should add code for this.

LarytheLord:
The notebook ends after displaying the prediction with `plt.imshow`, but most users will want to write the result to disk for downstream analysis.

This adds two cells after the existing visualization:

1. **Save probability map as TIFF** — `iio.imwrite("probabilities.tiff", prediction.values)` keeps the float data and all channels intact, which is what tools like Fiji/ImageJ and napari expect.

2. **Derive and save an argmax segmentation** — takes the most likely class per pixel, casts to `uint8`, writes it out, and shows a quick `tab10` colormap check so you can eyeball the result.

Both use `imageio.v3.imwrite` which is already imported in the notebook. No new dependencies.

Closes #2608

### DIFF
--- a/notebooks/pixel_classification_api/pixel-classification-api.ipynb
+++ b/notebooks/pixel_classification_api/pixel-classification-api.ipynb
@@ -52,6 +52,51 @@
     "# show the foreground channel:\n",
     "plt.imshow(prediction[..., 0])"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "export-heading",
+   "metadata": {},
+   "source": [
+    "### Saving the results\n",
+    "\n",
+    "The prediction is an `xarray.DataArray` with shape `(y, x, c)` where `c` is the number of label classes.\n",
+    "Each channel holds the probability (0\u20131) that a given pixel belongs to that class.\n",
+    "\n",
+    "Below are a few common ways to write results to disk."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "export-probabilities",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Save the full probability map as a multi-channel TIFF.\n",
+    "# TIFF handles float data natively and is widely supported\n",
+    "# by scientific imaging tools (Fiji / ImageJ, napari, etc.).\n",
+    "iio.imwrite(\"probabilities.tiff\", prediction.values)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "export-segmentation",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# For a simple segmentation, pick the most likely class per pixel\n",
+    "# and save as a single-channel image.\n",
+    "segmentation = numpy.argmax(prediction.values, axis=-1).astype(numpy.uint8)\n",
+    "iio.imwrite(\"segmentation.tiff\", segmentation)\n",
+    "\n",
+    "# Quick visual check\n",
+    "plt.imshow(segmentation, cmap=\"tab10\")\n",
+    "plt.title(\"Segmentation (argmax of probabilities)\")\n",
+    "plt.colorbar(label=\"class\")\n",
+    "plt.show()"
+   ]
   }
  ],
  "metadata": {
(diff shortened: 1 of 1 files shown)


notebooks/pixel_classification_api/pixel-classification-api.ipynb
```diff
@@ -52,6 +52,51 @@
...
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "export-heading",
+   "metadata": {},
+   "source": [
+    "### Saving the results\n",
+    "\n",
+    "The prediction is an `xarray.DataArray` with shape `(y, x, c)` where `c` is the number of label classes.\n",
```
btbest: > The prediction is an `xarray.DataArray` with shape `(y, x, c)`

Is it? What if the input data is 3D, multi-channel and/or time-series?

notebooks/pixel_classification_api/pixel-classification-api.ipynb
```diff
@@ -52,6 +52,51 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "export-probabilities",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Save the full probability map as a multi-channel TIFF.\n",
+    "# TIFF handles float data natively and is widely supported\n",
+    "# by scientific imaging tools (Fiji / ImageJ, napari, etc.).\n",
+    "iio.imwrite(\"probabilities.tiff\", prediction.values)"
```
btbest, line 81: Have you tried opening the tiff this writes in FIJI? How does it show up there?

notebooks/pixel_classification_api/pixel-classification-api.ipynb
```diff
@@ -52,6 +52,51 @@
...
+   "cell_type": "markdown",
+   "id": "export-heading",
+   "metadata": {},
+   "source": [
+    "### Saving the results\n",
+    "\n",
+    "The prediction is an `xarray.DataArray` with shape `(y, x, c)` where `c` is the number of label classes.\n",
+    "Each channel holds the probability (0\u20131) that a given pixel belongs to that class.\n",
+    "\n",
+    "Below are a few common ways to write results to disk."
```
btbest: > a few common ways 

The cell below shows exactly one :)

notebooks/pixel_classification_api/pixel-classification-api.ipynb
```diff
@@ -52,6 +52,51 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "export-segmentation",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# For a simple segmentation, pick the most likely class per pixel\n",
```
btbest: Would be better if there was an md cell first to explain what exactly happens here and why you'd do it.

btbest (changes requested):
Teaching / docs material like this can be tricky to get right, and we need to be sure what we guide the user towards actually does work in downstream tools (especially the ones mentioned).

Please include screenshot preview of what this looks like to the end user.

LarytheLord:
Thanks for the review. Addressed the inline comments:

- Shape description no longer claims a fixed layout. It now says dimensions depend on the trained project and input, with this 2D example noted specifically.
- Fiji note softened. No longer states definitively how it opens, since I haven't tested it in Fiji myself.
- Added .npy export as a second save method.
- Added markdown cell explaining argmax segmentation before the code.
- Added a note that axis=-1 assumes channel-last for this example.

For the screenshot, I don't have the full ilastik conda environment set up locally to run the notebook end-to-end. Happy to set it up if needed, or if you're able to run it that would confirm the output looks right.

btbest:
For docs notebooks like this it is critical to verify that the code runs as intended, so you will have to set up the environment and run ilastik.

We also generally consider interop with FIJI important for ilastik users, so making sure that TIFF outputs are read correctly by FIJI out-of-the-box would also be required.

Issue #3156, opened 2026-02-26
Cleanup (maybe rename) tiktorch menu in NN workflows


k-dominik:
Current menu:

<img width="208" height="118" alt="Image" src="https://github.com/user-attachments/assets/4eb9d91f-53fc-42ed-b593-ab0de532a60c" />

Currently these don't do anything, so they should be at least hidden. What would make sense to see here is

* _Open bioimageio cache folder_
* _Delete cache folder_

PR #3172, opened 2026-03-21
Cleanup TikTorch menu in NN workflows

bitwizard25:
Related to #3156
- Remove non-functional menu items (Set hyperparameters, Create TikTorch
  configuration, Validation Set) along with their dead dialog classes
  (ParameterDlg, ValidationDlg)
- Rename menu from "TikTorch" to "Neural Network"
- Add "Open bioimageio cache folder" — opens the cache dir in the system
  file manager; informs user if it doesn't exist yet
- Add "Delete cache folder" — confirms before deleting via shutil.rmtree

## Checklist

- [ ] Add tests.
- [ ] Add screenshots / screen recordings.

### DIFF
--- a/ilastik/applets/neuralNetwork/nnClassGui.py
+++ b/ilastik/applets/neuralNetwork/nnClassGui.py
@@ -20,12 +20,13 @@
 ###############################################################################
 import logging
 import os
+import pathlib
+import shutil
 import traceback
 from collections import OrderedDict
 from functools import partial
 
 import numpy
-import yaml
 from qtpy import uic
 from qtpy.QtCore import (
     QModelIndex,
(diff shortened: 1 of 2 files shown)


btbest:
Concerns:

- Relying on an explicitly internal module of bioimageio
- platformdirs fallback - is that actually bioimageio behaviour?
- tests with local import in each tests is a bit weird
- not sure about the testing approach patching the biomiageio dir when the main purpose of such a test would be to ensure the bioimageio dir hasn't diverged from where we're expecting it

k-dominik:
sorry @bitwizard25, closing this as we needed it earlier now.

The part about adding UI/Option to delete the cache for bioimageio is still relevant - but pending updates to bioimageio.

PR #3188, opened 2026-03-31
Fix: progressSignal formula

ASaha-os:
Changed the progressSignal formula causing error while batch Processing 
Issue : #3167

The fix maps p correctly: 100 * (a + (b - a) * p / 100).

Hopefully now with 310 files, finishing file 51 will show ~16.5%, file 155 will show ~50%, and so on linearly.


ASaha-os:
#3188 . need a change in progress formula

PR #3197, opened 2026-04-15
Remove setup.cfg, move contents to pyproject, move setup.py

### LINKED ISSUE #3195: ditch setup.py/cfg in favor of pyproject.toml
related to #3194 (and #2875).

In general a good idea to abandon the mixture of setup/pyproject.toml we have. But in particular when we specify oc/tracking plugins explicitly via entrypoints it's much better to have only a single and clear place to do that.

k-dominik:
pip handles install now completely without setup.py. It is still needed in the conda recipe (hence it got moved) in order to allow for loading the dynamic version and keeping single source of truth from tags.

this is done in preparation of moving the plugin system to entrypoints (yapsy won't work with py313).

closes #3195


conda-recipe/meta.yaml
```diff
@@ -53,7 +57,7 @@ outputs:
         - psutil
         - pydantic >2.10,<3
         - pyopengl
-        - pyqt 5.15.*
+        - pyside6
```
btbest: This looks like an accident for this PR

btbest (commented):
Happy to see this cleaned up a bit

conda-recipe/meta.yaml
k-dominik: spoiler alert :D

k-dominik:
Okay I'm going to merge this now - I don't think you absolutely _have_ to create a new environment now, but definitely for #3217 (which is still a bit in the works) @btbest

btbest:
Doesn't hurt every now and then :)

PR #3199, opened 2026-04-23
Quickfix: tracking export settings in UI

k-dominik:
cleanup of #3070, trying to do any export settings in tracking would fail because of changed function signature.


btbest (approved):
I believe the jargon for this is "an oopsie"

btbest:
> trying to do any export settings

I mean *technically* only when trying to change export settings while export source is Plugin :)

PR #3200, opened 2026-04-24
Fix main menu scaling

btbest:
These small changes return at least the main menu back to something bearable on my display (applet guis remain as awful as before, tbd).

Merge into main or into your branch if you're also working on the issue. I just wanted to push this while I work on other stuff.

### DIFF
--- a/ilastik/shell/gui/startShellGui.py
+++ b/ilastik/shell/gui/startShellGui.py
@@ -54,8 +54,12 @@ def startShellGui(workflow_cmdline_args, preinit_funcs, postinit_funcs):
         QApplication.setAttribute(Qt.AA_X11InitThreads, True)
 
     if platform.system() == "Windows":
+        from PyQt5.QtGui import QGuiApplication
+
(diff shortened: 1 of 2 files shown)


k-dominik:
Cool, would you be able to do a before/after screenshot?

btbest:
Before (window lt screen):

image

After

image

k-dominik:
Ah right, so just for me to understand - I assume you are on 150% and before it was rounding up to 200%? That's definitely better then :)

btbest:
Correct.

Also I forgot to put the pointer on the commit that this can most likely be removed when we go to qt6 because it's the default setting there...

k-dominik (approved):
Cool, fractional scaling it is.

The other size changes look a little bit arbitrary, the difference is noticeable only in direct comparison - tiny bit more compact now.

btbest:
Doesn't take a direct comparison on my display to see the difference

Without number tweaks:
image

With:
image

PR #3206, opened 2026-04-30
OpSubRegion: Clamp ideal_blockshape when cropping

btbest:
Not sure this makes any difference in practice tbh

- [ ] Add tests.
- [ ] Add screenshots / screen recordings.


btbest:
Might be a minor fix worth looking into at some point, but wasn't actually related to the issue I was looking into (which turned out to not be an issue at all, most likely anyway)

PR #3207, opened 2026-05-07
Bump tiktorch to 26.5.0

k-dominik:
Would be nice to have the latest version going for the next stable. Nothing to see here I guess, I tested manually, too.


btbest:
I had this in the env while testing #3211 and didn't run into any issues in NN or TDA, so should all be good :)

btbest (approved):

PR #3211, opened 2026-05-12
Don't block user from using network if spatial axis too small

### LINKED ISSUE #3198: NN: fix spatial axis compatibility check
Remove requirement for spatial axis size to be at least as big as minimum block size. We do padding, so even patches that are too small can be safely used in inference.

Relevant code:
https://github.com/ilastik/ilastik/blob/2edff03909280350d1f8810599a2c2049428c668/ilastik/utility/bioimageio_utils.py#L351-L353

CC: @elnazfazeli (thank you for the report)

k-dominik:
For spatial axes we can (and do) pad, this shouldn't result in a failure. Right now a warning will be added to the log.
Also decreased logging of setupLayers which spammed the log before.

Example warning from log
```
WARNING 2026-05-12 11:05:07,458 logger 83610 8789453888 bioimageio_utils.py:(354): UserWarning: UserWarning("Image shape for spacial axis with expected size=512 id='y' description='' type='space' unit=None scale=1.0 concatenable=False too small (480). Image will be padded.")
```

fixes #3198

## Checklist

- [na] Add tests.

### DIFF
--- a/ilastik/utility/bioimageio_utils.py
+++ b/ilastik/utility/bioimageio_utils.py
@@ -352,5 +352,5 @@ class InputValidator:
             else:
                 if target_axis_size < min_size:
                     warnings.warn(
-                        f"Image shape for spacial axis with expected {axis} too small ({target_axis_size}). Image will be padded."
+                        f"Image shape for spatial axis with expected {axis} too small ({target_axis_size}). Image will be padded."
                     )
--- a/ilastik/utility/bioimageio_utils.py
+++ b/ilastik/utility/bioimageio_utils.py
@@ -28,6 +28,7 @@ from __future__ import annotations
 from functools import partial
 import pathlib
 import tempfile
+import warnings
 from typing_extensions import TypeGuard, assert_never
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


ilastik/utility/bioimageio_utils.py
```diff
@@ -350,4 +351,6 @@ def check_single_image_min_shape(self, tensor_id: str, target_vigra_shape: Tagge
                     )
             else:
                 if target_axis_size < min_size:
-                    raise ValueError(f"Incompatible axis {axis}: {target_axis_size} < {min_size}")
+                    warnings.warn(
+                        f"Image shape for spacial axis with expected {axis} too small ({target_axis_size}). Image will be padded."
```
btbest:
```suggestion
                        f"Image shape for spatial axis with expected {axis} too small ({target_axis_size}). Image will be padded."
```
<3

btbest (approved):
Amazing, I've tripped over this myself enough times, and have seen it confusing users as well during courses. Who would've thought it was never necessary in the first place.

Unrelated to this, but since I found it now: If you can think of a microfix for https://github.com/ilastik/tiktorch/issues/245 that would be sweet (I'm thinking it should probably just not catch errors there)

PR #3213, opened 2026-05-12
Fix high-DPI UI ilastik-side

btbest:
In ilastik the primary problems were:

* Shell, splitter and viewer enforcing excessive minimum sizes
* Applet drawers specifying 256px minimum width, which is often more than necessary and didn't scale well
* Custom-built buttons loading direct PNGs instead of using the icon manager and QIcons
* Pixel classification: Main issue were the label list making itself a bit too wide and tall (so much time wasted finding this 😭) , and the custom styled buttons with hard-coded sizes.

I returned us to a narrower minimum drawer width, got rid of the drawer-internal scroll bars this caused, and switched to 32x32 icons from the icon manager in a bunch of places.

The main shell still enforces a minimum height that's larger than previously, but not problematically so. Technically, our previous minimum height was too small anyway (it allowed the ilastik logo on the main menu to get squished).

The label explorer handle remains a bit ugly but I assume that's just my display and looks fine on others.

One of these days maybe we'll find some SVGs or higher-res PNGs for those main labelling tool buttons eh?

Here's a bunch of screenshots, always with ilastik 1.4.2b9 (pre-high-dpi) on the left. I'm always showing the minimum possible UI size (except for tracking because it's such a tall stack)

<details>
  <summary>Non-high-DPI vs main</summary>
Main menu:
image
Pixel classification labelling:
image
Carving labelling:
image
NN model selection:
image
Structured tracking:
image
Layer stack footer:
image
</details>

<details>
  <summary>Non-high-DPI vs PR</summary>
Main menu:
image

Pixel classification labelling:

image

Carving labelling:

image

NN model selection:

image

Structured tracking:

image

Layer stack footer:

image
</details>

### DIFF
--- a/ilastik/applets/tracking/structured/structuredTrackingGui.py
+++ b/ilastik/applets/tracking/structured/structuredTrackingGui.py
@@ -1,8 +1,9 @@
 from __future__ import division
 from builtins import range
 from past.utils import old_div
-from qtpy import uic, QtWidgets
+from qtpy import uic
 from qtpy.QtGui import QColor
+from qtpy.QtWidgets import QAbstractSpinBox, QLineEdit, QMessageBox, QSizePolicy
 
 import os
 import logging
@@ -84,6 +85,10 @@ class StructuredTrackingGui(TrackingBaseGui, ExportingGui):
     def _loadUiFile(self):
         localDir = os.path.split(__file__)[0]
         self._drawer = uic.loadUi(localDir + "/drawer.ui")
+
+        # SpinBoxes set min-size by their max value, which is huge because one box has max value 2147483647
+        for input_widget in self._drawer.findChildren(QAbstractSpinBox) + self._drawer.findChildren(QLineEdit):
+            input_widget.setSizePolicy(QSizePolicy.Ignored, input_widget.sizePolicy().verticalPolicy())
 
         parameters = self.topLevelOperatorView.Parameters.value
         if "maxDist" in list(parameters.keys()):
@@ -684,4 +689,4 @@ class StructuredTrackingGui(TrackingBaseGui, ExportingGui):
 
     @threadRouted
     def postInformationMessage(self, prompt):
-        QtWidgets.QMessageBox.information(self, "Info:", prompt, QtWidgets.QMessageBox.Ok)
+        QMessageBox.information(self, "Info:", prompt, QMessageBox.Ok)
--- a/ilastik/applets/dataSelection/addFileButton.py
+++ b/ilastik/applets/dataSelection/addFileButton.py
@@ -22,10 +22,7 @@ from qtpy.QtCore import Signal, QModelIndex
 from qtpy.QtWidgets import QMenu, QPushButton
 from qtpy.QtGui import QIcon
 
-# this is used to find the location of the icon file
-import os.path
-
-FILEPATH = os.path.split(__file__)[0]
+from ilastik.shell.gui.iconMgr import ilastikIcons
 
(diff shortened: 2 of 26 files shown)


k-dominik:
I can already say that it looks good on a mac, too. I'll have to get used to a few change, but overall a definite improvement. I'll also try to test it tomorrow on a really hdi screen with high scale.

btbest:
I'll also have a look on the big screen tomorrow, and also check how it looks in a linux VM. Hopefully nothing major...

Edit: Windows big screen looks fine
Edit: Ubuntu is "fine" (except for volumina issue https://github.com/ilastik/volumina/pull/359)

ilastik/applets/tracking/structured/structuredTrackingGui.py
```diff
@@ -85,6 +86,10 @@ def _loadUiFile(self):
         localDir = os.path.split(__file__)[0]
         self._drawer = uic.loadUi(localDir + "/drawer.ui")
 
+        # SpinBoxes set min-size by their max value, which is huge because one box has max value 2147483647
+        for input_widget in self._drawer.findChildren(QAbstractSpinBox) + self._drawer.findChildren(QLineEdit):
+            input_widget.setSizePolicy(QSizePolicy.Ignored, input_widget.sizePolicy().verticalPolicy())
+
```
k-dominik, line 92: This doesn't have any effect for me - spin boxes are still large - or am I looking at the wrong applet?

with those lines:

Image

without those lines:
Image

k-dominik (approved):
Good cleanup of some hard coded widget sizes in ui files :)

Labellistview default sizing is now so much nicer!



nitpick: the update of the license headers probably doesn't deserve its own commit...

k-dominik:
I gave a couple of zoom levels a try and have now side by side: PR - 1.4.2rc1 - 1.4.1.post1

200 %(ish)

Screenshot 2026-05-13 114147

300 %

Screenshot 2026-05-13 122730

350 %

Screenshot 2026-05-13 123411

The PR is by far the best and makes ilastik usable at all these magnifications. 

My only concern is now that most font sizes/buttons don't seem to grow at all anymore so it might go against the intention on someone using a high scale level. But worlds better than what was there before and who knows, maybe in qt6 things work differently right out of the box.

ilastik/applets/tracking/structured/structuredTrackingGui.py
btbest: You're in the right place - this is exactly the issue that this line is supposed to solve :/

Edit: hold on no the Solver spinbox is missing; I must've missed another tracking applet

btbest:
> My only concern is now that most font sizes/buttons don't seem to grow at all anymore

The font sizes are actually identical in all of your screenshots across all three ilastik versions (20px base, 30px title)... a bit surprising but apparently consistent? One exception: the label layers from volumina use the base font size now rather than title size.

The elements do grow around the text labels across your 200/300/350 screenshots. I haven't measured but the increase seems to match +1/2 from 200 to 300 and +1/6 from 300 to 350:

image

Probably the "not scaling anymore" impression mainly comes from the chevrons blowing up the applet headers before... now they're not growing anymore because they are font-sized.

btbest:
Force-push to squash the license changes into other commits.

Also squashed the width fix for structured tracking into the "avoid scroll bars" commit (1.4.2b9 on the left, no high-dpi):

image

The input for the minimum size wasn't a width problem I think, I just found it rather strange that it's so large compared to all the other ones...

PR #3215, opened 2026-05-15
Update teaser screenshot -> video

k-dominik:
The screenshot we show there looks awfully out of date, I think we can do better :)

When doing the recording I took care not to use any keyboard shortcuts so it's clear what's happening - that's why it is a bit longer.

In order to see the video go to the "rich diff" view, or watch the video here for convenience:

https://github.com/user-attachments/assets/563829a6-6212-453a-9c14-43c92ac806eb


btbest (approved):
Looks great :)

PR #3214, opened 2026-05-15
Update "TikTorch" -> "NN Prediction" menu

k-dominik:
Maybe the last one for `1.4.2`? 😆 

This is bugging me too much. supersedes #3172 (sorry, but stale)

* remove all nonworking items (all)
* add single info item to display which device is running, which I think is useful :)

Opted against adding cache removal just now as there is no api in bioimageio to get the cache path, only to remove and I would to at least inform users about the location. (related: https://github.com/bioimage-io/spec-bioimage-io/issues/756)

fixes #3156

image

## Checklist

- [x] Format code and imports.
- [x] Add docstrings and comments.
- [no] Add tests.
- [x] Reference relevant issues and other pull requests.
- [x] Rebase commits into a logical sequence.
- [na] Update [user documentation](https://github.com/ilastik/ilastik.github.io).
- [x] Add screenshots / screen recordings.

PR #3216, opened 2026-05-18
Add menu option to open and delete bioimageio cache folder

k-dominik:
Since @FynnBe schooled me on where to find the bioimage io settings properly, we can still add the option to delete and inspect the model cache before the next release :)

image

* _Show model cache_ opens the default file browser at the cache location, if it exists, msgbox otherwise
* _Delete model cache_ clears the cache folder, or shows messagebox if it doesn't exit

Model cache directory can be missing if no model has been downloaded yet from the zoo

xref: #3214 

## Checklist

- [no] Add docstrings and comments.
- [no] Add tests.
- [no] Update [user documentation](https://github.com/ilastik/ilastik.github.io).

### DIFF
--- a/ilastik/applets/neuralNetwork/nnClassGui.py
+++ b/ilastik/applets/neuralNetwork/nnClassGui.py
@@ -24,6 +24,8 @@ import traceback
 from collections import OrderedDict
 from functools import partial
 
+from bioimageio.spec import settings as bioimageio_settings
+from bioimageio.spec.utils import empty_cache as empty_bioimageio_cache
 import numpy
 import yaml
 from qtpy import uic
@@ -31,12 +33,13 @@ from qtpy.QtCore import (
     QModelIndex,
     QPersistentModelIndex,
     QStringListModel,
+    QUrl,
     Qt,
     QTimer,
     Signal,
     Slot,
 )
-from qtpy.QtGui import QColor, QIcon
+from qtpy.QtGui import QColor, QDesktopServices, QIcon
 from qtpy.QtWidgets import (
     QAction,
     QColorDialog,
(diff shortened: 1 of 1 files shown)


FynnBe (approved):

ilastik/applets/neuralNetwork/nnClassGui.py
```diff
@@ -378,6 +381,38 @@ def menus(self):
...
+        def _delete():
+            if not model_cache_dir.exists():
+                QMessageBox.information(
+                    self,
+                    "Model Cache not found",
+                    f"Model cache directory at {model_cache_dir} not found. Likely no models have been downloaded yet.",
+                )
+                return
+            empty_bioimageio_cache()
+            logger.info(f"Deleted bioimageio model cache: {model_cache_dir}")
```
btbest: The cherry on top would be to add e.g. number of files deleted or space freed up in the log, but the simple path output is also sufficient.

btbest (approved):
Having tried it, it did feel a bit sudden for it to just go ahead and delete everything without asking for confirmation on an action that can't be undone. On the other hand, the model cache isn't particularly sensitive, so I'm ok with it.

I ended up checking the pattern across package managers anyway. conda has the nicest imo:

```
>>> conda clean
Will delete these files from /path/to/cache:
(...)
Total 2837 files (6.4 GB)
Delete? [Y / n]
>>> y
```

Some require an "intention flag" (`sudo apt-get clean`, `npm cache clean --force`). `pip cache purge` just deletes without asking for confirmation, but it does give you feedback on how much it deleted at least (unlike apt and npm).

btbest:
beautiful ✨

k-dominik:
Thank you @btbest and @FynnBe for taking the time to have a look and useful input!

PR #3217, opened 2026-05-19
Switch to plugins via entrypoints (away from yapsy)

### LINKED ISSUE #3194: Move away from yapsy
yapsy (our plugin system relies on that) is not maintained anymore, and does not play with Python 3.13 anymore.

Python itself can handle plugins (at least as far as we need them) via entrypoints. Switching to this setup will allow us a more concise plugin discovery. For the spherical textures I already added support for entrypoints, but this would make it the only way to add plugins. Breaking change: simply putting two files to add a plugin in the plugin folder will no longer work.

### LINKED ISSUE #3221: Tracking export "H5-Event-Sequence" is broken?
1. Changing the output file path in the dialog and then clicking OK -> reopen dialog again -> path reset to default (similar to https://github.com/ilastik/ilastik/issues/1864)

2. More importantly, the plugin seems to do nothing? When I click Export, it can run into the "file already exists" path if the regular export path happens to align with an existing file, but even if it doesn't, it doesn't seem to output any files

k-dominik:
This one seems dear to me - I have 5 branches with names suggesting to implement this reaching back to 2023 ;_). In any case, now that `yapsy` is no longer compatible with python versions >3.12 it's time to push through this.

With this PR plugins are expected to be python packages that define specific entrypoints (example pyproject.toml section):
```toml
# ilastik internal:
[project.entry-points."ilastik.objectfeature_plugins.default"]
[project.entry-points."ilastik.tracking_export_plugins.default"]

# external plugins:
[project.entry-points."ilastik.objectfeature_plugins"]
[project.entry-points."ilastik.tracking_export_plugins"]
```
Big advantage is that plugins can now properly declare dependencies and the solver can take those into account (particularly external plugins).

The new plugin manager class is very slim and can be used with types - huge win imo.

## Checklist

- [x] Finish hytra refactor #3220 
- [x] Publish new sphericaltexture plugin version https://github.com/KoehlerLab/ilastik-sphericaltexture/pull/3
- [x] rebase
- [x] make sure yapsy not in env anymore
- [x] Add tests.
- [x] remove the REMOVE... commit that introduces `ilastik-forge/label/noyapsy` for testing
- [x] move spherical_textures plugin to ilastik-forge main

fixes #3194
fixes #3221


ilastik/plugins/types.py
```diff
@@ -176,17 +237,17 @@ def _getFeatureNameTranslation(cls, category, name):
         :param name: The feature name string
         :returns: the long name of the feature
         """
-        from ilastik.plugins.manager import pluginManager
+        from ilastik.plugins.manager import plugin_manager
 
         all_props = None
```
btbest:
```suggestion
```

(immediately overwritten below)

docs/ilastik/plugin_system.rst
```diff
@@ -4,54 +4,47 @@
...
-    pluginManager.getAllPlugins()
+From code:
+----------
 
-Or only get plugins of a particular category, such as object feature
-plugins::
+Simply import the ilastik plugin manager::
 
-    pluginManager.getPluginsOfCategory('ObjectFeatures')
+    from ilastik.plugins import plugin_manager
```
btbest, line 27:
```
>>> from ilastik.plugins import plugin_manager
ImportError: cannot import name 'plugin_manager' from 'ilastik.plugins' (C:\Users\root\Code\ilastik-group\k-dominik-ilastik\ilastik\plugins\__init__.py)
```

(Needs an add to `__init__`)

ilastik/plugins/manager.py
```diff
@@ -18,46 +18,71 @@
...
-if sys.version_info < (3, 10):
-    from importlib_metadata import entry_points
-else:
-    from importlib.metadata import entry_points
+from .types import ObjectFeaturesPlugin, TrackingExportFormatPlugin
 
-from yapsy.PluginManager import PluginManager
 
-from ilastik.config import cfg
+class PluginNotFound(BaseException):
```
btbest:
```suggestion
class PluginNotFound(Exception):
```

[Inheriting from BaseException is a no-no](https://docs.python.org/3/library/exceptions.html#BaseException)

(more seriously: deriving from BaseException means `except Exception` doesn't catch `PluginNotFound`, which I doubt is intentional)

ilastik/applets/tracking/base/pluginExportOptionsDlg.py
```diff
@@ -120,7 +120,7 @@ def getAvailablePlugins(cls):
         Returns the list of available plugins
         """
         try:
-            exportPlugins = pluginManager.getPluginsOfCategory("TrackingExportFormats")
+            exportPlugins = plugin_manager.get_tracking_plugins()
             availableExportPlugins = [pluginInfo.name for pluginInfo in exportPlugins]
```
btbest:
```suggestion
            exportPlugins = plugin_manager.get_tracking_export_plugins()
            availableExportPlugins = [plugin.plugin_info.name for plugin in exportPlugins]
```

This triggers an error when trying to choose the Plugin export source in a tracking workflow.

ilastik/plugins/types.py
```diff
@@ -176,17 +237,17 @@ def _getFeatureNameTranslation(cls, category, name):
...
+            plugin = plugin_manager.get_object_feature_plugin_by_name("Standard Object Features")
         else:
-            plugin = pluginManager.getPluginByName(category, "ObjectFeatures")
-        if plugin:
-            plugin_feature_names = {name: {}}
-            all_props = plugin.plugin_object.fill_properties(plugin_feature_names)  # fill in display name and such
+            plugin = plugin_manager.get_object_feature_plugin_by_name(category)
+
+        plugin_feature_names = {name: {}}
+        all_props = plugin.fill_properties(plugin_feature_names)  # fill in display name and such
```
btbest, line 256: Minor: There's an edge case where in `opConservationTracking`, `category` can be "Cell Division Features", for which we don't actually use a real plugin.

This seems to be unreachable from the gui, because the checkbox that could send you here is only exposed in debug mode. But it probably wouldn't be bad to have a fallback here catching `PluginNotFound` and returning the unmodified `name` in that case. (Maybe this is what the `if plugin:` safeguard before this refactor might've been for?)

(Repro: In `OpConservationTracking._create_plugin_export_context`, set `with_merger_resolution = False`, then try to export the CSV-Table plugin)

```pytb
ERROR 2026-06-08 12:07:17,827 excepthooks 27304 11460 Unhandled exception in thread: 'DataExportThread'
ERROR 2026-06-08 12:07:17,863 excepthooks 27304 11460 Traceback (most recent call last):
  File "C:\Users\root\Code\ilastik-group\k-dominik-ilastik\ilastik\excepthooks.py", line 134, in run
    run_old(*args, **kwargs)
  File "C:\Users\root\.conda\envs\domilastik\Lib\threading.py", line 982, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Users\root\Code\ilastik-group\k-dominik-ilastik\ilastik\applets\dataExport\dataExportGui.py", line 416, in exportSync
    self.parentApplet.post_process_lane_export(lane_index, checkOverwriteFiles=False)
  File "C:\Users\root\Code\ilastik-group\k-dominik-ilastik\ilastik\applets\tracking\base\trackingBaseDataExportApplet.py", line 264, in post_process_lane_export
    return self._export_with_plugin(lane_index, checkOverwriteFiles, op.SelectedPlugin.value)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\root\Code\ilastik-group\k-dominik-ilastik\ilastik\applets\tracking\base\trackingBaseDataExportApplet.py", line 242, in _export_with_plugin
    status = self._pluginExportFunc(lane_index, filename, plugin, checkOverwriteFiles, argsSlot)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\root\Code\ilastik-group\k-dominik-ilastik\ilastik\workflows\tracking\conservation\conservationTrackingWorkflow.py", line 379, in _pluginExportFunc
    return self.trackingApplet.topLevelOperator.getLane(lane_index).exportPlugin(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\root\Code\ilastik-group\k-dominik-ilastik\ilastik\applets\tracking\conservation\opConservationTracking.py", line 702, in exportPlugin
    return plugin.export(filename, hypothesesGraph, plugin_export_context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\root\Code\ilastik-group\k-dominik-ilastik\ilastik\plugins_default\tracking_csv_export.py", line 99, in export
    featureName = self._getFeatureNameTranslation(category, feature).replace(" ", "_")
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\root\Code\ilastik-group\k-dominik-ilastik\ilastik\plugins\types.py", line 247, in _getFeatureNameTranslation
    plugin = plugin_manager.get_object_feature_plugin_by_name(category)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\root\Code\ilastik-group\k-dominik-ilastik\ilastik\plugins\manager.py", line 43, in get_object_feature_plugin_by_name
    raise PluginNotFound(
ilastik.plugins.manager.PluginNotFound: Could not find object feature plugin by name='Cell Division Features'. Available plugins: TestFeatures,Standard Object Features,2D Convex Hull Features,3D Convex Hull Features,2D Skeleton Features,Spherical Texture
```

btbest (commented):
Glad to see this finally coming together after such a long time and so many attempts :)

I love the additional typing; these feature dicts have always been super confusing.

Aside from the fix in `pluginExportOptionsDlg.py‎`, I couldn't get anything else to break, so good to go 🚀 the rest are minor notes.

Also not commented below because the file has no changes: `opConservationTracking.py` still has an `import hytra` that seems unneeded, should that also go?

ilastik/plugins/types.py
k-dominik: it's not that minor. This code path _can_ be hit from normal gui when going to subset features in the division classifier... so that needs some proper addressing.

k-dominik: okay I restored the previous pattern of allowing to miss certain plugins at runtime - i don't feel very comfortable with it, but it worked before...

k-dominik:
gonna move packages from noyapsy label to main now...

k-dominik:
thx @btbest for looking so thoroughly. The tracking (and object feature dialog) thing would've haunted us :) Will merge once CI is green

Note: New development environments need to be created after merging/pulling this.

PR #3218, opened 2026-05-21
Remaining 1.4.2 fixes

btbest:
Live Update button doesn't toggle to Pause icon when active:
image

@k-dominik What's your preference for tag/release after we merge this fix? Sneak `1.4.2` on the new head commit or `1.4.2.post1`?

Draft in case I find more stuff that should really be fixed for this release

## Checklist

- - Format code and imports.
- - Add docstrings and comments.
- - Add tests.
- - Rebase commits into a logical sequence.
- - Update [user documentation](https://github.com/ilastik/ilastik.github.io).
- - Add screenshots / screen recordings.

### DIFF
--- a/ilastik/applets/labeling/labelingGui.py
+++ b/ilastik/applets/labeling/labelingGui.py
@@ -252,7 +252,10 @@ class LabelingGui(LayerViewerGui):
         stylesheet = stylesheet_dark if is_qt_dark_mode() else stylesheet_light
         _labelControlUi.setStyleSheet(stylesheet)
         if hasattr(self.labelingDrawerUi, "liveUpdateButton"):
-            self.labelingDrawerUi.liveUpdateButton.setIcon(QIcon(ilastikIcons.Play))
+            icon = QIcon()
+            icon.addFile(ilastikIcons.Play, state=QIcon.Off)
+            icon.addFile(ilastikIcons.Pause, state=QIcon.On)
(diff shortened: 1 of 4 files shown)


k-dominik:
Bummer! But great that you found it. I'd be for moving the tag...

btbest:
~~Still checking a last issue that should be specific to my VM though. Putting this as ready since I think it should be a no-op.~~ (Yeah it was nothing) Mac still tbd as well but I'd be surprised to discover anything at this point. @k-dominik let me know any comments you might have before I merge and move the 1.4.2 tag!

btbest:
Minor: the Suggest Features button wasn't supposed to take up all of the horizontal space around it; looks a bit meh this way now. Will try if it's a super quick fix tomorrow and put it in if so.

btbest:
Alright here's using a QPushButton instead, width fixed width (left, 1.4.2 current; right, PR):

TDA labelling:
image

Pixel classification Suggest Features:
image

Kubuntu:

image


Mac:
🤦 Turns out Mac is the only platform where QPushButton actually looks really different from QToolButton. Idk, I don't feel particularly strongly about the styling change; just felt it was weird for Suggest Features to be such a huge button previously.

image

btbest:
I've reverted the styling changes and instead removed the side paddings to avoid the scroll bar at minimum width. Will let CI run now and then merge and move 1.4.2.

PR #3220, opened 2026-06-01
Hytra refactor

k-dominik:
This PR mostly tests the new hytra package. I removed all yapsy plugins there and made them modules. Let's see if CI still likes it (it does so locally, with solvers, too).

removed a bunch of hytra imports that mostly checked whether hytra was in the env - unecessary.

direct instantiation of the relevant class for the gmm merger resolver

```
# where one ditches anything related to plugins...
self.mergerResolverPlugin = pluginManager.getMergerResolver()

# and instantiate it directly, explicitly
self.gmmMergerResolver = GMMMergerResolver()
```

there are some indirect plugin invocations via `IlastikMergerResolver` which are made explicit on the hytra side.

xref: #3194 

- [na] Add tests.
- [x] move hytra package to main
- [x] remove last commit with `noyapsy` channel references

### DIFF
--- a/ilastik/applets/tracking/conservation/opConservationTracking.py
+++ b/ilastik/applets/tracking/conservation/opConservationTracking.py
@@ -32,7 +32,8 @@ from hytra.core.fieldofview import FieldOfView
 from hytra.core.ilastikmergerresolver import IlastikMergerResolver
 from hytra.core.probabilitygenerator import ProbabilityGenerator
 from hytra.core.probabilitygenerator import Traxel
-from hytra.pluginsystem.plugin_manager import TrackingPluginManager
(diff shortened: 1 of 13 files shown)


ilastik/applets/tracking/conservation/opConservationTracking.py
```diff
@@ -130,9 +124,7 @@ def __init__(self, parent=None, graph=None):
         self.RelabeledCachedOutput.connect(self._relabeledOpCache.Output)
 
         # Merger resolver plugin manager (contains GMM fit routine)
```
btbest:
```suggestion
```

btbest (approved):
lgtm

I couldn't figure out how to actually get the merger resolver to find any mergers, but it looks like it should just call the same code as before, so should be fine...

PR #3222, opened 2026-06-03
Add folder picker to multiscale dataset browser

btbest:
... and reduce the dialog's minimum and default size.

I guess I got tired of copy-and-pasting folder paths from Windows Explorer eventually :)

image

- ~~Add docstrings and comments.~~
- [X] Add tests.
- ~~Reference relevant issues and other pull requests.~~
- ~~Update [user documentation](https://github.com/ilastik/ilastik.github.io).~~
- [X] Add screenshots / screen recordings.

### DIFF
--- a/ilastik/applets/dataSelection/multiscaleDatasetBrowser.py
+++ b/ilastik/applets/dataSelection/multiscaleDatasetBrowser.py
@@ -59,6 +59,8 @@ class DefaultWidthComboBox(QComboBox):
         return size
 
     def minimumSizeHint(self):
+        """Prevent the dropdown from being as wide as the history.
+        OME-Zarr URLs have a habit of being very long, making the dialog huge by default"""
         size = super().minimumSizeHint()
         size.setWidth(0)
         return size
@@ -129,10 +131,11 @@ class MultiscaleDatasetBrowser(QDialog):
     def setup_ui(self):
         self.setMinimumSize(30 * line_height(), 16 * line_height())
         self.setWindowTitle("Select Multiscale Source")
+        self.setAcceptDrops(True)
         main_layout = QVBoxLayout()
 
         description = QLabel(self)
-        description.setText('Enter path or URL and click "Check".')
+        description.setText('Enter path or URL and click "Check". Or drag-and-drop a folder.')
         main_layout.addWidget(description)
 
         self.combo = DefaultWidthComboBox(self)
@@ -160,8 +163,8 @@ class MultiscaleDatasetBrowser(QDialog):
         self.browse_button.clicked.connect(self._browse_for_directory)
         self.check_button = QPushButton(self)
         self.check_button.setText("Check")
-        self.check_button.clicked.connect(self._validate_text_input)
-        self.combo.lineEdit().returnPressed.connect(self.check_button.click)
+        self.check_button.clicked.connect(lambda _: self._validate_text_input())
+        self.combo.lineEdit().returnPressed.connect(self._validate_text_input)
         combo_layout.addWidget(combo_label, 0, 0)
         combo_layout.addWidget(self.combo, 0, 1)
         combo_layout.addWidget(self.browse_button, 0, 2)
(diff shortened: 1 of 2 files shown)


ilastik/applets/dataSelection/multiscaleDatasetBrowser.py
```diff
@@ -38,15 +39,31 @@
...
+    def sizeHint(self):
+        size = super().sizeHint()
+        size.setWidth(self.DEFAULT_WIDTH * line_height())
+        return size
+
+    def minimumSizeHint(self):
+        size = super().minimumSizeHint()
+        size.setWidth(0)
+        return size
+
```
k-dominik, line 67: Maybe it mandates a comment that this is necessary for the combobox not to be as wide as the history. Otherwise maybe a bit weird to see here (at least for me)

ilastik/applets/dataSelection/multiscaleDatasetBrowser.py
```diff
@@ -171,6 +196,45 @@ def update_ok_button(current_entered_text):
...
+        path = pathlib.Path(text).expanduser()
+        if path.exists():
+            return str(path if path.is_dir() else path.parent)
+        if path.parent.exists():
+            return str(path.parent)
+
+        return str(default_dir)
+
+    def _browse_for_directory(self):
+        options = QFileDialog.Options(QFileDialog.ShowDirsOnly)
```
k-dominik: so, on windows this seems to have no effect to the docs

> The name of a directory. Both files and directories are displayed. However, the native Windows file dialog does not support displaying files in the directory chooser.

or, are the docs wrong?

on mac, at least on mine, it also doesn't have an effect (it will always also show files)...

So have you seen this do something on linux??

ilastik/applets/dataSelection/multiscaleDatasetBrowser.py
```diff
@@ -171,6 +196,45 @@ def update_ok_button(current_entered_text):
...
+            "Select OME-Zarr Dataset Folder",
+            self._starting_directory_for_folder_picker(),
+            options=options,
+        )
+        if not directory:
+            return
+
+        preferences.set(self.PREFERENCES_GROUP, self.PREFERENCES_SETTING, pathlib.Path(directory).as_posix())
+        self.combo.lineEdit().setText(directory)
+        self.check_button.click()
```
k-dominik: this is a very indirect way of doing the validation. It goes via the event system, whereas you can do it directly in a synchronous call:

```suggestion
        self._validate_text_input(_event=None)
```

is much clearer...

k-dominik (commented):
This is useful! I left a few comments.

Not sure what to think about the truncated entries now in the drop down - looks a bit sad when dropped down.

Aaand of course a feature request. How is drag and drop not a thing in this dialog? :)

ilastik/applets/dataSelection/multiscaleDatasetBrowser.py
btbest: Err... on Windows it works just fine for me 🤨 But I'm guessing here we see one possible reason why we never added a folder picker for n5...

k-dominik: from above:

> However, the native Windows file dialog does not support displaying files in the directory chooser.

so it shouldn't make a difference there either... 🤷

btbest: I'm not really following... neither this thread nor the docs. The whole point is to get a folder picker that does *not* display files, which this does. I'm not sure why "does not support displaying files in the directory chooser" would be an issue (why would you want to display files when you're explicitly showing a dir chooser?).

If "Both files and directories are displayed", then what's the point of the option at all? I guess on Mac it might display both dirs and files but only let you select a dir?

As said, on Win it does exactly what I would have hoped for:

image

k-dominik: i guess my point is that this option does not have an effect on mac, likely not on win. so why set it. You already get your directory picker with `QFileDialog.getExistingDirectory`

k-dominik: Maybe to reiterate this point. Windows, per default, doesn't show files with `QFileDialog.getExistingDirectory` (with default options), macos will show files (non-selectable) with `QFileDialog.getExistingDirectory` no matter what options. And for other platforms, I would also suggest to let the default of whatever linux distro people use prevail. In the end users should get what they are used to when interacting with system dialogs and not force the "default" of one of the platforms (what `options = QFileDialog.Options(QFileDialog.ShowDirsOnly)` could do if it worked). I am always annoyed when, e.g. fiji forces this directories only behavior as I am used to seeing the files as context.

btbest: Totally agree; I had somehow missed the point that the option is redundant 😅

btbest:
Drag and drop should work now :)

I made the whole dialog the drop target instead of just the text box, because it's the whole dialog's point, so why make it less convenient than it needs to be.

k-dominik (approved):
Looks good, test failure in `test_browse_button_selects_directory_and_checks` due to incomplete refactor after running validate directly not via click.

If tests are green it should go in :). I really like the drag an drop - feels so responsive...

btbest:
oops yes, I saw the failure last night and forgot to make the commit 😅

PR #3230, opened 2026-06-19
Python 3.13 upgrade

### LINKED ISSUE #3192: 2026 April dependency updates
Ultimately want to get to Python 3.13, numpy 2.0 - but it seems to be a bit more involved - so adding this issue to track related subissues

### LINKED ISSUE #3193: Update python to 3.13

k-dominik:
Updated dependencies/pins:

* `python` 3.13 (for the devenv)
* `numpy` 2.0
* `pytorch` 2.10
* renamed `ilastiktools` -> `ilastik-carving-tools`

nice side effect - we finally get rid of the "patched" label ilastik-forge channel, sad side effect: needed to host another rebuild of pyshtools since PR in the Feedstock is not getting any attention.

* `DataSelectionGuiTests` some tests never really did anything (via asserting a mock `called_once` instead of calling `assert_called_once` ... These tests run now and are fixed.
* Fix `os.path.isabs` behavior on windows with py313: since Python 3.13 os.path.isabs() returns False if the path is not starting with a drive letter. Reflect different expectations for paths on the different OSses.
* 

- [x] Add tests.
- [x] remove commit that installs volumina from branch (https://github.com/ilastik/volumina/pull/363)

fixes #3192 
fixes https://github.com/ilastik/ilastik/issues/3193

needs https://github.com/ilastik/volumina/pull/363

### DIFF
--- a/.github/workflows/core-conda-bld.yml
+++ b/.github/workflows/core-conda-bld.yml
@@ -154,5 +154,5 @@ jobs:
         run: |
           set VOLUMINA_SHOW_3D_WIDGET=0
           conda build --test --override-channels ^
-            -c ./pkgs -c ilastik-forge/label/patched-2 -c conda-forge -c ilastik-forge ^
+            -c ./pkgs -c conda-forge -c ilastik-forge ^
             ./pkgs/noarch/%ILASTIK_PACKAGE_NAME%
--- a/Readme.md
+++ b/Readme.md
@@ -41,7 +41,7 @@ ilastik is also available as a conda package on our `ilastik-forge` conda channe
 You can install it from the commandline using [conda][conda]:
 
 ```bash
- conda create -n ilastik --override-channels -c ilastik-forge/label/patched-2 -c conda-forge -c ilastik-forge ilastik
+ conda create -n ilastik --override-channels -c conda-forge -c ilastik-forge ilastik
 
 # activate ilastik environment and start ilastik
 conda activate ilastik
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -30,7 +30,7 @@ outputs:
         - setuptools_scm >=6.2
       run:
         - python >=3.10
-        - numpy >1.12,<1.27
+        - numpy >=1.25
         # aiohttp to enable zarr.storage.FSStore
         - aiohttp
         - cachetools
@@ -46,7 +46,7 @@ outputs:
         - hytra >=1.2.1
         - ilastik-feature-selection
         - ilastikrag >=0.1.4
-        - ilastiktools
+        - ilastik-carving-tools
         - jsonschema
         - mamutexport
         - marching_cubes
@@ -71,7 +71,7 @@ outputs:
         # There seems to be a problem with scikit-learn 1.7.0, test testContourMWTExport breaks
         - scikit-learn <=1.6.1
         - tifffile >=2022
-        - vigra 1.12.2
+        - vigra 1.12.3
         - xarray !=2023.8.0,!=2023.9.0,!=2023.10.0
         - z5py 2.*
         - zarr 2.*
@@ -103,7 +103,7 @@ outputs:
         - pytest.ini
 
       requires:
-        - python 3.11
+        - python 3.13
         - pytest >4
         - pytest-qt
         - volumina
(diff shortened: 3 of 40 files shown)


k-dominik:
probably best to start with https://github.com/ilastik/volumina/pull/363 @btbest

might carry along the [TEMP: try flang21-based pyshtools](https://github.com/ilastik/ilastik/pull/3230/commits/aec8c3656b8c2815e364bf0dcbc2a0c03eeb1116) commit a bit, until PR is merged in the upstream feedstock

btbest (commented):
When trying to open an existing tracking project (probably all of the parameters handled there should be converted to plain bool/int):

```pytb
  File "C:\Users\root\Code\ilastik-group\k-dominik-ilastik\ilastik\applets\tracking\base\trackingBaseGui.py", line 77, in __init__
    super(TrackingBaseGui, self).__init__(parentApplet, topLevelOperatorView)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\root\Code\ilastik-group\k-dominik-ilastik\ilastik\applets\layerViewer\layerViewerGui.py", line 225, in __init__
    self.initAppletDrawerUi()  # Default implementation loads a blank drawer from drawer.ui.
    ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "C:\Users\root\Code\ilastik-group\k-dominik-ilastik\ilastik\applets\tracking\conservation\conservationTrackingGui.py", line 159, in initAppletDrawerUi
    super(ConservationTrackingGui, self).initAppletDrawerUi()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "C:\Users\root\Code\ilastik-group\k-dominik-ilastik\ilastik\applets\tracking\base\trackingBaseGui.py", line 274, in initAppletDrawerUi
    self._drawer = self._loadUiFile()
                   ~~~~~~~~~~~~~~~~^^
  File "C:\Users\root\Code\ilastik-group\k-dominik-ilastik\ilastik\applets\tracking\conservation\conservationTrackingGui.py", line 80, in _loadUiFile
    self._drawer.trackletsBox.setChecked(parameters["withTracklets"])
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: setChecked(self, a0: bool): argument 1 has unexpected type 'numpy.bool'
```


Otherwise lgtm 🎉  and always nice to see a little bit of test cleanup / fixes 😇

k-dominik:
Hi @btbest and thank you for pointing this one out! I went back and forth all morning then, unsure how to best tackle it. What I got to now in the end is a little `qtcompat` module, that does nothing much, other than going from numpy (which we will get after deserialization at the latest) and qt expected signatures. I'm still torn :) So a bit of feedback from your side would be helpful.

It's basically

```python
from ilastik.utitilty.gui.qtcompat import ensure_bool

some_checkbox.setChecked(ensure_bool(some_val))
```

vs

```python
some_checkbox.setChecked(bool(some_val))
```

For both we (the ilastik developers) need to agree that we will add those type conversions now whenever there is a chance the value could be something from numpy. The former (to me) looks a bit more intentional than the latter, which has the advantage of not inventing anything new. With the `qtcompat` I currently chose to refuse some conversions with these functions (like `ensure_bool("some_str")` will fail... so... idk).

For  the sake of the discussion I only added it to `ilastik/applets/tracking/conservation/conservationTrackingGui.py` but would probably squash in more once we agree on an approach (same applies to all _values_ passed to volumina)...

btbest:
My guess about what Qt did internally to our plain values we were giving it before is that the "dumb" conversion to bool/int/float would be functionally identical to previously, so it would've been fine.

I think the additional validators are a good addition though. We won't be able to enforce that they are used, and I estimate there's a decent chance that we'll forget about them ourselves. But we can try :) Agree, I'd replace at least all the conversions in this PR with the new validators. And commit title "Ensure conversions for compat don't mask nonsense values" but that's just taste.

k-dominik:
Okay I added some tests (more for documenting the expected failure cases) and went through pretty much all applet drawers - and added the guards whenever values came from a slot. Feels good ;)

btbest (approved):
forgot to press the green check...

PR #3232, opened 2026-06-25
Extract clearscale

btbest:
This PR simplifies ilastik's OME-Zarr metadata handling using the new [`clearscale`](https://github.com/ilastik/clearscale) package.

This isn't a pure refactor though, there are some subtle changes in written OME-Zarr metadata (reflected in the test changes on this PR).

Requires:

```
pip install git+https://github.com/ilastik/clearscale.git@ilastik-review2
```

## Production changes

For those who like to read, and for my own record if necessary...

### `slot.meta.scales` is now `clearscale.Multiscale`

The most significant functional change across ilastik is this, `slow.meta.scales` throughout workflows before was
`{scale_key : TaggedShape}` .
Now it's a `clearscale.Multiscale` object, which works like
`{scale_key : Scale(shape=Shape(), pixel_size=PixelSize(), unit=Unit(), translation=Translation()) }`

clearscale's `Shape`, `PixelSize` etc. are all basically flavours of `TaggedShape` (axis -> value mappings).

All of the test changes besides the ones specified below are to accommodate this change.

### `1.0` scale (pixel size) metadata is normalised to `0.0` on vigra axistags

This is cleaner in the sense that vigra axistag resolution is `0.0` by default. Currently, reading `1.0` out of scale metadata puts that 1.0 into axistag resolution. `1.0` is effectively noop, but the vigra axistag is then technically non-default.

I've implemented the normalisation in clearscale: vigra default `0.0` maps to clearscale default `1.0` in `PixelSize.from_vigra` and backwards in `PixelSize.to_vigra`. This lets the axistag genuinely be default when the received value is the default.

This changes the assertion in `test_pixel_sizes::test_read_ome_zarr_v0_4`

### OME-Zarr `scale` metadata is now more consistent with 0.4 spec

* "multiscale-transforms" refers to meta under `multiscale["coordinateTransformations"]`
* "dataset-transforms" refers to meta under `multiscale["datasets"][]["coordinateTransformations"]`

(Each of which can have one `scale` and optionally one `translation`)

Older OME-Zarr versions had a documented convention where `scale` along *unscaled* axes like `t` is recorded on multiscale-transforms, while `scale` along *scaled* axes (typically `xyz`) is recorded on dataset-transforms. The respective other axes get the default `1.0` scale. The examples in the spec only show this for `t`, because the `xyz` are scaled axes:
```
// ...
"datasets": [
    {
        "path": "0",
        "coordinateTransformations": [{ // the voxel size for the first scale level (0.5 micrometer)
            "type": "scale", "scale": [1.0, 1.0, 0.5, 0.5, 0.5]
        }]
    },
    // ...
]
"coordinateTransformations": [{ // the time unit (0.1 milliseconds), which is the same for each scale level
    "type": "scale",   "scale": [0.1, 1.0, 1.0, 1.0, 1.0]
}],
// ...
```

Despite being a prominent convention, and shown in the spec's own examples, this directly contradict's the spec's own phrasing:

> Each dictionary in “datasets” MUST contain the field “coordinateTransformations” [...]. They MUST contain exactly one scale transformation that specifies the pixel size in physical units or time duration.

Very explicitly, pixel size = dataset-transforms, not multiscale-transforms, including time.

ilastik's writer took the contradictory convention all the way, applying the general rule "scale for unscaled axes -> multiscale-transforms. rest -> dataset-transforms". For the `single-scale OME-Zarr` export format, none of the axes are scaled, which means all `scale` values (pixel size) were written to multiscale-transforms, and `scale` values in dataset-transforms were all `1.0`. Example exerpt from a test:

```
    expected_multiscale_transform = [{"type": "scale", "scale": [0.4, 8.99991, 5.0, 0.3, 6.4]}]  # tczyx
    expected_dataset_transform = [{"type": "scale", "scale": [1.0, 1.0, 1.0, 1.0, 1.0]}]
```

The convention would have these split as `multiscale = [0.4, 1.0, 1.0, 1.0, 1.0]; dataset = [1.0, 8.99991, 5.0, 0.3, 6.4]` instead.
In other words, the current output for single-scale OME-Zarr agrees with neither the spec, nor the convention.

Not good. We should follow the spec's phrasing (pixel size -> dataset-scale), not its examples (t-scale -> multiscale-transforms), and we definitely shouldn't do something supported by neither.

clearscale does detect specifically the t-scale convention, and roundtrips it when found. Outside of that convention, we now always write the pixel size for all axes to dataset-transforms.

This changes the assertion in:
* `test_pixel_sizes::test_write_ome_zarr_single_scale`  -- pixel size goes in dataset-transforms now
* (The diff in `testOpExportSlot::test_ome_zarr_single_scale` just adds an assert on the unchanged dataset-scale.)
* `testOpExportSlot::test_ome_zarr_roundtrip_multiscale` (new test added specifically to ensure the t-scale split roundtrips)
*  `test_write_ome_zarr`
   * `test_port_ome_zarr_metadata_single_scale_export` -- this test was obsolete ever since the pixel size PR, since the metadata now "roundtrips" by becoming pixel size slot meta, and the export using the slot meta as its source. Replaced with the next two.
   * `test_unscaled_single_scale_export_writes_crop_to_multiscale_transforms` -- pixel size goes in dataset-transforms now,...
   * `test_unscaled_single_scale_export_round_trips_t_convention` -- ...unless the input multiscale had t-scale in multiscale-transforms, then roundtrip
   * `test_resized_single_scale_export` -- similarly sort of obsolete since the pixel size PR. Pixel size is now written in dataset-transforms.
   * `test_port_ome_zarr_metadata_multi_scale_export` -- t-scale now also in dataset-transforms (just like the test's input meta)

### OME-Zarr `translation` metadata now maintains a clean separation between roundtripped translation and added crop-translation

As above, this concerns the semantics of dataset-transformations vs. multiscale-transformations.

When the user exports a subregion, the `translation` of the multiscale-transforms is a good place to express the exported dataset's shift relative to its parent dataset.
This is already status-quo and isn't changed in this PR.

If the input already had its own translation, previously this would be added to the translation of the multiscale-transforms. Dataset translation was always 0. (And the summing into multiscale-transforms could be numerically wrong if the same axis also had a non-1.0 scale transform...)

This now round-trips more exactly: translation in dataset-transforms stays in dataset-transforms, translation in multiscale-transforms stays in multiscale-transforms. Even if multiscale-transforms are identity (which previously would be dropped in the export), that is now round-tripped.

`testOpExportSlot::test_ome_zarr_roundtrip_multiscale`, the new test for the t-scale split roundtrip, also demonstrates how translations roundtrip. dataset-translation -> dataset-translation (identity in the test), multiscale-translation -> multiscale-translation.

Where this gets complicated is when an OME-Zarr input was using the t-scale split convention. Having a translation next to the scale on the multiscale-transforms is not expected in this convention, and is semantically poorly defined. When a dataset follows this convention, it implies that the multiscale-transforms are part of the multiscale's own coordinate system (because the dataset-transforms alone would leave the t-axis scaled to 1.0 units, seconds or whatever, which is almost certainly wrong). There is no room for an additional transform to some external reference space in this case, so we have to choose between (1) maintaining the convention in our export and discarding a crop-translation if there was one; or (2) writing our crop-translation and composing the global t-scale into each dataset-transform, thereby breaking the written export out of the convention. I think this is a niche enough edge case that what exactly we choose is unlikely to matter in practice. I've figured that if a user is working with datasets that use this convention, they might be in a workflow with other tools that rely on this convention. Then it would be more important that the outputs conform to the convention, than adding metadata that express "this dataset is shifted relative to some unknown external reference".

* `testOpExportSlot`
   * `test_ome_zarr_multi_scale` -- dataset-translation stays dataset-translation now, identity multiscale-transforms stay identity multiscale-transforms
   * `test_ome_zarr_roundtrip_multiscale` (new test added for the t-scale split roundtrip also checks translation roundtrip)
*  `test_write_ome_zarr`
   * `test_port_ome_zarr_metadata_single_scale_export` -- obsolete, replaced with the next two.
   * `test_unscaled_single_scale_export_writes_crop_to_multiscale_transforms` -- crop offset becomes translation inside multiscale-transforms; dataset-transforms now maintain the source dataset's translation instead of adding that onto the crop-translation
   * `test_unscaled_single_scale_export_t_scale_convention_overrides_storing_crop` -- ...unless the input multiscale had t-scale in multiscale-transforms, then we do not write the crop translation at all
   * `test_port_ome_zarr_metadata_multi_scale_export` -- source dataset translation goes into every output dataset, crop translation goes into multiscale-transforms

## Checklist

- [X] Add tests.
- (-) Update [user documentation](https://github.com/ilastik/ilastik.github.io).
- (-) Add screenshots / screen recordings.
- [ ] After review: Publish clearscale to PyPI and conda-forge and update `dev/environment-dev.yml` and `meta.yaml`

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -21,6 +21,7 @@ outputs:
       script_env:
         - SETUPTOOLS_SCM_PRETEND_VERSION={{version}}
       script:
+        - {{PYTHON}} -m pip install git+https://github.com/ilastik/clearscale@ilastik-review
         - {{PYTHON}} -m pip install . --no-deps -vv
     requirements:
       host:
@@ -28,6 +29,7 @@ outputs:
         - pip
         - setuptools >=64.0
         - setuptools_scm >=6.2
+        - hatchling
       run:
         - python >=3.10
         - numpy >=1.25
--- a/dev/environment-dev.yml
+++ b/dev/environment-dev.yml
@@ -48,6 +48,9 @@ dependencies:
   - aiohttp
   - fsspec
   - s3fs >=2022.8.2
+  - pip
+  - pip:
+    - git+https://github.com/ilastik/clearscale@ilastik-review
 
   # Deep learning dependencies (neural net workflow, trainable domain adaptation)
   # can be changed to request gpu versions by uncommenting the following line:
--- a/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
+++ b/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
@@ -19,17 +19,15 @@
 # 		   http://ilastik.org/license.html
 ###############################################################################
 
-import json
 from collections import OrderedDict
 from functools import reduce
-from typing import Dict, List
-
+from typing import List, Set
 
 from qtpy.QtCore import Qt, QAbstractItemModel, QModelIndex
 from ilastik.utility import bind
 from ilastik.utility.gui import ThreadRouter, threadRouted
 from lazyflow.utility.helpers import bigintprod, eq_shapes
-from lazyflow.utility.io_util.multiscaleStore import Multiscale
+from lazyflow.utility.io_util.clearscale import Scale, Multiscale, Shape as ClearscaleShape
 from .opDataSelection import DatasetInfo
 from .dataLaneSummaryTableModel import rowOfButtonsProxy
 
@@ -45,12 +43,12 @@ class DatasetColumn:
     NumColumns = 7
 
 
-def _dims_to_display_string(dimensions: Dict[str, int], axiskeys: str, dtype: type) -> str:
+def _dims_to_display_string(scale: Scale, dtype: type) -> str:
     """Generate labels to put into the scale combobox / to display in the table.
-    XYZ dimensions will be reordered to match axiskeys."""
-    reordered_dimensions = [dimensions[axis] for axis in axiskeys if axis in "xyz"]
-    shape_string = " / ".join(str(size) for size in reordered_dimensions)
-    scale_file_size = bigintprod(reordered_dimensions) * dtype().nbytes
+    Aim: XYZ dimensions + data size, but in original axis order"""
+    xyz_shape = scale.shape.without_axes_except("xyz").to_list()
+    shape_string = " / ".join(str(size) for size in xyz_shape)
+    scale_file_size = bigintprod(xyz_shape) * dtype().nbytes
     size_labels = {1e15: "PB", 1e12: "TB", 1e9: "GB", 1e6: "MB", 1e3: "KB", 1: "B"}
     size_factor = [f for f in size_labels.keys() if scale_file_size >= f][0]
     size_string = f"{scale_file_size / size_factor:.1f} {size_labels[size_factor]}"
@@ -223,9 +221,7 @@ class DatasetDetailedInfoTableModel(QAbstractItemModel):
             return str(datasetInfo.drange or "")
         if index.column() == DatasetColumn.Scale:
             if datasetInfo.scales:
-                return _dims_to_display_string(
-                    datasetInfo.scales[datasetInfo.working_scale], datasetInfo.axiskeys, datasetInfo.laneDtype
-                )
+                return _dims_to_display_string(datasetInfo.scales[datasetInfo.working_scale], datasetInfo.laneDtype)
             return UninitializedDisplayData[index.column()]
 
         if index.column() == DatasetColumn.PixelSize:
@@ -275,8 +271,8 @@ class DatasetDetailedInfoTableModel(QAbstractItemModel):
         # is the first option in the drop-down box
         return OrderedDict(
             {
-                key: _dims_to_display_string(tagged_shape, datasetInfo.axiskeys, datasetInfo.laneDtype)
-                for key, tagged_shape in reversed(datasetInfo.scales.items())
+                key: _dims_to_display_string(scale, datasetInfo.laneDtype)
+                for key, scale in reversed(datasetInfo.scales.items())
             }
         )
 
@@ -286,16 +282,16 @@ class DatasetDetailedInfoTableModel(QAbstractItemModel):
         Return indices of the options from `get_scale_options` that have matching scales in all other roles.
         """
 
-        def shapes_intersection(shapes1: List[Dict[str, int]], shapes2: List[Dict[str, int]]):
-            return [s1 for s1 in shapes1 if any(eq_shapes(s1, s2) for s2 in shapes2)]
+        def shapes_intersection(shapes1: Set[ClearscaleShape], shapes2: Set[ClearscaleShape]):
+            return {s1 for s1 in shapes1 if any(eq_shapes(s1, s2) for s2 in shapes2)}
 
         # Checking if ready etc. shouldn't be necessary here because this should only be called
         # while building the scale dropdown (by which point everything should be ready).
-        scale_shapes_per_role: List[List[OrderedDict[str, int]]] = [
+        scale_shapes_per_role: List[Set[ClearscaleShape]] = [
             (
-                list(role_slot.value.scales.values())
+                set(role_slot.value.scales.keys_by_shape)
                 if role_slot.value.scales
-                else [dict(zip(role_slot.value.axistags.keys(), role_slot.value.laneShape))]
+                else {ClearscaleShape(zip(role_slot.value.axistags.keys(), role_slot.value.laneShape))}
             )
             for role_slot in self._op.DatasetGroupOut[laneIndex]
             if role_slot.ready()
@@ -304,8 +300,8 @@ class DatasetDetailedInfoTableModel(QAbstractItemModel):
 
         scale_options: Multiscale = self._op.DatasetGroupOut[laneIndex][self._roleIndex].value.scales
         indices_of_common_scales = []
-        for i, shape in enumerate(reversed(scale_options.values())):
-            if any(eq_shapes(shape, common_shape) for common_shape in common_scale_shapes):
+        for i, scale in enumerate(reversed(scale_options.values())):
+            if any(eq_shapes(scale.shape, common_shape) for common_shape in common_scale_shapes):
                 indices_of_common_scales.append(i)
         return indices_of_common_scales
 
--- a/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
+++ b/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
@@ -23,11 +23,11 @@ from collections import OrderedDict
 from functools import reduce
 from typing import List, Set
 
+from clearscale import Scale, Multiscale, Shape as ClearscaleShape
 from qtpy.QtCore import Qt, QAbstractItemModel, QModelIndex
 from ilastik.utility import bind
 from ilastik.utility.gui import ThreadRouter, threadRouted
 from lazyflow.utility.helpers import bigintprod, eq_shapes
-from lazyflow.utility.io_util.clearscale import Scale, Multiscale, Shape as ClearscaleShape
 from .opDataSelection import DatasetInfo
 from .dataLaneSummaryTableModel import rowOfButtonsProxy
 
--- a/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
+++ b/ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
@@ -23,7 +23,7 @@ from collections import OrderedDict
 from functools import reduce
 from typing import List, Set
 
-from clearscale import Scale, Multiscale, Shape as ClearscaleShape
+import clearscale
 from qtpy.QtCore import Qt, QAbstractItemModel, QModelIndex
 from ilastik.utility import bind
 from ilastik.utility.gui import ThreadRouter, threadRouted
@@ -43,7 +43,7 @@ class DatasetColumn:
     NumColumns = 7
 
 
-def _dims_to_display_string(scale: Scale, dtype: type) -> str:
+def _dims_to_display_string(scale: clearscale.Scale, dtype: type) -> str:
     """Generate labels to put into the scale combobox / to display in the table.
     Aim: XYZ dimensions + data size, but in original axis order"""
     xyz_shape = scale.shape.without_axes_except("xyz").to_list()
@@ -282,23 +282,23 @@ class DatasetDetailedInfoTableModel(QAbstractItemModel):
         Return indices of the options from `get_scale_options` that have matching scales in all other roles.
         """
 
-        def shapes_intersection(shapes1: Set[ClearscaleShape], shapes2: Set[ClearscaleShape]):
+        def shapes_intersection(shapes1: Set[clearscale.Shape], shapes2: Set[clearscale.Shape]):
             return {s1 for s1 in shapes1 if any(eq_shapes(s1, s2) for s2 in shapes2)}
 
         # Checking if ready etc. shouldn't be necessary here because this should only be called
         # while building the scale dropdown (by which point everything should be ready).
-        scale_shapes_per_role: List[Set[ClearscaleShape]] = [
+        scale_shapes_per_role: List[Set[clearscale.Shape]] = [
             (
                 set(role_slot.value.scales.keys_by_shape)
                 if role_slot.value.scales
-                else {ClearscaleShape(zip(role_slot.value.axistags.keys(), role_slot.value.laneShape))}
+                else {clearscale.Shape(zip(role_slot.value.axistags.keys(), role_slot.value.laneShape))}
             )
             for role_slot in self._op.DatasetGroupOut[laneIndex]
             if role_slot.ready()
         ]
         common_scale_shapes = reduce(shapes_intersection, scale_shapes_per_role)
 
-        scale_options: Multiscale = self._op.DatasetGroupOut[laneIndex][self._roleIndex].value.scales
+        scale_options: clearscale.Multiscale = self._op.DatasetGroupOut[laneIndex][self._roleIndex].value.scales
         indices_of_common_scales = []
         for i, scale in enumerate(reversed(scale_options.values())):
             if any(eq_shapes(scale.shape, common_shape) for common_shape in common_scale_shapes):
--- a/lazyflow/utility/helpers.py
+++ b/lazyflow/utility/helpers.py
@@ -22,7 +22,7 @@
 
 from functools import reduce
 from operator import mul
-from typing import Iterable, Tuple, Type, Union, Dict
+from typing import Iterable, Tuple, Type, Union, Mapping
 import sys
 import numbers
 import numpy
@@ -165,8 +165,10 @@ def get_ram_per_element(dtype: Union[Type[object], numpy.dtype]) -> int:
         return sys.getsizeof(None)
 
 
-def eq_shapes(test: Dict[str, int], ref: Dict[str, int]) -> bool:
+def eq_shapes(test: Mapping[str, int], ref: Mapping[str, int]) -> bool:
     """Check if two tagged shapes are equal. Ignore channel. Additional singleton axes are ok."""
+    if hasattr(test, "without_axes") and hasattr(ref, "without_axes"):
+        return test.without_axes("c").without_singletons() == ref.without_axes("c").without_singletons()
     common_axes = set(test.keys()) & set(ref.keys())
     extra_axes = set(test.keys()) ^ set(ref.keys())
     common_match = all(test[a] == ref[a] for a in common_axes if a != "c")
--- a/lazyflow/utility/io_util/OMEZarrStore.py
+++ b/lazyflow/utility/io_util/OMEZarrStore.py
@@ -233,10 +233,10 @@ class OMEZarrTranslations:
         if dataset is None and self.multiscale_translation is None:
             return Translation.identity(axiskeys)
         if dataset is None:
-            return Translation(self.multiscale_translation).reorder(axiskeys)
+            return Translation(self.multiscale_translation).with_axes(axiskeys)
         if self.multiscale_translation is None:
-            return Translation(dataset).reorder(axiskeys)
-        return (Translation(self.multiscale_translation) + Translation(dataset)).reorder(axiskeys)
+            return Translation(dataset).with_axes(axiskeys)
+        return (Translation(self.multiscale_translation) + Translation(dataset)).with_axes(axiskeys)
 
 
 def _axistags_from_multiscale(multiscale: OME_ZARR_MULTISCALE) -> vigra.AxisTags:
--- a/lazyflow/utility/io_util/OMEZarrStore.py
+++ b/lazyflow/utility/io_util/OMEZarrStore.py
@@ -25,6 +25,8 @@ import os
 from typing import Dict, List, Optional, Literal, Tuple, Any
 from urllib.parse import unquote_to_bytes
 
+from clearscale import Multiscale
+from clearscale.ome_zarr import make_fake_shapes
 import s3fs
 import vigra
 from aiohttp import ClientConnectorError, ClientResponseError
@@ -35,8 +37,6 @@ from zarr.storage import FSStore, LRUStoreCache
 
 from lazyflow import rtype
 from lazyflow.utility import Timer, Memory
-from lazyflow.utility.io_util.clearscale import Multiscale
-from lazyflow.utility.io_util.clearscale.ome_zarr import make_fake_shapes
 from lazyflow.utility.io_util.multiscaleStore import MultiscaleStore, DEFAULT_SCALE_KEY
 
 logger = logging.getLogger(__name__)
--- a/lazyflow/utility/io_util/OMEZarrStore.py
+++ b/lazyflow/utility/io_util/OMEZarrStore.py
@@ -26,7 +26,7 @@ from typing import Dict, List, Optional, Literal, Tuple, Any
 from urllib.parse import unquote_to_bytes
 
 from clearscale import Multiscale
-from clearscale.ome_zarr import make_fake_shapes
+from clearscale.ome_zarr import make_proportional_shapes
 import s3fs
 import vigra
 from aiohttp import ClientConnectorError, ClientResponseError
@@ -281,7 +281,7 @@ def _fetch_and_validate_ome_zarr_spec(uri: str, sort_uri: Optional[str] = None) 
     valid_ms = []
     for ms_dict in spec["multiscales"]:
         try:
-            valid_ms.append(Multiscale.from_ome_zarr(ms_dict, get_shape=make_fake_shapes(ms_dict)))
+            valid_ms.append(Multiscale.from_ome_zarr(ms_dict, shape_source=make_proportional_shapes(ms_dict)))
         except ValueError:
             continue
     if not valid_ms:
@@ -389,7 +389,7 @@ class OMEZarrStore(MultiscaleStore):
                 }
             return zarray.shape
 
-        multiscale = Multiscale.from_ome_zarr(self._multiscale_spec, get_shape=get_shape_keep_zarray)
+        multiscale = Multiscale.from_ome_zarr(self._multiscale_spec, shape_source=get_shape_keep_zarray)
         dtype = next(iter(self._scale_data.values()))["zarray"].dtype.type
         axistags = vigra.defaultAxistags("".join(multiscale.axes()))
         super().__init__(
--- a/lazyflow/utility/io_util/OMEZarrStore.py
+++ b/lazyflow/utility/io_util/OMEZarrStore.py
@@ -39,7 +39,8 @@ from zarr.storage import FSStore, LRUStoreCache
 
 from lazyflow import rtype
 from lazyflow.utility import Timer, Memory
-from lazyflow.utility.io_util.clearscale import Scale, Shape, Spacing, Unit
+from lazyflow.utility.io_util.clearscale import Scale, Shape, Spacing, Unit, Translation
+from lazyflow.utility.io_util.clearscale._axis_values import Axes
 from lazyflow.utility.io_util.multiscaleStore import MultiscaleStore, DEFAULT_SCALE_KEY
 
 logger = logging.getLogger(__name__)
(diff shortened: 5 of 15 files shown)


btbest:
Looks like I built myself a footgun with the clever simplification that transforms do on composition. Should stop having clever thoughts entirely. Needs to be fixed on clearscale's side.

btbest:
Rebased on main after py3.13

ilastik/applets/dataSelection/datasetDetailedInfoTableModel.py
```diff
@@ -45,12 +43,12 @@ class DatasetColumn:
     NumColumns = 7
 
 
-def _dims_to_display_string(dimensions: Dict[str, int], axiskeys: str, dtype: type) -> str:
+def _dims_to_display_string(scale: Scale, dtype: type) -> str:
```
k-dominik: Not from the current changeset really, but while at it....
```suggestion
def _dims_to_display_string(scale: Scale, dtype: "Type[numpy.number]") -> str:
```

with
```Python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    import numpy
```
or something like this is probably more exact - since

tests/test_lazyflow/test_operators/test_ioOperators/testOpInputDataReader.py
```diff
@@ -387,28 +382,20 @@ def ome_zarr_store_on_disc(
...
         expected_images = [image_original, image_scaled]
-        # Scales metadata for GUI
-        expected_multiscale = OrderedDict(
-            [
-                (path0, OrderedDict(zip("cyx", dataset_shape))),
-                (path1, OrderedDict(zip("cyx", scaled_shape))),
-            ]
+        expected_multiscale = clearscale.Multiscale.from_ome_zarr(
+            correct_multiscale_zattrs,
+            shape_source=lambda path: tuple({path0: dataset_shape, path1: scaled_shape}[path]),
```
k-dominik, line 387:
```suggestion
            shape_source={path0: dataset_shape, path1: scaled_shape},
```
if you define `dataset_shape` and `scaled_shape` (and `chunk_size`) as tuples right away (no reason not to as far as I can see), then this is much more readable. Or is there some benefit in having a callable here (clearscale code seems to upgrade shape_source to a callable anyway).

tests/test_lazyflow/test_operators/test_ioOperators/testOpExportSlot.py
```diff
@@ -297,6 +295,140 @@ def test_ome_zarr_roundtrip(self):
...
+                "coordinateTransformations": [
+                    {"scale": [0.3, 1.0, 1.0], "type": "scale"},
+                    {"translation": [2.0, 0.6, 0.0], "type": "translation"},
+                ],
+            }
+        ]
+        # Expected written meta is the same as input, but:
+        # - tczyx
+        # - no name
+        # - multiscale-global transformations are folded into datasets (similar to ome-zarr 0.6 convention)
```
k-dominik, line 343: How sure are you about the folding of existing transformation. I'm not against it, but I would only change the metadata layout if there is a grounding in spec. From the 0.6rc0 spec I currently get:

> If applications require additional transformations, each multiscales object MAY contain the field coordinateTransformations, describing transformations that are applied to all resolution levels in the same manner. The following constraints apply...

The way I read it is that there are legit use-cases for `multiscales > coordinateTransformations` and these should probably not be touched?

In any case, you're enforcing transformation folding already in 0.4 (which again, I could not find this terminology in 0.6, but I have spent way less time than you with the spec) so I think it's fair to ask for a reference.

lazyflow/utility/helpers.py
```diff
@@ -165,8 +165,10 @@ def get_ram_per_element(dtype: Union[Type[object], numpy.dtype]) -> int:
         return sys.getsizeof(None)
 
 
-def eq_shapes(test: Dict[str, int], ref: Dict[str, int]) -> bool:
+def eq_shapes(test: Mapping[str, int], ref: Mapping[str, int]) -> bool:
     """Check if two tagged shapes are equal. Ignore channel. Additional singleton axes are ok."""
+    if hasattr(test, "without_axes") and hasattr(ref, "without_axes"):
```
k-dominik, line 170: lazyflow already depends on clearscale - why not do a proper instance check here and type annotations?

btbest:
Ok the first review point from this morning regarding always using `import clearscale` has revealed that clearscale wasn't exposing the `ome_zarr` submodule on the top-level API and that's breaking the tests :) I've pushed the [simple fix](https://github.com/ilastik/clearscale/commit/ab78eb70613b11ecd041653f16adadbf26beda1c) to clearscale but I thought I'd leave the tag where it is now to avoid adding further disruptions to your review 😅 let me know if you'd rather have me move the tag

tests/test_lazyflow/test_utility/test_io_util/test_write_ome_zarr.py
```diff
@@ -195,7 +182,8 @@ def test_port_ome_zarr_metadata_single_scale_export(tmp_path, tiny_5d_vigra_arra
                 {
                     "path": "source_scale",
                     "coordinateTransformations": [
-                        {"type": "scale", "scale": "should not be accessed"},
+                        # Bad metadata: Ensure it's not copied
+                        {"type": "scale", "scale": [2.4, 1.3, 3.7, 6.9]},
```
k-dominik, line 186: For you this is probably obvious - for me not at all - not sure why this metadata is "bad".

Related: is the comment for this test function still correct? (e.g. "multiscale["coordinateTransformations"] should report scale for t-axis only")? It doesn't look like it. But this is also related to your decision to "fold" transformations into the datasets.

lazyflow/utility/io_util/write_ome_zarr.py
```diff
@@ -425,8 +206,8 @@ def write_ome_zarr(
             "Appending to an existing OME-Zarr store is not yet implemented."
             f"\nPath: {abs_export_path}."
         )
-    export_offset: TaggedShape = (
-        ODict(zip(image_source_slot.meta.getAxisKeys(), export_offset)) if export_offset else None
+    export_offset = (
```
k-dominik, line 209: my type checker doesn't like overwriting the variable with clearscale.PixelOffset...

lazyflow/utility/io_util/write_ome_zarr.py
```diff
@@ -435,37 +216,38 @@ def write_ome_zarr(
...
+        input_multiscale = reordered_source.meta.get("scales")
+        input_scale = input_multiscale[input_scale_key] if input_multiscale and input_scale_key else None
         interpolation_order = OpResize.semantics_to_interpolation[
             reordered_source.meta.get("data_semantics", ImageTypes.Intensities)
         ]
 
         if target_scales is None:  # single-scale export
             single_target_key = input_scale_key if input_scale_key else SINGE_SCALE_DEFAULT_KEY
-            target_scales = ShapesByScaleKey({single_target_key: export_shape})
+            target_scales = clearscale.BlueprintShapes({single_target_key: export_shape})
```
k-dominik, line 230: again, assigning with a type that is not matching the type definition

k-dominik (commented):
Looks mostly good - some nitpicks are expected with bigger code changes :). In general it's very nice to see all the multiscale stuff encapsulated. I find the code very readable once one is a bit familiar with the design decisions in clearscale.

The only thing I am not sure about is the folding of multiscale trafos into datasets. First of all behavior before was to keep the `multiscale > transformations`. No you're folding. I'm not sure I could find this in the 0.4 spec (which we're writing currently) and also not in the future 0.6 spec. I would opt to keep the original behavior. It's great that you have transformation concatenation in your library and I'm sure it will have its use-cases. But in general keeping the data layout most faithful to the original is what users would hope for (if they care). The metadata is also readable by humans, and then messing with transformations that way could get confusion - non obvious.

btbest:
Ok well, I guess I'm giving up trying to find a workaround for the version parsing for circleci and appveyor. I've updated the PR description with the details of how OME-Zarr metadata round-trips now.

dev/environment-dev.yml
```diff
@@ -49,6 +49,9 @@ dependencies:
   - aiohttp
   - fsspec
   - s3fs >=2022.8.2
+  - pip
+  - pip:
+    - git+https://github.com/ilastik/clearscale@ilastik-review2
```
k-dominik, line 54: I know this is only a temporary thing but, you used hatchling for packaging - the way you set it up it requires something that looks like a version. Something like `v0.0.1+ilastik.review2` would've worked... (I had to get a working local environment, too ;))

tests/test_lazyflow/test_operators/test_ioOperators/testOpInputDataReader.py
```diff
@@ -387,28 +382,20 @@ def ome_zarr_store_on_disc(
...
-                (path0, OrderedDict(zip("cyx", dataset_shape))),
-                (path1, OrderedDict(zip("cyx", scaled_shape))),
-            ]
+        expected_multiscale = clearscale.Multiscale.from_ome_zarr(
+            correct_multiscale_zattrs,
+            shape_source={path0: dataset_shape, path1: scaled_shape},
         )
-        # Singleton the error so that OMEZarrMultiscaleMeta objects can be eq compared
-        err_placeholder = InvalidTransformationError()
-        monkeypatch.setattr(InvalidTransformationError, "__new__", lambda _: err_placeholder)
```
k-dominik, line 399: with that the `monkeypatch` arg is in the function is no longer needed

tests/test_lazyflow/test_operators/test_ioOperators/test_pixel_sizes.py
```diff
@@ -769,6 +770,9 @@ def test_read_ome_zarr_v0_4_no_units(graph):
     }
     array_mock = mock.Mock()
     array_mock.shape = (3, 2, 11, 10, 9)
+    # 1.0 is the default scale in OME-Zarr. It will be mapped to the vigra default resolution 0.0
+    expected_resolution = dataset_scale.copy()
+    expected_resolution["t"] = 0
```
k-dominik, line 775: I'm not sure I get this reasoning. The default in vigra is 0.0 indicating unset. I looked at the clearscale code and it seems the to_vigra method will always map 1.0 to 0.0 (`if  ... self[tag.key] != self._default()`) - which is not the same. The scale could very well be 1.0 and in this case interpreting it to "unset" is wrong. One experiment with 1.0 for t - you get 0 in ilastik, next experiment with 1.0001 for t - you get 1.0001. I think when going from something multiscale to vigra, it should always be set. On the other hand loading from vigra going into a multiscale, the conversion should be done. If you want round-tripping then you'd have to find another way of tracking whether something was defaulted then comparing to a plausible and allowed value for that scale.

lazyflow/utility/io_util/OMEZarrStore.py
```diff
@@ -498,22 +195,17 @@ def _fetch_and_validate_ome_zarr_spec(uri: str, sort_uri: Optional[str] = None)
...
-            "or reports no multiscale datasets. Details below."
-            f"\nMetadata read from {uri}/.zattrs."
-            f"\n\nProblematic metadata entry: {e.json_path}"
-            f"\nProblem: {e.message}"
-            f"\nRequired properties: {e.schema}"
-            f"\nFull metadata received:\n{spec}"
+    ome_group = clearscale.OmeZarrGroup.from_attrs(spec, shape_source="singletons")
+    if not ome_group.multiscales:
+        if not ome_group.children:
+            raise NoOMEZarrMetaFound(f"No valid multiscale entry found in: {spec!r}")
```
k-dominik, line 201: Maybe a nicer interface would be if the `ome_group` object would should know if it's a valid multiscale? Otherwise one would have to know about these conventions here...

lazyflow/utility/io_util/OMEZarrStore.py
```diff
@@ -587,49 +279,46 @@ def __init__(self, uri: str, target_scale: Optional[str] = None):
...
                     "zarray": zarray,
                     "chunks": zarray.chunks,
                     "shape": zarray.shape,
                 }
-                logger.info(f"Initializing scale {scale_key} took {timer.seconds()*1000} ms.")
-        self.ome_zarr_translations = OMEZarrTranslations.from_multiscale_spec(self._multiscale_spec)
+            return zarray.shape
+
+        multiscale = clearscale.Multiscale.from_ome_zarr(self._multiscale_spec, shape_source=get_shape_keep_zarray)
+        dtype = next(iter(self._scale_data.values()))["zarray"].dtype.type
```
k-dominik, line 313:
```suggestion
        dtype = self._scale_data.first_value["zarray"].dtype.type
```

🤷 or is there a reason you're not using this method?

lazyflow/utility/io_util/write_ome_zarr.py
```diff
@@ -20,182 +20,102 @@
...
+    chunk_shape: clearscale.Shape,
     export_dtype,
 ) -> zarr.Array:
     """Creates folders and zarr-internal (not OME) metadata files."""
-    assert len(chunk_shape) == len(scale_shape), "chunk and image shape must have same dimensions"
+    assert list(chunk_shape.keys()) == list(scale_shape.keys()), "chunk and image shape must have same axes"
     store = FSStore(abs_export_path, mode="w", **OME_ZARR_V_0_4_KWARGS)
-    zarray = zarr.creation.zeros(scale_shape, store=store, path=scale_key, chunks=chunk_shape, dtype=export_dtype)
+    zarray = zarr.creation.zeros(
+        scale_shape.to_tuple(), store=store, path=scale_key, chunks=chunk_shape.to_tuple(), dtype=export_dtype
```
k-dominik, line 117: Feedback for the interface: I found the behavior of `to_tuple` surprising for a dict. `value_tuple` or `to_value_tuple`, or tuple(scale.values()) :D but yeah. But you've looked at this longer...

Edit: After having it seen now in so many places maybe I warmed up to it? :)

lazyflow/utility/io_util/write_ome_zarr.py
```diff
@@ -364,44 +149,55 @@ def _get_scaling_method_metadata(export_scalings: ScalingsByScaleKey, interpolat
...
-    if multiscale_transformations:
-        ome_zarr_multiscale_meta["coordinateTransformations"] = multiscale_transformations
-
-    scaling_meta = _get_scaling_method_metadata(export_scalings, interpolation_order)
-    if scaling_meta:
-        ome_zarr_multiscale_meta["metadata"] = scaling_meta
+    # Exporter versions:
+    # 2: after adding the `multi-scale OME-Zarr` format
+    # 3: after extracting clearscale
+    ilastik_signature = {"name": "ilastik", "version": ilastik_version, "ome_zarr_exporter_version": 3}
```
k-dominik, line 163: This could have maybe a more prominent place - a constant maybe?

lazyflow/utility/io_util/write_ome_zarr.py
```diff
@@ -410,62 +206,64 @@ def write_ome_zarr(
...
+    export_offset_raw: Union[ShapeTuple, None],
+    target_scales: Optional[clearscale.BlueprintShapes] = None,
 ):
     pc = PathComponents(export_path)
     if pc.internalPath:
         raise ValueError(
             f'Internal paths are not supported by OME-Zarr export. Received internal path: "{pc.internalPath}"'
         )
-    abs_export_path = pc.externalPath
+    abs_export_path = str(pc.externalPath)
```
k-dominik, line 217: is this _not_ a string?

lazyflow/utility/io_util/write_ome_zarr.py
k-dominik, line 147: not part of the diff, but if the version here is tracking skimage, then it's already behind (current is 0.26). skimage has a version attr (`skimage.__version__`) but probably better to use importlib...

```
from importlib.metadata import version
skimage_version = version("scikit-image")
```

tests/test_lazyflow/test_operators/test_ioOperators/testOpExportSlot.py
```diff
@@ -297,6 +314,141 @@ def test_ome_zarr_roundtrip(self):
...
+                "coordinateTransformations": [
+                    {"scale": [0.3, 1.0, 1.0], "type": "scale"},
+                    {"translation": [2.0, 0.6, 0.0], "type": "translation"},
+                ],
+            }
+        ]
+        # Expected written meta is the same as input, but:
+        # - tczyx
+        # - no name
+        # - "s1" transformations are discarded. The exported "s1" is a newly generated downscale.
```
k-dominik, line 362: :+1:

tests/test_lazyflow/test_operators/test_ioOperators/testOpExportSlot.py
```diff
@@ -297,6 +314,141 @@ def test_ome_zarr_roundtrip(self):
...
+        assert len(written_scale) == len(expected_scale)
+        assert written_scale == pytest.approx(expected_scale, abs=1e-15)
+        if len(expected_transforms) == 1:
+            return
+        assert "translation" in written_transforms[1]
+        written_transl = written_transforms[1]["translation"]
+        expected_transl = expected_transforms[1]["translation"]
+        assert len(written_transl) == len(expected_transl)
+        assert written_transl == pytest.approx(expected_transl, abs=1e-15)
+
```
k-dominik, line 451: maybe to be safe right now raise if len(expected_transforms) > 2, as this function doesn't handle it and who knows what the future brings ;)

tests/test_lazyflow/test_operators/test_ioOperators/testOpInputDataReader.py
btbest: No benefit, just leftover from a previous version where shape_source only took callables :)

tests/test_lazyflow/test_operators/test_ioOperators/testOpExportSlot.py
btbest: No more folding going on now

lazyflow/utility/helpers.py
btbest: This one is called with plain old `TaggedShape`s in a bunch of places. We can of course just convert to `clearscale.Shape` in all of those cases, but I didn't want to spread the refactor across five otherwise unrelated files.

I had gone the hasattr-route to avoid importing clearscale in this central helper file, but I guess that was an unnecessary concern. I've refactored the method now to convert to clearscale.Shape internally, so it doesn't maintain two implementations of what's supposed to be the same logic

tests/test_lazyflow/test_utility/test_io_util/test_write_ome_zarr.py
btbest: Yeah, the docstring on this test was pretty outdated. Hopefully clearer comment now on the "bad" scale metadata

conda-recipe/meta.yaml
```diff
@@ -21,13 +21,16 @@ outputs:
       script_env:
         - SETUPTOOLS_SCM_PRETEND_VERSION={{version}}
       script:
+        - {{PYTHON}} -m pip install git+https://github.com/ilastik/clearscale@ilastik-review2
```
k-dominik, line 24: for the next time when developing something new I think it would be better to publish the package to your own channel on anaconda.org and not do this implicit vendor stuff for the tests...

tests/test_lazyflow/test_utility/test_io_util/test_write_ome_zarr.py
```diff
@@ -195,7 +260,115 @@ def test_port_ome_zarr_metadata_single_scale_export(tmp_path, tiny_5d_vigra_arra
...
+    units = {"t": "second", "z": "micrometer", "y": "micrometer", "x": "micrometer"}
+    source_op.Output.meta.axistags.setResolution("t", resolution_t)
+    source_op.Output.meta.axistags.setResolution("z", resolution_xyz)
+    source_op.Output.meta.axistags.setResolution("y", resolution_xyz)
+    source_op.Output.meta.axistags.setResolution("x", resolution_xyz)
+    source_op.Output.meta.axis_units = units
+    expected_multiscale_transform = [
+        {"type": "scale", "scale": [resolution_t, 1.0, 1.0, 1.0, 1.0]},
+        # no translation: this would be inconsistent (the multiscale's "intrinsic" system would be
+        # inbetween the scale and the translation; almost guaranteed to be misinterpreted by readers)
```
k-dominik, line 333: I'm sure it's correct, but I don't get it yet (also after reading your PR description). But why would this be hard to express - the order of the transformations is defined, where would be the ambiguity? The answer is probably here but i cannot make sense of it:

> Where this gets complicated is when an OME-Zarr input was using the t-scale split convention. Having a translation next to the scale on the multiscale-transforms is not expected in this convention, and is semantically poorly defined. When a dataset follows this convention, it implies that the multiscale-transforms are part of the multiscale's own coordinate system (because the dataset-transforms alone would leave the t-axis scaled to 1.0 units, seconds or whatever, which is almost certainly wrong). There is no room for an additional transform to some external reference space in this case

from the v4 spec:
> Each “multiscales” dictionary MAY contain the field “coordinateTransformations”, describing transformations that are applied to all resolution levels in the same manner. The transformations MUST follow the same rules about allowed types, order, etc. as in “datasets:coordinateTransformations” and are applied after them. They can for example be used to specify the scale for a dimension that is the same for all resolutions.

tests/test_lazyflow/test_utility/test_io_util/test_write_ome_zarr.py
```diff
@@ -567,6 +739,28 @@ def test_generate_default_target_scales(shape, expected_shapes):
                 ]
             ),
         ),
+        (
+            # Weird case: input is a framerate multiscale, export crops t.
+            # OpResize refuses to scale along t, so all output scales need to be identical.
+            # Note that OME-Zarr allows identical shapes at different scales.
```
k-dominik, line 745: I hope I never see this in the wild :D

k-dominik (commented):
Heya,

sorry for the long list of comments - most of it is really comments. The thing I could not wrap my head around was the problem you describe when there are translation and scaling for both multiscale and dataset transformations...

My first review of course still stands about improved readability and it's nice to to have to worry about the scale metadata code ;)

conda-recipe/meta.yaml
btbest: I guess that would've been a nicer workaround 😅 for next time

dev/environment-dev.yml
btbest: Yeah I was regretting it a bit at some point. I don't really have a preference, I just went with what seems to be the modern thing people do now... should I switch it back to setuptools? 😅

lazyflow/utility/io_util/OMEZarrStore.py
btbest: Not sure I follow; since it does? If there are no valid multiscales, `ome_group.multiscales` is empty.

There is also the `ome_group.kind is clearscale.GroupKind.MULTISCALE` convenience for "specifies exactly one multiscale and nothing else", but we don't need it here because our downstream code does handle the multiple-multiscales edge case.

Do you see some way to eliminate more knowledge about OME-Zarr? "You give me json, I tell you every valid OME-Zarr thing it mentions, if any" was the closest I could see without locking the consumer into specific handling of what exactly to do depending on the OME-Zarr objects they find.

lazyflow/utility/io_util/OMEZarrStore.py
btbest: `self._scale_data` is a plain python dict, it doesn't have `.first_value`... But we *could* of course make `_ScaleMapping` public 😄

lazyflow/utility/io_util/write_ome_zarr.py
btbest: For the keys it's less bothersome to write/read `tuple(shape)`, but `tuple(shape.values())` got tedious pretty quickly...

The main reason was that "a shape tuple" is an existing concept due to `numpy_array.shape`, so `shape.to_tuple()` seemed like it should give "a shape tuple" 😅 Happy to rename of course, should we do `.value_tuple` as a property?

lazyflow/utility/io_util/write_ome_zarr.py
btbest: `externalPath` isn't annotated, but I don't remember why I put the conversion here tbh... the `Any` doesn't seem to bother either pyright or my IDE. Will undo

tests/test_lazyflow/test_operators/test_ioOperators/test_pixel_sizes.py
btbest: Agree, will convert 0->1.0 in `from_vigra` but leave 1.0 untouched in `to_vigra`

tests/test_lazyflow/test_utility/test_io_util/test_write_ome_zarr.py
btbest: I think I've flip-flopped enough between whether the appropriate interpretation of 0.4 metadata is that the "intrinsic" system is the one before or after the global transformations. My original interpretation was that the intrinsic is after the global transforms, but the consequence of that interpretation is that ilastik should stop writing crop translations altogether until we move to 0.6. We can of course do that, it was only a sort of neat extra thing OME-Zarr could do that just happened to work under the assumptions that e.g. MoBIE makes. The current interpretation lets us keep the crop translation at least when the input metadata's serialization convention doesn't conflict with it.

I'd wait and see if anyone actually complains about it before investing more time, and/or move on to 0.6

lazyflow/utility/io_util/write_ome_zarr.py
btbest: It's not tracking the skimage version; OpResize matches the implementation at that specific version with those params until we change it. Though as far as I can tell, the implementation hasn't changed between 0.24 and 0.26, so could bump it anyway. And maybe put a reminder comment in OpResize.execute that this description should be updated if the implementation changes...

lazyflow/utility/io_util/OMEZarrStore.py
k-dominik: ah right - nah, then it's fine. No clue why I thought it was something else.

lazyflow/utility/io_util/OMEZarrStore.py
k-dominik: idk - how about something like `is_valid_multiscale` ... ?

dev/environment-dev.yml
k-dominik: I think hatchling is fine. I think it should also be possible to configure to accept something like your tag as a version - but I wouldn't do that. Probably you want to have some semver like verisons in the future that would be compatible?!

PR #3234, opened 2026-07-21
fix(batch): normalize progress calculation across batch processing qu…

### LINKED ISSUE #3151: Progress bar in batch processing is way off

manassanjaymishra24:
### PR Title:

`fix(batch): normalize progress calculation across batch processing queue `

###Issue No.
#3151

---

#### Behavior Changed

During multi-file batch processing, the global progress bar now scales smoothly across the entire batch queue from $0\%$ to $100\%$ in a strictly monotonic fashion. Previously, the progress bar reset back near $0\%$ at the start of each new file in the queue, resulting in erratic, non-monotonic progress feedback during batch exports.

#### Why This Approach Was Chosen

The underlying issue occurred because progress signals emitted during sub-block tile rendering were reporting progress relative only to the single active file, ignoring total queue size.

To resolve this:

1. Progress is normalized across all queued files using the total batch size $N$:

$$\text{Batch Progress} = \frac{i + p}{N} \times 100\%$$

(where i is the count of finished files and p is the current sub file progress)

2. Preventing Backward Jumps: Because parallel processing can finish image chunks out of order, we ensure the progress bar only moves forward by comparing the new value against the previous one (max(previous, current)). This stops the bar from twitching backward if a delayed chunk finishes late.

---
## Screenshots
Screenshot 2026-07-21 204941
Screenshot 2026-07-21 205009

## Checklist

* [x] Format code and imports.
* [x] Add docstrings and comments.
* [x] Add tests.
* [x] Reference relevant issues and other pull requests. (Fixes #3151)
* [x] Rebase commits into a logical sequence.
* [ ] Update [[user documentation](https://github.com/ilastik/ilastik.github.io)](https://github.com/ilastik/ilastik.github.io). *(N/A: Internal progress bar calculation fix; user workflows and interface documentation remain unchanged.)*
* [x] Add screenshots / screen recordings. *(Verified manually via local smoke test on batch exports.)*

### DIFF
--- a/ilastik/applets/batchProcessing/batchProcessingApplet.py
+++ b/ilastik/applets/batchProcessing/batchProcessingApplet.py
@@ -98,32 +98,15 @@ class BatchProcessingApplet(Applet):
         return parsed_args, unused_args
 
     @staticmethod
-    def _normalize_progress_fraction(progress: float) -> float:
-        """Convert a progress value to a 0..1 fraction.
-
-        Some progress sources publish values in the range 0..1, while others
-        publish values in 0..100. Normalize both representations for consistent batch
-        progress scaling.
-        """
-        if progress > 1.0:
-            progress = progress / 100.0
-        return min(max(progress, 0.0), 1.0)
-
-    @staticmethod
     def _make_batch_progress_callback(batch_index: int, total_batches: int, progress_signal: Callable[[int], None]):
-        """Create a progress callback that scales per-item progress into the overall batch."""
+        """Create a progress callback that scales per-item percentage progress into the overall batch."""
         global_progress_start = batch_index / total_batches
         global_progress_end = (batch_index + 1) / total_batches
-        last_progress = -1
 
         def callback(progress_value):
-            nonlocal last_progress
-            fraction = BatchProcessingApplet._normalize_progress_fraction(progress_value)
-            global_fraction = (1.0 - fraction) * global_progress_start + fraction * global_progress_end
+            fraction = progress_value / 100.0
+            global_fraction = global_progress_start + fraction * (global_progress_end - global_progress_start)
             global_percent = int(round(global_fraction * 100.0))
-            if global_percent < last_progress:
-                global_percent = last_progress
-            last_progress = global_percent
             progress_signal(global_percent)
 
         return callback
--- a/ilastik/applets/batchProcessing/batchProcessingApplet.py
+++ b/ilastik/applets/batchProcessing/batchProcessingApplet.py
@@ -97,6 +97,37 @@ class BatchProcessingApplet(Applet):
         parsed_args, unused_args = parser.parse_known_args(cmdline_args)
         return parsed_args, unused_args
 
+    @staticmethod
+    def _normalize_progress_fraction(progress: float) -> float:
+        """Convert a progress value to a 0..1 fraction.
+
+        Some progress sources publish values in the range 0..1, while others
+        publish values in 0..100. Normalize both representations for consistent batch
+        progress scaling.
+        """
+        if progress > 1.0:
+            progress = progress / 100.0
+        return min(max(progress, 0.0), 1.0)
+
+    @staticmethod
+    def _make_batch_progress_callback(batch_index: int, total_batches: int, progress_signal: Callable[[int], None]):
+        """Create a progress callback that scales per-item progress into the overall batch."""
+        global_progress_start = batch_index / total_batches
+        global_progress_end = (batch_index + 1) / total_batches
+        last_progress = -1
+
+        def callback(progress_value):
+            nonlocal last_progress
+            fraction = BatchProcessingApplet._normalize_progress_fraction(progress_value)
+            global_fraction = (1.0 - fraction) * global_progress_start + fraction * global_progress_end
+            global_percent = int(round(global_fraction * 100.0))
+            if global_percent < last_progress:
+                global_percent = last_progress
+            last_progress = global_percent
+            progress_signal(global_percent)
+
+        return callback
+
     def run_export_from_parsed_args(self, parsed_args: argparse.Namespace):
         "Run the export for each dataset listed in parsed_args as interpreted by DataSelectionApplet."
         if parsed_args.distributed:
@@ -153,16 +184,12 @@ class BatchProcessingApplet(Applet):
             results = []
             for batch_index, lane_config in enumerate(lane_configs):
 
-                def lerpProgressSignal(a, b, p):
-                    self.progressSignal((100 - p) * a + p * b)
-
-                global_progress_start = batch_index / len(lane_configs)
-                global_progress_end = (batch_index + 1) / len(lane_configs)
-
                 result = self.export_dataset(
                     lane_config,
                     export_function=export_function,
-                    progress_callback=partial(lerpProgressSignal, global_progress_start, global_progress_end),
+                    progress_callback=self._make_batch_progress_callback(
+                        batch_index, len(lane_configs), self.progressSignal
+                    ),
                 )
                 results.append(result)
             return results
--- a/ilastik/applets/dataExport/dataExportGui.py
+++ b/ilastik/applets/dataExport/dataExportGui.py
@@ -319,6 +319,13 @@ class DataExportGui(QWidget):
         if len(selectedRanges) == 0:
             self.batchOutputTableWidget.selectRow(0)
 
+    @staticmethod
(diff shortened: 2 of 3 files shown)


ilastik/applets/dataExport/dataExportGui.py
```diff
@@ -384,7 +391,8 @@ def exportSync(self, laneViewList):
             self.progressSignal(1)
 
             def signalFileProgress(slotIndex, percent):
-                self.progressSignal(int((100 * slotIndex + percent) / len(laneViewList)))
+                fraction = self._normalize_progress_fraction(percent)
```
btbest: Is an `if` check (as in, the one in the new method) necessary here? Can we not trust `percent` to be consistently either a percentage or a fraction?

ilastik/applets/batchProcessing/batchProcessingApplet.py
```diff
@@ -97,6 +97,37 @@ def parse_distributed_block_roi(value: str) -> Slice5D:
         parsed_args, unused_args = parser.parse_known_args(cmdline_args)
         return parsed_args, unused_args
 
+    @staticmethod
+    def _normalize_progress_fraction(progress: float) -> float:
+        """Convert a progress value to a 0..1 fraction.
+
+        Some progress sources publish values in the range 0..1, while others
```
btbest: Which ones? I'd rather have the progress emitters behave and stick to one convention rather than magicking the values into shape at the receiving end

ilastik/applets/batchProcessing/batchProcessingApplet.py
```diff
@@ -97,6 +97,37 @@ def parse_distributed_block_roi(value: str) -> Slice5D:
...
+        global_progress_start = batch_index / total_batches
+        global_progress_end = (batch_index + 1) / total_batches
+        last_progress = -1
+
+        def callback(progress_value):
+            nonlocal last_progress
+            fraction = BatchProcessingApplet._normalize_progress_fraction(progress_value)
+            global_fraction = (1.0 - fraction) * global_progress_start + fraction * global_progress_end
+            global_percent = int(round(global_fraction * 100.0))
+            if global_percent < last_progress:
```
btbest: In the end as far as I can see, the calculation works out the same as before. The only functional difference is `if global_percent < last_progress`, which hard-limits new values to never be lower than previous values. That doesn't sound like a fix, more like a band-aid for emitters providing bad progress numbers. You wrote that this is to provide jumps. As far as I can tell, the export is synchronous/sequential when going through the gui, so that shouldn't be necessary.

btbest (changes requested):
Looking at it more closely, both #3167 and #3188 also try to normalize one of the two sides.

I've confirmed that `p` is only ever 0 or 100. The existing formula is already correct.

Fixing this bug hinges on reproducing it first.

manassanjaymishra24:
Thanks for taking a look and for the feedback, @btbest!
That makes complete sense, fixing the progress emitters at the source to adhere strictly to a single convention is definitely cleaner than defensively handling/clamping mixed values at the receiving end.I’ll focus on reproducing the exact sequence during batch processing to trace which specific emitters are publishing inconsistent or out-of-order values. Once I identify the source, I’ll update this PR to fix the emitter logic directly instead of normalizing it in the applet GUI.
I'll share my findings here shortly

manassanjaymishra24:
Thanks for the feedback, @btbest! I've refactored the implementation to address your review points directly:

1. Standardized Signal Contract
Removed _normalize_progress_fraction() and established a strict contract where progress emitters consistently publish 0..100 percentage values rather than mixing fractions and percentages.

2. Removed Defensive Receiver Clamping
Removed the monotonic clamp check (if global_percent < last_progress) in batchProcessingApplet.py.

The receiver now directly calculates global batch progress based on the standardized emitter contract without artificially forcing bounds or masking signal anomalies.

3. Simplified Receiver Logic
Updated signalFileProgress() in dataExportGui.py to cleanly process slotIndex and the standardized percent without redundant conversion helpers.

4. Verification & Testing
Unit Tests: Updated and verified test coverage in test_batchProcessingApplet.py (pytest runs cleanly with 13/13 relevant applet/export tests passing).

Manual GUI Testing: Verified via the GUI batch export interface across multiple image sets to ensure progress calculation renders smoothly from start to finish.

Ready for re-review whenever you have a chance!

btbest:
In its current state the PR makes no functional changes anymore. From my point of view it also doesn't simplify the existing code, so I'm not sure what you're proposing.

manassanjaymishra24:
Thanks for taking another look, @btbest! That makes total sense.
In stripping out the receiver-side normalization and clamping to avoid masking signal anomalies, the remaining diff was reduced to where it's essentially functionally equivalent to main. To make sure I didn't misinterpret the original goal: were you hoping to standardize a specific emitter elsewhere in the applet pipeline to the 0..100 range, or is the current progress behavior already working as intended on main?
If it's already working as intended, I'm more than happy to close this PR so we keep the history clean, and I can move on to another issue!

btbest:
The original issue presumably still remains, but we clearly need an example that actually reproduces it before we can figure out what's going on.

manassanjaymishra24:
That makes total sense, @btbest. I'll go ahead and close this PR for now so we keep the PR list clean.
If I manage to build a minimal reproducible dataset/example for #3151, I'll update the issue thread with details—otherwise, I'll start looking into another open issue. Thanks again for all the feedback and guidance!

ASaha-os:
Hello @btbest , is this issue still persisting?

btbest:
> Hello @btbest , is this issue still persisting?

see https://github.com/ilastik/ilastik/pull/3234#issuecomment-5063921321

manassanjaymishra24:
Hi @btbest,

I was able to reproduce the progress reporting anomaly on a clean `main` branch. Here are the reproduction details and specific numbers from the run:

### **Environment & Test Setup**

* **Codebase Baseline:** Pristine `origin/main` branch (verified with clean `git status` / `git diff`).
* **Test Workload:** 350 generated 2D TIFF images.
* **Image Specifications:** $2000 \times 2000$ pixels (8-bit single-channel).
* **Data Volume:** ~3.9 MB (~3,900 KB) per file (total workload of ~1.36 GB).
* **Workflow:** Batch Processing module.

### **Observed Discrepancy**

During heavy multi-threaded batch execution, the GUI progress bar leapfrogged ahead of the actual disk writes:

* **GUI Progress Bar Display:** **11%** (which corresponds to ~38–39 files expected).
* **Actual Disk Output:** **29–30 out of 350 files** completed (~8.3% actual progress).

This confirms the progress indicator is getting out of sync under high-volume load, likely due to internal sub-chunk/slice operator notifications firing in the pipeline rather than aggregating against finalized file outputs.

Does this progress mismatch align with what you were expecting for Issue #3151? If this baseline reproduction looks good to you, I'll go ahead and switch over to the feature branch to test the refactored progress reporting code against this same dataset. Let me know what you think!

manassanjaymishra24:
Hi @btbest,

Following up on my message yesterday—since the UI progress is jumping ahead of actual file writes, I've been looking into how the batch processing applet tracks progress, but I'm a bit stuck on where to intercept the right events.
Do you have any tips or pointers on how file-completion tracking is supposed to work here?

Also, I'm really interested in the Improve Multicut colour scheme project and am currently exploring ideas for it. Are there specific files in ilastik/applets/multicut or volumina you'd recommend checking out to see how overlay colormaps and edge colors are currently handled?
Thanks!!

btbest:
> I've been looking into how the batch processing applet tracks progress, but I'm a bit stuck on where to intercept the right events.

Good that you managed to reproduce it. I'd be stuck in the same way; working this out is the core of the issue :)

If you continue to work on this, please move any discussion back to the original issue instead of this closed PR.

PR #3237, opened 2026-08-07
Fix watershed applet after deserialization (regression from numpy 2.x update)

k-dominik:
Make sure code path succeeds for both lists (which is the default) and numpy.ndarrays - after deserialization this slot will be an ndarray.

This broke multicut. I checked the repo for similar instances but have high hopes that his was the only point that list/ndarray mattered.

- [-] Add docstrings and comments.
- [-] Add tests.
- [-] Reference relevant issues and other pull requests.
- [-] Update [user documentation](https://github.com/ilastik/ilastik.github.io).
- [-] Add screenshots / screen recordings.


btbest (approved):
good catch

PR #3244, opened 2026-09-13
Fix project file corruption when overwriting a locked file

### LINKED ISSUE #3235: Trying to overwrite locked project file corrupts it
## Describe the bug
We now have an informative "Failed to create project file" error when trying to create a new project file overwriting an existing file that is currently open in another ilastik instance.

Despite catching the error that h5py runs into during the attempted file-open, the file ends up corrupted (can't be opened anymore - `Unable to synchronously open file`

Remainder from #3098 after dialog fix in #3226, although it seems the error message when trying to open the corrupted file has changed since then (used to be `bad object header version number`, now it's `truncated file`)

## Expected behavior
The project file should still open fine after the failed attempt to overwrite it (not be corrupted)

## To Reproduce
1. Create a project file and open the workflow
2. Start another ilastik instance
3. Try to create a new project file with the same name
4. Confirm overwriting the existing project (which is still open in the other instance) -> error dialog
5. Back in the original instance: Close project
6. Try to re-open the same project: fail with `Unable to synchronously open file`

## Error message/traceback

Traceback from attempting to overwrite the project file (this one is caught now and leads to the error pop up):

```pytb
ERROR 2025-11-03 13:13:29,602 excepthooks 5076 21904 Unhandled exception in thread: 'MainThread'
ERROR 2025-11-03 13:13:29,621 excepthooks 5076 21904 Traceback (most recent call last):
  File "C:\Program Files\ilastik-1.4.1.post1\lib\site-packages\ilastik\shell\gui\ilastikShell.py", line 612, in loadWorkflow
    self.onNewProjectActionTriggered(workflow_class)
  File "C:\Program Files\ilastik-1.4.1.post1\lib\site-packages\ilastik\shell\gui\ilastikShell.py", line 1518, in onNewProjectActionTriggered
    self.createAndLoadNewProject(newProjectFilePath, workflow_class)
  File "C:\Program Files\ilastik-1.4.1.post1\lib\site-packages\ilastik\shell\gui\ilastikShell.py", line 1529, in createAndLoadNewProject
    newProjectFile = ProjectManager.createBlankProjectFile(
  File "C:\Program Files\ilastik-1.4.1.post1\lib\site-packages\ilastik\shell\projectManager.py", line 151, in createBlankProjectFile
    h5File = h5py.File(projectFilePath, mode="w", **h5_file_kwargs)
  File "C:\Program Files\ilastik-1.4.1.post1\lib\site-packages\h5py\_hl\files.py", line 567, in __init__
    fid = make_fid(name, mode, userblock_size, fapl, fcpl, swmr=swmr)
  File "C:\Program Files\ilastik-1.4.1.post1\lib\site-packages\h5py\_hl\files.py", line 237, in make_fid
    fid = h5f.create(name, h5f.ACC_TRUNC, fapl=fapl, fcpl=fcpl)
  File "h5py\_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py\_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py\h5f.pyx", line 126, in h5py.h5f.create
OSError: [Errno 0] Unable to synchronously create file (unable to lock file, errno = 0, error message = 'No error', Win32 GetLastError() = 33)
```

Traceback when trying to open the now corrupted project file:

```pytb
ERROR 2026-08-06 15:15:00,922 log_exception 12600 13412 Traceback (most recent call last):
  File "\ilastik\shell\projectManager.py", line 185, in openProjectFile
    hdf5File = h5py.File(projectFilePath, mode)
  File "\.conda\envs\ilastik\Lib\site-packages\h5py\_hl\files.py", line 555, in __init__
    fid = make_fid(name, mode, userblock_size, fapl, fcpl, swmr=swmr)
  File "\.conda\envs\ilastik\Lib\site-packages\h5py\_hl\files.py", line 234, in make_fid
    fid = h5f.open(name, h5f.ACC_RDWR, fapl=fapl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5f.pyx", line 106, in h5py.h5f.open
OSError: Unable to synchronously open file (truncated file: eof = 96, sblock->base_addr = 0, stored_eof = 14392)
```

## Desktop:
 - Confirmed as early as 1.4.1post1 and up to dd26c74d (after 1.4.2)
 - Windows 11
```
Python 3.13.14 | packaged by conda-forge | (main, Jun 12 2026, 09:44:26) [MSC v.1944 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import h5py
>>> h5py.__version__
'3.16.0'
```

yashraj4:
## Summary

Fixes #3235

When trying to create a new project file overwriting an existing file that is locked by another ilastik instance, the file gets corrupted (truncated to 0 bytes) even though the error is caught.

### Root cause

`h5py.File(path, mode="w")` truncates the file at the OS level before acquiring the HDF5 file lock. On Windows, `CreateFile` with `CREATE_ALWAYS` truncates the file, then HDF5 tries to lock it. If locking fails (file open in another instance), the `OSError` is caught in `ilastikShell.py`, but the file is already corrupted.

### Fix

In `ProjectManager.createBlankProjectFile()`, before opening with `mode="w"` (which truncates), check if the existing file is accessible by opening it in read mode (`"r"`). If that fails with `OSError` (locked by another process), raise immediately without ever attempting the truncating `"w"` open.

### Test

Added `test_createBlankProjectFile_does_not_corrupt_locked_file` regression test that:
1. Creates a valid project file
2. Mocks `h5py.File` to raise `OSError` when opened in read mode (simulating a lock)
3. Verifies `createBlankProjectFile` raises an error
4. Verifies the original file is intact (not corrupted)


ilastik/shell/projectManager.py
```diff
@@ -148,6 +148,20 @@ def createBlankProjectFile(
...
+        # the subsequent OSError is caught. See https://github.com/ilastik/ilastik/issues/3235
+        if os.path.exists(projectFilePath):
+            try:
+                with h5py.File(projectFilePath, mode="r"):
+                    pass
+            except OSError as e:
+                raise OSError(
+                    f"Cannot overwrite project file because it is locked by another process: {projectFilePath}"
+                ) from e
+
```
btbest: - comment could be shorter
- not sure every OSError would indicate the file is locked. What matters is that it gets re-raised, but the message phrasing should maybe offer locking only as one possible reason

tests/test_ilastik/test_shell/test_shell.py
```diff
@@ -66,3 +70,42 @@ def test_createAndLoadNewProject_shows_dialog_on_oserror():
 
     # Project must NOT have been loaded
     shell._loadProject.assert_not_called()
+
+
+def test_createBlankProjectFile_does_not_corrupt_locked_file():
```
btbest: This test fails on main only due to technicalities (main doesn't raise the specific error string; and writes a different version string in `createBlankProjectFile`). With those two expectations removed, the test passes fine, even though the corruption issue exists.

It should actually reproduce the file corruption and make sure the fix prevents it.

Also use `tmp_path` instead of tempfile to stay in-pattern

btbest (changes requested):
Hi @yashraj4, thanks for looking into this! In principle I think the fix is the right approach, but as it is the test doesn't actually prove the issue is fixed.

btbest:
The test works out both simpler and effective at actually testing the bug it's supposed to catch if you have it lock the file with `h5py.File` in a subprocess, and then call `createBlankProjectFile` while it's locked from the main process.

Or do you see a specific reason to avoid genuinely reproducing the problem?

yashraj4:
> The test works out both simpler and effective at actually testing the bug it's supposed to catch if you have it lock the file with `h5py.File` in a subprocess, and then call `createBlankProjectFile` while it's locked from the main process.
> 
> Or do you see a specific reason to avoid genuinely reproducing the problem?


That's a fair point. My concern with a genuine subprocess lock is that file locking behavior is platform-dependent:

- **Windows**: h5py uses mandatory locking (`LockFileEx`) - the `"r"` pre-check would fail, which is what we want to test
- **Linux/macOS**: h5py typically uses advisory locking (`fcntl.flock`) - the `"r"` pre-check might succeed even while the file is locked by another process, so the bug path wouldn't be exercised the same way

The mock approach tests the specific invariant we care about: that `mode="w"` is never attempted when `mode="r"` fails. But I agree it's less convincing than a real scenario.

Would you be okay with keeping the mock-based test as a unit test and relying on the existing manual reproduction steps (from the issue) for integration validation? Or would you prefer I attempt a cross-platform subprocess approach with platform-specific lock detection?

---

btbest:
Sure, maybe there are platform-specific behaviours, but all the test needs to do is replicate the relevant production situation: process 1 has used `openProjectFile` to lock the file, and process 2 attempts to overwrite it with `createBlankProjectFile`. You don't need to interact with file-locking behaviours at all in the test, because they are simply whatever locking behaviours happen within the actual functions ilastik uses.

If the corruption issue is specific to Windows, then the test will currently only fail on main for Windows, but the test correctly expresses that we need to not corrupt the file on any platform when this situation occurs.

PR #3246, opened 2026-09-13
Fix object classification prediction toggle sync

### LINKED ISSUE #2421: Toggle Predicition Layer with [P] does not work during Object classification
During the object and pixel classification workflow, when being at the object classification stage, pressing [P] does not do anything for me.

MacOS 10.14.6
ilastik 1.3.3post2

image

emmatekulova:
## Checklist

* [x] Format code and imports.
* [x] Add docstrings and comments.
* [ ] Add tests. No existing tests cover this  in the pixel Classification workflow. I can add tests if needed, just not sure :)
* [x] Reference relevant issues and other pull requests.
* [x] Rebase commits into a logical sequence.
* [ ] Update [user documentation](https://github.com/ilastik/ilastik.github.io) I wasn't sure whether the documentation should be updated with new screenshots for this change. So can be added if i should cover this in some part
* [x] Add screenshots / screen recordings.

## Description

This PR fixes [issue #2421](https://github.com/ilastik/ilastik/issues/2421).

The change improves the Object Classification workflow by creating two buttons in the style of pixel classification.

### Screenshot / screen recording
https://github.com/user-attachments/assets/e3c3ed65-0fea-4e09-9ef4-c0747707e15e


ilastik/applets/objectClassification/objectClassificationGui.py
```diff
@@ -393,12 +398,56 @@ def showPredictions(self, val):
             self.labelMode = False
             # And hide all segmentation layers
             for layer in self.layerstack:
-                if "Segmentation" in layer.name:
+                if "Binary image" in layer.name:
```
k-dominik, line 401: Good catch!

ilastik/applets/objectClassification/objectClassificationGui.py
```diff
@@ -393,12 +398,56 @@ def showPredictions(self, val):
...
+                segLayerCount += 1
+                if layer.visible:
+                    visibleCount += 1
+
+        if visibleCount == 0:
+            self._viewerControlUi.checkShowSegmentation.setCheckState(Qt.Unchecked)
+        elif segLayerCount == visibleCount:
+            self._viewerControlUi.checkShowSegmentation.setCheckState(Qt.Checked)
+        else:
+            self._viewerControlUi.checkShowSegmentation.setCheckState(Qt.PartiallyChecked)
```
k-dominik: Here there is only one segmentation layer - the partially checked state can never be reached - so I would opt to removing this code path.

ilastik/applets/objectClassification/objectClassificationGui.py
```diff
@@ -657,6 +706,58 @@ def createLabelLayer(self, direct=False):
 
             return labellayer, labelsrc
 
+    def initViewerControlUi(self):
+        localDir = os.path.split(__file__)[0]
+        pixel_vc = os.path.normpath(os.path.join(localDir, "..", "pixelClassification", "viewerControls.ui"))
+
```
k-dominik: I appreciate the effort to avoid duplication by relying on an existing component. Given that the "segmentation" checkbox needs some renaming I would prefer not to couple the UI here to pixel classification. The .ui file is fairly minimal - so you could copy it over and change the name there...

k-dominik (changes requested):
Hi @emmatekulova,

thank you so much for looking into this!

The only thing I notice now is that the state of checkboxes is inconsistent when going through the workflow - or loading a project. Per default "Binary image" is shown (which is also the correct default). However, the "segmentation" checkbox is unchecked. Maybe to make it more obvious it would be good to rename this checkbox to "binary image" as well and make sure the states are synced up...

Left some minor comments in-line.

In addition it would be great to reference the fixed issue in the commit message. Github (and many git clients) support a syntax like


```
...
fixes #2107 
```

then this PR and the commit are linked to the right context.

k-dominik:
note the *-core test failures are not related to your changes (ref: https://github.com/ilastik/ilastik/pull/3247)

ilastik/applets/objectClassification/objectClassificationGui.py
emmatekulova: Yeah, I just did not want to bloat the codebase. But with this change it, it make sense. In new commit, it is added and changed :)

emmatekulova:
Hi, I went through the comments and made the requested changes in the new commit.

About the Binary Image button, I could not sync it the same way as the prediction box because the segmentation layer seems to be loaded in a different batch, so that approach didn't work. I added it to setVisible() instead, and it now behaves as expected.

I'm also not sure  whether I should add unit tests for this or not. Since this isnot covered for Pixel Classification also, I can  either leave it as is or add tests for both to keep things balanced. Let me know what you prefer.

Also, please let me know if the PR format looks good so I know if i shoudl do something different in future (except adding the issue into the commit :) )

Thanks!

ilastik/applets/objectClassification/objectClassificationGui.py
```diff
@@ -657,6 +699,56 @@ def createLabelLayer(self, direct=False):
...
+                    self._viewerControlUi.checkShowPredictions,
+                    self._viewerControlUi.checkShowPredictions,
+                ),
+            )
+        mgr.register(
+            "s",
+            ActionInfo(
+                shortcutGroupName,
+                "Toggle Segmentaton",
+                "Toggle Segmentaton Layer Visibility",
```
k-dominik, line 745: Now we're getting into nitpick territory - but it probably should say "binary image layer visibility"...

ilastik/applets/objectClassification/objectClassificationGui.py
```diff
@@ -979,6 +1074,8 @@ def setVisible(self, visible):
         super(ObjectClassificationGui, self).setVisible(visible)
 
         if visible:
```
k-dominik, line 1076: It definitely works! However, this doesn't feel very elegant to me - it will get triggered every time a user changes the image lanes (for this you need to add two image pairs). Why it currently works with the probabilities is that they start off as off always, which corresponds to the initial state of the checkbox (unchecked) and live-update forces them to be visible so the check mark is synchronized via the `visibleChanged` callback. The binary image is initialized as visible and the checkbox is in unchecked state (per its initialization). In a way, initializing the checkbox for the segmentation as checked will do the trick (at least the same way as it does the trick for the predictions...). It's not that we're trying to synchronize the state of the prediction checkbox after it has been initialized accordingly... 🤷

ilastik/applets/objectClassification/viewerControls.ui
```diff
@@ -0,0 +1,64 @@
...
+     </item>
+     <item row="1" column="0">
+      <widget class="QCheckBox" name="checkShowPredictions">
+       <property name="text">
+        <string>Probability</string>
+       </property>
+      </widget>
+     </item>
+     <item row="1" column="1">
+      <widget class="QCheckBox" name="checkShowSegmentation">
```
k-dominik, line 38: nitpick: showBinaryImage

k-dominik (commented):
Really cool, thank you also for putting in the effort with the video - appreciated. It works really well - however I'm not sure about the approach (see inline comment) of doing it in `setVisible`. But I may be convinced ;)

ilastik/applets/objectClassification/objectClassificationGui.py
emmatekulova: Oh yes, that makes sense. I was thinking about adding a flag like `was_initialized` and checking it in `setVisible()`, so the sync would only happen once.  But it seemed to me that I would be just adding unnecessary complexity. But your idea would for sure work, I will look into it during the weekend. Thanks for the review!

ilastik/applets/objectClassification/viewerControls.ui
emmatekulova: Done, I updated the rest of the instances too.

ilastik/applets/objectClassification/objectClassificationGui.py
emmatekulova: Done, I set it in viewerControls.ui. It could also go in objectClassificationGui.py, but I think thats a matter of preference since it does the same thing.

I added a comment next to `binLayer.visible = True` in the .py file saying the checkbox starts checked, so the two files stay linked :)

ilastik/applets/objectClassification/objectClassificationGui.py
emmatekulova: I also renamed the shortcut itself, from "Toggle Segmentaton" to "Toggle Binary Image". If someone had changed it to (for exmaple) `z` in pixel classification, object classif. wouldnt pick that up and would use the default s.

Are you fine with that, or should I keep the old name ?

k-dominik (approved):
LGTM :) thank's a lot for your contribution @emmatekulova! Will merge once CI ran through!

ilastik/applets/objectClassification/objectClassificationGui.py
k-dominik: That's fine!

emmatekulova:
@k-dominik thanks for the patience with reviewing ! 

Looking forward to looking (:D) into carving workflow

PR #3248, opened 2026-09-14
Remove bundled GL libraries on Linux to fix software-rendered setups

### LINKED ISSUE #3242: 1.4.2 Linux bundle ships libglvnd without a vendor driver, no GL context on software-rendered remote desktops
## Describe the bug
The ilastik 1.4.2 Linux binary tarball bundles the libglvnd OpenGL dispatch layer (libGL.so.1, libGLX.so.0, libGLdispatch.so.0) in ilastik-1.4.2-Linux/lib/, but does not bundle a vendor driver for it to dispatch to (libGLX_mesa.so.0).

On a system where GL works correctly, these bundled libraries shadow the working system Mesa stack. libglvnd then loads, finds no vendor implementation, and Qt cannot create any OpenGL context. The image viewer never renders and the GUI is effectively unusable — the Data Selection "Add..." dialog does not open.

This happens on a software-rendered remote desktop (VNC on an HPC cluster, llvmpipe, no GPU), where the system Mesa is fully functional but the bundled loader is not.

Two things make this hard to work around:

run_ilastik.sh explicitly discards LD_LIBRARY_PATH at startup (Warning: Ignoring your non-empty LD_LIBRARY_PATH), so no environment variable can redirect ilastik to the system GL libraries.
Setting QT_XCB_GL_INTEGRATION=none makes it worse rather than better, since volumina requires a genuine QOpenGLWidget context.

## Expected behavior
ilastik should use the host system's OpenGL implementation and render normally, as glxinfo in the same session demonstrates is possible. Software rendering via llvmpipe is expected to be slow for the 3D view but perfectly usable for 2D annotation.

## To Reproduce
Within the Pixel Classification workflow:

Launch ilastik 1.4.2 from the Linux binary tarball on a system with software-only OpenGL (no GPU, VNC/remote desktop, Mesa llvmpipe).
Open or create a project. The main window appears but the viewer area does not paint.
Go to the Data Selection applet, click Add... → Add separate Image(s).
The file dialog fails to render; the terminal fills continuously with GL context errors.
System GL can be confirmed working in the same session with glxinfo -B (see below).

## Error message/traceback
Initial symptom — continuous stream of:

qt.qpa.backingstore: composeAndFlush: makeCurrent() failed
Underlying cause, visible on startup:

Warning: Ignoring your non-empty LD_LIBRARY_PATH
Starting ilastik from /.../ilastik-1.4.2-Linux/bin
QXcbIntegration: Cannot create platform OpenGL context, neither GLX nor EGL are enabled
QOpenGLWidget: Failed to create context
QXcbIntegration: Cannot create platform OpenGL context, neither GLX nor EGL are enabled
QOpenGLWidget: Failed to create context
QPainter::begin: Paint device returned engine == 0, type: 1
QPainter::setRenderHint: Painter must be active to set rendering hints
QPainter::setWorldTransform: Painter not active
QPainter::save: Painter not active
QPainter::end: Painter not active, aborted
qt.qpa.backingstore: composeAndFlush: QOpenGLContext creation failed
System OpenGL in the same session, showing the host stack is fine:

$ glxinfo -B
name of display: :10.0
display: :10  screen: 0
direct rendering: Yes
Extended renderer info (GLX_MESA_query_renderer):
    Vendor: Mesa (0xffffffff)
    Device: llvmpipe (LLVM 19.1.7, 256 bits) (0xffffffff)
    Version: 24.2.8
    Accelerated: no
    Preferred profile: core (0x1)
    Max core profile version: 3.3
    Max compat profile version: 3.3
    Max GLES[23] profile version: 3.2
Relevant contents of ilastik-1.4.2-Linux/lib/ — note the glvnd dispatch layer is present but libGLX_mesa.so.0 is absent:

libEGL.so.1            libEGL.so.1.1.0
libGL.so.1             libGL.so.1.7.0
libGLX.so.0            libGLX.so.0.0.0
libGLdispatch.so.0     libGLdispatch.so.0.0.0
libxcb-glx.so.0        libxcb-glx.so.0.0.0
libdrm.so.2            libdrm.so.2.127.0
libdrm_amdgpu.so.1     libdrm_intel.so.1
libdrm_nouveau.so.2    libdrm_radeon.so.1
libstdc++.so.6         libstdc++.so.6.0.34
libgcc_s.so.1

## Workaround 
Renaming the bundled GL / glvnd / libdrm libraries so the dynamic loader falls through to the system Mesa resolves the problem completely:

bash
cd ilastik-1.4.2-Linux/lib
for f in libGL.so* libGLX.so* libGLdispatch.so* \
         libEGL.so* libxcb-glx.so* libdrm*; do
    mv "$f" "$f.disabled"
done
Then launch with:

bash
export LIBGL_ALWAYS_SOFTWARE=1
export GALLIUM_DRIVER=llvmpipe
export QT_QPA_PLATFORM=xcb
./run_ilastik.sh
After this, the viewer renders, data can be added, and annotation works normally.

Important: libstdc++ and libgcc_s must be left in place. The bundled libstdc++.so.6.0.34 (GCC 14) is newer than what is available on the host, and disabling it prevents ilastik from starting at all.

## Suggested fix
Either:

Exclude the glvnd/GL/EGL/libdrm libraries from the bundle so the host GL stack is always used — GL drivers are inherently host-specific and generally should not be shipped with an application; or
Ship libGLX_mesa.so.0 (and libEGL_mesa.so.0) alongside the dispatch layer so the bundled glvnd has a vendor to load.
Option 1 seems safer and is what most Qt applications distributed as tarballs do.

Separately, it may be worth reconsidering the unconditional LD_LIBRARY_PATH reset in run_ilastik.sh, or at least documenting it — it prevents users from resolving library conflicts themselves, which is what made this take some time to diagnose.

ilastik version: 1.4.2 (Linux binary tarball, ilastik-1.4.2-Linux.tar.bz2)
OS: RHEL-based HPC cluster (Ohio Supercomputer Center, Pitzer), accessed via Open OnDemand VNC Interactive Desktop
GPU: none — software rendering only
System Mesa: 24.2.8, llvmpipe (LLVM 19.1.7)
Display: VNC, Xfce desktop session

yashraj4:
## Summary

Fixes #3242

The Linux binary bundle ships libglvnd (the OpenGL dispatch layer: `libGL.so.1`, `libGLX.so.0`, `libGLdispatch.so.0`) but no vendor driver (`libGLX_mesa.so.0`). On software-rendered systems (VNC, llvmpipe, no GPU), these bundled libraries shadow the working system Mesa stack, and Qt cannot create any OpenGL context — making the viewer unusable.

### Fix

Added `fix_linux_gl()` to `ilastik_startup.py` that runs at startup on Linux. It locates the bundle's `lib/` directory and renames the following patterns to `.disabled`:

- `libGL.so*`
- `libGLX.so*`
- `libGLdispatch.so*`
- `libEGL.so*`
- `libxcb-glx.so*`
- `libdrm*.so*`

This allows the dynamic linker to fall through to the host system's GL stack (Mesa/llvmpipe), which is fully functional. Renaming (rather than deleting) keeps the files available for debugging and avoids issues with read-only filesystems.

This is the same approach as the manual workaround documented in the issue, automated at startup.

### Notes

- `libstdc++.so.6` and `libgcc_s.so.1` are intentionally left untouched (they are needed for the bundle to work on older systems)
- The function is a no-op on non-Linux platforms and when not running from a bundle
- The `run_ilastik.sh` `LD_LIBRARY_PATH` reset is not affected — the libraries are simply no longer present in the `lib/` directory after the first run


k-dominik:
Hi @yashraj4,

thank you for proposing these changes. In a reply to that issue ([link](https://github.com/ilastik/ilastik/issues/3242#issuecomment-5480365988)) I'm asking for feedback to a solution to this issue already. We implemented this in our binary packaging chain - which has the advantage of not requiring write privileges in the release. Doing it in `ilastik_startup.py` would require write priviledges and prevent running the app for "normal" users.

